PR #6, opened 2015-11-30
Fixed broken return path to job from notification page

yahoopete:


antimega:
Smallest. Pull request. Ever.

PR #11, opened 2015-12-01
Implement register

servingUpAces:
Implement the register endpoints.
Creates a new user, using role 'platform_admin' as a default. State is defaulted to pending.
Include validation on registration form. 
Email address must be gov.uk.
Password is at least 10 characters long.
Mobile number is a +44 number
Form validation errors will be added to the template in another story.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,6 +1,6 @@
 from flask_wtf import Form
 from wtforms import StringField, PasswordField
-from wtforms.validators import DataRequired, Email, Length
+from wtforms.validators import DataRequired, Email, Length, Regexp
 
 
 class LoginForm(Form):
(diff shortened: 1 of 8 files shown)


app/main/forms.py
```diff
@@ -12,3 +12,24 @@ class LoginForm(Form):
     password = PasswordField('Password', validators=[
         DataRequired(message='Please enter your password')
     ])
+
+
+gov_uk_email = "(^[^@^\\s]+@[^@^\\.^\\s]+(\\.[^@^\\.^\\s]*)*.gov.uk)"
+mobile_number = "^\\+44[\\d]{10}$"
```
minglis: We could do with a strong parsing library for phone numbers - allowing anything and sorting it out. Maybe need to spin up a new story.

app/main/views/register.py
```diff
@@ -0,0 +1,33 @@
...
+                    email_address=form.email_address.data,
+                    mobile_number=form.mobile_number.data,
+                    password=form.password.data,
+                    created_at=datetime.now(),
+                    role_id=1)
+        try:
+            users_dao.insert_user(user)
+            return redirect('/two-factor')
+        except Exception as e:
+            return jsonify(database_error=e.message), 400
```
minglis: We need to watch these as e.message can contain passwords I think - when it dumps the query

app/main/forms.py
servingUpAces: Good point, Let's put that in a new story then

PR #15, opened 2015-12-04
3 Factor authentication

servingUpAces:
The post register endpoint will send a random 5 digit code via sms and another via email.
 If either code fails to send, the user will not be created and the person can register again.
 The codes are saved to the session cookie, and expire in 1 hour.
    
 Another iteration of this story will save the codes to a database.

### DIFF
--- a/config.py
+++ b/config.py
@@ -19,7 +19,7 @@ class Config(object):
     SESSION_COOKIE_SECURE = True
 
     NOTIFY_DATA_API_URL = os.getenv('NOTIFY_API_URL', "http://localhost:6001")
-    NOTIFY_DATA_API_AUTH_TOKEN = os.getenv('NOTIFY_API_TOKEN', "pLuj5kat5auC9Ve")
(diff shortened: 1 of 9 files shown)


config.py
```diff
@@ -11,6 +13,14 @@ class Config(object):
     MAX_FAILED_LOGIN_COUNT = 10
     PASS_SECRET_KEY = 'secret-key-unique-changeme'
 
+    SESSION_COOKIE_NAME = 'notify_admin_session'
+    SESSION_COOKIE_PATH = '/admin'
+    SESSION_COOKIE_HTTPONLY = True
+    SESSION_COOKIE_SECURE = True
+
+    NOTIFY_DATA_API_URL = os.getenv('NOTIFY_API_URL', "http://localhost:6001")
+    NOTIFY_DATA_API_AUTH_TOKEN = os.getenv('NOTIFY_API_TOKEN', "pLuj5kat5auC9Ve")
```
minglis: Do we still want this in here?

PR #18, opened 2015-12-07
Verify flow

servingUpAces:
Implementing the verify flow. When a person registers with a valid email and mobile number, they are sent a code for each. When they post a correct code on the verify page they will be redirected to the /add-service page and the user.state will be set to active.

### DIFF
--- a/app/main/dao/users_dao.py
+++ b/app/main/dao/users_dao.py
@@ -30,3 +30,10 @@ def increment_failed_login_count(id):
     user = User.query.filter_by(id=id).first()
     user.failed_login_count += 1
     db.session.commit()
+
+
+def activate_user(id):
+    user = get_user_by_id(id)
+    user.state = 'active'
+    db.session.add(user)
+    db.session.commit()
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -39,5 +39,7 @@ class RegisterUserForm(Form):
 
 
 class VerifyForm(Form):
-    sms_code = IntegerField(DataRequired(message='SMS code can not be empty'))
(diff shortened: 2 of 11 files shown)


app/main/forms.py
```diff
@@ -36,3 +36,10 @@ class RegisterUserForm(Form):
                              validators=[DataRequired(message='Please enter your password'),
                                          Length(10, 255, message='Password must be at least 10 characters'),
                                          Blacklist(message='That password is blacklisted, too common')])
+
+
+class VerifyForm(Form):
+    sms_code = IntegerField("Text message confirmation code",
```
minglis: Do we have any max length validations? maxlength on form element, and validators?

app/main/views/verify.py
```diff
@@ -0,0 +1,32 @@
...
+    return render_template('verify.html', form=VerifyForm())
+
+
+@main.route('/verify', methods=['POST'])
+def process_verify():
+    form = VerifyForm()
+    if form.validate_on_submit():
+        valid_sms = checkpw(form.sms_code.data, session['sms_code'])
+        valid_email = checkpw(form.email_code.data, session['email_code'])
+        if valid_sms is False:
```
minglis: Does this need to be a single response with both states?

app/main/dao/users_dao.py
```diff
@@ -30,3 +30,10 @@ def increment_failed_login_count(id):
     user = User.query.filter_by(id=id).first()
     user.failed_login_count += 1
     db.session.commit()
+
+
+def activate_user(id):
```
minglis: Guess we still need to have the conversation about the best place for DB exceptions / Error pages and so on. Maybe not this story, but if we capture it in pivotal.

PR #24, opened 2015-12-10
Use correct port to connect to API

quis:
The API runs on port `6011`, as set here:
https://github.com/alphagov/notifications-api/blob/761e7b4444f45948182e9fa6a14e01955188786a/application.py#L11


quis:
That was not the API I was looking for.

PR #27, opened 2015-12-11
Put elements on pages for ‘send text messages’ flow

quis:
https://www.pivotaltracker.com/story/show/109814690

--

## Move all pages into their own directory

There are a lot of pages. It seems like a good idea to keep them in one place, so they don’t get mixed up with partials and layouts.

## Get SASS compilation working

Main thing that was missing was including the main CSS file in the template.

There are a few hacky bits here, like moving the whole of toolkit inside the stylesheets directory.

Would arguably be cleaner using something that isn’t Flask Assets, but that’s something for later.

## Add a template filter for message placeholders …
As a first guess placeholders can be added to messages with the `((placeholder))` syntax.

This commit adds a Jinja template filter to convert strings containing said-formatted strings into HTML, which can then be styled to highlight which parts will be substituted in messages.

## Hard code some data in the Python …
This commit adds some stubbed data for the pages. The structure of the data is just a proposal, but it gives the templates something to work with for now.

## Add GOV.UK Elements as a submodule

Since GOV.UK Elements is versionned now it makes sense to bring it in as a dependency. This enforces a separation between what generic stuff we’re using from Elements and what is specific to our app.

The benefit is that when the generic stuff changes it will be easy to bring those changes in.

This commit also bumps GOV.UK frontend toolkit to the latest version (v4.5.0).

## Add a non-JS template picker

image

We could do something with Javascript to only show the selected template. For now this is something that works without Javascript.

This means we can put off getting the build and testing pipeline for Javascript set up, which is a bigger and more unknown piece of work.

## Add table of recipients

image

Uses an adapted version of the ['summary table' pattern from Digital Marketplace frontend toolkit](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/summary-table.html)

## Use Flask for routing
- replace links that look like buttons with forms and submit buttons
- split the view code for SMS into its own file
- move the routing into the Python by adding handling for `post` requests
- use Flask’s `url_for` to generate URLs, rather than hard coding them (so that it’s easier to change the URLs)
- changes the URLs for sending text messages


quis:
Trying out block labels with the message templates beside them:

image

quis:
## Replace CSV preview with rendered messages

If there are less than 7 messages, show them all.

If there are more than 7, show the first and last three, and a link to the remaining x.

image

quis:
## Improvements to send SMS page
- make messages align horizontally
- change wording of step 2 (from Lorena)

image

quis:
## Simplify _check and confirm_ page

Based on a conversation with Lorena:
- remove the row numbering so it’s easier to scan the list of phone numbers
- add subheadings for 'first three' and 'last three'
- put the 'see all' link at the end

--

image

yahoopete:
Wow,  boss pull request!

PR #31, opened 2015-12-15
Add service

servingUpAces:
https://www.pivotaltracker.com/story/show/110067722

Implementation of the add-service endpoints. When a service is added it is mapped to the user that is logged in.


app/main/views/index.py
```diff
@@ -35,12 +35,6 @@ def dashboard():
...
 
 
-@main.route("/add-service")
-@login_required
-def addservice():
-    return render_template('views/add-service.html')
-
-
 @main.route("/email-not-received")
 def emailnotreceived():
```
minglis: def email_not_received(): 

;-)

servingUpAces: Everything added in the index.py is just the initial template. When it's implemented it is moved to a file like email_not_received. This method is being moved and implemented now.

PR #33, opened 2015-12-17
Use a Node-based tools for handling assets

quis:
## …or how to move a bunch of things from a bunch of different places into `app/static`

There are three main reasons not to use Flask Assets:
- It had some strange behaviour like only recompiling the assets when `main.scss` is changed, not when its `@include`’d files are changed
- It was based on Ruby SASS, which is slower to get new features than libsass, and meant depending on Ruby, _and_ having the SASS Gem globally installed—so you’re already out of being a ‘pure’ Python app
- Martyn and I have experience of doing it this way on Marketplace, and we’ve ironed out the kinks

The specific technologies this introduces, all of which are Node-based:
- Gulp – like a Makefile written in Javascript
- NPM – package management, used for managing Gulp and its related dependencies
- Bower – also package management, and the only way I can think to have GOV.UK template as a proper dependency

…speaking of which, GOV.UK template is now a dependency. This means it can’t be modified at all (eg to add a global `#content` wrapper), so every page now inherits from a template that has this wrapper. But it also means that we have a clean upgrade path when the template is modified.

Everything else (toolkit, elements) I’ve kept as submodules for now but moved them to a more logical place (`app/assets` not `app/assets/stylesheets`, because they contain more than just SASS/CSS).

This means that Flask assets and all the Ruby stuff can be (and have been) removed.

### Heroku

Set the right buildpacks:
```shell
heroku buildpacks:set heroku/python
heroku buildpacks:add --index 1 heroku/nodejs
```

This is tested working, eg http://murmuring-beyond-8173.herokuapp.com

### Amazon 

Probably needs an `npm install` adding somewhere, cc @minglis


minglis:
Can we hold off this for a bit till I sort out deployment.

PR #35, opened 2015-12-17
Implementation for did not receive email or sms code

servingUpAces:
This is the implementation behind the 'I did not receive an email' and 'I did not receive a text message' link on the verify page. 
When used, a new verify code is created and sent to the email/sms. The user email address or mobile number is also updated, to correct any bad data for the person.
All the codes are valid until on if them is used, then all the outstanding codes are marked as used.
There is also a fix to user.is_active() method which was considering a 'pending' user as active.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -95,21 +95,18 @@ class AddServiceForm(Form):
 
 def validate_codes(field, code_type):
     codes = verify_codes_dao.get_codes(user_id=session['user_id'], code_type=code_type)
-    for code in codes:
-        if validate_code(field, code):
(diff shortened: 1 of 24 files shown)


app/main/forms.py
```diff
@@ -93,3 +91,25 @@ def validate_service_name(self, a):
...
+
+
+def validate_code(field, code):
+    if field.data is not None:
+        if check_hash(field.data, code.code) is False:
+            return False
+        else:
+            return True
+    else:
+        return False
```
minglis: if field.data and check_hash(field.data, code.code):
    return True
return False

??

PR #36, opened 2015-12-17
Service activity flow

quis:
_Branched off #34, which should be merged first. Changes introduced in this pull request: https://github.com/alphagov/notifications-admin/compare/quarter-width-nav...service-activity-flow_

Story [109814822](https://www.pivotaltracker.com/story/show/109814822)

--

image

image

## Add messages to the current job’s history

This mocks out a data structure for a job’s messages, and renders this data:
- on the notification page, as a table, which links through to…
- …the page for an individual message

## Add some new components
- ‘Big number’ to show the totals
- ‘Banner’ to show the status of your job
- The ‘history’ sidebar to the SMS message component

## Add unit tests for activity (and SMS) flows

These are basic tests to make sure that the pages stay stitched together.


quis:
Make data consistent from dashboard → job
- adds a table of recent activity to the job page, which is shared by the dashboard page
- uses the same filename and message template as the first job shown on the dashboard
- sets the time of file uploads/message delivery/message sending to always match ‘just now’

image

quis:
## Add big numbers to dashboard

A suggestion of what high-level information users might need on the dashboard page.

image

## 	Remove log of activity from single notification

Having the full history of the message is more information than is necessary. We should only show what stage the message is at, and the time that it reached that stage.

We can do research later on to find out if users understand or care about the different stages.

Before | After
--- | ---
image | image

PR #39, opened 2016-01-04
Refactor add-service form such that the dao is not exposed in the forms.

servingUpAces:


app/templates/govuk_template.html
```diff
@@ -1,3 +1,4 @@
+{%- from "macros.html" import render_field %}
```
quis: We shouldn’t be adding stuff to GOV.UK template directly because it will break our ability to easily upgrade it in the future.

app/templates/macros.html
```diff
@@ -0,0 +1,12 @@
+{% macro render_field(field) %}
```
quis: Can this file be named something like `textbox.html` rather than `macros` generically? I’ve started added a bunch of macros in later pull requests and it’s going to get vast if they’re all in one file.

mannickutd:
Looks good.

PR #40, opened 2016-01-05
Form views refactor

mannickutd:
Refactor for views, forms and templates trying to use more of the flask functionality.


app/main/forms.py
```diff
@@ -42,6 +48,16 @@ class RegisterUserForm(Form):
                                          Length(10, 255, message='Password must be at least 10 characters'),
                                          Blacklist(message='That password is blacklisted, too common')])
 
+    def validate_email_address(self, field):
+        # Validate email address is unique.
+        if field.data in self.existing_emails:
+            raise ValidationError('Email address already exists')
+
+    def validate_mobile_number(self, field):
```
servingUpAces: We do not have a requirement to keep the mobile number unique.

mannickutd: I have commented out the validation check, to be added in later when we need this check.

mannickutd:
Should be good now working with the latest build :).

app/main/forms.py
```diff
@@ -42,6 +48,18 @@ class RegisterUserForm(Form):
...
+        # Validate mobile number is unique
+        # Code to re-added later
+        # if field.data in self.existing_mobiles:
+        #    raise ValidationError('Mobile number already exists')
+        pass
+
 
 class TwoFactorForm(Form):
     sms_code = StringField('sms code', validators=[DataRequired(message='Please enter your code'),
```
servingUpAces: Let's use 'Enter verification code' for the label

app/main/views/register.py
```diff
@@ -28,16 +33,16 @@ def process_register():
...
-            return jsonify(database_error='encountered database error'), 400
-    else:
-        return jsonify(form.errors), 400
-    return redirect('/verify')
+        users_dao.insert_user(user)
+        # TODO possibly there should be some exception handling
+        # for sending sms and email codes.
+        # How do we report to the user there is a problem with
+        # sending codes apart from service unavailable?
+        # at the moment i believe http 500 is fine.
```
servingUpAces: Yes we do need more exception handling, we also need to handle exceptions from the db better. There is a story for that.

app/main/forms.py
mannickutd: Fixed.

PR #44, opened 2016-01-06
Provide logout link

mannickutd:


app/templates/admin_template.html
```diff
@@ -29,6 +29,11 @@
   <div class="phase-banner-beta">
     <strong class="phase-tag">BETA</strong>
   </div>
+  {% if current_user.is_authenticated %}
+    <div class="">
```
quis: This top navigation should be passed to the `proposition_header` block in the GOV.UK template. There’s [some documentation here](https://github.com/alphagov/govuk_template/#propositional-title-and-navigation).

mannickutd:
Please have a look, there was a security hole discovered after implementation of the sign-out link was created. The first part of the authentication now sets the session['user_email'] rather than the session['user_id'].

mannickutd:
You can now use the TestClient for logging in and logging out as well :).

tests/app/main/views/test_sign_out.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_sign_out_user(notifications_admin,
+                       notifications_admin_db,
+                       notify_db_session,
+                       mocker):
+    with notifications_admin.test_request_context():
+        _set_up_mocker(mocker)
+        email = 'valid@example.gov.uk'
+        password = 'val1dPassw0rd!'
+        user = User(email_address=email,
```
servingUpAces: If you are interested there is a create_test_user function in test/__init__.py

PR #46, opened 2016-01-07
Add some useful owasp suggested headers

ashimali:
Added owasp suggested useful headers to all responses. Did not add content-security policy yet as there may be some external resources used by this app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,7 +67,3 @@ node_modules
 bower_components
 app/templates/govuk_template.html
 npm-debug.log
-
-# private environment variables
-environment-private.sh
-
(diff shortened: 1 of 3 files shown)


mannickutd:
Looks good

.gitignore
```diff
@@ -67,3 +67,7 @@ node_modules
 bower_components
 app/templates/govuk_template.html
 npm-debug.log
+
+# private environment variables
+environment-private.sh
```
quis: It’s better not to pollute `.gitignore` with things that are local or editor-specific. You can [ignore these things globally](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) instead.

PR #49, opened 2016-01-07
Error pages

ashimali:
Basic error pages for 401, 404, 500 errors

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -118,11 +118,12 @@ def useful_headers_after_request(response):
     response.headers.add('X-XSS-Protection', '1; mode=block')
     return response
 
-def register_errorhandlers(app):
+
+def register_errorhandlers(application):
     def render_error(error):
         # If a HTTPException, pull the `code` attribute; default to 500
         error_code = getattr(error, 'code', 500)
         return render_template("error/{0}.html".format(error_code)), error_code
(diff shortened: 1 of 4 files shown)


app/__init__.py
```diff
@@ -116,3 +117,13 @@ def useful_headers_after_request(response):
...
+
+
+def register_errorhandlers(application):
+    def render_error(error):
+        # If a HTTPException, pull the `code` attribute; default to 500
+        error_code = getattr(error, 'code', 500)
+        return render_template("error/{0}.html".format(error_code)), error_code
+    for errcode in [401, 404, 500]:
+        application.errorhandler(errcode)(render_error)
+    return None
```
mannickutd: None is redundant

ashimali: Good point

app/templates/error/404.html
```diff
@@ -0,0 +1,10 @@
+{% extends "admin_template.html" %}
+{% block page_title %}Page Not Found{% endblock %}
```
quis: > Page **n**ot **f**ound

quis:
Other than one remarkably nitpicky comment, this looks good to me.

ashimali:
@quis no not nitpicky, I think you are right on that one :)

PR #50, opened 2016-01-07
Forgot password and Reset password

servingUpAces:
This implement the forgot-password and reset-password endpoints.

When the user requests a reset password link, using '/forgot-password', the user.state is set to request_password_reset. Which means the user will only be able to reset their password and not sign-in.
Once the user resets the password successfully, by using the link in the email, the user state is set to active once more. The user is redirected to the 'two-factor' page and is send a sms verification code to sign in with.

If the link is used a second time they will be redirected to the index page with a message
that the link in the email is not longer valid.


app/main/views/__init__.py
```diff
@@ -31,3 +35,33 @@ def send_email_code(user_id, email):
...
+def send_change_password_email(email):
+    try:
+        link_to_change_password = url_for('.new_password', token=generate_token(email), _external=True)
+        admin_api_client.send_email(email_address=email,
+                                    from_str='notify@digital.cabinet-office.gov.uk',
+                                    message=link_to_change_password,
+                                    subject='Reset password for GOV.UK Notify',
+                                    token=admin_api_client.auth_token)
+    except:
+        traceback.print_exc()
```
mannickutd: Not production code, don't have traceback in production.
Capturing all exceptions is not good design, be specific of which exception will be thrown.

app/main/views/__init__.py
```diff
@@ -31,3 +35,33 @@ def send_email_code(user_id, email):
...
+
+def check_token(token):
+    from itsdangerous import TimestampSigner, SignatureExpired
+    signer = TimestampSigner(current_app.config['SECRET_KEY'])
+    try:
+        email = signer.unsign(token, max_age=current_app.config['TOKEN_MAX_AGE_SECONDS'])
+        return email
+    except SignatureExpired as e:
+        current_app.logger.info('token expired %s' % e)
+        return None
```
mannickutd: By default python returns 'None' so the statement 'return None' is redundant.

app/main/views/forgot_password.py
```diff
@@ -0,0 +1,19 @@
+from flask import render_template, flash
+from app.main import main
+from app.main.dao import users_dao
+from app.main.forms import ForgotPasswordForm
+from app.main.views import send_change_password_email
+
+
+@main.route('/forgot-password', methods=['GET', 'POST'])
+def forgot_password():
+    form = ForgotPasswordForm()
```
mannickutd: I would prefer validation be done in the form, so pass in the email addresses that are available rather than do another separate check after you have already validated that the form is valid. This will look different compared to the other forms.

app/templates/views/forgot-password.html
```diff
@@ -12,15 +12,13 @@ <h1 class="heading-xlarge">Create a new password</h1>
...
+	  {{ render_field(form.email_address, class='form-control-2-3') }}
       <p>
-      <label class="form-label" for="email">Email address</label>
-      <input class="form-control" id="email" type="text" value="">
-      </p>
-
-
-      <p>
-      <a class="button" href="sign-in" role="button">Send email</a>
+      <button class="button" href="sign-in" role="button">Send email</button>
```
mannickutd: Remove the href here, the form should be posting to the same endpoint as the get.

tests/app/main/views/test_new_password.py
```diff
@@ -0,0 +1,63 @@
...
+                                                                         mocker):
+    _set_up_mocker(mocker)
+    with notifications_admin.test_request_context():
+        with notifications_admin.test_client() as client:
+            user = create_test_user('active')
+            token = generate_token(user.email_address)
+        response = client.post('/new-password/{}'.format(token),
+                               data={'new_password': 'a-new_password'})
+        assert response.status_code == 302
+        assert response.location == 'http://localhost/two-factor'
```
mannickutd: Use url_for please. Use the _external parameter for hostname adding to the url.

tests/app/main/views/test_new_password.py
```diff
@@ -0,0 +1,63 @@
...
+                                                                                     notify_db_session):
+    with notifications_admin.test_request_context():
+        with notifications_admin.test_client() as client:
+            notifications_admin.config['TOKEN_MAX_AGE_SECONDS'] = -1000
+            user = create_test_user('active')
+            token = generate_token(user.email_address)
+        response = client.post('/new-password/{}'.format(token),
+                               data={'new_password': 'a-new_password'})
+        assert response.status_code == 302
+        assert response.location == 'http://localhost/forgot-password'
```
mannickutd: Use url_for please.

tests/app/main/views/test_new_password.py
```diff
@@ -0,0 +1,63 @@
...
+        notifications_admin.config['TOKEN_MAX_AGE_SECONDS'] = 86400
+
+
+def test_should_return_raise_no_data_found_exception_when_email_address_does_not_exist(notifications_admin,
+                                                                                       notifications_admin_db,
+                                                                                       notify_db_session):
+    with notifications_admin.test_request_context():
+        with notifications_admin.test_client() as client:
+            token = generate_token('doesnotexist@it.gov.uk')
+            try:
```
mannickutd: Why are you using a try, catch here? All url request should always have a http response returned.

tests/app/main/views/test_forgot_password.py
```diff
@@ -0,0 +1,25 @@
...
+           in response.get_data(as_text=True)
+
+
+def test_should_redirect_to_password_reset_sent(notifications_admin,
+                                                notifications_admin_db,
+                                                mocker,
+                                                notify_db_session):
+    _set_up_mocker(mocker)
+    create_test_user('active')
+    response = notifications_admin.test_client().post('/forgot-password',
```
mannickutd: Please use url_for.

tests/app/main/views/test_forgot_password.py
```diff
@@ -0,0 +1,25 @@
+from tests.app.main import create_test_user
+
+
+def test_should_render_forgot_password(notifications_admin, notifications_admin_db, notify_db_session):
+    response = notifications_admin.test_client().get('/forgot-password')
```
mannickutd: Please use url_for

app/main/views/new_password.py
```diff
@@ -0,0 +1,29 @@
...
+        if email_address:
+            user = users_dao.update_password(email_address.decode('utf-8'), form.new_password.data)
+            send_sms_code(user.id, user.mobile_number)
+            return redirect(url_for('main.two_factor'))
+        else:
+            flash('expired token request again')
+            current_app.logger.info('we got here')
+            return redirect(url_for('.forgot_password'))
+    else:
+        return render_template('views/new-password.html', toke=token, form=form)
```
mannickutd: 'toke=token'??

app/main/views/new_password.py
```diff
@@ -0,0 +1,29 @@
...
+    form = NewPasswordForm()
+    if form.validate_on_submit():
+        email_address = check_token(token)
+        if email_address:
+            user = users_dao.update_password(email_address.decode('utf-8'), form.new_password.data)
+            send_sms_code(user.id, user.mobile_number)
+            return redirect(url_for('main.two_factor'))
+        else:
+            flash('expired token request again')
+            current_app.logger.info('we got here')
```
mannickutd: Remove log message.

app/main/views/__init__.py
servingUpAces: I have removed the exception. However the application still needs to handle error and render error pages.

app/main/views/forgot_password.py
servingUpAces: This validation was not done in the form intentionally. Firstly to add the flash message for the use. Secondly the total number of users in the system will be passed into the form (is this ok?). Thirdly is this really a form validation error? The email is not recognised by the system but the user entered a valid email address. How the message appears is a question for GDS style guide.

PR #51, opened 2016-01-08
Add pages for service settings flow

quis:
https://www.pivotaltracker.com/story/show/111149916

# Add pages for ‘service settings’ flow

Adds the pages and wires them together, so that it’s possible to click through them.

The wording is not quite English, but attempts to be an rough description of what the consequences are for each of the four actions.

image

image

# Add confirm loop

For actions where:
- we want you to be sure that you want to do what you’re about to do
- we want to be sure it’s you trying to do the thing

This adds a second step that asks you to confirm your password.

image

# Rename `submit_form` component to `page_footer`

Submit form was
- a confusing name in itself
- not descriptive, because it also has an optional ‘back’ link

This commit also puts this component in as many pages as possible, stripping out any hard-coded buttons and links.

It replaces any usage of buttons for ‘back’, because these are links, not buttons (ie they don’t change any data).


quis:
image

PR #55, opened 2016-01-11
First slice of csv upload of phone numbers for sending messages.

ashimali:
At the moment the file contents are not persisted but checked in memory.
The first and last record are shown if all are valid.

If there are invalid rows, they are reported and the user is prompted
to go back and sort out the upload file.

The storing of upload result (i.e. validation of file) in session will
be removed in next story which is about persisting of file for offline
processing.

### DIFF
--- a/app/templates/views/check-sms.html
+++ b/app/templates/views/check-sms.html
@@ -21,6 +21,7 @@
 		<p><a href="{{url_for('.sendsms')}}" class="button">Go back and resolve errors</a></p>
 
 	{% else %}
+
 		<h2 class="heading-medium">Check and confirm</h2>
 		<form method="POST" enctype="multipart/form-data">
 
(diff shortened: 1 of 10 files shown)


app/templates/views/check-sms.html
```diff
@@ -10,60 +10,35 @@
...
+		{% endfor %}
+		<p><a href="{{url_for('.sendsms')}}" class="button">Go back and resolve errors</a></p>
+	{% else %}
+		<h2 class="heading-medium">Check and confirm</h2>
+		<form method="POST" enctype="multipart/form-data">
+			{{ submit_form(
+			    "Send {} text messages".format(upload_result.valid|count),
+			    url_for(".sendsms")
+			)}}
+	    <h3 class="heading-small">First message in file</h3>
```
quis: What’s the reason for changing from first three/last three messages to first/last?

ashimali: That was in the story in pivotal, so just following orders:) Happy to make it first and last three, or even configurable if you like.

quis: I think stick to first and last three, that’s what I discussed with @henryhadlow when I worked on these pages.

PR #70, opened 2016-01-13
Mobile number validation

quis:
_https://www.pivotaltracker.com/story/show/109262348_

_Reviewing: I’d be especially interested to know if there are any [test cases](https://github.com/alphagov/notifications-admin/blob/mobile-number-validation/tests/app/main/test_phone_number_form_field.py#L13-L54) I’ve missed._

***

## Add form field for a UK mobile phone number

This field does two things:
- validates the format of the phone number
- outputs a consistent representation of the phone number

Because of this I think it’s better represented as a new field type, rather than individual validators.

I also think that it’s better to do this without regular expression(s), because it makes returning the specific error easier.

This commit also adds basic pass/fail test for a series of valid/invalid phone numbers.

## Test for specific error messages

This commit:
- improves the tests to check for specific error messages, rather than just pass/fail
- makes the error messages more human, and more indicative of what the user needs to do to fix the error

## Add the new field to the application

This commit replaces the previous `StringField` used for collecting mobile phone numbers with the `UKMobileNumber` field.

This means changing a few of the preexisting tests to have more realistic mobile numbers so that they still pass.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,3 +1,4 @@
+import re
 from flask_wtf import Form
 
 from wtforms import (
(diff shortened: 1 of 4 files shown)


app/main/forms.py
```diff
@@ -24,11 +25,48 @@ def email_address():
         Regexp(regex=gov_uk_email, message='Enter a gov.uk email address')])
 
 
+class UKMobileNumber(StringField):
+
```
mannickutd: Why not use a regex validation would be faster and more intuitive to the developer. A google search resulted in http://stackoverflow.com/questions/16405187/regular-expression-for-uk-mobile-number-python
Is there a reason not to use this?

quis: 1. Using a regex would make it harder to tell the user what the specific error is.

2. The regex you linked to won’t match all the different variations mentioned in [the story](https://www.pivotaltracker.com/story/show/109262348). In order to match all of those variations it would get really long, and I don’t think long regexes are particularly easy to understand.

3. I think we also need a custom field in order to format the number after validating it (see https://github.com/alphagov/notifications-admin/pull/70/files#diff-e618e3e19b02f84ab65f8d0cef43212bR64).

PR #73, opened 2016-01-13
Add page to choose services

quis:
Because a user can have multiple services, they need a way to navigate between them. Normally they can use the ▶ Switcher to do this, except when:

- they first sign in
- they are on a page which isn’t associated with a service (eg user profile) in which case we can’t use the switcher because it won’t know what the ‘current’ service is

So this commit adds a new page with a (fake) list of services.

***

image

### DIFF
--- a/app/main/views/choose_service.py
+++ b/app/main/views/choose_service.py
@@ -5,5 +5,5 @@ from app.main import main
 
 @main.route("/services")
 @login_required
-def chooseservice():
+def choose_service():
     return render_template('views/choose-service.html')
--- a/app/main/views/choose_service.py
+++ b/app/main/views/choose_service.py
@@ -1,0 +1,9 @@
+from flask import render_template
+from flask_login import login_required
+from app.main import main
+
+
+@main.route("/services")
+@login_required
+def chooseservice():
+    return render_template('views/choose-service.html')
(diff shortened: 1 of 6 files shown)


app/main/views/choose_service.py
```diff
@@ -0,0 +1,9 @@
+from flask import render_template
+from flask_login import login_required
+from app.main import main
+
+
+@main.route("/services")
+@login_required
+def chooseservice():
```
mannickutd: Just a side note, good to separate function names by words with an '_'

tests/app/main/views/test_two_factor.py
```diff
@@ -28,7 +28,7 @@ def test_should_login_user_and_redirect_to_dashboard(notifications_admin, notifi
                                    data={'sms_code': '12345'})
 
             assert response.status_code == 302
-            assert response.location == url_for('main.dashboard', service_id=123, _external=True)
+            assert response.location == 'http://localhost/services'
```
mannickutd: Please use url_for function to create urls.

app/main/views/choose_service.py
quis: @mannickutd See f15a59b

tests/app/main/views/test_two_factor.py
quis: In tests I like to confirm that I’m getting the expected URL. Overkill you think?

mannickutd: Not overkill at all, it is about consistency and updates, if that endpoint changes in the future we don't have to update hundreds of tests for a specific string. The actual url should be ambiguous we are testing a view endpoint.

quis: OK, happy to go with that, changed in 0a46a10

PR #76, opened 2016-01-13
Upload csv s3

ashimali:
This slice of csv upload, persists file to /tmp and if file contents are valid user can submit, which puts file to s3 bucket. 

Tasks left are clearup of files in /tmp and creation of job via api.

### DIFF
--- a/app/main/views/sms.py
+++ b/app/main/views/sms.py
@@ -1,5 +1,8 @@
 import csv
 import re
+import os
+
+from datetime import date
 
 from flask import (
     request,
@@ -12,9 +15,11 @@ from flask import (
 )
 
 from flask_login import login_required
+from werkzeug import secure_filename
 
 from app.main import main
 from app.main.forms import CsvUploadForm
+from app.main.uploader import s3upload
 
 # TODO move this to the templates directory
 message_templates = [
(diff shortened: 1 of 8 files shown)


app/main/views/sms.py
```diff
@@ -69,33 +74,66 @@ def sendsms(service_id):
 @login_required
 def checksms(service_id):
     if request.method == 'GET':
-
-        recipients_filename = request.args.get('recipients')
-        # upload results in session until file is persisted in next story
-        upload_result = session.get(recipients_filename)
+        filename = request.args.get('recipients')
```
mannickutd: Might be good to handle the situation where 'recipients' isn't available, and throw a 400? At the moment this will throw a 500.

ashimali: Yes good point

tests/app/main/views/test_sms.py
```diff
@@ -1,16 +1,22 @@
...
+    _setup_mocker_for_empty_file(mocker)
     with notifications_admin.test_request_context():
         with notifications_admin.test_client() as client:
             user = create_test_user('active')
             client.login(user)
-
             upload_data = {'file': (BytesIO(''.encode('utf-8')), 'emtpy.csv')}
-            response = client.post('/services/123/sms/send', data=upload_data,
-                                   follow_redirects=True)
+            response = client.post('/services/123/sms/send',
```
mannickutd: Please use url_for

PR #77, opened 2016-01-14
Manage templates page

quis:
image

***

# Add manage templates link to main navigation

Because it’s important

# Add a prototype email template

image

If the templates page contains text messages and emails then there’s two ways it could be structured:
- into two sections, all text messages first, then all emails
- emails and text messages interleaved, sorted by date

I think the second one is better. Imagine a situation where you mostly do emails but have a few text messages. You’d have to scroll past the text messages to get to your emails. Every time.

I reckon that the most commonly accessed templates will be the most recent ones.

# Simplify template picker to be just radio buttons

Includes making the selection buttons work properly

image

# Add link to delete a template

image

image

This is a link not a button because:
- it’s less prominent—delete is an infrequent action
- it’s a two-step process, and only the second part changes any data (so it has a button)

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,7 +1,7 @@
 import os
 import re
 
-from flask import Flask, session, Markup, render_template
+from flask import Flask, session, Markup, escape, render_template
 from flask._compat import string_types
 from flask.ext.sqlalchemy import SQLAlchemy
 from flask_login import LoginManager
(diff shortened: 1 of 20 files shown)


mannickutd:
Please use url_for, otherwise it all looks good.

quis:
@mannickutd Done: 9f3621b

PR #107, opened 2016-01-22
Fix keyboard mashing

quis:


quis:
Seems to be fixed already.

PR #110, opened 2016-01-25
Show correct default state on service settings page

quis:
By default a service should be active (ie keys not suspended). For some reason the API is returning the opposite.

This commit reverses the logic to make it look right for hack day.


quis:
Closing this because it was only for hack day, but we’re not doing any more deploys today.

PR #117, opened 2016-01-27
Password validation for service changes

mannickutd:
Sorry in advance to because this should of been multiple pull requests. Snowball......


app/main/views/two_factor.py
```diff
@@ -5,19 +5,27 @@
...
+    form = TwoFactorForm(_check_code)
 
     if form.validate_on_submit():
-        verify_codes_dao.use_code_for_user_and_type(user_id=user.id, code_type='sms')
+        user = users_dao.get_user_by_id(user_id)
+        # Check if coming from new password page
+        if 'password' in session['user_details']:
+            user.set_password(session['user_details']['password'])
+            users_dao.update_user(user)
+        del session['user_details']
```
ashimali: Maybe a try finally around here to make sure the details get delete from session?

app/main/views/sign_in.py
```diff
@@ -6,37 +6,29 @@
...
-    except:
-        import traceback
-        traceback.print_exc()
-        abort(500)
+    form = LoginForm()
+    if form.validate_on_submit():
+        user = _get_and_verify_user(form.email_address.data, form.password.data)
+        if user:
+            users_dao.send_verify_code(user.id, 'sms')
+            session['user_details'] = {"email": user.email_address, "id": user.id}
```
ashimali: Do we still need email address in the session?

app/main/views/verify.py
```diff
@@ -20,13 +20,11 @@ def verify():
     # TODO there needs to be a way to regenerate a session id
     # or handle gracefully.
     user_id = session['user_details']['id']
```
ashimali: Could probably clear this from session once form submitted and user logged in?

app/main/views/sign_in.py
mannickutd: The 'user_details'  structure is used in other places, consistency was the reason for having both email and id.

ashimali: I think email was used at some point, but not sure it is anymore. I may be mistaken. And user_details was to no clash with flask login which uses user_id in session.

ashimali:
Everything else looks fine to me.

PR #124, opened 2016-01-29
Uploading csv creates job via api

ashimali:
This slice creates job. Next and last slice is showing jobs after submit.

### DIFF
--- a/app/main/views/sms.py
+++ b/app/main/views/sms.py
@@ -1,6 +1,5 @@
 import csv
 import re
-import os
 
 from datetime import date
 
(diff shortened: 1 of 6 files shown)


app/main/views/sms.py
```diff
@@ -67,18 +67,23 @@ def check_sms(service_id, upload_id):
...
                 ((name)), we’ve received your ((thing)). We’ll contact you again within 1 week.
             ''',
             service_id=service_id
         )
     elif request.method == 'POST':
-        # TODO create the job with template, file location etc.
+        file_name = request.form['original_file_name']
+
+        # TODO need a real template id picked from form
+        template_id = 1
```
mannickutd: Is this going to be fixed?

PR #146, opened 2016-02-03
Updated main nav to group things more sensibly

yahoopete:
Adding more verbs than nouns too.... so it's clear what you do at each view.

No ticket for this, as off-piste.

Naughty.

Will look like:

screenshot 2016-02-03 10 50 48


quis:
Clearer I think. Still might have the problem of text message templates vs email templates but let’s deal with that when we come to it.

PR #147, opened 2016-02-03
Split ‘send SMS’ page into two pages

quis:
This pull request just splits the existing page into two. It doesn’t do any substantive changes to how the two parts of the page work.

It also changes the wording on the blue banner—we can ignore any mention of trial/test/mode if we just talk about ‘not live’. It also feels sensible to link through to the page where you can make that
change, rather than dig through the navigation.

image

image


quis:
## Don’t allow user to send texts without template

image

PR #150, opened 2016-02-04
Send text flow revised

quis:
## Replace message previews on check page with table

The first 3/last 3 messages didn’t test well, it wasn’t immediately obvious what was going on.

This commit replaces it with just a preview of the first message, and a table showing the details of the subsequent messages.

image

## Updates to display of jobs

This commit:
- adds the template to the jobs page (and puts it at the top of the send SMS page) so that it consistently appears in the same place throughout the journey
- put the real data about a job on the jobs page and on the dashboard

image

image

## Updates to send and check SMS pages …

Adds a back button instead of a back link (more prominent, you’re likely to go back on these pages if you’ve made a mistake)

Tweaks to wording

image

## Fix spacing of first page headers

This commit customises the margins of the first header on each page so that it lines up with the navigation.

image

## Add two steps to onboarding banner

It’s not obvious what the first steps are when you’ve just signed up.

This commit changes the banner on the dashboard to make it obvious.

image


quis:
## Updates to ‘send SMS’ page

Based on discussion with Pete.

Make the blue banner an ‘important’ banner (copied from Register to Vote, used because it’s not as boxy and fits on the page better).

Remove the back button because you haven’t changed any data yet. If you need to go back you can just press back or start again.

Make the filename stand out more.

Remove the ‘download example’ link. Will need to revisit the best way of doing this.

Make text messages consistently 2/3rd width.

image

## Fake phone number on job page

To show what kind of data we want to surface here.

## Make first page of ‘send texts’ use links not form

This doesn’t need to be a form—it’s not changing any data.

And having the primary action on the page as ‘Use this template’ it makes it clear what the page is for.

image

PR #156, opened 2016-02-05
Skip ‘choose service’ page if user has one service

quis:
We used to do this by redirecting on the choose service page. However when we lost the dropdown and this page also became the page for adding a new service (in 3617f2e936aadbdf71d582e58d88a16629ee5ca2) the redirect was removed.

This commit re-adds the redirect on the two factor page, so that it only happens on first login.

So the flows are:

**Multiple services**
`Sign in` → `Enter two factor code` → `Choose service` → `Service dashboard`

**One service**
`Sign in` → `Enter two factor code` → `Service dashboard`

**No services (you’ve deleted all your services)**
`Sign in` → `Enter two factor code` → `Choose service` → `Add new service`

### DIFF
--- a/app/main/views/two_factor.py
+++ b/app/main/views/two_factor.py
@@ -5,7 +5,7 @@ from flask import (
 from flask_login import login_user
 
 from app.main import main
-from app.main.dao import users_dao
+from app.main.dao import users_dao, services_dao
 from app.main.forms import TwoFactorForm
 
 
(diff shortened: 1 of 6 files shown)


quis:
## Add mocked `get_services` to `test_client.login`

Because the redirect after logging in checks the number of services a user has, this now needs to be mocked.

Right now this means adding `mock_get_login` to any tests that need a login. This must be one of the first mocks, so that it can be overridden by any use of `mock_get_services`, for tests that specifically want to rely on a quantity of mocked services, or their contents.

This is a bit fragile, but there’s already a TODO in the code to make it better so ¯\_(ツ)_/¯

PR #165, opened 2016-02-09
Updated wsgi so that it can read the environment from the files prese…

minglis:
…nt on AWS instances


quis:
LGTM. Holding off on merging until the file is on the boxes.

PR #171, opened 2016-02-10
Make URLs for assets cache-proof

quis:
https://www.pivotaltracker.com/story/show/113448149

***

This commit adds a query string to assets URLs which is generated from a hash of the file contents. When asset files are changed they will now be served from a different URL, which means they won’t be loaded from browser cache.

This is similar to how GOV.UK template adds its version number as a querystring parameter for its assets.

This is mostly copied [from Digital Marketplace utils](https://github.com/alphagov/digitalmarketplace-utils/pull/102). They have it in a shared codebase, we only have one frontend app so don’t need to do that.

Usage in a template:
``` jinja
{{ asset_fingerprinter.get_url('stylesheets/application.css') }}
```

Output:
```
static/stylesheets/application.css?418e6f4a6cdf1142e45c072ed3e1c90a
```

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -18,6 +18,7 @@ from app.notify_client.user_api_client import UserApiClient
 from app.notify_client.job_api_client import JobApiClient
 from app.notify_client.status_api_client import StatusApiClient
 from app.its_dangerous_session import ItsdangerousSessionInterface
+from app.asset_fingerprinter import AssetFingerprinter
 import app.proxy_fix
 from config import configs
 from utils import logging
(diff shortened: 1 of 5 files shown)


quis:
## Cache static files for a year

image

We only want static files to not come from the browser cache when they have changed. The best way to do this is by cache busting the URLs.

Otherwise, we want static files to be cached for a long time. This commit sets the `Expires` HTTP header to 1 year in the future.

Previously it was set to 12 hours, the default.

From the Flask docs:
> Default cache control max age to use with send_static_file() (the default
> static file handler) and send_file(), in seconds. Override this value on a
> per-file basis using the get_send_file_max_age() hook on Flask or Blueprint,
> respectively. Defaults to 43200 (12 hours).
– http://flask.pocoo.org/docs/0.10/config/

PR #180, opened 2016-02-17
Use Template util to replace, highlight and validate CSV files

quis:
_Depends on alphagov/notifications-utils#1_

***

## Use `Template` to replace/highlight placeholders

This commit brings in the `Template` util.

It also does a fair bit of tiying up, which I’ve unfortunately squashed into one massive commit. The main change is moving 404 handling into the templates dao, so that every view isn’t littered with `try: … except(HTTPError)`.

It also adds new features, in a prototype sort of way, which are:
- download a prefilled example CSV
- show all the columns for your template on the 'check' page

image

## Check CSV files match the template

This commit adds a first stab at checking whether a CSV file has the right data to fill the placeholders.

The UI is very much first bash, but I’d like to get this merged and see how it feels. The main thing is that we’ve got all the bit in place now to do this logic.

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -19,6 +19,7 @@ from app.notify_client.job_api_client import JobApiClient
 from app.notify_client.status_api_client import StatusApiClient
 from app.its_dangerous_session import ItsdangerousSessionInterface
 from app.asset_fingerprinter import AssetFingerprinter
+from app.utils import validate_phone_number, InvalidPhoneError
 import app.proxy_fix
 from config import configs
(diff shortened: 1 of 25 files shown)


quis:
## Validate CSVs fully

This commit extends the existing function to validate each row’s phone number to also validate that all the required data is present.

It does this using the checking that the `Template` class can do when given a template and a `dict` of values.

screen shot 2016-02-17 at 15 51 36

quis:
When you upload a valid CSV:

image

PR #187, opened 2016-02-22
Add pages to invite, edit, and delete users 

quis:
This takes the original prototype version of this page, and, using the same
fake data (ie nothing is wired up):
- adds an invite users page
- adds an edit (and delete) user page

Both these pages allow the user to set another user’s permissions.

This commit adds images for the ticks and crosses, so we have control over their appearance.

image

image

image

***

The yes/no pattern owes some debt to this page from GOV.UK Verify:

image

***

## Also
**Inline images in CSS**
Because this commit’s parent added a few new images, we are now serving at least a handful of images, therefore a few additional HTTP requests. It’s better to combine multiple HTTP requests into one for performance reasons (up to a point).

This commit adds an extra step to the preprocessing of SASS files which takes any images it finds, base64 encodes them and inlines them into the distributed CSS files.

It also modifies the content security policy to allow inline images.

**Compress GOV.UK Template assets**
While looking at front end performance I noticed that the GOV.UK Template CSS and Javascript files weren’t being minified, so were being served with (for example) lots of comments still in the code.

This changes the Gulp task from simply copying the files to also minfiying them, where appropriate.

**Updates to styleguide**
Brings in changes to the email message and table patterns.


app/main/views/sms.py
```diff
@@ -127,6 +127,7 @@ def check_sms(service_id, upload_id):
         template_id = upload_data.get('template_id')
         raw_template = templates_dao.get_service_template_or_404(service_id, template_id)['data']
         upload_result = _get_rows(contents, raw_template)
+        print(upload_result)
```
mannickutd: please remove this print statement

tests/app/main/views/test_add_service.py
```diff
@@ -31,6 +32,46 @@ def test_should_add_service_and_redirect_to_next_page(app_,
...
+    mock_get_services,
+    api_user_active,
+    mock_get_user,
+    mock_get_user_by_email
+):
+    with app_.test_request_context():
+        with app_.test_client() as client:
+            client.login(api_user_active)
+            with client.session_transaction() as session:
+                print('session: {}'.format(session))
```
mannickutd: Please remove the print statement

quis:
## Fix users page for research

For the lab sessions, users should just see themselves as the only user, not all the fake users (it might confuse them).

## Remove the ‘manage templates’ page

The ‘manage templates’ page was almost identical to the ‘send text messages’ page.

This commit consolidates them into one and makes them all hang together.

Part of this means tweaks to the javascript so that files upload as soon as you’ve chosen them.

PR #200, opened 2016-02-25
Consolidate page templates for emails + texts

quis:
#### _Depends on: https://github.com/alphagov/notifications-utils/pull/4_

***

Use one page template for each of:
- choosing a message template
- adding recipients
- checking an email message
- looking at a job

This commit consolidates the two templates into one, and adds logic to show the SMS message pattern or the email message pattern depending on the type of template.

It also gives email templates a bit more width, because the body and the from address tend to be quite long.

## Also: Fix example CSV for email templates

It should populate the CSV file with the user’s email address, not their phone number.


app/main/views/send.py
```diff
@@ -143,15 +157,19 @@ def check_messages(service_id, upload_id):
             values=upload_result['rows'][0] if upload_result['valid'] else {},
             drop_values={'to'}
         )
+        print(template.additional_data)
```
ashimali: errant print

quis: Removed and squashed into f5d19ec.

quis:
## Prefix text messages with service name

Brings in: alphagov/notifications-utils#4

PR #211, opened 2016-02-29
Connect api keys permission

mannickutd:


app/__init__.py
```diff
@@ -35,6 +36,7 @@
 status_api_client = StatusApiClient()
 invite_api_client = InviteApiClient()
 asset_fingerprinter = AssetFingerprinter()
```
minglis: Why so many clients? Is it not one client?

app/main/views/dashboard.py
```diff
@@ -23,6 +23,8 @@ def service_dashboard(service_id):
     try:
         service = get_service_by_id(service_id)
         session['service_name'] = service['data']['name']
+        print(service['data']['id'])
```
minglis: print!

app/__init__.py
mannickutd: pivotal task created, to be cont..

PR #213, opened 2016-03-01
Accept CSVs with 'email address' or 'phone number'

quis:
### _Depends on https://github.com/alphagov/notifications-api/pull/109_
***

CSV files currently have ‘to’ as the recipient column. This is changing in https://github.com/alphagov/notifications-api/pull/109

The admin app also has to validate that the CSV files have the right columns, because the API expects any CSV that it’s given to have been checked (also we want things to actually work).

This commit is the minimum code change needed. In the future it should reuse the same code as the API for processing CSV files. This will need more thinking.

image

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -27,6 +27,7 @@ from app.main.dao import templates_dao
 from app.main.dao import services_dao
 from app import job_api_client
 from app.utils import validate_recipient, InvalidPhoneError, InvalidEmailError
+from utils.process_csv import first_column_heading
 
 page_headings = {
     'email': 'Send emails',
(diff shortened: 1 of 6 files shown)


app/main/views/send.py
```diff
@@ -174,7 +176,7 @@ def check_messages(service_id, upload_id):
             upload_result=upload_result,
             template=template,
             page_heading=page_headings[template.template_type],
-            column_headers=['to'] + list(template.placeholders_as_markup),
+            column_headers=[first_column_heading[template.template_type]] + list(template.placeholders_as_markup),
```
mannickutd: Nothing big here, just something to keep in mind in the future.
The list object has an 'extend' method, that uses the iterable interface. Using the 'extend' method would save you type casting 'template.placeholders_as_markup' as a list.

quis: Didn’t know about `[].extend()`, cool. 

Done some digging and it doesn’t work very cleanly here because it doesn’t return the new array, eg:
```python
l = [1, 2, 3].extend([4, 5, 6])
print(l)
>>> None
```

mannickutd: Yep this is fine

mannickutd:
Get all tests passing and we can get this merged.

PR #217, opened 2016-03-01
Add send permission

mannickutd:


app/notify_client/models.py
```diff
@@ -81,8 +81,12 @@ def permissions(self):
     def permissions(self, permissions):
         raise AttributeError("Read only property")
 
-    def has_permissions(self, service_id, permissions):
-        return True
+    def has_permissions(self, service_id, permissions, or_=False):
```
quis: This method is neat, but it needs unit tests.

app/templates/main_nav.html
```diff
@@ -8,6 +8,12 @@ <h2 class="navigation-service-name">
     <li><a href="{{ url_for('.choose_template', service_id=service_id, template_type='email') }}">Send emails</a></li>
     <li><a href="{{ url_for('.letters_stub', service_id=service_id) }}">Send letters</a></li>
   </ul>
+  {% elif current_user.has_permissions(session.get('service_id', ''), ['manage_templates']) %}
+  <ul>
+    <li><a href="{{ url_for('.choose_template', service_id=service_id, template_type='sms') }}">Text message templates</a></li>
+    <li><a href="{{ url_for('.choose_template', service_id=service_id, template_type='email') }}">Text email templates</a></li>
```
quis: “email templates”, not “Text email templates”

app/templates/views/choose-template.html
```diff
@@ -49,7 +55,9 @@ <h1 class="heading-large">{{ page_heading }}</h1>
     {% endif %}
 
     <p>
+      {% if current_user.has_permissions(session.get('service_id', ''), ['manage_templates']) %}
```
quis: It’s a bit repetitive having to pass the `service_id` in every time. Could we try [writing a template filter](http://jinja.pocoo.org/docs/dev/api/#writing-filters) instead? eg

```jinja
{% if current_user|has_permission('manage_templates') %}
```

quis: Could work exactly the same as this method: https://github.com/alphagov/notifications-admin/pull/217/files#diff-5609c27c7d83fbaa57c74f5491ef464bR99

PR #225, opened 2016-03-03
Revalidate file upload

servingUpAces:
If the file was invalid and Upload a CSV file was clicked, the job was created, then the send would fail when sending the file, trying to replace a placeholder that didn't exist.

This commit calls send_messages again if the files exist on the request.


app/main/views/send.py
```diff
@@ -197,21 +197,25 @@ def check_messages(service_id, upload_id):
...
-                raise e
-
-        flash('We’ve started sending your messages', 'default_with_tick')
-        return redirect(
-            url_for('main.view_job', service_id=service_id, job_id=upload_id)
-        )
+        if request.files:
+            # The csv was invalid, validate the csv again
+            return send_messages(service_id, template_id)
+        else:
```
quis: Don’t need the `else:` here because of the short-circuit `return`.

quis:
Should probably have a test for this. It’s a pretty important bit of routing.

quis:
:+1: to merge once Travis is dealt with.

PR #227, opened 2016-03-03
Make job page poll for updates

quis:
This is a first go at having the job page update without refreshing.

The approach I’ve taken is to do all the rendering of HTML on the server side, rather than use a Javascipt templating engine like mustache. This ensures that we don’t have to maintain two sets of templates.

So the approach is to split the job page into partials. These partials can then:
- be included in the job page to render the whole page
- be rendered indivudually and then returned as a blob of HTML inside a JSON response

Then I’ve added a Javascript module which looks for areas of the page that should be reloaded. For each area of the page it will poll a URL and re-render that section of the page when it gets new HTML. It implements some throttling so that API calls will never happen more frequently than 0.67 times/second.

## Remove tick banner

I’d like to see if we can get away without this now that the page updates. In research the ‘We’ve started’ part of the message confused people, especially when they’d only sent one message.


tests/app/main/views/test_jobs.py
```diff
@@ -41,4 +45,39 @@ def test_should_show_page_for_one_job(app_,
...
+    file_name = job_data['original_file_name']
+
+    with app_.test_request_context():
+        with app_.test_client() as client:
+            client.login(api_user_active)
+            response = client.get(url_for('main.view_job_updates', service_id=service_id, job_id=job_id))
+
+        assert response.status_code == 200
+        content = json.loads(response.get_data(as_text=True))
+        print(content)
```
mannickutd: Please remove print statement

PR #228, opened 2016-03-03
Fix permission names, all tests passing.

mannickutd:


quis:
:+1: to merge it yourself when Travis passes.

PR #237, opened 2016-03-07
Filter out accepted invites from manage users page

ashimali:
Also added basic test of invite client.

### DIFF
--- a/app/notify_client/invite_api_client.py
+++ b/app/notify_client/invite_api_client.py
@@ -48,6 +48,7 @@ class InviteApiClient(BaseAPIClient):
     def _get_invited_users(self, invites):
         invited_users = []
         for invite in invites:
-            invited_user = InvitedUser(**invite)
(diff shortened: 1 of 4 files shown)


app/notify_client/invite_api_client.py
```diff
@@ -48,6 +48,7 @@ def accept_invite(self, service_id, invited_user_id):
     def _get_invited_users(self, invites):
         invited_users = []
         for invite in invites:
-            invited_user = InvitedUser(**invite)
-            invited_users.append(invited_user)
+            if invite['status'] != 'accepted':
```
minglis: Could this be done at point of use? Could the template show what it cares about rather than never have access to accepted invites in the application?

Would we ever need to know about accepted invites in the UI?

ashimali: @minglis sure, i'll move to template.

PR #238, opened 2016-03-07
Show invite email

ashimali:
Register from invite now shows invited user email in disabled field


quis:
Print the email address on the page in a `<p>` if it’s not editable. `<input disabled…` is non-obvious.

ashimali:
@quis will do

PR #241, opened 2016-03-08
Give the user better error messages for CSV files

quis:
_Depends on:_
- [x] _alphagov/notifications-utils#9_
- [ ] _alphagov/notifications-api#134_

screencapture-localhost-6012-services-604d66ec-72d4-4145-9948-51e4fe19377d-check-c46f9769-1404-4b42-aa52-ebd3e24c6056-1457427932751

This pull request strips out a lot of the complex stuff that the views and templates in this app were doing. There is now a cleaner separation of concerns:

- utils returns the number and type of errors in the csv
- `get_errors_for_csv` helper in this app maps the number and type of errors onto human-friendly error messages
- the view and template just doing the glueing-together of all the pieces

This is (hopefully) easier to understand, definitely makes the component parts easier to test in isolation, and makes it easier to give more specific error messages.

## Store phone number as the user entered it

It’s confusing to the user to have their phone number played back to them in a format that they didn’t enter it. We’ve seen multiple times that people enter 0781… and then don’t recognise their own phone number when it’s played back as +44781…

The API can handle phone numbers in any format as of alphagov/notifications-api#134

So there is no need to reformat the user’s phone number before storing it now.

***

[#115026867](https://www.pivotaltracker.com/story/show/115026867)


app/main/views/send.py
```diff
@@ -185,107 +193,77 @@ def send_message_to_self(service_id, template_id):
...
+
+    try:
+        job_api_client.create_job(
+            upload_id,
+            service_id,
+            upload_data.get('template_id'),
+            upload_data.get('original_file_name'),
+            upload_data.get('notification_count')
+        )
+    except HTTPError as e:
```
mannickutd: We don't want to abort a 404 here, we should be throwing 500 if the api fails.

quis: Seems legit. Removed the `try`/`except` in c33bb0b.

ashimali:
:thumbsup:

PR #243, opened 2016-03-08
Forgot password

servingUpAces:
story: https://www.pivotaltracker.com/story/show/114940405

The password reset functionality was broken when the notifications-python-client stopped taking content as an option in the send_email method. The forgot-password endpoint will call the newly created /user/reset-password api endpoint to send an email contain the forgot password url. Updates to new-password were also required to make this work.

### DIFF
--- a/tests/app/main/views/test_two_factor.py
+++ b/tests/app/main/views/test_two_factor.py
@@ -1,7 +1,5 @@
 from flask import url_for
 
-from tests import create_test_user
-
 
 def test_should_render_two_factor_page(app_,
                                        api_user_active,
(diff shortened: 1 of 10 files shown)


tests/app/main/views/test_two_factor.py
```diff
@@ -109,3 +107,30 @@ def test_remember_me_set(app_,
...
+                    'email': api_user_active.email_address,
+                    'password': 'changedpassword'}
+
+            response = client.post(url_for('main.two_factor'),
+                                   data={'sms_code': '12345'})
+
+            assert response.status_code == 302
+            assert response.location == url_for(
+                'main.service_dashboard',
+                service_id="596364a0-858e-42c8-9062-a8fe822260eb",
```
mannickutd: service = mock_get_services_with_one_service()
service['id']
instead of using the constant

PR #245, opened 2016-03-08
Exact permissions added.

mannickutd:


quis:
There’s a lot of boilerplate in the tests here. Could perhaps have one test, and [parameterize](https://pytest.org/latest/parametrize.html) it with a list of arguments to `user_has_permissions()` and a list of exception(s) that should be [suppressed](https://docs.python.org/3/library/contextlib.html#contextlib.suppress).

mannickutd:
Yep ok boilerplate can be fixed up but we need this fix now  :).

PR #246, opened 2016-03-09
Display email address that invitee will be registered with.

ashimali:
Also add flash message for users who already have an account.


minglis:
I've seen this: 👍

PR #247, opened 2016-03-09
Email flow fixes

quis:
## Fix bug which prevented viewing an email job

image

The template for viewing a job was not getting all the variables it needed in order to display an email template. Hadn’t noticed this before, because email templates require more variables than SMS templates.

This commit fixes that bug.

## Make text of ‘send’ button say ‘email’ or ‘text…’

image

Previously the send button on the ‘Check and confirm’ page always said ‘Send x messages’, irrespective of whether you were sending emails or text messages.

This commit makes it say one of
- Send 1 email
- Send 29 emails
- Send 1 text message
- Send 999 text messages

***

[#115326437](https://www.pivotaltracker.com/story/show/115326437)

### DIFF
--- a/app/assets/stylesheets/components/email-message.scss
+++ b/app/assets/stylesheets/components/email-message.scss
@@ -17,6 +17,7 @@
       @include core-19;
       border-bottom: 0;
       border-top: 1px solid $border-colour;
+      vertical-align: top;
     }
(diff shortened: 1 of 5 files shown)


minglis:
I've seen this 👍

PR #250, opened 2016-03-09
Default service limit reduced to 50

minglis:
Reduced to 50

- question is, should this not be done in the API?


quis:
:+1: (for the time being)

PR #251, opened 2016-03-09
View function for request go live was always calling update service.

ashimali:
Actually this should be no op until whatever workflow will take place
in the real world is implemented.

For the moment just display flash message to say request being
processed and do nothing.


mannickutd:
:+1: When you are ready

PR #252, opened 2016-03-09
Fix for duplicate email registration

ashimali:
That front end got stack trace and not flash error as expected.

The error message does not specify the error is a duplicate email
address so as not to reveal which emails are current user accounts.

### DIFF
--- a/app/main/views/register.py
+++ b/app/main/views/register.py
@@ -35,6 +35,7 @@ def register():
             return redirect(url_for('main.verify'))
         else:
             flash('There was an error registering your account')
+            return render_template('views/register.html', form=form), 400
 
     return render_template('views/register.html', form=form)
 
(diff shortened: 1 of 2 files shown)


app/main/views/register.py
```diff
@@ -46,7 +50,11 @@ def register_from_invite():
     if form.validate_on_submit():
         if form.service.data != invited_user['service'] or form.email_address.data != invited_user['email_address']:
             abort(400)
-        return _do_registration(form)
+        registered = _do_registration(form)
+        if registered:
+            return redirect(url_for('main.verify'))
+        else:
+            flash('There was an error registering your account')
```
minglis: Should this have a status code of 400 or 401 or something?

minglis: Allows us to see it in logs

minglis:
👍

In lieu of bigger fix coming down the pipe

PR #254, opened 2016-03-09
Fix bug with ‘ghost’ links in template textbox

quis:
This is a bug with the code that highlights placeholders.

It was:

1. taking the value of the textbox
2. copying it straight into the HTML of the layer that contains the blue lozenges

This meant that any HTML that the user typed into the textbox was rendered by the browser.

This commit fixes the bug by:

1. taking the contents of the textbox
2. copying it to the _text_ (not inner HTML) of a temporary `<div>`
3. taking the inner HTML of that `<div>` (whose text has been encoded with HTML entities in step 2., eg `>` becomes `&gt;`)
4. using that for the HTML content of the layer with the blue lozenges

Before | After
--- | ---
screen shot 2016-03-09 at 16 52 11 | screen shot 2016-03-09 at 16 52 46

### DIFF
--- a/app/assets/javascripts/highlightTags.js
+++ b/app/assets/javascripts/highlightTags.js
@@ -33,7 +33,7 @@
     };
 
     this.update = () => this.$backgroundMaskForeground.html(
-      this.$textbox.val().replace(
+      $('<div/>').text(this.$textbox.val()).html().replace(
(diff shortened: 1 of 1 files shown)


mannickutd:
:+1:

PR #255, opened 2016-03-09
Added error message if no permissions chosen

ashimali:
If no permissions chosen in invite, render validation errors.

Custom error message to over ride that from WTF forms radio field required creation of a custom field.


quis:
Going to pick this up tomorrow and add some CSS.

PR #256, opened 2016-03-10
Use ‘name 1’ when sending yourself a message

quis:
We reckon this is better than ‘test name’, ‘test date’, etc.


mannickutd:
:+1:

PR #260, opened 2016-03-10
Error handling

servingUpAces:
- Refactored register_errorhandlers so that it handles HTTPError
- Remove most cases where we catch HTTPError
- Handle HTTPError from API and routing errors from flask.
- Removed catching the HTTPError where applicable.


ashimali:
Seen by me :+1:

PR #262, opened 2016-03-11
Make preview of uploaded CSV files clearer

quis:
Before | After
--- | ---
image | screen shot 2016-03-11 at 08 53 41 image

## Start row numbers from 2
Because row 1 will be the column headers when you look at a CSV file in Excel.

## Don’t mix errors with valid rows

Brings in (and depends on): https://github.com/alphagov/notifications-utils/pull/11

Changes the number of rows shown to be at most 15 (either 15 rows with errors or 15 valid rows).


quis:
Demo’d to @minglis

PR #264, opened 2016-03-11
Fix to sign in test and user model

servingUpAces:
- Noticed that the api was being called when running the unit tests.
- This also lead me to find the the failed_login_in count was always returning 0.


ashimali:
all good :+1:

PR #265, opened 2016-03-11
Bug fix for passing list of empty string if no permissions selected for

ashimali:
invite.

Send empty list instead.

### DIFF
--- a/app/notify_client/models.py
+++ b/app/notify_client/models.py
@@ -129,7 +129,10 @@ class InvitedUser(object):
         if isinstance(permissions, list):
             self.permissions = permissions
         else:
-            self.permissions = permissions.split(',')
+            if permissions:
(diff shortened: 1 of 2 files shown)


mannickutd:
:+1:

app/notify_client/models.py
```diff
@@ -129,7 +129,10 @@ def __init__(self, id, service, from_user, email_address, permissions, status, c
         if isinstance(permissions, list):
             self.permissions = permissions
         else:
-            self.permissions = permissions.split(',')
+            if permissions:
+                self.permissions = permissions.split(',')
```
quis: Does this case (comma separated string) ever happen?

ashimali: @quis yep that's exactly what it expects. comma separated values or empty string.

PR #266, opened 2016-03-11
Fix bugs and gripes with adding and editing templates

quis:
## Remove ‘cancel’ buttons from edit template pages

Since cancel just takes you back to the list of templates, which is accessible from the left nav, it’s only adding clutter to the page.

## Reword the help text about adding placeholders

In research we’ve seen users struggle with this. Often they’d type ((Joe Bloggs)) or ((XX/XX/XXXX)) instead of a sensible name.

Lets see if explaining it differently helps.

screen shot 2016-03-11 at 13 38 54

## Fix innaccurate headings on add/edit template page

The same `.html` file is shared between adding a template and editing a template.

The page heading needs to be contextual to the URL, either ‘add’ or ‘edit’. Somewhere along the way this got lost; this commit reinstates it.

## Move ‘new thing’ buttons alongside heading

On the send messages and manage team pages we have big green buttons for adding/inviting a new template or team member.

On the add template page it was at the bottom, and often got missed.

On the manage team page it was at the top, but maybe too prominent because it’s big and green.

This commit tries putting it in the top right of the page instead (except when the template page is empty, in which case it’s unchanged).

screen shot 2016-03-11 at 11 07 01

screen shot 2016-03-11 at 11 09 50

### DIFF
--- a/app/assets/stylesheets/_grids.scss
+++ b/app/assets/stylesheets/_grids.scss
@@ -2,7 +2,6 @@
   @include grid-column(3/4);
 }
 
-
 .column-one-eighth {
   @include grid-column(1/8);
 }
@@ -18,4 +17,12 @@
 
 .bottom-gutter-2-3 {
   margin-bottom: $gutter * 2/3;
+}
+
+.align-with-heading {
+  display: block;
+  text-align: center;
+  margin-top: 45px;
+  padding-left: 2px;
+  padding-right: 2px;
 }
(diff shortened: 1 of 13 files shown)


quis:
## Make nav and page headings reflect permissions

Fixes [#115488955](https://www.pivotaltracker.com/story/show/115488955)

> If you have the Edit templates permission but not the Send messages permission, the navigation should read
>
> - Text message templates
> - Email templates
> - Letter templates
>
> The page headings should also read
>
> - Text message templates
> - Email templates
> - Letter templates
> 
> respectively

mannickutd:
:+1:

PR #267, opened 2016-03-11
Add page explaining how to send template using API

quis:
Would like to test something like this and see how well it works.

Intention of having this page is so:
- template IDs are discoverable (https://www.pivotaltracker.com/story/show/115404593)
- it’s obvious there’s an ‘automated’ way to send messages, as well as the CSV way (we’ve seen people oblivious to this in research)

screen shot 2016-03-11 at 09 57 11
screen shot 2016-03-11 at 10 41 01


minglis:
👍

PR #268, opened 2016-03-11
Add contact service manager message

mannickutd:


quis:
This will show to any user who doesn’t have templates. Doesn’t seem right—you should only see this message if you don’t have permission to create templates.

PR #269, opened 2016-03-11
Change new invite registration flow to only need

ashimali:
sms for verification.

This may change again soon with story to split 2 factor
pages, but for now is correct.


quis:
What is going on with the tests on this PR?

ashimali:
@quis oops unintended

mannickutd:
:+1:

PR #273, opened 2016-03-14
added message if not templates and doesn't have ability to add templa…

mannickutd:
…tes.

### DIFF
--- a/app/templates/views/choose-template.html
+++ b/app/templates/views/choose-template.html
@@ -17,7 +17,7 @@
     {% if current_user.has_permissions(['manage_templates']) %}
       <a href="{{ url_for('.add_service_template', service_id=service_id, template_type=template_type) }}" class="button">Add a new template</a>
     {% else %}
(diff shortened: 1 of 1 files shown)


mannickutd:
Demoed to @servingUpAces

app/templates/views/choose-template.html
```diff
@@ -16,6 +16,8 @@ <h1 class="heading-large">{{ page_heading }}</h1>
 
     {% if current_user.has_permissions(['manage_templates']) %}
       <a href="{{ url_for('.add_service_template', service_id=service_id, template_type=template_type) }}" class="button">Add a new template</a>
+    {% else %}
+      <div class="column-two-thirds">You need to ask your service manager to add templates before you can send messages</div>
```
quis: This isn’t inside a `.grid-row` so the alignment will be off. Better to just make it a paragraph with no class.

quis:
:+1:

PR #275, opened 2016-03-14
Add email list

mannickutd:


config.py
```diff
@@ -64,6 +65,7 @@ class Development(Config):
     WTF_CSRF_ENABLED = False
     REMEMBER_COOKIE_SECURE = False
     SESSION_COOKIE_SECURE = False
+    EMAIL_DOMAIN_LIST = os.getenv('EMAIL_DOMAIN_LIST', 'gov.uk,notify.gov.uk,cabinet.digital-office.gov.uk').split(',')
```
quis: @**digital**._cabinet_-office.gov.uk

Also, …@gov.uk and …@notify.gov.uk will never exist, so why are they in here?

mannickutd: It was just for development and testing but I will change the default. and not on development will have to pull from the environment

PR #276, opened 2016-03-14
The verify view was not passing along the next param to the two factor

ashimali:
view.

Now if it is passed and it is a url on the same domain that request
originates from then it is used.


servingUpAces:
:+1:

PR #283, opened 2016-03-15
Accept invite if user already part of service bug fix

ashimali:
Additional check needed to see if user was already a user for the service that they were invited to.


mannickutd:
:+1:

PR #284, opened 2016-03-15
Do not send email in case of invite.

ashimali:
The user does not have to validate the email token, but it
was still being sent.


mannickutd:
:+1:

PR #286, opened 2016-03-16
Fix to forgot password

servingUpAces:
Should show the sent password reset link page when the user is not found.


quis:
Demo’d :+1:

PR #290, opened 2016-03-17
Platform admin user 

servingUpAces:
- A platform admin user is able to see a list of all services.
- Each service on the list is linked to the dashboard page of the service.
- The platform admin user can see/edit templates, see/invite users, see/edit service settings.
- The platform admin user can not send messages, see/edit api keys and developer docs.
- Renamed notifications_api_client to service_api_client

### DIFF
--- a/app/notify_client/models.py
+++ b/app/notify_client/models.py
@@ -13,6 +13,7 @@ class User(UserMixin):
         self._failed_login_count = fields.get('failed_login_count')
         self._state = fields.get('state')
         self.max_failed_login_count = max_failed_login_count
+        self.platform_admin = fields.get('platform_admin')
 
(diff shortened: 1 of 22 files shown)


app/notify_client/models.py
```diff
@@ -91,6 +94,10 @@ def has_permissions(self, permissions, service_id=None, or_=False):
             return set(self._permissions[service_id]) >= set(permissions)
         return False
 
+    def has_platform_admin_permissions(self):
```
mannickutd: Think this method can be removed.

mannickutd:
Some route urls permissions would be good.
It should be fairly easy to create the tests for the each route you want to test.

PR #292, opened 2016-03-17
Start aligning Admin app with config styles used elsewhere

minglis:
- no config overrides - now all set in environment
- use different files for staging and live too allow for differently named env variables
- updates to run_app and run_tests scripts to set correct environment (test/development) so correct config picked up
- use environment file on deployed environments to pick correct config


ashimali:
:+1:

PR #293, opened 2016-03-17
Changed registration flow to first send email verification link that

ashimali:
On first registration user enters details and on submitting is sent to page saying verification link has
been sent emailed to them.

When user clicks on email verification link in email app sends sms code for second step of
account verification and user forwarded to page to enter sms code.

Successfully entering email sends user to usual create service page.

Moved user dao calls that simply proxied calls to client to calling user client directly.

Note that users can still be sent an email verification code to their email if they update email address.

This pr depends on https://github.com/alphagov/notifications-api/pull/165


servingUpAces:
:+1:

PR #305, opened 2016-03-18
Api service permissions

mannickutd:


mannickutd:
:+1: demoed to @henryhadlow

PR #309, opened 2016-03-21
Move documentation into footer, not behind login

quis:
https://www.pivotaltracker.com/story/show/115775183

image


servingUpAces:
:+1:

PR #310, opened 2016-03-21
Remove ‘delete this service’ link

quis:
> We have an option to delete a service, which looks good, but then doesn't delete the service.
>
> Until we have the ability to delete, let's get rid of the link.

https://www.pivotaltracker.com/story/show/115775329


servingUpAces:
:+1:

PR #312, opened 2016-03-21
Only prefix SMS templates with service name

quis:
Implements: https://github.com/alphagov/notifications-utils/pull/13

Fixes: https://www.pivotaltracker.com/story/show/115879891


servingUpAces:
:+1:

PR #313, opened 2016-03-21
    Do not throw 404 if email address is not found on sign in.

servingUpAces:
https://www.pivotaltracker.com/story/show/115947639


ashimali:
:+1: demoed to me

PR #314, opened 2016-03-21
In registration flow If new sms requested then check and send redirected to two factor

ashimali:
In case user is pending they need to be sent back to verify not two factor.


servingUpAces:
:+1:  demo looks good

PR #317, opened 2016-03-21
Revert "In registration flow If new sms requested then check and send redirected to two factor"

ashimali:
Reverts alphagov/notifications-admin#314

I'm not sure why travis is all upset about this. I'd like to revert and create new branch and try again.


servingUpAces:
:+1:

PR #318, opened 2016-03-21
Platform admin

servingUpAces:
- A platform admin user is able to see a list of all services.  
- Each service on the list is linked to the dashboard page of the service.
- The platform admin user can see/edit templates, see/invite users, see/edit service settings.
- The platform admin user can not send messages, see/edit api keys.


ashimali:
demoed to me :+1:

PR #320, opened 2016-03-21
In registration flow re send sms code

ashimali:
In registration flow If new sms requested then check and send redirected to two factor but should
have redirected to verify.


servingUpAces:
:+1:

PR #321, opened 2016-03-21
Only added 'Get Started' if you have the appropriate permissions.

mannickutd:


app/templates/views/service_dashboard.html
```diff
@@ -25,20 +25,22 @@
...
-      {% endcall %}
+      {% if current_user.has_permissions(['manage_templates', 'send_texts', 'send_emails', 'send_letters'], or_=True) %}
+        {% call banner_wrapper(subhead='Get started', type="tip") %}
+          <ol>
+            {% if current_user.has_permissions(['manage_templates']) %}
+              <li>
+                <a href='{{ url_for(".add_service_template", service_id=service_id, template_type="sms") }}'>Add a template</a>
+              </li>
+            {% endif %}
+            {% if current_user.has_permissions(['send_texts', 'send_emails', 'send_letters']) %}
```
quis: This (and the one above) should be `send_texts` only, because right now the link is “Send yourself a text message”.

quis:
:+1:

PR #322, opened 2016-03-21
Use real API endpoints in documentation

quis:
There is no need to make the API endpoints environment-specific. But it is still useful to know what the endpoint is.

Also fixes a broken link on the integration page.


quis:
Demo’d to @servingUpAces

PR #323, opened 2016-03-21
Add Terms of use page

henryhadlow:


henryhadlow:
@quis I need your help with two things – linking to the `_status` page (I've put a placeholder link) and adding some tests – is that necessary?

quis:
@henryhadlow The URL for that status page would be `/_status`. Not sure [how much value it adds at this stage](https://www.notifications.service.gov.uk/_status)…

app/templates/views/terms-of-use.html
```diff
@@ -0,0 +1,207 @@
...
+{% block maincolumn_content %}
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+
+    <h1 class="heading-large">
+      Terms of use
+    </h1>
+
+    <p>To accept these terms, you must be the service manager for {{ session.get('service_name', 'Service') }}. If you’re not the service manager, you’ll need to <a href="{{ url_for('.invite_user', service_id=service_id) }}">invite them</a>.</p>
```
quis: This needs to work for users who are logged in and those who aren’t. I can help with this.

app/templates/views/terms-of-use.html
```diff
@@ -0,0 +1,207 @@
...
+
+      <p>You agree not to use GOV.UK Notify to send marketing messages.</p>
+
+      <p>If you attempt to use GOV.UK Notify for marketing, your templates won’t pass our content review.</p>
+
+      <h3 class="heading-small" id="you-agree-to-send-messages-consistent-with-our-guidelines">
+        You agree to send messages consistent with our design patterns, style guide and information security guidelines
+      </h3>
+
+      <p>Your messages must follow our <a href="https://designpatterns.hackpad.com/Notifications-5vuitmNqIjZ" rel="external">design patterns</a>, <a href="https://www.gov.uk/topic/government-digital-guidance/content-publishing" rel="external">style guide</a> and <a href="https://docs.google.com/document/d/15-OjaEqDBy31uDU7nLZCpYIQOnzSCJR63-cp3cQI9G8" rel="external">information security guidelines</a>.</p>
```
quis: External links are missing, I’ll fix that.

quis: Fixed in 			bce535b

PR #332, opened 2016-03-22
Set permissions with checkboxes, not yes/no inputs

quis:
Before | After 
--- | ---
screen shot 2016-03-22 at 10 35 00 | image

The yes/no pattern didn’t work too well, because:
- it didn’t read naturally as a question and answer
- often users left them completely unclicked if they didn’t want to set the permission (rather than clicking no)

This commit changes both the invite and edit user pages to use checkboxes to set permissions. If also rewords these pages to read more naturally, and explain what the permissions mean.

This meant changing some of the view logic around invites and permissions, and I ended up refactoring a bunch of it because I found it hard to understand what was going on.


quis:
image

:boom:

quis:
Demo’d to @henryhadlow

PR #333, opened 2016-03-22
Fix for trying to verify email on partially complete registration

ashimali:
When user clicks on verification link but doesn't complete verification, if they try to use link 
again the code will have been used. 

Therefore they will need a new email with new link to use for verification.


mannickutd:
:+1:

PR #334, opened 2016-03-22
Remove user from service

mannickutd:


mannickutd:
Demoed to @henryhadlow

quis:
Can you squash the merge commits? Having extra merges in our history messes up these numbers:

image

app/main/views/manage_users.py
```diff
@@ -106,6 +109,51 @@ def edit_user_permissions(service_id, user_id):
...
 
+@main.route("/services/<service_id>/users/<user_id>/delete", methods=['GET', 'POST'])
+@login_required
+@user_has_permissions('manage_users', admin_override=True)
+def remove_user_from_service(service_id, user_id):
+    user = user_api_client.get_user(user_id)
+    service = get_service_by_id(service_id)
+    # Need to make the email address read only, or a disabled field?
+    # Do it through the template or the form class?
+    form = PermissionsForm(**{
```
quis: I rewrote this to be way cleaner yesterday: https://github.com/alphagov/notifications-admin/blob/master/app/main/views/manage_users.py#L88-L90

And if it’s going to be in two places it should probably be factored into a helper function.

app/templates/flash_messages.html
```diff
@@ -5,7 +5,7 @@
       {{ banner(
         message,
         'default' if ((category == 'default') or (category == 'default_with_tick')) else 'dangerous',
-        delete_button="Yes, delete" if 'delete' == category else None,
+        delete_button="Yes, {}".format(category) if category in ['delete', 'remove'] else None,
```
quis: Nice.

PR #335, opened 2016-03-23
Dashboard updates

quis:
aj

## Make dashboard stats update dynamically

https://www.pivotaltracker.com/story/show/115874485

This uses the same component as on the jobs page to make a partial section of the page update itself periodically.

## Allow custom interval on AJAX sections of page

Some pages with AJAX should update quickly, because the data is likely to be changing quickly, and be finished changing sooner. Other pages we want to have tick over a bit slower.

This commit adds an optional ‘interval’ parameter to the `updateContent` modules, which sets how often the page should ping the server for an update.

It then sets the interval for the dashboard page to be 10 seconds, rather than the default 1.5 seconds.

## Improve display of failure rates

https://www.pivotaltracker.com/story/show/116052359

- add absolute numbers for failures
- make percentages accurate to 1 decimal place (50.0%)
- make error colour appear if failures go above 3.0%
- make error colour boolean (don’t interpolate between the colours)

### DIFF
--- a/app/assets/javascripts/updateContent.js
+++ b/app/assets/javascripts/updateContent.js
@@ -1,12 +1,10 @@
 (function(GOVUK, Modules) {
   "use strict";
 
-  const interval = 1500; // milliseconds
-
   GOVUK.timeCache = {};
   GOVUK.resultCache = {};
 
-  let getter = function(resource, render) {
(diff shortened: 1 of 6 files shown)


mannickutd:
:+1:

PR #337, opened 2016-03-23
Functionality added and all tests working.

mannickutd:
Update correct use of permissions form.


mannickutd:
demoed to @henryhadlow

PR #342, opened 2016-03-23
Clean up config a bit. Get values for config from environment.

ashimali:
Removed some old redundant config.


environment_test.sh
```diff
@@ -0,0 +1,6 @@
+export NOTIFY_ADMIN_ENVIRONMENT='config.Test'
+export ADMIN_CLIENT_SECRET='dev-notify-secret-key'
+export ADMIN_CLIENT_USER_NAME='dev-notify-admin'
+export API_HOST_NAME='http://localhost:6011'
+export DANGEROUS_SALT='dev-notify-salt'
+export SECRET_KEY='dev-notify-secret-key'
```
mannickutd: We need debug set to True for tests.

mannickutd:
:+1:

PR #347, opened 2016-03-24
Strip trailing CSV rows, relax phone validation

quis:
Implements (and depends on):
- alphagov/notifications-utils#14
- alphagov/notifications-utils#15


mannickutd:
:+1:

PR #350, opened 2016-03-29
Only show ‘Choose service’ if multiple services

quis:
When the user clicks ‘GOV.UK Notify’ in the header, they should, by default, be redirected to the dashboard for their service. They should only see the ‘Choose service’ page if they have multiple services.

This also allows some logic to be factored out of the template, so one route now handles all this redirection.

In the future we might want to keep the last-used service in the session, and always redirect to that. But for now, this should fix most of the confusion for first-time users.

https://www.pivotaltracker.com/story/show/115878087


mannickutd:
:+1:

PR #351, opened 2016-03-29
service_dao file has been removed and replaced with just using the se…

mannickutd:
…rvice_api_client.


app/main/views/templates.py
```diff
@@ -24,7 +24,7 @@
 @user_has_permissions('manage_templates', admin_override=True)
 def add_service_template(service_id, template_type):
 
-    service = sdao.get_service_by_id_or_404(service_id)
+    service = service_api_client.get_service(service_id)
```
quis: This changes the behaviour because it won’t 404 if the service doesn’t exist.

The intention of this line (does the service exist) would be better served by a decorator I think.

quis: I was wrong about the above—the 404 will be caught here: https://github.com/alphagov/notifications-admin/blob/35e093ee34d0ccd3b23ea63e2ec2cad0381e1c7f/app/__init__.py#L184

app/main/views/service_settings.py
```diff
@@ -72,7 +66,13 @@ def _check_password(pwd):
     if form.validate_on_submit():
         service['name'] = session['service_name_change']
         try:
-            update_service(service)
+            service_api_client.update_service(
```
quis: Could we have an alternate method on the client here that accepts a service object, rather than having to unpack it every time?

quis:
I’ve reviewed all the code on this PR and left a couple of comments.

We talked in person about how we could use the `user_has_permissions` decorator in a lot of these cases instead of making an explicit API call to get the service then not use the return value, eg here:
https://github.com/alphagov/notifications-admin/pull/351/files#diff-e5c9a0476eea9e898f6077cd948778cdR54

But this requires work elsewhere, so should wait for now.

PR #352, opened 2016-03-29
If user is pending it means they have not verified email yet

ashimali:
The check that was in place sent them to verify which was wrong.


minglis:
👍

PR #353, opened 2016-03-29
Removed template_dao file and replaced with straight use of the servi…

mannickutd:
…ce_api_client calls.


app/notify_client/api_client.py
```diff
@@ -137,3 +139,32 @@ def delete_service_template(self, service_id, template_id):
         """
         endpoint = "/service/{0}/template/{1}".format(service_id, template_id)
         return self.delete(endpoint)
+
+
+class TemplatesBrowsableItem(BrowsableItem):
```
quis: We don’t use this class anywhere. Kill it with :fire:?

PR #355, opened 2016-03-29
Visually align headings inside grid rows

quis:
When the first heading on the page is inside a grid row, it doesn’t vertically align properly with the navigation. This is because it isn’t targeted by the selector that does this for pages without an initial grid row.

This commit:
- adds an extra selector to target these headings
- makes the ‘add new thing’ buttons maintain their alignment with the page heading

Before | After
--- | ---
image | image


ashimali:
:+1:

PR #356, opened 2016-03-29
After send sms or email to self or batch back button bug

ashimali:
The back button to check and confirm page returns error as session data no longer available.

Now going back from sent page does not take user to check, but rather the start of sending
either sms or email.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -98,6 +98,12 @@ def choose_template(service_id, template_type):
 @user_has_permissions('send_texts', 'send_emails', 'send_letters')
 def send_messages(service_id, template_id):
 
+    service = services_dao.get_service_by_id_or_404(service_id)
+    template = Template(
+        templates_dao.get_service_template_or_404(service_id, template_id)['data'],
+        prefix=service['name']
+    )
+
     form = CsvUploadForm()
     if form.validate_on_submit():
         try:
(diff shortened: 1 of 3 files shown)


app/main/views/send.py
```diff
@@ -98,6 +98,12 @@ def choose_template(service_id, template_type):
 @user_has_permissions('send_texts', 'send_emails', 'send_letters')
 def send_messages(service_id, template_id):
 
+    service = services_dao.get_service_by_id_or_404(service_id)
```
quis: Use the API client rather than the DAO here, else you’re going to have conflicts when  #351 and #353 get merged.

tests/app/main/views/test_send.py
```diff
@@ -46,31 +46,6 @@ def test_send_test_message_to_self(
     api_user_active,
     mock_login,
     mock_get_service,
-    mock_get_service_template,
-    mock_s3_upload,
-    mock_has_permissions
-):
-
-    expected_data = {'data': ['phone number', '+4412341234'], 'file_name': 'Test run'}
```
quis: Why are we deleting this test? Seems valid that we should test the send message to self route works for both phone and email (but we shouldn’t have the tests named the same!)

quis:
Seen in person, comments addressed, :+1:

PR #358, opened 2016-03-30
Add view_activity permissions

servingUpAces:
To ensure that only members of the service have access to pages like the dashboard, view activity, view team member, the user_has_permissions decorator was added to the pages. However, a user with no permissions on the service (thought of as default permission to view only access) would not have permission to view the page. The platform admin user also was unable to see the dashboard page for all the services. This pull request add a new permission, view_activity. All users will get this permission as a default. The pull request on notifications-api to update the data should be merge before this one, as it will update all existing users with the new permission. 

The access_developer_docs was also removed as it is not a required permission. Everyone can see the documentation.


app/main/views/manage_users.py
```diff
@@ -66,7 +66,7 @@ def invite_user(service_id):
             email_address,
             ','.join(
                 role for role in roles.keys() if request.form.get(role) == 'y'
-            )
+            ).join('view_activity')
```
mannickutd: Probably should put a comment here as to why we are joining this 'view_activity' string to the list.

app/main/views/manage_users.py
```diff
@@ -98,7 +98,7 @@ def edit_user_permissions(service_id, user_id):
             user_id, service_id,
             permissions=set(chain.from_iterable(
                 permissions for role, permissions in roles.items() if form[role].data
-            ))
+            )) | {'view_activity'}
```
mannickutd: Another comment here as to what 'view_activity' is.

environment_test.sh
```diff
@@ -1,6 +1,6 @@
 export NOTIFY_ADMIN_ENVIRONMENT='config.Test'
 export ADMIN_CLIENT_SECRET='dev-notify-secret-key'
 export ADMIN_CLIENT_USER_NAME='dev-notify-admin'
-export API_HOST_NAME='http://localhost:6011'
+export API_HOST_NAME='http://localhost:6311'
```
mannickutd: Why is this changed?

tests/app/main/views/test_manage_users.py
```diff
@@ -305,45 +274,46 @@ def test_no_permission_manage_users_page(app_,
...
+                    user_id=active_user_with_permissions.id))
             assert response.status_code == 302
             assert response.location == url_for(
-                'main.manage_users', service_id=service['id'], _external=True)
+                'main.manage_users', service_id=service_one['id'], _external=True)
+            mock_remove_user_from_service.assert_called_once_with(service_one['id'],
+                                                                  str(active_user_with_permissions.id))
+
+
+def _login_user(user, client, mocker, service):
```
mannickutd: Why is there another way to login now?
Why not just adjust the original way to login with this new way?
Now we have 2 ways to login a test user.

app/main/views/manage_users.py
quis: I don’t think this is doing what you think it’s doing:
```
In [6]: ",".join(['one', 'two', 'three']).join('four')
Out[6]: 'fone,two,threeoone,two,threeuone,two,threer'
```

quis:
```python
','.join(list(role for role in roles.keys() if request.form.get(role) == 'y') + 'view_activity')
```

Would do it.

environment_test.sh
servingUpAces: There has been a few times where the test pass when there are missing mocks, so set it to something other than the port we are using.

quis: This is a good idea.

mannickutd:
:+1:

PR #361, opened 2016-03-30
Removed some un needed flash messages raised as bugs.

ashimali:
In the process found a couple of edge cases of incorrect use of invitation links by other users which are now handled.

### DIFF
--- a/tests/app/main/views/test_accept_invite.py
+++ b/tests/app/main/views/test_accept_invite.py
@@ -16,8 +16,7 @@ def test_existing_user_accept_invite_calls_api_and_redirects_to_dashboard(app_,
                                                                           mock_check_invite_token,
                                                                           mock_get_user_by_email,
                                                                           mock_get_users_by_service,
-                                                                          mock_add_user_to_service,
-                                                                          mock_accept_invite):
+                                                                          mock_add_user_to_service):
 
     expected_service = service_one['id']
     expected_redirect_location = 'http://localhost/services/{}/dashboard'.format(expected_service)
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_accept_invite.py
```diff
@@ -61,9 +59,9 @@ def test_existing_user_with_no_permissions_accept_invite(app_,
             assert response.status_code == 302
```
servingUpAces: Should we assert the mock_add_user_to_service is being called?

tests/app/main/views/test_accept_invite.py
```diff
@@ -122,15 +119,13 @@ def test_existing_signed_out_user_accept_invite_redirects_to_sign_in(app_,
...
-            mock_accept_invite.assert_called_with(expected_service, sample_invite['id'])
 
             assert response.status_code == 200
             page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
             assert page.h1.string.strip() == 'Sign in'
             flash_banners = page.find_all('div', class_='banner-default')
-            assert len(flash_banners) == 2
+            assert len(flash_banners) == 1
             assert flash_banners[0].text.strip() == 'Please log in to access this page.'
-            assert flash_banners[1].text.strip() == 'You already have an account with GOV.UK Notify. Sign in to your account to accept this invitation.'  # noqa
```
servingUpAces: Should we check the mocks are being called appropriately?

servingUpAces:
Let's add some assertions for the mocks in the unit tests, then it should be ready to go.

tests/app/main/views/test_accept_invite.py
ashimali: it is :)

servingUpAces:
:-1:

PR #363, opened 2016-03-31
Fix logic around team on navigation

quis:
In the navigation you should see either:
- manage team
- view team members

This depends on which permissions you have. You shouldn’t see both at once. There was a bug which meant you could see both. This commit fixes that bug.

https://www.pivotaltracker.com/story/show/116608291


mannickutd:
:+1:

PR #366, opened 2016-03-31
Email from uniqueness

servingUpAces:
- Update the service name validation in the ServiceNameForm and AddServiceForm to check the email_safe version of the name against a list of all service email_from fields.
- Update find_all_service_names to find_all_service_email_from, which returns the email_from of all services.
- Send email_from when creating the service
- Update email_from when changing the service_name


ashimali:
:+1: seen the demo

PR #367, opened 2016-04-01
Make ‘get started’ work for emails and texts

quis:
image

***

image

***

Previously the ‘get started’ block on the dashboard pushed users into starting with a text message. This led to users writing email content into a text message template because it didn’t match their expectations.

Also, the ‘send yourself a test’ link from the dashboard:
- wasn’t much used, because users responded to the similar call to action on the choose template page instead
- was confusing if you had created an email template because it presupposed that you’d created a text message template

So this commit changes the ‘get started’ block to be a choice between creating an email template or a text message template. We reckon that the language of ‘set up’ will help establish that templates are not
equivalent to individual messages. This language is now repeated on the ‘choose template’ page when you don’t have templates as well.

## Remove `service_dashboard.html`

This template wasn’t being used. It was replaced by `dashboard/dashboard.html` in b38ae08


mannickutd:
:+1:

PR #368, opened 2016-04-01
Don’t name SASS files with underscores

quis:
Underscores at the start of filenames are a convention that’s supposed to tell a preprocessor to ignore a file. Since we’re explicitly specifying which files need to be preprocessed, this isn’t necessary.

Also, it makes SASS lint throw a warning.


mannickutd:
Sorry my bad i think.
:+1:

PR #370, opened 2016-04-04
Make first time use clearer

quis:
# Show a tour when users first create a service

This commit adds a 3 screen tour, similar to those used on GOV.UK Verify and Passports.

We guerilla tested this on Friday, and it really helped users to build a mental model of how Notify works, so that when they’re playing around with it they have a greater sense of what they’re aiming to do. This
makes concepts like templates and placeholders click more quickly.

https://www.pivotaltracker.com/story/show/116710119

screen shot 2016-04-04 at 09 26 11
screen shot 2016-04-04 at 09 26 15
screen shot 2016-04-04 at 09 26 21
screen shot 2016-04-04 at 09 26 24
screen shot 2016-04-04 at 09 26 30

# Enhance the ‘how to do placeholders’ hint

This commit makes the ‘how to do placeholders’ box part of the tour,with the same blue background.

It also adds some Javascript enhancement so that:
- it responds to the contents of the message template
- has a ‘show me’ link which inserts ‘Dear ((name))’ into the template contents textbox

We’ve found that this has helped people understand what placeholders are, and how to do them.

placehint

# Make add service page much simpler

We reckon that most of the text on this page wasn’t helping.

image


app/main/forms.py
```diff
@@ -202,7 +202,7 @@ class SMSTemplateForm(Form):
         validators=[DataRequired(message="Template name cannot be empty")])
 
     template_content = TextAreaField(
-        u'Message',
+        u'Message content',
         validators=[DataRequired(message="Template content cannot be empty")])
```
mannickutd: Maybe the error message needs to reflect the field name?

quis: Good point. Went through and reviewed all the error messages in 953be15.

mannickutd:
:+1:

PR #371, opened 2016-04-04
Tweak some wording

quis:
## Remove the verbs from the navigation

Because ‘Send text messages’ isn’t very helpful if you’re looking to edit a template.

It also helps front-load the navigation, ie ‘Team’ is the first word, rather than the more generic ‘Manage’.

## Add a hint to ‘Template name’ text input

> Your recipients won’t see this

We’ve seen lots of users unsure what the ‘template name’ field is for, especially on first use. Is it the subject line of the text message? Will it be who the message comes from? Answer: neither.

So this commit adds a hint to make that clearer.

## Replace ‘send a batch’ with ‘send from a CSV file’ 

This is better. Not done enough testing to know if it’s the best, but let’s try it.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -29,18 +29,6 @@ from app import (job_api_client, service_api_client)
 from app.utils import user_has_permissions, get_errors_for_csv
 
 
-send_messages_page_headings = {
-    'email': 'Send emails',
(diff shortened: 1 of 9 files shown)


mannickutd:
:+1:

PR #372, opened 2016-04-04
Improve csv examples

quis:
## Put two rows in example CSV file

We reckon that this will help people understand that the first row is headers and the subsequent rows are the users’ data.

## Use context managers for StringIO

It’s possible that leaving a StringIO stream open could cause a memory leak. Using a context manager closes it automatically when it’s finished with.

## Show an example of a CSV on the ‘send’ page

image

This should help:
- understanding of what the CSV should contain
- understanding what ‘download example’ will do

## Use template name as file name for example CSV

`11.csv` or `9a1d8a7f-c202-43b9-9239-68939d34fec6.csv` is not very friendly.

`Two week reminder.csv` is much nicer.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -49,6 +49,18 @@ def get_page_headings(template_type):
     }[template_type]
 
 
+def get_example_csv_rows(template):
+    return [
+        [
+            {
+                'email': current_user.email_address,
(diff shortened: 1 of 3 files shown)


mannickutd:
:+1:

PR #376, opened 2016-04-04
Show error if restricted service uploads real CSV

quis:
Implements: https://github.com/alphagov/notifications-utils/pull/16

Also changes a bunch of the mocked phone numbers to be real-looking so that the tests are checking for the right kind of error.


mannickutd:
:+1:

PR #377, opened 2016-04-04
Split Help page into Trial Mode and Pricing pages

henryhadlow:
- Add routes
- Add footer links
- Make heading size consistent on Cookies page


app/templates/views/trial-mode.html
```diff
@@ -0,0 +1,80 @@
...
+      <li>
+        you can only send 50 messages per day
+      </li>
+    </ul>
+
+    <p>
+      When you’re ready we can remove these restrictions.
+    </p>
+
+    <!-- <h2 class="heading-medium">To remove these restrictions</h2>
```
quis: Just remove it. If it ever needs to come back it’s still in the Git history.

quis:
Checked out locally, :+1:

PR #380, opened 2016-04-04
Use SVG images for tour

quis:
SVG images will look sharper and scale more cleanly. However they are not supported on older browsers[1].

> The <picture> element allows for fallback images when the browser doesn't support a specified image format.

`<picture>` is supported in the latest versions of Chrome, Firefox, IE (Edge), Safari and Android, so all these browsers will get the SVG version[2].

Older browsers will fall back to the PNG version.

[1] http://caniuse.com/#feat=svg
[2] https://css-tricks.com/a-complete-guide-to-svg-fallbacks/#fallback-svg-as-img-picture
[3] http://caniuse.com/#feat=picture

This pull request also adds page `<title>`s to tour pages.


mannickutd:
:+1:

PR #382, opened 2016-04-05
Some proposed changes to the docs

ashimali:
@catherineheywood here are a few suggestions. You are by no means obliged to accept of course.


mannickutd:
:+1:

PR #384, opened 2016-04-05
Fix incorrect heading on send page

quis:
Accidentally reverted here: https://github.com/alphagov/notifications-admin/commit/f84797e6d538d1aec15c0ce324c8b76563236df7#diff-f7abe60b3b6c55859d983b3e19008838L14

Matches the link on the choose template page.


mannickutd:
:+1:

PR #387, opened 2016-04-05
Template statistics now surfaced on dashboard.

ashimali:
Job list removed.

Template statistics retrieved at same time as
notification stats.


quis:
Should have tests for the client new API client, eg: https://github.com/alphagov/notifications-admin/blob/master/tests/app/main/notify_client/test_notification_client.py

ashimali:
@quis test added

app/main/views/dashboard.py
```diff
@@ -36,7 +45,8 @@ def service_dashboard(service_id):
         spent_this_month='0.00',
         statistics=add_rates_to(statistics),
         templates=templates,
-        service_id=str(service_id))
+        service_id=str(service_id),
```
quis: You don’t need to pass this through any more, do `current_service.id` in the template instead.

app/templates/views/dashboard/template-statistics.html
```diff
@@ -0,0 +1,20 @@
+{% from "components/table.html" import list_table, field, right_aligned_field_heading %}
+{% call(item) list_table(
+  template_statistics,
+  caption="Recent templates used",
+  empty_message='You haven’t sent any batch messages yet',
+  field_headings=['Template', 'Type', right_aligned_field_heading('Usage')]
+) %}
+  {% call field() %}
+    <a href="{{ url_for('.edit_service_template', service_id=service_id, template_id=item.template.id) }}">
```
quis: `service_id` is missing here when this is rendered from the `.json` endpoint.

servingUpAces:
:+1:

quis:
:+1: :+1:

PR #394, opened 2016-04-05
When a user accepts an invite, show them the tour

quis:
It’s jarring to be sent right into the dashboard.

We think the tour makes things less jarring.


mannickutd:
:+1:

PR #397, opened 2016-04-06
Make it easier to deal with phone numbers in spreadsheets

quis:
## Make phone numbers human readable
- in the example table on the ‘send’ page
- in the example CSV files

Implements: alphagov/notifications-utils#18

## Add row numbers to tables

When a table is showing the contents of a CSV file, it should look something like a spreadsheet.

The minimally skeuomorphic way to do this is by adding row numbers.

image

This commit doesn’t
- make the row numbers monospace (it’s barely noticeable and doesn’t reflect what actual spreadsheets do)
- make the first column heading ‘Row’ (again, doesn’t reflect how actual spreadsheets work, and takes up more valuable space)

https://www.pivotaltracker.com/story/show/116966855

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -73,6 +73,10 @@
 
   }
 
+  &-index {
+    width: 15px;
+  }
+
 }
 
 .table-field-heading {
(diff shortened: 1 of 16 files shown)


mannickutd:
:+1:

PR #408, opened 2016-04-07
Host image used in email template in app

quis:
Right now the crown logo in email templates is hosted on Github. Github is not a CDN.

For now, hosting it in the app is a better solution. At some point we should have a CDN for all assets on the app, which would be even better.

https://www.pivotaltracker.com/story/show/116952911


mannickutd:
:+1:

PR #412, opened 2016-04-07
Aggregation of template stats in admin app.

ashimali:
Individual templates stats items fetched from api and aggregated in admin app.


mannickutd:
:+1:

PR #415, opened 2016-04-07
Change wording of form hints on register page

quis:
We’ve seen in research that people can be reticent to give their real phone number. Telling them that it will be used for something should help (ie we’re not just collecting it for marketing).

This also rewords the other form hints on this page to be less computery because we haven’t looked at them in aaaages.

Before | After
--- | ---
image | image


mannickutd:
:+1:

PR #416, opened 2016-04-07
Make the truncation of rows clearer on the check page

quis:
## eg

> Only showing the first 50 rows

-

> Only showing rows with errors

-

> Only showing the first 50 rows with errors

## Change wording when there are hidden rows

This caused some anxiety about why the rows were being hidden. Were there problems with them?

This commit reframes the wording to talk about the rows that are shown instead.

## Increase number of rows shown

We probably shouldn’t hide the contents of the CSV when people are just testing the app, or if they’re starting off with small jobs.

A limit of 15 rows displayed was awkwardly on the cusp between just testing and sending a small batch.

This commit increases the limit to 50; I reckon that over 50 recipients no-one will be wanting to check them all individually.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -227,8 +227,8 @@ def check_messages(service_id, template_type, upload_id):
         contents,
         template_type=template.template_type,
         placeholders=template.placeholders,
-        max_initial_rows_shown=15,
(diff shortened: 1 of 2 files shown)


mannickutd:
:+1:

PR #417, opened 2016-04-07
Revert "Use tablular numbers for ‘big number’ pattern"

quis:
Reverts alphagov/notifications-admin#169

> they are ugly :P

– @henryhadlow


mannickutd:
:+1:

PR #418, opened 2016-04-07
More helpful error when signed in and trying to accept an invite

quis:
> You’re signed in as test@user.gov.uk. This invite is for another email address. Sign out and click the link again to accept this invite.


mannickutd:
:+1:

PR #421, opened 2016-04-07
Change banners on home page if no permissions

quis:
## Change banners on home page if no permissions

(except view activity, natch)

> No permissions – Remove trial mode banner

> No permissions – Add blue banner saying 'You only have permission to view this service'

## Shorten empty message on choose template

You may have permission to add templates but not to send messages. The previous error message was misleading.


mannickutd:
👍

PR #422, opened 2016-04-07
Don’t allow commas in placeholders

quis:
## Don’t allow commas in placeholders

> If a user tries to save a template containing something like ((name,date)) we should give a validation error.

This is because it causes havoc with the column headers in CSV files.

https://www.pivotaltracker.com/story/show/117043389

image

## Make border width of field layers match textbox

We put a border on text boxes that have errors. This means that they take up more space than the layers above them which don’t have borders.

This makes the layers that highlight the fields in a template misalign.

This commit adds a line to make the borders match, which fixes this.

image

### DIFF
--- a/app/assets/javascripts/highlightTags.js
+++ b/app/assets/javascripts/highlightTags.js
@@ -46,9 +46,10 @@
 
       this.initialHeight = this.$textbox.height();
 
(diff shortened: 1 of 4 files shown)


mannickutd:
👍

PR #424, opened 2016-04-08
Add cancel button to file upload pattern

quis:
We’ve seen users try to click the upload button again when their file doesn’t upload in time. This is because we’re giving them no feedback that something is happening.

I reckon that changing the button to a cancel button is the easiest way to show that something is happening, and also means that (unlike disabling the send button) the user can’t get stuck if the connection drops completely.

upload


mannickutd:
👍

PR #425, opened 2016-04-11
Removed unused files and directory.

servingUpAces:


scripts/run_tests.sh
```diff
@@ -34,6 +34,6 @@ display_result $? 2 "Front end code style check"
 py.test --cov=app --cov-report=term-missing tests/
 display_result $? 3 "Code coverage"
 
-export NOTIFY_ADMIN_ENVIRONMENT='config.Test'
-py.test -v
-display_result $? 4 "Unit tests"
+#export NOTIFY_ADMIN_ENVIRONMENT='config.Test'
```
quis: Did you mean to remove this? It’s an improvement to not run the tests twice, but if that the intent then delete it don’t comment it out.

PR #427, opened 2016-04-11
Update for db rebase

mannickutd:
[DO NOT MERGE] Wait for approval from @servingUpAces

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -92,7 +92,7 @@ def choose_template(service_id, template_type):
     )
 
 
-@main.route("/services/<service_id>/send/<int:template_id>", methods=['GET', 'POST'])
+@main.route("/services/<service_id>/send/<template_id>/csv", methods=['GET', 'POST'])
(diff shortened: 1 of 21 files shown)


app/main/views/send.py
```diff
@@ -92,7 +92,7 @@ def choose_template(service_id, template_type):
     )
 
 
-@main.route("/services/<service_id>/send/<int:template_id>", methods=['GET', 'POST'])
+@main.route("/services/<service_id>/send/<template_id>/csv", methods=['GET', 'POST'])
```
servingUpAces: Why did we change the url here?

mannickutd: Flask was having a hard time matching up the url, with the view method.
Didn't want to hold up the pull request because this was failing.
The change puts it more in line with the other view method names.

quis: You could try `<uuid:template_id>` instead by using https://github.com/wbolster/flask-uuid

PR #428, opened 2016-04-11
Make choose template page clearer

quis:
This pull request is mainly about improving the choose template page, especially for email templates.

It makes a few changes to other pages for consistency’s sake, which are described in the commits.

The main differences can be seen in this demo:

email-toggle


mannickutd:
👍

PR #431, opened 2016-04-12
Sort template stats by usage

quis:
Template stats should show the most-used template first.

This commit:
- re-writes the `aggregate_usage` function to use `itertools.groupby`, which can do aggregation, and can return data in a structure that’s easy to sort on
- uses generators so that we’re not keeping lots of rows of template stats in memory

https://www.pivotaltracker.com/story/show/117348893


mannickutd:
:+1:

PR #434, opened 2016-04-12
Add view for a single template

quis:
> At the moment, we have an all email templates page, and an edit an
> individual page.
>
> This gets messy when we refer to templates like the dashboard and the
> activity views. We solve this currently by using anchor links to the
> list page, but this is clunky.
>
> So lets add it, then update the links on the dash and activity to the
> new view page.
>
> Should be a link from the view a single template page, to the template
> hub page.

https://www.pivotaltracker.com/story/show/117349227


mannickutd:
👍

PR #437, opened 2016-04-13
Don’t lint SCSS files that have come from outside

quis:
We should (and do) keep exact copies of SCSS files that have come from
elsewhere so that we can easily upgrade them. But sometimes they don’t
always pass our linting rules, or throw a lot of warnings, which is
noisy.

This commit:
- moves such files into their own subdirectory
- tells SCSS Lint to ignore files in this directory


mannickutd:
:+1:

PR #438, opened 2016-04-13
Allow filtering of notifications by type, status or both

quis:
activity-filter


mannickutd:
:+1: Code looks good, just fix the tests.

PR #439, opened 2016-04-13
Wrap long words in email template previews

quis:
If you put, for example, a URL in an email template it can be very long.
This can cause it to overflow its container. This commit forces it to
wrap instead.

Before | After 
--- | ---
image | image

https://www.pivotaltracker.com/story/show/117233785


mannickutd:
👍

PR #443, opened 2016-04-13
Make activity page use new nomenclature for status

quis:
Statuses used to be:
- failed
- complaint
- bounce
- sent
- delivered

Now they are:
- sent
- sending
- delivered

This change broke the notifications page on the admin app.

It also made me realise that we should be ignoring ‘sending’ messages in the history page—it should only show messages we’ve tried to deliver.

The code for this is a bit of a bodge, but it will get things working again for now.


mannickutd:
:+1:

PR #445, opened 2016-04-14
Use a different module for doing nice timestamps

quis:
There is a bug in Babel[1] which means that it throws an exception if it can’t find the locale it wants to use.

This is not a problem when developing locally, because OS X comes with lots of locales. The AWS machines, however, only have one locale, which is not the one that Babel is looking for. Hence the ‘Activity’ page doesn’t work.

This commit changes to using Humanize[2] instead, which is much less vast and hopefully less hungry in its requirements.

That’s a morning we won’t get back…

1. https://github.com/python-babel/babel/issues/137
2. https://pypi.python.org/pypi/humanize


mannickutd:
:+1:

PR #447, opened 2016-04-14
Remove human-friendly timestamps from activity page

quis:
The fix introduced by e6490e1 did not fix anything.

So lets remove the ‘3 days ago’ thing completely and just say the date and time instead.

image


mannickutd:
:+1:

PR #451, opened 2016-04-14
Show email subject on choose template page

quis:
Removed this before as part of truncating email previews in
https://github.com/alphagov/notifications-admin/commit/3a5b76ce2a672ba5d42b4013a0274f64742eb5d7#diff-b5f54dc364655c298fd119e3bc148cc6R45

But actually trying to use the app it’s a weird inconsistency to not have the subject show everywhere.

So this commit reinstates it.

***

image


servingUpAces:
:+1:

PR #456, opened 2016-04-15
Put the new documentation in the app

quis:
## Get the docs rendering in the app

This commit replaces the old _API Documentation_ page with the Markdown version that Catherine has been working on.

I’ve checked that there’s nothing obviously wrong or placeholder-y still in there, so I think we’re good to go.

## Add some basic typography, tweak the Markdown

This commit adds some basic type sizes, weights and spacing for HTML elements scoped inside a `.documentation` container.

It also tweaks some of the Markdown so that it renders the same as it does on Github.

***

screencapture-localhost-6012-documentation-1460718700400


servingUpAces:
:+1:

PR #458, opened 2016-04-15
Reverse the order of the filters on activity page

quis:
So that they match the order of the page title, eg ‘Failed text messages’

image


mannickutd:
:+1:

PR #459, opened 2016-04-15
Wrap long words in SMS template previews

quis:
If you put, for example, a URL in an SMS template it can be very long. This can cause it to overflow its container. This commit forces it to wrap instead.

Related: https://github.com/alphagov/notifications-admin/pull/439

image


mannickutd:
:+1:

PR #460, opened 2016-04-15
Update deployment files for aws

servingUpAces:


quis:
This is the same result I got trying to copy these changes from the live branch.

:+1:

PR #463, opened 2016-04-15
Make placeholder hint work for SMS messages

quis:
The changes introduced in 54e42a2021e26785f892a10dcdb0accca4185c4f didn’t get applied to the add/edit SMS template page. This commit fixes that.


mannickutd:
:+1:

PR #464, opened 2016-04-18
Fix innacurate error message on CSV check

quis:
We only know if a row contains missing cell(s), not how many cells are missing data, eg

image

This commit changes the error message to be more accurate.


mannickutd:
:+1:

PR #467, opened 2016-04-19
Put a link to the tour in the footer

quis:
https://www.pivotaltracker.com/story/show/116960703

## Make tour work for non-logged-in users

There’s no content in the tour that’s specific to a service, and we have no good reason to only restrict these pages to users with accounts.

## Make new service the current service on first use

When you add a new service, it’s probably the one you want to do stuff with.

When you get invited, the service you’ve been invited to is probably the one you want to use.

## Put a link to the tour in the footer

So you can:
- see it again
- see it if you don’t have an account yet

***

image

### DIFF
--- a/app/main/views/add_service.py
+++ b/app/main/views/add_service.py
@@ -43,7 +43,7 @@ def add_service():
                                                        user_id=session['user_id'],
                                                        email_from=email_from)
 
-        return redirect(url_for('main.tour', service_id=service_id, page=1))
+        return redirect(url_for('main.tour', page=1))
     else:
         return render_template(
             'views/add-service.html',
--- a/app/main/views/add_service.py
+++ b/app/main/views/add_service.py
@@ -42,7 +42,7 @@ def add_service():
                                                        restricted=True,
                                                        user_id=session['user_id'],
                                                        email_from=email_from)
-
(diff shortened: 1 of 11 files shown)


app/main/views/add_service.py
```diff
@@ -42,8 +42,8 @@ def add_service():
                                                        restricted=True,
                                                        user_id=session['user_id'],
                                                        email_from=email_from)
-
-        return redirect(url_for('main.tour', service_id=service_id, page=1))
+        session['service_id'] = service_id
```
mannickutd: Shouldn't need to do this. The after request function should be doing this.

app/main/views/dashboard.py
```diff
@@ -34,7 +34,8 @@ def service_dashboard(service_id):
 
     if session.get('invited_user'):
         session.pop('invited_user', None)
-        return redirect(url_for("main.tour", service_id=service_id, page=1))
+        session['service_id'] = service_id
```
mannickutd: See comment above.

mannickutd:
Apart from the comments looks good. :+1:

quis:
Talked about the comments IRL. These two cases are maybe the only ones where we can’t rely on `after_request`.

PR #470, opened 2016-04-19
Make the flow of using templates better

quis:
For users who:
- want to send messages from a template
- want to edit templates

For developers:
- who need to get the ID of a template

This commit mainly cleans up the choose template page so there are less options, and the options that are there are less wordy.

This means:
- moving ‘send yourself a test’ onto the send messages page, and making it a button
- stripping a lot of stuff out of the ‘send from API’ page, so it’s more obvious what the template ID is

image

image

image

image


ashimali:
👍

PR #471, opened 2016-04-19
Show the last 7 days of statistics on the dashboard

quis:
**Depends on:**
- [x] **alphagov/notifications-api#241**

***

image

We want to align all our stats to be for the last 7 days.

This means summing up the stats response from the API to make the Big Number. Previously the big number was counting sent notifications as successful. This commit changes it to only look at delivered
notifications.

The upshot of this is that we can link from the dashboard to the activity page when there are failures.


ashimali:
👍

PR #477, opened 2016-04-20
Show template usage for all time on it’s own page

quis:
## Limit templates usage on dashboard to last 7 days

Previous the table of templates on the dashboard was for all time. This commit uses the `limit_days` parameter of the API endpoint to only show template usage from the last 7 days, aligning with the big numbers shown above.

image

## Show template usage for all time on it’s own page

> We show the last weeks template usage on the dashboard, which is great, but if you're looking for longer term trends, you're out of luck.
> 
> So, let's let you see more on a more detailed page (linked from the dashboard). Initially this should just show you all templates that you have used ever and the count for each. Order same as dashboard, most  popular first.

https://www.pivotaltracker.com/story/show/117614585

> In the United Kingdom the financial year runs from 1 April to 31 March for government financial statements.

https://en.wikipedia.org/wiki/Fiscal_year#United_Kingdom

image

## Say ‘sent’ rather than ‘processed’

Hopefully users understand that sent `!=` delivered

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -38,13 +38,8 @@ def service_dashboard(service_id):
 
     return render_template(
         'views/dashboard/dashboard.html',
-        statistics=add_rates_to(
-            statistics_api_client.get_statistics_for_service(service_id, limit_days=7)['data']
-        ),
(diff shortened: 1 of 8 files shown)


mannickutd:
:+1:

PR #488, opened 2016-04-22
Show error if you try to send too many messages

quis:
image

***

https://www.pivotaltracker.com/story/show/117630691

There is a limit of 50 messages per day in trial mode. Right now, we don’t tell you this, we just start failing your messages.

This commit adds an error message if you upload a CSV file that has too many rows in it.

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -22,7 +22,11 @@
         {% if statistics.emails_requested or statistics.sms_requested %}
           <p>
             You can only send {{ current_service.message_limit }}
-            messages per day in <a href="{{ url_for('.trial_mode')}}">trial mode</a>.
+            messages per day
+            {%- if current_service.restricted %}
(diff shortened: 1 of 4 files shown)


app/templates/views/check.html
```diff
@@ -13,7 +13,31 @@
 
 {% block maincolumn_content %}
 
-  {% if errors %}
+  {% if count_of_recipients > (current_service.message_limit - statistics.get('emails_requested', 0) - statistics.get('sms_requested', 0)) %}
```
servingUpAces: This check happens whether a service is restricted or not. If the csv that they load contains more messages then the services.message_limit (currently set to 25000) then they will get a message about being in trial mode and exceeding the limit.

quis: @servingUpAces Removed the trial mode wording for live services in 5f8a5ed

mannickutd:
:+1:

PR #491, opened 2016-04-22
Update the department id so tickets for Deskpro hit the right bucket

yahoopete:


yahoopete:
You beat me to it @servingUpAces!

quis:
👍

PR #493, opened 2016-04-25
Reword homepage

quis:
image

## Remove some words from the home page
So that the proposition is clearer.

## Don’t talk about letters on the homepage

We don’t have letters yet. No point in someone setting up an account hoping they can send letters.


yahoopete:
👍

PR #494, opened 2016-04-25
If not logged in, go to homepage when clicking GOV.UK 

quis:
It’s a bit of a weird experience to be taken to the sign in screen when you click GOV.UK in the header. It’s doubly weird if you take the tour before creating an account, and at the end of the tour you get prompted
to sign in.

This commit adds some extra logic to take you to the homepage instead, which I think is more what you’d expect.

### Also, add test for redirecting to your last-used service

If you click GOV.UK and have:
- multiple services
- a service in your session

…we take you to the dashboard for that service. This works great, but wasn’t tested. This commit adds a test for it.

### DIFF
--- a/app/main/views/choose_service.py
+++ b/app/main/views/choose_service.py
@@ -16,8 +16,11 @@ def choose_service():
 
 
 @main.route("/services-or-dashboard")
-@login_required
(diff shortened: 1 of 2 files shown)


mannickutd:
:+1:

PR #497, opened 2016-04-25
API keys page improvements

quis:
image

## Use the API key pattern for showing service ID

When we show an identifier, like an ID or a key, we have a pattern which adds a ‘copy to clipboard’ button. We weren’t using this pattern for service ID.

This commit also moves service ID underneath the table, so hopefully people will be less likely to confuse the service ID for an API key when scanning down the page.

## Use the same pattern as elsewhere for ‘new thing’

On the team and templates pages we have a pattern for adding a new ‘thing’, which is a green button in the top right.

This commit changes the API key page to follow the same pattern.

### DIFF
--- a/app/templates/views/api-keys.html
+++ b/app/templates/views/api-keys.html
@@ -20,8 +20,12 @@
   </div>
 
   <p>
-    To connect to the API you will need to create an API Key. Each service can have multiple API Keys to allow
-    for test and live environments.
+    To connect to the API you need to create an API Key.
+  </p>
+  <p>
+    Each service can have multiple API Keys. This allows you to integrate a
+    number of systems, each with its own key. You can also have separate
+    keys for your development and test environments.
   </p>
 
   <p>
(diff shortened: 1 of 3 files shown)


app/templates/views/api-keys.html
```diff
@@ -1,15 +1,23 @@
...
+        API keys
+      </h1>
+    </div>
+    <div class="column-one-third">
+      <a href="{{ url_for('.create_api_key', service_id=current_service.id) }}" class="button align-with-heading">Create new API key</a>
+    </div>
+  </div>
 
   <p>
     To connect to the API you will need to create an API Key. Each service can have multiple API Keys to allow
```
yahoopete: We really should say "Each service can have multiple API Keys. This allows you to integrate a number of systems, each with its own key. You can also have separate keys for your development and test environments."

yahoopete:
Comment aside, 👍

PR #499, opened 2016-04-25
Fix the big numbers on the dashboard

quis:
## Make sure the stats client passes params through
This was the cause of the dashboard showing statistics from all time instead of the last 7 days.

## Separate thousands with comma for big numbers
Makes it easier to see just how big the numbers are.

image


mannickutd:
:+1:

PR #506, opened 2016-04-26
Failed login count set to zero with password reset.

mannickutd:


mannickutd:
Demoed to @henryhadlow

PR #507, opened 2016-04-26
Fix anchor tag in flash message.

servingUpAces:
see https://www.pivotaltracker.com/story/show/117513779 for instructions to recreate the bug.


ashimali:
👍

PR #508, opened 2016-04-26
Put a form on the request to go live page

quis:
_N.B. all of the functional changes are in the second commit_

***

When someone click the big green button, we will get a support ticket that looks something like:

```
From Test User <test@user.gov.uk> on behalf of Test Service
(6ce466d0-fd6a-11e5-82f5-e0accb9d11a6)

---

We’ll send about 1000 text messages in the first month, and then 10,000
text messages per month after that. Usage of our service is about 50%
higher in March, at the end of the tax year.
```

***

image


quis:
## Ask specifically for numbers of emails/texts

We want to know this, so hopefully this will prompt users to tell it to us.

image

yahoopete:
:+1:

app/main/views/service_settings.py
```diff
@@ -86,15 +89,47 @@ def _check_password(pwd):
...
 def service_request_to_go_live(service_id):
-    if request.method == 'GET':
-        return render_template(
-            'views/service-settings/request-to-go-live.html'
+
+    form = RequestToGoLiveForm()
+
+    if form.validate_on_submit():
+
+        print(current_service)
```
servingUpAces: print can go

app/main/views/service_settings.py
```diff
@@ -86,15 +89,47 @@ def _check_password(pwd):
...
-    elif request.method == 'POST':
-        flash('Thanks your request to go live is being processed', 'default')
-        # TODO implement whatever this action would do in the real world
+        if resp.status_code != 201:
+            current_app.logger.error(
+                "Deskpro create ticket request failed with {} '{}'".format(
+                    resp.status_code,
+                    resp.json())
+                )
+            abort(500, "Feedback submission failed")
```
servingUpAces: Should the error message be Submission to go live failed?

PR #509, opened 2016-04-26
Update Slack link to point at channel, not archive

quis:
https://ukgovernmentdigital.slack.com/messages/govuk-notify/

not

https://ukgovernmentdigital.slack.com/archive/govuk-notify/


yahoopete:
:+1:

PR #510, opened 2016-04-26
Removed reference to fail over rates on pricing page

yahoopete:


app/templates/views/pricing.html
```diff
@@ -20,22 +20,16 @@ <h2 class="heading-medium">Text messages</h2>
...
       <li>Free allowance: 250,000 text messages per service, per financial year</li>
-      <li>If your text messages are long, it will count as 2 or 3 messages depening on length:
+      <li>If your messages are long, they will count as 2 or 3 text messages, depending on length:
       <ul class="list list-bullet">
-        <li>Up to 160 characters = 1 message</li>
-        <li>Up to 306 characters = 2 messages</li>
-        <li>Up to 459 characters = 3 messages</li>
+        <li>Up to 160 characters = 1 text message</li>
+        <li>Up to 306 characters = 2 text messages</li>
+        <li>Up to 459 characters = 3 text messages</li>
```
servingUpAces: Do we need to let them know messages over 459 characters are not allowed?

PR #511, opened 2016-04-26
Put app environment in support tickets

quis:
## Put URL to service in request to go live ticket

So that:
- we know what environment the request has come from
- a platform admin user can jump straight to that service to check it out

## Put environment URL in feedback tickets

So that:
- we know what environment a request has been sent from


mannickutd:
:+1:

PR #512, opened 2016-04-27
Patch update sections of the page on AJAX

quis:
Currently, when we update a section of the page with AJAX we replace the entire HTML of the section with the new HTML. This causes problems:
- if you’re trying to interact with that section of the page, eg by inpecting it, clicking or hovering an element
- (probably) for screenreaders trying to navigate a page which is changing more than is necessary

This commit replaces the call to `.html()` with a pretty clever library called diffDOM[1]. DiffDOM works by taking a diff of the old element and the new element, then doing a patch update, ie only modifying the parts that have changed.

This is similar in concept to React’s virtual DOM, while still allowing us to render all markup from one set of templates on the server-side.

1. https://github.com/fiduswriter/diffDOM

### DIFF
--- a/app/assets/javascripts/updateContent.js
+++ b/app/assets/javascripts/updateContent.js
@@ -4,6 +4,8 @@
   GOVUK.timeCache = {};
   GOVUK.resultCache = {};
 
+  var dd = new diffDOM();
+
   let getter = function(resource, interval, render) {
 
     if (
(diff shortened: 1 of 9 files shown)


mannickutd:
React style nice!

mannickutd:
:+1:

PR #515, opened 2016-04-27
Show sent, not pending numbers on the dashboard

quis:
The dashboard should only talk about notifications that are *successful* or *failed*. It should not count notifications that are still queued.

This will stop:
- a situation like ‘0 emails, 14.1% failed’
- the big numbers and the template statistics numbers not adding up to the same total


mannickutd:
:+1:

PR #517, opened 2016-04-27
Add events to temp history page for story sign off

ashimali:


quis:
You don’t need any of the `.grid-row` `div`s on this page.

But functionally, 👍

PR #518, opened 2016-04-28
Count only requested and failed messages on dashboard

quis:
Requested, delivered and failed are the three states stored in the notification statistics table. They are not discrete, eg a message can be counted in delivered and failed.

`requested` is incremented **when a notification is created**, and has no chance of being incremented twice for the same notification.

The template statistics are incremented **when a notification is created only**.

Therefore the only way to make the numbers line up is to count:
- messages sent as being `requested`
- failure rate being `failed`/`requested` **not** `failed`/`failed`+`delivered`

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -99,19 +99,14 @@ def add_rates_to(delivery_statistics):
         delivery_statistics
     )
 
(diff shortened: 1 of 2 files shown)


ashimali:
👍

PR #519, opened 2016-04-28
Change activity to view to request a week of data

ashimali:
Cleaned up filters a bit so that if you want both templates email and sms is passed. 

Added mock assertion to tests.

There is pr on api to filter by a day limit. The default with be a week + today (8 days)

https://github.com/alphagov/notifications-api/pull/274


servingUpAces:
👍 demoed and reviewed

PR #520, opened 2016-04-28
Redirect to dashboard if this isn't the first service

pixeltrix:
There's no need to show the tour again if the user has previously created a service so just redirect to the dashboard instead.

https://www.pivotaltracker.com/story/show/116960671


quis:
Checked out and tested locally. 👍

PR #524, opened 2016-04-29
Add usage to the dashboard

quis:
Takes the number of emails and SMS fragments sent from:
https://github.com/alphagov/notifications-api/pull/273

Using these numbers it’s possible to show:
- how much of your allowance is left
- or how much you have spent

For now the allowance and rates are hard coded.

Only for users that have manage service.

image

image

image

image


mannickutd:
Lots of short term fixes here, lots of hard coded values. But if that is to be fixed later than :+1:

PR #527, opened 2016-04-29
Only check on csv is the file extension.

mannickutd:
Only check on csv is the file extension.

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -20,8 +20,8 @@ class CsvFileValidator(object):
         self.message = message
 
     def __call__(self, form, field):
-        if not form.file.data.mimetype == 'text/csv':
-            raise ValidationError(self.message)
+        if not field.data.filename.lower().endswith('.csv'):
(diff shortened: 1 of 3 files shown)


quis:
Tested locally.

Made a little tweak: 
image

👍

app/main/validators.py
```diff
@@ -16,12 +16,12 @@ def __call__(self, form, field):
...
 class CsvFileValidator(object):
 
-    def __init__(self, message='Not a csv file'):
+    def __init__(self, message='{} is not a CSV file'):
         self.message = message
 
     def __call__(self, form, field):
-        if not form.file.data.mimetype == 'text/csv':
-            raise ValidationError(self.message)
+        if field.data.filename.rsplit('.')[1].lower() != 'csv':
```
quis: This gives an `IndexError` if you upload a file with no extension.

I think `field.data.filename.lower().endswith('.csv')` would work instead.

mannickutd: very true

quis:
👍

PR #528, opened 2016-04-29
Link up the last 7 days

quis:
## Make big numbers on the homepage linkable

The big numbers on the homepage relate directly to the notifications on the notification page. So let’s link them. With a _hyper_ link.

This commit actually adds two links, one of which is semantically correct, and one of which is visually correct, ie makes the whole black area of the box clickable/hoverable.

## Fix page titles on activity page

We were getting some weirdness like ‘Failed both’.

This commit fixes the problem, and adds some tests for the page headings to make sure they don’t break again.

***

big-clicks

## Remove activity from the nav

We saw lots of people in the lab clicking activity hoping for… something. But it’s not really where you go to do a thing, so they weren’t finding what they were looking for.

Since you can now get to the activity from the dashboard, let’s remove the link in the nav, to make thing less ambiguous.

image

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -16,13 +16,38 @@
   @extend %big-number;
   background: $text-colour;
   color: $white;
+  position: relative;
 
   .big-number {
     padding: 15px;
+    position: relative;
+  }
+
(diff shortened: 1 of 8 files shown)


mannickutd:
:+1:

PR #533, opened 2016-05-03
Rename ‘this year’

quis:
Make it clear that this section is about what you’ve used, ie your usage

Click ‘breakdown’ to see how the maths is done, and what time period it covers.


yahoopete:
👍

PR #536, opened 2016-05-04
Ignore case and spacing in CSV column headers

quis:
Implements: https://github.com/alphagov/notifications-utils/pull/28


minglis:
👍

PR #539, opened 2016-05-04
Updated flask-login to version 0.3.2

ashimali:
It took so long as lots of tests broke and it all boiled down to turning off session protection for dev and test configs. I wish I'd known that all along.


mannickutd:
:+1:

PR #541, opened 2016-05-05
Introduce replacement of placeholders to send yourself a test

quis:
Hopefully, by getting the user to replace their placeholders when they send themselves a test message, it will reinforce the idea of how placeholders work.

Full changes detailed in the commits.

***

image

***

image


mannickutd:
:+1:

PR #552, opened 2016-05-11
Added provider management pages in.

minglis:
Depends on: https://github.com/alphagov/notifications-api/pull/295

Adds platform admin pages to mange priority of the providers.

- list all providers
- update the priority

### DIFF
--- a/app/main/views/providers.py
+++ b/app/main/views/providers.py
@@ -1,0 +1,40 @@
+from flask import (
+    render_template, url_for)
+
+from flask_login import (
+    login_required,
+)
+from werkzeug.utils import redirect
+from app.main import main
+from app.main.forms import ProviderForm
+from app.utils import user_has_permissions
+from app import provider_client
+
+
+@main.route("/providers")
+@login_required
+@user_has_permissions(admin_override=True)
+def view_providers():
+    providers = provider_client.get_all_providers()['provider_details']
+    email_providers = [email for email in providers if email['notification_type'] == 'email']
+    sms_providers = [sms for sms in providers if sms['notification_type'] == 'sms']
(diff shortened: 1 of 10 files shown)


app/main/views/providers.py
```diff
@@ -0,0 +1,40 @@
...
+@main.route("/provider/<provider_id>", methods=['GET', 'POST'])
+@login_required
+@user_has_permissions(admin_override=True)
+def view_provider(provider_id):
+
+    provider = provider_client.get_provider_by_id(provider_id)['provider_details']
+
+    form = ProviderForm(active=provider['active'], priority=provider['priority'])
+
+    print(form)
```
mannickutd: Please remove print statements.

tests/app/main/views/test_providers.py
```diff
@@ -0,0 +1,263 @@
...
+        {
+            'active': True,
+            'priority': 2,
+            'display_name': 'second_email_provider',
+            'identifier': 'second_email',
+            'id': '0bd529cd-a0fd-43e5-80ee-b95ef6b0d51b',
+            'notification_type': 'email'
+        }
+    ]
+}
```
mannickutd: These would make great functions in the tests/init.py
def provider_json(<include arguments>)

Bonus for having these as test fixtures.

app/main/views/providers.py
minglis: doh

minglis: fixed

tests/app/main/views/test_providers.py
minglis: Thinking of leaving these as these are only relevant to these tests?

PR #557, opened 2016-05-11
Notifications display updated_at instead of created_at.

mannickutd:


app/templates/partials/jobs/notifications.html
```diff
@@ -27,7 +27,7 @@
         align='right',
         status='error' if item.status == 'Failed' else 'default'
       ) %}
-        {{ item.status|title }} at {{ item.sent_at|format_time }}
+        {{ item.status|title }} at {{ item.updated_at|format_time }}
```
minglis: Is there a test somewhere that checks we're rendering the correct fields?

mannickutd: Yep added check now in existing test.

PR #558, opened 2016-05-12
Make template usage on dashboard easier to scan

quis:
image

The dashboard looked a bit table-y. This commit makes four main changes:

- show a bar chart (drawn in CSS) for template usage (only shown if you’ve used more than one template recently)
- only break down template usage by template name, not template type (because that’s happening with the big numbers)
- change the style of the ‘show more’ links under each section so that they are all consistent, and a little less busy (one less keyline)
- remove the ‘recent templates‘ title so that the first two sections of the page group under ‘in the last 7 days’

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -1,19 +1,29 @@
 %big-number,
 .big-number {
 
-  @include bold-48;
+  display: block;
+
+  &-number {
+    @include bold-48;
+    display: block;
+  }
 
   &-label {
     @include core-19;
-    display: block;
-    padding-bottom: $gutter-half;
+    display: inline-block;
+    padding-bottom: 10px;
   }
 
 }
 
 .big-number-smaller {
+
   @extend %big-number;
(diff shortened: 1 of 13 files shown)


yahoopete:
Functionally :+1:

henryhadlow:
Few things:

- Can we have `<h2 class="heading-small">Templates</h2>` above the Templates module
- If there's only one template, it should show the module with one template listed and a volume, but no bar
- We need to say if the templates are email or text message (grey type?)

quis:
## Include template type in bar graph label

It’s possible that users will have email and SMS templates with similar names, and will send them depending on their users’ contact preferences.

So it’s useful to be able to compare how many emails vs SMS you’re sending, even if the template names are similar.

template-bars


## Show template stats even if there’s only 1 row

If you’re only ever sending one template it’s really useful to be able to jump straight to that template from the dashboard. So this commit:
- shows the template stats even if there’s only one row
- hides the bar chart if there’s only one row (because it will always be 100%, and won’t be obvious what it is without its siblings)

image

PR #560, opened 2016-05-12
Give the <code> example in docs better CSS

quis:
image

***

- adds a background colour, to words wrapped in `<code>` tags, like we
  have for whole snippets of code
- reduces the font size of all code blocks
  a) to differentiate them further
  b) to fit more on the screen


yahoopete:
👍

PR #565, opened 2016-05-16
Accept common spreadsheet formats, not just CSV

quis:
### This is the non-spikey version of #544

***

We require users to export their spreadsheets as CSV files before uploading them. But this seems like the sort of thing a computer should be able to do.

So this pull request adds a wrapper class which:
- takes a the uploaded file
- normalises the line breaks (CSV and TSV only) or reads it using [Pyexcel](https://github.com/pyexcel/pyexcel)
- gives the data back in CSV format

This allows us to accept `.csv`, `.xlsx`, `.xls` (97 and 95), `.ods`, `.xlsm` and `.tsv` files. We can upload the resultant CSV just like normal, and process it for errors as before.

Testing
---

To test this I’ve added a selection of common spreadsheet files as test data. They all contain the same data, so the tests look to see that the resultant CSV output is the same for each.

Error handling
---

There shouldn’t be a case where we see a `ValueError` on upload any more. Our file handling should be robust enough to deal with whatever is thrown at it.

So to make sure of this, this pull request:
- adds test files with bad data (PNG files with their extensions changed to look like spreadsheets)
- catches whatever exceptions are raised by trying to parse these files
- returns a helpful flash message to the user

Anything else should raise a `500`, eg if the file can’t be uploaded to S3.

UI changes
---

This commit doesn’t change the UI, apart from to give a different error message if a user uploads a file type that we still don’t understand.

I intend to do this as a separate pull request, in order to fulfil https://www.pivotaltracker.com/story/show/119371637


minglis:
👍 saw this. looks fine

PR #578, opened 2016-05-17
Convert timezones using pytz.

minglis:


quis:
Seen working on Martyn’s machine, 👍

PR #579, opened 2016-05-17
On page load, hide the instructions for uploading a file

quis:
details

***

> On the send pages, we have a call to action in the middle of the page with lots of words and placeholder buttony looking things above and below it....
>
> You'll also see this every time you use this page, even though you probably get it after a single use.
>
> So let's wrap it up under a usefully titled (?) help link that expands to reveal all the things.

This commit implements the above.

It also changes the words to talk about various spreadsheet formats, not just CSV, now that https://github.com/alphagov/notifications-admin/pull/565 is merged.

## Add polyfill for HTML5 `<details>` element

The details element is fairly new. It doesn’t work in all browsers, and isn’t keyboard/screenreader accessible even in some new browsers.

This commit adds the Polyfill script [from GOV.UK elements](http://govuk-elements.herokuapp.com/typography/#typography-hidden-text) which polyfills these features for browsers that don’t support them.

***

https://www.pivotaltracker.com/story/show/119371637

### DIFF
--- a/app/assets/javascripts/detailsPolyfill.js
+++ b/app/assets/javascripts/detailsPolyfill.js
@@ -1,0 +1,198 @@
+// From
+// https://github.com/alphagov/govuk_elements/blob/4926897dc7734db2fc5e5ebb6acdc97f86e22e50/public/javascripts/vendor/details.polyfill.js
+//
+// ---
+//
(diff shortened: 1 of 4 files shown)


yahoopete:
Gold :+1:

PR #582, opened 2016-05-17
Make ‘Too many recipients’ error lower priority

quis:
image

This commit makes two main changes to what happens when a user is in trial mode and they upload some email addresses belonging to other people.

1. Add a specific banner error telling the user about trial mode

2. Make this error higher priority, eg it will show up before the  error about having too many recipients in your file

This means making some changes to the tests so that the example CSV files include the user’s phone number, then making them invalid by omitting data required by the templates.

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/34

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -13,31 +13,20 @@
 
 {% block maincolumn_content %}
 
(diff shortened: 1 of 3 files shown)


mannickutd:
:+1:

PR #594, opened 2016-05-20
Setting the style for the flash message on feedback page

yahoopete:
Was using the default of dangerous - which it shouldn't. Obvs.


quis:
Locally: 
image

quis:
Something more active maybe?

> We’ve received your feedback

Your call, 👍 either way.

PR #596, opened 2016-05-20
Use correct key to set team in DeskPro

quis:
For the `POST` ticket endpoint, the parameter is `agent_team_id`:

https://manuals.deskpro.com/html/developer-apps/api/api-tickets.html#id2

***

We should also rename the Credstash key… _later_.


mannickutd:
:+1:

PR #598, opened 2016-05-20
Explain why we need your phone number

quis:
The previous hint text was sort of half way there. Told you that you’d need your phone handy, but not _why_. This came up multiple times in user research. Also, the term ‘handy’ is not as familiar to users as
I anticipated.

This commit changes the message to explain exactly what we use your phone number for.

***

image


mannickutd:
:+1:

PR #601, opened 2016-05-23
Download csv for job added.

mannickutd:
Download button now available on the job page.
Allows you to only download notifications for that job.


leohemsted:
👍

PR #602, opened 2016-05-23
delete template button POSTs archived:true

leohemsted:
i notice we weren't even using `_attach_current_user` beforehand so killed it but idk if we want to record who archived a template?

Yes we do want to know who archived the template.


mannickutd:
Yes we do want to know who archived the template.

leohemsted:
@mannickutd updated

mannickutd:
:+1:

PR #603, opened 2016-05-24
Hide ‘created at…’ on the templates page

quis:
I don’t think that if there’s only one version of the template that it’s useful to see the created at date.

The auditing stuff only becomes relevant once someone a template has been changed.


mannickutd:
:+1:

PR #604, opened 2016-05-24
Improve the ‘download a CSV’ from the job page

quis:
## Put ‘download CSV’ on job in logical place

image

## Give the job CSV a helpful filename

If you’re downloading a bunch of reports from your jobs then it’s useful to be able to differentiate between them. This commit makes it easy to do so by naming the file with:
- the name of the template
- when the job was created


mannickutd:
:+1:

PR #608, opened 2016-05-24
New platform admin page

leohemsted:
image

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -1,0 +1,14 @@
+from flask import render_template
+from flask_login import login_required
+
+from app.main import main
+from app.utils import user_has_permissions
+
+
+@main.route("/platform-admin")
+@login_required
+@user_has_permissions(admin_override=True)
+def platform_admin():
+    return render_template(
+        'views/platform-admin.html'
+    )
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -46,6 +46,11 @@
             <li>
               <a href="{{ url_for('main.choose_service') }}">Switch service</a>
             </li>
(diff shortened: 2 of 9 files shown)


app/main/views/platform_admin.py
```diff
@@ -0,0 +1,15 @@
+from flask import render_template, abort
+from flask_login import login_required, current_user
+
+from app.main import main
+
+
+@main.route("/platform-admin")
+@login_required
+def platform_admin():
+    if not current_user.platform_admin:
```
mannickutd: There is a decorator for permissions.
If that doesn't have the functionality can you please update it.

app/templates/admin_template.html
```diff
@@ -46,6 +46,11 @@
             <li>
               <a href="{{ url_for('main.choose_service') }}">Switch service</a>
             </li>
+            {% if current_user.platform_admin %}
```
mannickutd: Same here there is a permission method which we use to check if users have permissions for links.

tests/app/main/views/test_platform_admin.py
```diff
@@ -0,0 +1,34 @@
...
+        with app_.test_client() as client:
+            response = client.get(url_for('main.platform_admin'))
+            assert response.status_code == 302
+            assert url_for('main.index', _external=True) in response.location
+
+
+def test_should_403_if_not_platform_admin(app_, active_user_with_permissions, mocker):
+    with app_.test_request_context():
+        with app_.test_client() as client:
+            mocker.patch('app.user_api_client.get_user', return_value=active_user_with_permissions)
```
mannickutd: There are already mocks for get_user, where possible and appropriate it is good to use the py.test fixtures

mannickutd:
:+1:

PR #609, opened 2016-05-24
Put uploaded files on the dashboard

quis:
image

This commit depends on and uses the data returned by:
- [x] https://github.com/alphagov/notifications-api/pull/345
- [ ] a pull request which will allow us to limit this data to the last 7 days

It puts the jobs on the dashboard.

It also:
- links to the jobs page
- makes the numbers on the jobs page consistent with the dashboard
- makes the numbers on an individual job consistent with the appearance of the dashboard

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -202,7 +202,7 @@ def format_datetime_normal(date):
 
 
 def format_datetime_short(date):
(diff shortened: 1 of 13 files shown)


mannickutd:
:+1:

quis:
Closing this because I forgot we don’t have the failure rate in yet.

PR #613, opened 2016-05-25
Make the tour interactive

quis:
_The code for this is quite hacky and light on tests. But I’d really like to get
it in the app for the research tomorrow to see how well the feature works._

This commit changes the tour from being a set of static screens to some help
which guides you through the process of sending your first test message.

The theory behind this is that what users are really struggling with is the
concept of a variable, rather than the relationship between the placeholders and
the column headers. And like learning to program, the best way to learn is by
taking an example and modifying it to your own needs.

This means that when someone adds their first service we set them up an
example email template and an example text message template. Then there is a
guided, three step process where _all_ the user can do is send a test message to
themselves.

Once the message is sent, the user still has the example templates which they
can edit, rather than having to remember what they’re supposed to be doing.

***

image

***

image

***

image

***

image


app/main/views/add_service.py
```diff
@@ -44,11 +44,29 @@ def add_service():
                                                        email_from=email_from)
         session['service_id'] = service_id
 
-        services = service_api_client.get_services({'user_id': session['user_id']}).get('data', [])
-        if (len(services) > 1):
+        if (len(service_api_client.get_services({'user_id': session['user_id']}).get('data', [])) > 1):
```
mannickutd: Looks like a java if statement

mannickutd:
Tested the code locally and it seems to work fine.

mannickutd:
:+1:

PR #614, opened 2016-05-25
Put uploaded files on the dashboard

quis:
image

This commit depends on and uses the data returned by:
- [x] alphagov/notifications-api#345
- [x] alphagov/notifications-api#347
- [x] #612

It:
- puts jobs from the last 7 days on the dashboard
- links to the jobs page
- makes the numbers on the jobs page consistent with the dashboard
- makes the numbers on an individual job consistent with the appearance of the dashboard

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -202,7 +202,7 @@ def format_datetime_normal(date):
 
 
 def format_datetime_short(date):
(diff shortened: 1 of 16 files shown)


minglis:
👍

PR #620, opened 2016-05-26
Remove the get started banner from the dashboard

quis:
We cover this stuff in the tour now. It can also show up if you delete all your templates.


mannickutd:
:+1:

PR #624, opened 2016-05-27
Fix 403s when viewing a job as platform admin

quis:
While battling the bank holiday traffic on my way home this evening I realised why looking at the job page was causing lots of 403 errors in the Chrome console (a spike which was reflected on the Graphite screen).

There is a separate `.json` endpoint for getting updates to a job. This endpoint did not have the `admin=True` option set in its permissions decorator. So when viewing a service as a platform admin you could see the page, but not receive any AJAX-powered updates to it.


mannickutd:
:+1:

PR #625, opened 2016-05-31
Job notification download now includes all notifications.

mannickutd:


quis:
Seen working on @mannickutd’s machine 👍

PR #630, opened 2016-05-31
Platform admin page - per service stats

leohemsted:
headline stats are global, table shows stats per service (since the beginning of the day) image


quis:
👍

PR #631, opened 2016-05-31
Warn users when a template change is going to break things

quis:
image

When a user adds or removes placeholders in their template we should consider this a ‘breaking change’ and warn them accordingly.

Implementing this mostly relies on using alphagov/notifications-utils#37

I’m temporarily storing the new template until the user confirms that they want to make the changes in done using hidden fields. This is a bit hacky, but the complexity of making sessions interact with WTForms was just too much to handle.

***

This commit also makes some styling changes to various tables, details in the commits.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1,4 +1,4 @@
-import string
+from string import ascii_uppercase
 
 from flask import request, render_template, redirect, url_for, flash, abort, session
 from flask_login import login_required
@@ -127,7 +127,7 @@ def edit_service_template(service_id, template_id):
                 'views/templates/breaking-change.html',
                 template_change=template_change,
                 new_template=new_template,
-                column_headings=list(string.ascii_uppercase[:len(new_template.placeholders) + 1]),
+                column_headings=list(ascii_uppercase[:len(new_template.placeholders) + 1]),
                 example_rows=[
                     [first_column_heading[new_template.template_type]] + list(new_template.placeholders),
                     get_example_csv_rows(new_template),
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1,12 +1,16 @@
-from flask import request, render_template, redirect, url_for, flash, abort
+import string
+
+from flask import request, render_template, redirect, url_for, flash, abort, session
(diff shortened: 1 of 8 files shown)


leohemsted:
not sure how much we care but if you change the case of the placeholder (eg `((name))` to `((NAME))`) it complains despite placeholders being case insensitive

quis:
Discussed IRL, case sensitivity is something we should fix, but not in this PR (needs a utils update).

app/main/views/templates.py
```diff
@@ -109,14 +113,36 @@ def edit_service_template(service_id, template_id):
     form = form_objects[template['template_type']](**template)
 
     if form.validate_on_submit():
+        subject = form.subject.data if getattr(form, 'subject', None) else None
```
leohemsted: it's a bit more correct to use `hasattr` rather than `getattr`

quis: Changed in acfe87e

app/templates/views/templates/breaking-change.html
```diff
@@ -0,0 +1,72 @@
+{% extends "withnav_template.html" %}
+{% from "components/banner.html" import banner_wrapper %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/table.html" import list_table, text_field, index_field, index_field_heading %}
+
+{% macro list_of_placeholders(placeholders, oxford_comma=False) %}
+  {% for placeholder in placeholders %}
+    {% if loop.last and loop.length > 1 %}{% if oxford_comma %},{% endif %} and {% endif %}
+    <span class="placeholder">(({{ placeholder }}))</span>{% if not loop.last and loop.length != 2 %},{% endif %}
```
leohemsted: oxford commas not working (`You added ((A)), ((B)), and ((C))`)

leohemsted:
👍  apart from minor comment about oxford commas - if you dont care then feel free to just merge (but i'd rather we kill the oxford comma line entirely since if i edit template and turn it on on it prints two commas 
image
)

quis:
Removed the commas in 53ec44a

image

PR #640, opened 2016-06-03
Change wording around the example on the send page

quis:
‘How to format your file’ sounds complicated and incidental.

‘See an example’ sounds like something that might be useful if you’re feeling stuck.

Also makes the list of file formats a list, because it’s a list.

***

image


servingUpAces:
👍

PR #641, opened 2016-06-03
Added invite user button for platform admins on any service

minglis:
Implements: https://www.pivotaltracker.com/story/show/120678143

Puts invite user button on any service for platform admins, so we can manage the use case that a service loses all logins (people leave / move on / domain changes etc)

Also allows us to manage the notify root service that is coming.


tests/app/main/views/test_manage_users.py
```diff
@@ -273,6 +271,7 @@ def test_no_permission_manage_users_page(app_,
             client.login(api_user_active, mocker, service_one)
             response = client.get(url_for('main.manage_users', service_id=service_one['id']))
             resp_text = response.get_data(as_text=True)
+            print(resp_text)
```
servingUpAces: Oops - print statement left in

servingUpAces:
👍 
I tried it out and it works.

PR #646, opened 2016-06-03
Prioritise CSV errors to help you match placeholders to column headers

quis:
There’s more detail in the commits, but broadly this is what’s changing:

## The order of errors is (only one displayed at a time):

1. No phone number/email column
2. Column headers don’t match placeholders
3. Missing or bad data
4. Trial mode
5. Daily limit reached

## Row-level errors are only shown if all the columns are there

Row-level errors are:

- bad phone number/email address
- missing data

I think it’s distracting to show these on the page if there’s something more fundamentally wrong with the file, eg placeholders don’t match.

***

image

***

image

***

image

***

image

***

image


mannickutd:
:+1:

PR #647, opened 2016-06-06
Make placeholders look more consistent and editable

quis:
Depends on:
- [x] alphagov/notifications-utils#40

***

In research we’ve noticed two problems with the appearance of placeholders:

1. We are inconsistent about when we display the ((double brackets)). Sometimes we show them, sometimes we don’t. This doesn’t help user’s understanding about where the column names in their CSV file come from, or how they can edit the template to fix any errors.

2. Because they look so different from normal `<textarea>` text, it’s not immediately obvious that they can be edited just like normal text. They look more like something that can be dragged/inserted.

So this commit:

1. Makes the brackets always-visible.

2. Makes the text colour of the placeholder `$text-colour`, and only highlights the name of the ‘variable’, not the brackets themselves. This means using a different colour (yellow) for the background, because the colour contrast between `$light-blue` and `$text-colour` does not pass WCAG AAA.

***

yellow-p

image

image

image

***

image

***

image

### DIFF
--- a/app/assets/javascripts/highlightTags.js
+++ b/app/assets/javascripts/highlightTags.js
@@ -15,16 +15,14 @@
         .wrap(`
           <div class='textbox-highlight-wrapper' />
         `)
-        .after(this.$backgroundMaskForeground = $(`
+        .after(this.$background = $(`
           <div class="textbox-highlight-background" aria-hidden="true" />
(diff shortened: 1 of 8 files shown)


yahoopete:
Functionally :+1:

mannickutd:
:+1:

PR #648, opened 2016-06-06
Redirect to single template view after edit

quis:
When you edit a template, you’re probably going to do something with it straight afterwards, eg send yourself a test.

We could make it easy to find the template you’ve just saved by putting it at the top of the pile, but this gets confusing for other reasons (order of templates will constantly shift about).

So this commit changes the flow to take you to the single page for the template you’ve just edited.


mannickutd:
:+1:

PR #652, opened 2016-06-06
Expand permissions before post to api

ashimali:
Expand permissions to all possible values on admin before posting to api. 

For example instead of posting 'send_messages' to be stored against invited user, post 'send_emails', 'send_texts', 'send_letters'. 

As the latter values are save from the outset against the invite user, then when the user is created, no mapping needs to be done.

This makes admin template work for both existing and invited users without change.

This fixes:
https://www.pivotaltracker.com/story/show/120829927

This PR will enable this following on API.
https://github.com/alphagov/notifications-api/pull/384

### DIFF
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -1,5 +1,4 @@
 from itertools import chain
-from collections import OrderedDict
 from flask import (
     request,
     render_template,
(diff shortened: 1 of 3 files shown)


servingUpAces:
👍

ashimali:
Will merge on 07.06.2016 as user testing happening at the moment and this needs to be deployed to all environments.

PR #654, opened 2016-06-06
Fix 500 on job page when notifications are sending

quis:
Sending notifications don’t have an `updated_at`. This causes the time formatting to throw a wobly, because it doesn’t expect `None`.

This commit changes the template to also look for the `created_at`, which all notifications have.


servingUpAces:
👍

PR #655, opened 2016-06-06
Two factor page bug

ashimali:
In case user details were not in session the redirect did not use url_for to redirect to sign in.

This fixes: https://www.pivotaltracker.com/story/show/120823303


servingUpAces:
👍

PR #656, opened 2016-06-06
Bug fixed when sending yourself a test the back button links to the c…

mannickutd:
…orrect url.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -205,7 +205,6 @@ def send_from_api(service_id, template_id):
 @login_required
 @user_has_permissions('send_texts', 'send_emails', 'send_letters')
 def check_messages(service_id, template_type, upload_id):
-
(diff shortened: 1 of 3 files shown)


app/main/views/send.py
```diff
@@ -205,71 +205,75 @@ def send_from_api(service_id, template_id):
...
+            ),
+            original_file_name=session['upload_data'].get('original_file_name'),
+            upload_id=upload_id,
+            form=CsvUploadForm(),
+            statistics=statistics,
+            back_link=back_link
+        )
+    except:
+        import traceback
+        traceback.print_exc()
```
quis: Is this meant to be here or just for debugging?

mannickutd: dammit rookie error

quis:
👍

PR #657, opened 2016-06-07
Make activity consistent

quis:
image

***

_This pull request does a fair bit of refactoring (in separate commits) but the main functional changes are:_

## Remove template type filter from activity

This commit splits the activity page into two pages, one for emails and one for SMS.

Technically this means moving from having template type in the querystring and putting in it the URL, eg:

Before | After
--- | ---
`/services/abc/notifications/?template_type=sms` | `/services/abc/notifications/sms`

## 	Add filters for ‘processed’ and sending states  …

- _Processed_ is all the notifications that we know about, ie sending, failed and delivered (can come up with a better word for this later)

- _Sending_ is notifications that we have either put into a queue or are waiting to hear back from the provider about.

The big numbers on the dashboard are a count of all the messages we’ve processed. So when you click them, the table of notifications you see on the dashboard should contain that number of notifications.

This also gets the activity page one step closer to being like the job page:

         | Before                     | After
---------|----------------------------|---------------------------------
Activity | Sending, failed, both      | Processed, sending, failed, delivered
Job page | Sending, failed, delivered | Sending, failed, delivered

***

_Also, here’s how to make sure you’re writing nice atomic commits:_

image

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -72,20 +72,6 @@ def view_job(service_id, job_id):
                                                        version=job['template_version'])['data']
     notifications = notification_api_client.get_notifications_for_service(service_id, job_id)
     finished = job['status'] == 'finished'
-    if 'download' in request.args and request.args['download'] == 'csv':
-        csv_content = generate_notifications_csv(
-            notification_api_client.get_notifications_for_service(
-                service_id,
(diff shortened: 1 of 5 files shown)


mannickutd:
:+1:

PR #661, opened 2016-06-08
Change when invite gets marked as accepted.

ashimali:
Previously it was done on sign in but can be done as soon as invite link is clicked and user is added to service.


mannickutd:
:+1:

PR #673, opened 2016-06-13
Link back to all template versions from 1 version

quis:
If you get linked to a single version of a template, you’re at a dead end. Let’s add a link to go back up a level to where you can understand the current version in context.

image


leohemsted:
👍

PR #674, opened 2016-06-13
Improve AJAX behaviour when server is slow/erroring

quis:
## Stop AJAX updates queuing requests

The `updateContent` module updates a section of the page based on an AJAX request.

The current implementation will poll every `x` seconds. If the server takes a long time to respond, this results in a the browser queuing up requests. If a particular endpoint is slow, this can result in one
client making enough requests to slow down the whole server, to the point where it gets removed from the loadbalancer, eventually bringing the whole site down.

This commit rewrites the module so that it queues up the render operations, not the requests.

There is one queue per endpoint, so for `http://example.com/endpoint.json`:

### 1. Queue is empty
```javascript
{
  'http://example.com/endpoint.json': []
}
```

### 2. Inital re-render is put on the queue…
```javascript
{
  'http://example.com/endpoint.json': [
    function render(){…}
  ]
}
```

### …AJAX request fires
```
GET http://example.com/endpoint.json
```

### 3. Every `x` seconds, another render operation is put on the queue
```javascript
{
  'http://example.com/endpoint.json': [
    function render(){…},
    function render(){…},
    function render(){…}
  ]
}
```

### 4. AJAX request returns queue is flushed by executing each queued render function in sequence
```javascript
render(response); render(response); render(response);
```
```javascript
{
  'http://example.com/endpoint.json': []
}
```

### 5. Repeat

This means that, at most, the AJAX requests will never fire more than once every `x` seconds, where `x` defaults to `1.5`.

## Handle AJAX errors when updating content

With the change in implementation above, any error (eg server responds with `500`) would cause the page to not be updated again.

This is better than the previous implementation, whereby the browser would re-request as fast as it could until it got a successful response.

This commit handles errors by clearing the render queue if the server returns an error. So:

- Any updates that would have been performed based on this request are dropped
- Subsequent updates will be attempted as if it was the first load of the page, ie after a delay of `x` seconds

## 	Revert "Comment out ajax queries to update content on frontend."

This reverts commit 651584d.

Should be safe to turn the AJAX back on now it’s not going to denial-of-service any slow pages.

### DIFF
--- a/app/assets/javascripts/updateContent.js
+++ b/app/assets/javascripts/updateContent.js
@@ -17,11 +17,16 @@
     while(queue.length) queue.shift()(response);
   };
 
+  var clearQueue = queue => (queue.length = 0);
+
   var poll = function(renderer, resource, queue, interval) {
 
-    if (queue.push(renderer) === 1) $.get(
-      resource,
+    if (queue.push(renderer) === 1) $.ajax(
+      resource
+    ).done(
       response => flushQueue(queue, response)
+    ).fail(
+      () => clearQueue(queue)
     );
 
     setTimeout(
--- a/app/assets/javascripts/updateContent.js
+++ b/app/assets/javascripts/updateContent.js
@@ -1,48 +1,44 @@
-(function(GOVUK, Modules) {
+(function(Modules) {
   "use strict";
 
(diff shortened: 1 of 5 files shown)


leohemsted:
👍

PR #675, opened 2016-06-13
Fix table display

quis:
## Fix display of spreadsheet tables

a9f79bc made all tables have a `fixed` layout. This causes issues with the spreadsheet-looking tables.

This commit treats tables with half-width first columns as the exception, not the rule, and makes other tables display as before.

Before | After
--- | ---
image | image

## Make job counters smaller

With sending, delivered and failed all on one line there’s not much space. When these numbers get relatively big (in the 000s) they can start mushing into each other.

This commit makes them smaller so that they remain separate.

image

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -26,6 +26,16 @@
 
 }
 
+.big-number-smallest {
+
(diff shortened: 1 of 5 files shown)


leohemsted:
👍

PR #681, opened 2016-06-13
Make dashboard and activity pages consistently clickable

quis:
This pull request is mostly cosmetic changes. Details of each are in the commits, but the end result is pages that look like this:

image

image

image


ashimali:
👍

PR #682, opened 2016-06-14
Resend verification code

ashimali:
This PR relates to https://www.pivotaltracker.com/story/show/120370831

Requesting a resend of verify code for a user that has not completed registration will allow user to check and modify mobile number.

Registered (active) users will only be able to request resend to their existing registered number.


servingUpAces:
👍  Tried it locally and it works a treat

PR #685, opened 2016-06-14
fix bug - calling wrong template_statistic endpoint causing 500 error…

leohemsted:
… on delete

- also unrolled a test with a for loop into a parametrized test


ashimali:
👍

PR #686, opened 2016-06-14
Stop page jumping on first load with a long email

quis:
Before | After
--- | ---
jump | jump-after

A long email message needs to be collapsed to only show the first few lines. The problem is that we were doing this by adding a class with Javascript, meaning that the email wasn’t being collapsed until the
script in the footer ran.

This caused a jump in the page because the browser was painting the whole email message, then repainting it once it was collapsed.

This commit takes advantage of the `.js-enabled` class added to the `<body>` by a script in the `<head>` of GOV.UK template.

This means that the email message is collapsed with CSS before the first paint of the page, so no jump.

This introduces some complexity in how we determine which emails get the expander toggle. Because they’re already collapsed we can’t get their height and work out if they’re long enough to need collapsing.

So we need to take a copy of the message, put it off-screen, expand it, get its height, then remove it from the DOM. Bit of a faff.

Because of this there’s still a quick flash of the toggle if you see an email message that’s too short to need collapsing. I think this is the lesser of two evils—very short email messages will be few and far
between in the real world.

### DIFF
--- a/app/assets/javascripts/expandCollapse.js
+++ b/app/assets/javascripts/expandCollapse.js
@@ -7,20 +7,16 @@
 
       this.$component = $(component);
 
-      if (this.$component.height() < this.$component.data('max-height')) return;
-
-      this.$component      
-        .append(`
-          <div class='toggle' tabindex='0'>...<span class='visually-hidden'>show full email</span></div>
-        `)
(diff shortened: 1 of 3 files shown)


servingUpAces:
👍  demoed and reviewed

PR #687, opened 2016-06-14
Highlight failing jobs on the dashboard

quis:
> When we have jobs that have over 3% failure rates we should highlight
> those so that peoples attention is drawn to deal with the failure.
>
> They would then go to the job view to see what the details are where
> they could filter by failure, but that's a different story...
>
> This is just about calculating and highlighting those that need their
> attention.

— https://www.pivotaltracker.com/story/show/121206123

This commit:

- calculates the failure rate for each job
- makes jobs with a failure rate of > 3% go red on the dashboard

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -19,7 +19,7 @@ from app import (
     service_api_client,
     template_statistics_client
 )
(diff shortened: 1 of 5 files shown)


leohemsted:
👍

PR #692, opened 2016-06-16
Python 3.4 unpacking

quis:
## Fix unpacking syntax for Python 3.4

Python 3.5 introduced the ability to do unpacking before keyword arguments, eg
```python
my_method(**dictionary, foo=bar)
```

This was introduced in https://www.python.org/dev/peps/pep-0448/

This is not valid syntax in Python 3.4, which is what’s running on our boxes.

This commit changes an instance of this syntax which was introduced in 56d9c29

## Make Travis use same Python version as our boxes

image

Our boxes are running Python 3.4.3. Travis was set to use 3.5. This is bad because it means that tests can pass on Travis but then the app can fail to start.


ashimali:
👍

PR #697, opened 2016-06-17
Usage limit bug fix

ashimali:
This fixes: https://www.pivotaltracker.com/story/show/121284677

Move check_messages in admin over to using get notification stats for day.

This relies on https://github.com/alphagov/notifications-api/pull/434 which should be merged first.

### DIFF
--- a/app/notify_client/statistics_api_client.py
+++ b/app/notify_client/statistics_api_client.py
@@ -25,10 +25,12 @@ class StatisticsApiClient(BaseAPIClient):
     def get_statistics_for_service_for_day(self, service_id, day):
         url = '/service/{}/notifications-statistics/day/{}'.format(service_id, day)
         try:
-            self.get(url=url)['data']
+            return self.get(url=url)['data']
         except HTTPError as e:
             if e.status_code == 404:
                 return None
+            else:
+                raise e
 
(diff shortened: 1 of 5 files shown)


app/notify_client/statistics_api_client.py
```diff
@@ -21,6 +22,14 @@ def get_statistics_for_service(self, service_id, limit_days=None):
             params=params
         )
 
+    def get_statistics_for_service_for_day(self, service_id, day):
+        url = '/service/{}/notifications-statistics/day/{}'.format(service_id, day)
+        try:
+            self.get(url=url)['data']
+        except HTTPError as e:
+            if e.status_code == 404:
+                return None
```
leohemsted: This function will always return None - either it'll return None here, or implicitly return none at the end. I guess you meant to put `return self.get(....)`.

However I wonder what you want to do if the status isn't 404 - if it's 500, we should definitely re-raise for example. I think we should probably re-raise if it's anything except for 404 since that's expected.

Also definitely add a couple of tests (i'd probably add one making sure `check_messages` is using the return value, one making sure that when it returns 404 `check_messages` handles correctly, and one making sure that if it 500s we don't lose the exception)

PR #698, opened 2016-06-17
when not logged in, redirect to sign-in

leohemsted:
parts of the initial setup/login stages were throwing 500s if user
not already in process (ie: user directly navigated to url):

* /resend-email-verification
* /text-not-received
* /send-new-code
* /verify


ashimali:
👍

PR #699, opened 2016-06-17
Reorder the breaking change page

quis:
The table on this page looked too much like something you were supposed to interact with. To understand what’s expected of you from this page, you need to know:

- what you’ve done to make this page happen (you changed some stuff)
- what actions are available now (confirm, back)

The rest of the information is supplementary—it helps you make a decision, but it’s not key to understanding what the decision _is_.

This also matches what we do on the upload file page, where the non-essential example is under the button.

***

image


leohemsted:
👍

PR #700, opened 2016-06-17
Redirect to the templates page after the tour

quis:
When you’re dropped straight into the ‘edit template’ page it’s still a bit confusing what the thing you’re typing is—the mental model isn’t quite there yet.

I think it will help (rather than redirect to the dashboard) to redirect to the choose template page. You can then choose to edit the example or add your own template.

This should help people understand that the example is one of a number of templates that you create.


leohemsted:
👍

PR #701, opened 2016-06-17
Add some words to the template statistics on the dashboard

quis:
image

## Make template statistics on the dashboard a table

It was a `<dl>` before which is kinda weird. Especially when the jobs table was a real `<table>`.

It also means we can give it column headings so that new and invited users have a better idea of what it is.

## Say templates ‘used’ instead of templates ‘sent’ 

You _use_ a template to _send_ a message


leohemsted:
👍

PR #705, opened 2016-06-20
Filter test messages from jobs on the dashboard

quis:
While test messages technically have a file and are a job, there’s not much reason to ever revisit them. So all they end up doing is cluttering the dashboard and making it harder to find the actual files you’ve
actually uploaded from your computer.

So this commit:

- abstracts the name of test messages into config
- filters out any files whose filename represents a test message
- adds some more thorough tests for the jobs on the dashboard

### DIFF
--- a/tests/app/main/views/test_dashboard.py
+++ b/tests/app/main/views/test_dashboard.py
@@ -2,6 +2,7 @@ import copy
 from flask import url_for
 
 from bs4 import BeautifulSoup
+from freezegun import freeze_time
 
 from tests import validate_route_permission
 from tests.conftest import SERVICE_ONE_ID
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_dashboard.py
```diff
@@ -140,6 +141,49 @@ def test_should_show_all_templates_on_template_statistics_page(
...
+    for index, filename in enumerate((
+        "export 1/1/2016.xls",
+        "all email addresses.xlsx",
+        "applicants.ods",
+        "thisisatest.csv",
+    )):
+        for string in (filename, 'Uploaded 1 January at 11:09'):
+            assert string in table_rows[index].find_all('th')[0].text
+        for column_index, count in enumerate((1, 0, 0)):
+            assert table_rows[index].find_all('td')[column_index].text.strip() == str(count)
```
leohemsted: can you `pytest.mark.parametrize` these

quis: Don’t think so. But the first for loop is overkill. Changed in b7e58b0

leohemsted:
👍

PR #711, opened 2016-06-20
Give users more help about what to put in their templates

quis:
new-edit-template-email

***

new-edit-template-sms

***

image


quis:
Updated based on discussion with @yahoopete 

image

yahoopete:
:+1:

PR #712, opened 2016-06-20
Daily limit bugfix

ashimali:
Move check_messages in admin over to using get notification stats for day.

This is for https://www.pivotaltracker.com/story/show/121284677


servingUpAces:
👍 tested and reviewed

PR #715, opened 2016-06-20
Make it clearer what Notify is for from homepage

quis:
Uses some graphic design to communicate that Notify is:

- for sending messages to lots of people
- which can be personalised

image


quis:
## Add a reference number to the example texts

To show that the personalisation can include any kind of information, not just name.

image

PR #717, opened 2016-06-21
Add live/trial flag on platform-admin page

leohemsted:
Show 'Live' or 'Trial' under each service to easily identify services

image


quis:
Made some tweaks: 

image

quis:
👍 to @leohemsted’s bit

PR #720, opened 2016-06-22
Use tabular numbers for all big numbers

quis:
Reimplements https://github.com/alphagov/notifications-admin/pull/169

> Tabular numbers have numerals of a standard fixed width. As all
> numbers have the same width, sets of numbers may be more easily
> compared. We recommend using them where different numbers are likely
> to be compared, or where different numbers should line up with each
> other, eg in tables.

— https://github.com/alphagov/govuk_frontend_toolkit/blob/5f38012f94dc2dcf676b565204126d056b794e90/docs/mixins.md#tabular-numbers

image

image

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -4,7 +4,7 @@
   display: block;
 
   &-number {
-    @include bold-48;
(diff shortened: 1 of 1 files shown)


leohemsted:
👍

PR #721, opened 2016-06-22
Remove email from address on the job page

quis:
- it’s just taking up space
- it’s too late to change it if it’s wrong
- it’s not accurate if you’ve changed your service name since creating the job


yahoopete:
:+1:

PR #725, opened 2016-06-23
Play back the user’s email address

quis:
screen shot 2016-06-23 at 15 42 45

When you register you type in your email address. If you don’t get the email there’s no way of knowing it’s because you’ve mistyped it.

If we play back the email address, you can double check it.

This commit also removes the ⅔ column on this page to make sure a long email address doesn’t wrap.


ashimali:
👍

PR #726, opened 2016-06-23
Platform admin: Highlight service in research mode

quis:
image

***

Research mode trumps live.

Copies the same style we use in the page footer.


leohemsted:
👍

PR #727, opened 2016-06-24
Fix ‘help’ appearing when it shouldn’t

quis:
party2

Steps to reproduce:
- make a template with a placeholder
- click ‘send yourself a test’
- leave fields blank
- click ‘check’
- see error, click ‘back’

Expected: previous page

Actual: previous page with blue help sidebar

When the URL contains `help=0`, `request.args.get('help')` returns '0'. Doing `if '0':` is the same as doing any `if <non empty string>:` which returns `True`.

So we should only display the help when the help query parameter is:

- not missing
- AND a string that isn’t `'0'`

***

This commit also does a bit of tidying up, detailed in subsequent commits.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -243,7 +243,7 @@ def check_messages(service_id, template_type, upload_id):
     )
 
     if request.args.get('from_test') and len(template.placeholders):
-        extra_args = {'help': 1} if request.args.get('help') else {}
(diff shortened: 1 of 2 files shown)


leohemsted:
damn your pr description gif game is ridiculously on point

leohemsted:
👍 

though we should probably sort out this help malarkey at some point it just feels gross

PR #730, opened 2016-06-27
send key_type = normal when creating api keys

leohemsted:
~~depends on https://github.com/alphagov/notifications-api/pull/473~~

also cleaned up the test_api_keys file a little - removed a couple of unnecessary fixtures

### DIFF
--- a/app/notify_client/api_key_api_client.py
+++ b/app/notify_client/api_key_api_client.py
@@ -2,6 +2,11 @@ from notifications_python_client.base import BaseAPIClient
 from app.notify_client import _attach_current_user
 
 
+# must match key types in notifications-api/app/models.py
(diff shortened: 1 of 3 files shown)


app/notify_client/api_key_api_client.py
```diff
@@ -2,6 +2,11 @@
 from app.notify_client import _attach_current_user
 
 
+# must match key types in notifications-api/app/models.py
+KEY_TYPE_NORMAL = 'normal'
+KEY_TYPE_TEAM = 'team'
```
leohemsted: not sure if there's a more elegant way of sharing constants between admin and api?

servingUpAces:
👍

PR #731, opened 2016-06-27
Use password_changed_at field from User object

servingUpAces:
The user profile date was displaying a hard coded date, this PR fixed that.

https://www.pivotaltracker.com/story/show/122205615
screen shot 2016-06-28 at 10 41 39

Was previously showing 1 January 2016

### DIFF
--- a/app/templates/views/user-profile.html
+++ b/app/templates/views/user-profile.html
@@ -14,7 +14,7 @@
       {'label': 'Name', 'value': current_user.name, 'url': url_for('.user_profile_name')},
       {'label': 'Email address', 'value': current_user.email_address, 'url': url_for('.user_profile_email')},
       {'label': 'Mobile number', 'value': current_user.mobile_number, 'url': url_for('.user_profile_mobile_number')},
-      {'label': 'Password', 'value': 'Last changed 1 January 2016, 10:00AM', 'url': url_for('.user_profile_password')},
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -458,7 +457,8 @@ def api_user_active(fake_uuid):
                  'state': 'active',
                  'failed_login_count': 0,
                  'permissions': {},
-                 'platform_admin': False
+                 'platform_admin': False,
+                 'password_changed_at': str(datetime.utcnow())
```
leohemsted: should probably add this line to the other fake users too

app/templates/views/user-profile.html
```diff
@@ -14,7 +14,7 @@ <h1 class="heading-large">Your profile</h1>
       {'label': 'Name', 'value': current_user.name, 'url': url_for('.user_profile_name')},
       {'label': 'Email address', 'value': current_user.email_address, 'url': url_for('.user_profile_email')},
       {'label': 'Mobile number', 'value': current_user.mobile_number, 'url': url_for('.user_profile_mobile_number')},
-      {'label': 'Password', 'value': 'Last changed 1 January 2016, 10:00AM', 'url': url_for('.user_profile_password')},
+      {'label': 'Password', 'value': current_user.password_changed_at |format_datetime_short, 'url': url_for('.user_profile_password')},
```
leohemsted: is it worth keeping the 'Last changed' label - it feels a bit mislabeled to have a field that simply says

> Password: 28 June at 09:18

leohemsted:
👍

PR #734, opened 2016-06-28
Refactor markup behind the AJAX bits of the page

quis:
This pull request does some refactoring with the aim of:

- templates being split up in a more logical way
- less repetition, so rendering a part of the page is always done through the same code path
- deleting some code

The main changes are:

## Better structure for pages with AJAX components

Before 
```
dashboard.html
|_ today.html
   |_ some random html
   |_ a few things split into partials
```

After

```
dashboard.html
|_ partial
|_ partial
|_ …
```

## Derive markup of pages from rendered partials

This means that each section of the page is rendered independently. It then either:

- gets returned as the value of a key in the JSON (when going via AJAX)
- gets spat out somewhere on it’s parent page (for first page load)


leohemsted:
👍

PR #741, opened 2016-07-01
Revert questionable design decisions in send messages flow

quis:
## Move ‘send test’ link back to template page and reword ‘send emails’ to ‘upload recipients’

image
image

This changes it back to how it was when we first introduced this feature: #181

It’s kind of lost on the page where you upload a file, which is a shame because it’s a good learning opportunity.

‘Send emails’ doesn’t speak to you if you already have the idea of a file or address book in mind. ‘Upload’ better describes what you’re going to do on the next page.

Also makes all the links regular weight, because having the first one bold looked like a heading.

## Don’t hide the example on the send page

image

People missed this example because it was hidden in a `<details>` element. Perhaps the wording of the link wasn’t ideal, but we’ve tried two different variations of it (#640).

Better not to hide things than try to think of some text which will make people want to show them.

Also adds some headings to chunk up this page a bit.

## Rename check and confirm to preview

image

We’ve seen people land on this page and expect the message to be on their phone already.

‘Check and confirm’ sounds a lot like ‘check your phone’, which is language that we use earlier on when we _have_ sent a message.

Hopefully ‘preview’ is a better indication that it’s not sent yet.

### DIFF
--- a/app/assets/stylesheets/components/message.scss
+++ b/app/assets/stylesheets/components/message.scss
@@ -29,10 +29,6 @@
 
     }
 
-    .primary {
-      @include bold-19;
-    }
-
   }
 
 }
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -242,11 +242,16 @@ def check_messages(service_id, template_type, upload_id):
         ) if current_service['restricted'] else None
     )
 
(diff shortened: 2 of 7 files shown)


yahoopete:
:+1:

PR #742, opened 2016-07-01
Don’t show send/edit links for deleted template

quis:
Since you can’t really send or edit a deleted template we should show a message telling you that the template has been deleted.

This is important because deleted templates still show up in the template statistics.

image


leohemsted:
👍

PR #743, opened 2016-07-01
Hide download CSV link when in the tour

quis:
There is really no point to downloading the CSV file of delivery info when you’re in the tour. It’s just distracting at this point.

### DIFF
--- a/app/templates/partials/jobs/notifications.html
+++ b/app/templates/partials/jobs/notifications.html
@@ -4,7 +4,7 @@
   <div class="dashboard-table">
 {% endif %}
 
-  {% if notifications %}
+  {% if notifications and request.args.get('help', '0') == '0' %}
     <p class="bottom-gutter">
       <a href="{{ url_for('.view_job_csv', service_id=current_service.id, job_id=job.id, status=status) }}" download="download" class="heading-small">Download as a CSV file</a>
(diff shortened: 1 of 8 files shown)


app/templates/partials/jobs/notifications.html
```diff
@@ -4,7 +4,7 @@
   <div class="dashboard-table">
 {% endif %}
 
-  {% if notifications %}
+  {% if notifications and request.args.get('help', '0') == '0' %}
```
ashimali: Why not pass this to partial as arg e.g. request.args.get('help', '0') == '0' becomes is_tour_page or true or false.

quis: Jinja macros can’t return a value unfortunately.

ashimali: Not quite what I meant. Just as notifications is set in the view function and passed to template you could have a variable "is_tour_page" set to result of request.args.get('help', '0') == '0' and pass that to template.  request.args.get('help', '0') == '0'  is a little opaque in this context.

quis: I follow you now. Changed in 29d9333.

yahoopete:
:+1:

PR #747, opened 2016-07-01
Set text message sender

ashimali:
Users can set a value that appears as the sender of a text message.

It can be up to eleven characters alpha numeric, no special characters
allowed.


servingUpAces:
👍 reviewed and demoed

PR #751, opened 2016-07-04
Fix get started banner on dashboard

quis:
This banner was always being shown because the template was never getting sent the service’s templates from the API.

This commit fixes this to only show the banner when a service has no templates, and adds some tests to make sure it doesn’t happen again.

Also fixes the link that took you to the email page instead of the text message page.


leohemsted:
👍

PR #753, opened 2016-07-04
Better flash messages for setting text message sender.

ashimali:


tests/app/main/views/test_service_settings.py
```diff
@@ -735,3 +735,37 @@ def test_if_sms_sender_set_then_form_populated(app_,
         assert response.status_code == 200
         page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
         assert page.find(id='sms_sender')['value'] == 'elevenchars'
+
+
+def test_set_text_message_sender_flash_messages(
```
quis: I think this test would be better [parametrized](http://pytest.org/latest/parametrize.html), eg:

```python
@pytest.mark.parametrize("sender, expected_flash_message", [
    ("elevenchars", 'Text message sender set to elevenchars'),
    ('', 'Text message sender removed')
])
```

quis:
👍

PR #757, opened 2016-07-05
Make tests run faster by using multiple CPU cores

quis:
We can make the tests run slightly faster by parallelizing them across multiple CPU cores:

- from some casual testing locally, 2 cores gives the optimum speedup
- Travis container-based builds [have 2 CPU cores available](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments)
- the net gain is about 20%, or 2 seconds
- unfortunately we can’t do this on the API because each test is still using the same instance of the database

This pull request also tidies up the `run_tests.sh` script a little.


leohemsted:
👍

PR #763, opened 2016-07-06
Api key types

leohemsted:
Can now set api keys on the front end. Tada! 🎉  🎉  🎉 

image

image

@quis - revoking a key leaves you with two columns of quite long content  - do we want to look at that or is that fine?

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -111,7 +111,12 @@
 
 }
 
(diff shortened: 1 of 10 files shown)


quis:
@leohemsted good point. I reckon this works better (8b9704d):

image

quis:
And a less verbose datetime helps (d214bae):

image

servingUpAces:
👍

PR #764, opened 2016-07-06
mmg rates set to 1.65p

leohemsted:
also moved a default parameter to a plain ol' local variable to reduce ambiguity about whether it is overriden by any callers (it's not)

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -178,7 +178,10 @@ def get_dashboard_partials(service_id):
     }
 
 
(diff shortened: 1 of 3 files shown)


leohemsted:
unsure if i should fix the rates on the front end being hard coded now or not - it's pretty low prio IMO since no-one's sent >250k SMS yet, but it's something we definitely need to do, especially since we've got the provider_rates in the DB and are never actually using the buggers

quis:
👍

PR #772, opened 2016-07-08
Add api client call to email users that are already registered

servingUpAces:
This method will need to be called from the register endpoint. 
There is a pull request to add the new api endpoint. 

https://www.pivotaltracker.com/story/show/123740331


ashimali:
👍 seen demo

PR #775, opened 2016-07-11
Make ‘temporary failure’ clearer in frontend and CSV downloads

quis:
image

‘Phone number doesn’t exist’ as opposed to ‘Permanent failure’ has tested well because it talks in terms of things people understand.

We should do the same for ‘Temporary failure’, because users are unclear:
- why this is happening
- if it’s temporary, is Notify going to retry it (it’s not)

We think that ‘Phone/inbox not currently accepting messages’ answers makes these things clearer.

I’ve reworded it slightly to:

- Inbox not accepting messages right now
- Phone not accepting messages right now

## Add formatted notification status to CSV

This commit makes the CSV download use the same language for failure reasons as the frontend.

It also adds a test around this stuff, which was patchily tested before.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -235,16 +235,16 @@ def format_notification_status(status, template_type):
         'email': {
             'failed': 'Failed',
             'technical-failure': 'Technical failure',
-            'temporary-failure': 'Temporary failure',
-            'permanent-failure': 'Email address does not exist',
+            'temporary-failure': 'Inbox not accepting messages right now',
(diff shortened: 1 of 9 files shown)


yahoopete:
We should update the https://www.notifications.service.gov.uk/delivery-and-failure page too as part of this....

quis:
@yahoopete 

screencapture-localhost-6012-delivery-and-failure-1468242153028

yahoopete:
:+1:

PR #787, opened 2016-07-14
Increase daily message limit for unrestricted services

ashimali:
Set daily message limit for live services to 250,000

This pr is for https://www.pivotaltracker.com/story/show/126230157


quis:
👍

PR #794, opened 2016-07-15
Rename api_client to service_api_client

leohemsted:
also moved some test files to accurately mimic the real app folder structure


quis:
👍

PR #795, opened 2016-07-18
Make ‘Available for…’ text on job page relative

quis:
#### Reviewers: all the substantive changes in this pull request are in cfadead and 9a5fbca. Other changes are detailed in the commits; they’re either refactoring or making the tests more thorough.

***

When we say ‘delivery information is available for 7 days’ you have to infer _when_ the seven days starts. When you come back to the page it still says ‘available for 7 days’ even if you only have a day left to download it. This is confusing.

This pull request changes the text to be relative to now, eg ‘available for 7 days’, ‘available for 1 day’.

***

image

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -320,19 +320,23 @@ def get_job_partials(job):
                 status=request.args.get('status', '')
             ),
             help=get_help_argument(),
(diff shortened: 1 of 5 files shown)


leohemsted:
👍

PR #796, opened 2016-07-19
Really hide the ‘download CSV’ link in the tour

quis:
Tried to do this in https://github.com/alphagov/notifications-admin/pull/743

Didn’t account for it reappearing when the AJAX fired, because the AJAX request didn’t account for whether or not the user was in the tour.


leohemsted:
👍

PR #797, opened 2016-07-19
Deploy to staging

quis:


quis:
Checked my stuff on preview, 👍 to continue

PR #802, opened 2016-07-20
Remove notification-statistics endpoints from totals boxes 

leohemsted:
( depends on https://github.com/alphagov/notifications-api/pull/540 )

hoo boy. Lots of blood sweat and tears went into this one.

* Some refactoring of existing `statistics_utils` functions to pull out some shared functionality
* adding `mock_get_detailed_service` to any test that hits the dashboard - `test_new_invited_user_verifies_and_added_to_service` needed a bunch of other mocks for reasons I can't decipher
* changing the partial `_totals.html` to use the slightly different, more nested new statistics structure fresh from the detailed service endpoint

(#803 waiting on this)

PR #803, opened 2016-07-20
Remove stats from job page

leohemsted:
~~( depends on  alphagov/notifications-api#540 and #802 )~~

much easier than #802 - little to no extra work required, and got to remove the first statistics endpoint!


ashimali:
👍

PR #804, opened 2016-07-21
Too many messages

leohemsted:
A little clean-up ahead of removing statistics from this page - moved the "csv row count goes over your daily limit" warning to its own partial file and cleaned it up a bit - it was previously only showing your total allowed messages and trial mode warning if you had not sent any messages that day - which doesn't really make sense. Now it shows that bit every time, but only shows the "how many messages you have remaining" if you've sent messages.

#### to see the changes to the template, look here: https://github.com/alphagov/notifications-admin/pull/804/commits/afce715a3d59b76392093b33ffeebe08817d42c4?w=1

It now looks like:

image

image


ashimali:
👍

PR #809, opened 2016-07-25
don't hit statistics endpoints from send

leohemsted:
i've made three functions in service_api_client but i'm not 100% sold on that - should I just have one function with defaults. up for debate, i'm happy either way.

blocked by ~~#804~~, ~~https://github.com/alphagov/notifications-utils/pull/56~~ and ~~https://github.com/alphagov/notifications-api/pull/559~~


app/templates/partials/check/too-many-messages.html
```diff
@@ -0,0 +1,24 @@
...
+    </h1>
+    <p>
+      You can only send {{ current_service.message_limit }} messages per day
+      {%- if current_service.restricted %}
+        in <a href="{{ url_for('.trial_mode')}}">trial mode</a>
+      {%- endif -%}
+      .
+    </p>
+    <p>
+      {% if current_service.message_limit != remaining_messages %}
```
minglis: Does this work? Should be >=? Can this miss?

tests/app/notify_client/test_service_api_client.py
```diff
@@ -21,17 +23,17 @@ def test_client_posts_archived_true_when_deleting_template(mocker):
...
-
-
-def test_client_gets_service_with_detailed_params(mocker):
+@pytest.mark.parametrize(
+    'function,params', [
+        (ServiceAPIClient.get_service, {}),
+        (ServiceAPIClient.get_detailed_service, {'detailed': True}),
+        (ServiceAPIClient.get_detailed_service_for_today, {'detailed': True, 'today_only': True})
+    ],
+    ids=lambda x: x.__name__
```
quis: Nifty.

app/templates/partials/check/too-many-messages.html
```diff
@@ -0,0 +1,24 @@
...
+        You can still send {{ remaining_messages }} messages today, but
+      {% endif %}
+      ‘{{ original_file_name }}’ contains
+      {{ count_of_recipients }} {{ recipients.recipient_column_header }}
+      {%- if count_of_recipients != 1 -%}
+        {{ 'es' if 'email address' == recipients.recipient_column_header else 's' }}
```
minglis: Nasty, does this help? http://jinja.pocoo.org/docs/dev/templates/#i18n

tests/app/notify_client/test_service_api_client.py
```diff
@@ -21,17 +23,17 @@ def test_client_posts_archived_true_when_deleting_template(mocker):
...
+)
+def test_client_gets_service(mocker, function, params):
     client = ServiceAPIClient()
-    mock_post = mocker.patch('app.notify_client.service_api_client.ServiceAPIClient.get')
+    mock_post = mocker.patch.object(client, 'get')
```
quis: This variable should be called `mock_get`.

app/templates/views/check.html
```diff
@@ -103,8 +103,8 @@ <h1 class='banner-title'>
       {% endcall %}
     </div>
 
-  {% elif count_of_recipients > (current_service.message_limit - statistics.get('emails_requested', 0) - statistics.get('sms_requested', 0)) %}
-    {% include "partials/check/too_many_messages.html" %}
+  {% elif recipients.more_rows_than_can_send %}
```
quis: So much better.

quis:
Tested locally. Looked at code. Other than the naming comment, 👍 .

app/templates/partials/check/too-many-messages.html
leohemsted: in theory never, but happy to change this just to be safe

leohemsted: scratch that, this is already a defensive if statement (ie: if we had to chose i'd prefer it to trip because it contains additional information) so i'll keep in the !=

app/templates/partials/check/too-many-messages.html
leohemsted: not really - that's about "here's two phrases, pick one of them based on whether `x > 1` - whereas we want to modify the phrase based on whether the word is 'email address' or 'phone number'.

PR #811, opened 2016-07-28
Add beta banner to homepage

quis:
image

Because the homepage of the app is discoverable to the public, it makes “gives the wrong impression” for it not to have a beta banner.

Also makes the strip at the top red only when you’re logged in. Mainly because the red/beta orange/blue combo looks gross. And also because you’re not in admin mode until you’re logged in.


leohemsted:
👍

PR #812, opened 2016-07-28
Update environment configs to override blue header

quis:
Because the header bar now defaults to blue instead of red, live needs to be explicitly set to have a red header bar.

This also updates the staging config, which used to inherit from the preview one (and get an orange header) but now inherits from the base config, and so needs its header colour setting explicitly as well.


leohemsted:
👍

PR #820, opened 2016-08-01
Upgrade GOV.UK template

quis:
Full changes:
https://github.com/alphagov/govuk_template/compare/v0.17.1...v0.18.0


leohemsted:
👍

PR #821, opened 2016-08-01
Make page <title> on API info match page heading

quis:
Otherwise you might get it mixed up with the API integration page when you have a bunch of tabs open.


quis:
@leohemsted d673cb1

leohemsted:
👍

PR #827, opened 2016-08-02
Move guidance underneath

quis:
## Move template guidance underneath

Two problems with having it on the side:
- some users didn’t see it at all
- there wasn’t space to have additional guidance about Markdown-style formatting

## Add Gwen’s initial guidance about formatting

This is based on some work Gwen did for Civil Service Digital. Let’s get it in for now so that we have a starting point from which to improve.

This specifically doesn’t reference ‘optional’ placeholders because I don’t know how best to explain those yet.

## Text message templates

image

## Email templates

image


leohemsted:
👍

PR #832, opened 2016-08-03
Make numbers on a job less confusing

quis:
## Summary

The counts on the job page:
- were stuff we care about more than stuff that a user should care about
- changed in quite a confusing way, eg sending moved up and down and was basically a count of Notify’s throughput, which is not a user-friendly concept

This PR changes the counts to:
- be less abstract
- only ever move in one direction

It also changes the download link to only show when processing has finished.

## Before

img_5717

## After

img_5718

## In action

final countdown 9

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -2,6 +2,7 @@
 import ago
 import time
 import dateutil
+from orderedset import OrderedSet
 from datetime import datetime, timedelta, timezone
 from itertools import chain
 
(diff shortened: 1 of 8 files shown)


app/main/views/jobs.py
```diff
@@ -48,10 +48,11 @@ def _parse_filter_args(filter_dict):
...
 def _set_status_filters(filter_args):
     all_failure_statuses = ['failed', 'temporary-failure', 'permanent-failure', 'technical-failure']
-    all_statuses = ['sending', 'delivered'] + all_failure_statuses
+    all_sending_statuses = ['created', 'sending']
+    all_statuses = all_sending_statuses + ['delivered'] + all_failure_statuses
     if filter_args.get('status'):
-        if 'processed' in filter_args.get('status') or not filter_args.get('status'):
-            filter_args['status'] = all_statuses
+        if 'sending' in filter_args.get('status'):
+            filter_args['status'].extend(all_sending_statuses[:1])
```
leohemsted: i think code reads a lot better if we just say `filter_args['status'] = all_sending_statuses` here rather than slicing a two-element array?

quis:
@leohemsted 01bb2ad and b85b929

## Don’t let _set_status_filters mutate original list

Mutating stuff is scary and prone to problems. Better for it to return a new list instead.

## Lose the confusing list slicing

Slicing was to avoid duplicate items in the list. A more idiomatic way to avoid duplicate items in a list is to use a `set` instead.

The order of the list doesn’t really matter, but it’s a lot easier to test for if the order is consistent.

leohemsted:
👍

PR #833, opened 2016-08-03
Stop AJAX updater consuming ever-increasing memory

quis:
The pages with AJAX on were feeling quite sluggish, and it felt like they were making the whole browser slow down.

Looking at the performance stuff in Chrome, the number of DOM nodes was going up and up, which is weird because the number of things on the page wasn’t changing. This was causing the page to consume more and more memory in order to store all these nodes.

screen shot 2016-08-03 at 08 12 06

This is kinda beyond my understanding, but I tried a few things and it looks like the browser was having a hard time garbage collecting the temporary bits of DOM used to update the page.

By assigning these bits of DOM to variables before using them it seems that the browser has an easier time cleaning them up.

screen shot 2016-08-03 at 08 10 29

Note that memory consumption and number of nodes are staying a lot more stable in this second screenshot.

### DIFF
--- a/app/assets/javascripts/updateContent.js
+++ b/app/assets/javascripts/updateContent.js
@@ -3,11 +3,16 @@
 
   var queues = {};
   var dd = new diffDOM();
+  var timer;
 
-  var getRenderer = $component => response => dd.apply(
-    $component.get(0),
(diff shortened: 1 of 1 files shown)


leohemsted:
👍  😕  i guess this is all magic to me too

PR #837, opened 2016-08-03
Add department crests for email sub branding

quis:
We need to host these images somewhere. Let’s start off doing it in the admin app, like we do for other email assets.

Eventually we should have a CDN in front of this.

Images lifted from https://github.com/alphagov/whitehall/tree/76ff2e0b6d50c56c8c86379ec413ecd4f85dc988/app/assets/images/crests


leohemsted:
👍

PR #838, opened 2016-08-03
Don’t show banner on dashboard to API users

quis:
The "you only have permission to view this service" banner sort of makes sense if you don’t have _any_ permissions, but it doesn’t if you have permission to create API keys. If you can create API keys you can do a lot more than just view the service.


leohemsted:
👍

PR #844, opened 2016-08-04
Fix deskpro user email

leohemsted:
that five minute fix turned out to be about an hour cos of some edge cases.

also i cleaned up behind martyn's changes to the environment files 😉 

### in deskpro: 

image

### the user sees responses like:
image


quis:
👍

PR #845, opened 2016-08-05
Make it clear who you’re sending the example to

quis:
It’s not one of your colleagues. You’re not inviting someone. We’ve seen people get confused about this.

image


minglis:
👍

PR #847, opened 2016-08-05
Make ‘total’ the default status on a job

quis:
The blue bar on the job page works out which is the current status from the query parameters, not the parameters that the API gets called with.

## Before

image

## After

image


leohemsted:
👍

PR #851, opened 2016-08-08
Add spacer image for email template

quis:
Forgot to include this before.

The template expects it in `./images` but it should probably live in `./images/email-template`.

This commit puts it in both places; we can clean up later.


leohemsted:
👍

PR #855, opened 2016-08-09
Deploy to staging

leohemsted:


quis:
@leohemsted I’ve added my changes now, deploy away 🚢

PR #856, opened 2016-08-09
ᵂᴵᴾ Show upcoming jobs on the dashboard

quis:
This commit is only a prototype of the UI, it doesn’t make any functional changes.

image

- adds a new ‘in the next 24 hours’ section to the dashboard which lists upcoming jobs
- tweaks some spacing on the dashboard so that it doesn’t look like too much of a mess


quis:
Closing in favour of #903

PR #857, opened 2016-08-09
ᵂᴵᴾ Show scheduled time and cancel button on job page

quis:
This commit is only a prototype of the UI, it doesn’t make any functional changes.

image

If you’ve scheduled a job:

- you need to know what time it will start sending
- you need to be able to cancel it

This commit replaces the notification list (which we won’t have for scheduled jobs anyway) with some UI that allows the above.

Once implemented the counts will look like this for a scheduled job:

```
3128    0          0           0
total   sending    delivered   failed
```


quis:
Closing in favour of #903

PR #859, opened 2016-08-10
Link to client documentation, not API documentation

quis:
image

image

## Replace documentation link with links to clients

In research we found that developers orientate themselves around the API clients rather than the documentation.

We should get them to the client documentation as quickly as possible.

We currently link to the API documentation in three places:
- API integration page
- global footer
- template ‘API info’ page

For the first two this commit:
- removes the link to the documentation
- adds links to each of the 5 clients

For the last one it just removes the link entirely.

## Move terms of use into help

Terms of use is non-technical, it should sit alongside the other non-technical things like pricing.

## Rename documentation to API documentation

Everything in this column of the footer is specifically to do with the API now.

## Remove example personalisation from template page

This was of dubious value, and the syntax probably isn’t accurate for all languages.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -193,13 +193,9 @@ def send_from_api(service_id, template_id):
         service_api_client.get_service_template(service_id, template_id)['data'],
         prefix=current_service['name']
     )
-    personalisation = {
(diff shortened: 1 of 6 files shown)


leohemsted:
👍

PR #861, opened 2016-08-10
Talk about trial mode on API keys page

quis:
When you make a ‘normal’ API key it won’t work as described until your service is live.

We should make this clear at the point at which you choose this key.

image


leohemsted:
👍

PR #863, opened 2016-08-11
Allow partial service updates from service API client

quis:
cat-looks-shocked

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -66,17 +66,12 @@ def service_name_change_confirm(service_id):
     form = ConfirmPasswordForm(_check_password)
 
     if form.validate_on_submit():
-        current_service['name'] = session['service_name_change']
-        current_service['email_from'] = email_safe(session['service_name_change'])
         try:
             service_api_client.update_service(
                 current_service['id'],
-                current_service['name'],
-                current_service['active'],
-                current_service['message_limit'],
-                current_service['restricted'],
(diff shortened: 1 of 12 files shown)


app/main/views/service_settings.py
```diff
@@ -277,14 +259,8 @@ def service_set_sms_sender(service_id):
...
             current_service['id'],
-            current_service['name'],
-            current_service['active'],
-            current_service['message_limit'],
-            current_service['restricted'],
-            current_service['users'],
-            current_service['email_from'],
-            current_service['reply_to_email_address'],
-            sms_sender=form.sms_sender.data if form.sms_sender.data else None)
+            sms_sender=form.sms_sender.data if form.sms_sender.data else None
```
leohemsted: `form.sms_sender.data or None`

app/notify_client/service_api_client.py
```diff
@@ -73,38 +73,36 @@ def get_services(self, *params):
...
+        disallowed_attributes = set(data.keys()) - {
+            'name',
+            'users',
+            'message_limit',
+            'active',
+            'restricted',
+            'email_from',
+            'reply_to_email_address',
+            'sms_sender',
+            'created_by'
```
leohemsted: should probably not allow updating of fields we specifically don't expect the admin page to update (so `created_by` and `users` i think)

leohemsted:
👍

PR #864, opened 2016-08-11
Add a page to set branding and organisation

quis:
Platform admin only.

Adds radio buttons to choose one of:
- three hard-coded branding options
- organisations from a list provided by the API

image

Depends on:
- [ ] https://github.com/alphagov/notifications-api/pull/585

Branched off: https://github.com/alphagov/notifications-admin/pull/863 ~~(will need rebasing onto master)~~.

To do:

- [x] pre-populate the currently selected option
- [x] make validation validate valid choices of organisation

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -45,6 +45,7 @@ from app.notify_client.template_statistics_api_client import TemplateStatisticsA
 from app.notify_client.user_api_client import UserApiClient
 from app.notify_client.events_api_client import EventsApiClient
 from app.notify_client.provider_client import ProviderClient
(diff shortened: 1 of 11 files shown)


app/__init__.py
```diff
@@ -293,6 +298,7 @@ def load_service_before_request():
             _request_ctx_stack.top,
             'service',
             service_api_client.get_service(service_id)['data'] if service_id else None)
+        print('before_request', request.url)
```
servingUpAces: Let's remove the print stmt

servingUpAces:
👍

PR #870, opened 2016-08-16
Create Docker build image, build project with Docker

bandesz:


app/version.py.dist
```diff
@@ -1,3 +1,4 @@
 __travis_commit__ = ""
-__time__ = "2016-07-05:14:44:52"
+__time__ = ""
 __travis_job_number__ = ""
+__travis_job_url__ = ""
```
leohemsted: should this file be gitignored?

bandesz: no

PR #878, opened 2016-08-22
Handle the API migration

minglis:
If it;s the new format API hide the results

- Temp fix until new Admin code ships. Should be live for 5 mins.


leohemsted:
👍

PR #879, opened 2016-08-22
Changed some more instances of "confirmation" code to "security" code

minglis:


quis:
👍 

I think that keeping as it ‘Text message code‘ as the form label is fine, different context.

PR #886, opened 2016-08-22
For the brief moment when we have incompatible API versions out,

minglis:
We won't call template stats by template id as the format has changed. So just ask the question "do you want to delete" without the context.


servingUpAces:
👍

PR #891, opened 2016-08-23
from alphabetical to desc co…

minglis:
the order was alphabetical by template name, and this was wrong and inconsistant from what was before. Change to be desc usage.


leohemsted:
👍

PR #894, opened 2016-08-23
Fix admin app putting service into research mode

quis:
We changed the `update_service` method to only update individual attributes of a service, and only allow it to update specified attributes in 0cfe106.

We neglected to specify `research_mode` as one of the allowed attributes.

This broke the app’s ability to put a service in or out of research mode.

This commit:
- makes sure the tests cover this eventuality
- fixes the bug by specifying `research_mode` as one of the allowed attributes

_See only the non-whitespace changes in 0f6a090_


leohemsted:
👍

PR #899, opened 2016-08-24
Handle new jobs api

minglis:
Admin changes to allow https://github.com/alphagov/notifications-api/pull/627

Must ship ahead of the above PR.

Adds a translation layer into the jobs client to parse out the new form job statistics into the old format. Doesn't change the existing HTML, only translates new data into the old form.

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -78,13 +78,14 @@ def view_job(service_id, job_id):
     filter_args = _parse_filter_args(request.args)
     filter_args['status'] = _set_status_filters(filter_args)
 
+    finished = (
+        bool(job.get('notification_count', 0) and ((
+            job.get('notifications_delivered', 0) +
+            job.get('notifications_failed', 0)
+        ) == job.get('notification_count', 0))))
     return render_template(
         'views/jobs/job.html',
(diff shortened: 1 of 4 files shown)


app/main/views/jobs.py
```diff
@@ -78,13 +78,14 @@ def view_job(service_id, job_id):
     filter_args = _parse_filter_args(request.args)
     filter_args['status'] = _set_status_filters(filter_args)
 
+    finished = (
+        bool(job.get('notification_count', 0) and ((
+            job.get('notifications_delivered', 0) +
+            job.get('notifications_failed', 0)
+        ) == job.get('notification_count', 0))))
```
leohemsted: can we re-write this a bit to make it easier to understand? it's a bit lispy here

i'm thinking something like 
```
finished = job.get('notification_count', 0) > 0 and (
    job.get('notifications_delivered', 0) + job.get('notifications_failed', 0) == job['notification_count']
)
```

(just fyi `+` is always higher precedence than `==` - `1+2 == 3` and `3 == 1 + 2` are equivalent)

leohemsted:
other than that 1 comment looks good

leohemsted:
👍 much nicer

PR #904, opened 2016-08-24
Make settings page a table and do a better job of explaining the reply to address/text message sender

quis:
Shouldn’t be merged before:
- [x] https://github.com/alphagov/notifications-api/pull/624
- [x] https://github.com/alphagov/notifications-admin/pull/905

***

image

image

image


yahoopete:
👍

PR #907, opened 2016-08-24
Platform admin stats rework

leohemsted:
* [x] https://github.com/alphagov/notifications-api/pull/628 - api PR

Rewrote front-end to deal with already formatted statistics

Also added tests for research mode and trial mode flags


app/main/views/platform_admin.py
```diff
@@ -1,50 +1,66 @@
...
+    return stats
+
+
+def format_stats_by_service(services):
+    for service in services:
+        stats = service['statistics'].values()
+        yield {
+            'id': service['id'],
+            'name': service['name'],
+            'sending': sum((stat['requested'] - stat['delivered'] - stat['failed']) for stat in stats),
```
minglis: What does this mean for API vs Job notifications. What does requested mean? I guess this is in the API?

minglis:
👍

app/main/views/platform_admin.py
leohemsted: no difference between them - job rows that haven't been processed yet aren't included in this figure (for now)

PR #910, opened 2016-08-25
Removed check on notifications-sent, that was used to check whether i…

minglis:
…t was old or new style API.

Now new style is deployed it's not needed.

Tests updated to reflect this,


leohemsted:
👍

quis:
👍 to the merge

PR #912, opened 2016-08-26
Fix dashboard page

minglis:
Fix for admin app to show job data on dashboard page
- uses the new stats block in the API to build the stats expected by the dashboard page
- all work done in client


leohemsted:
seems reasonable 👍

PR #916, opened 2016-08-30
Base ‘% complete’ on notifications requested

quis:
A job is complete when all notifications are delivered or failed.

The ‘report’ is complete once we have all the notifications in the database.

This commit changes the meaning of the percentage from the former to the latter. This is how it was before we removed the aggregate stats for jobs.

This is important, because once the job processing is complete then you can accurately reconcile the report with the CSV file that you’ve uploaded.

### DIFF
--- a/app/notify_client/job_api_client.py
+++ b/app/notify_client/job_api_client.py
@@ -18,15 +18,14 @@ class JobApiClient(BaseAPIClient):
     @staticmethod
     def __convert_statistics(job):
         results = defaultdict(int)
-        if 'statistics' in job:
-            for outcome in job['statistics']:
-                if outcome['status'] in ['failed', 'technical-failure', 'temporary-failure', 'permanent-failure']:
-                    results['failed'] += outcome['count']
-                if outcome['status'] in ['sending', 'pending', 'created']:
-                    results['sending'] += outcome['count']
-                if outcome['status'] in ['delivered']:
-                    results['delivered'] += outcome['count']
-                results['requested'] += outcome['count']
+        for outcome in job['statistics']:
+            if outcome['status'] in ['failed', 'technical-failure', 'temporary-failure', 'permanent-failure']:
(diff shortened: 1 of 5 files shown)


app/notify_client/job_api_client.py
```diff
@@ -16,11 +17,7 @@ def init_app(self, app):
 
     @staticmethod
     def __convert_statistics(job):
-        results = {
-            'sending': 0,
-            'delivered': 0,
-            'failed': 0
-        }
+        results = defaultdict(int)
```
leohemsted: ah nice

app/notify_client/job_api_client.py
```diff
@@ -16,11 +17,7 @@ def init_app(self, app):
...
         if 'statistics' in job:
```
leohemsted: possibly dumb question but is this ever not the case? it looks like this was a switch put in to prevent breaking when deploying api/admin separately

leohemsted:
👍  save for that question about 'statistics' ever being not present

app/notify_client/job_api_client.py
quis: It’s always assigned: https://github.com/alphagov/notifications-api/blob/bcfa83de79e11856f12a7581b6519ec4a66dad36/app/job/rest.py#L48

And the structure should always be the same, even if the counts are zero because they’re generated from this query: https://github.com/alphagov/notifications-api/blob/668e6c9716cfffdb987cbf8abe20c14576864a1c/app/dao/jobs_dao.py#L11-L22

This line suggests that it’s a hangover from the aggregate tables: https://github.com/alphagov/notifications-admin/commit/8c159da3eaf3a8100bdc2603b4addf25b9fe573d#diff-9886486bf41b0680d23588b190c252eaL24

Going to remove it.

quis: Changed in 40e22d8

PR #917, opened 2016-08-30
Make placeholder highlighting align at all size

quis:
Because the placeholder highlighting was defined in pixels it got slightly out of line when it was used at larger type sizes, eg inside a heading.

By using `em`s it will scale with the size of the type.

## Before
screen shot 2016-08-30 at 15 33 21

## After
screen shot 2016-08-30 at 15 32 59


leohemsted:
mad css shit 👍

PR #919, opened 2016-08-31
Revert "Revert "Remove check for statistics not being on job""

quis:
## Make sure create job client can handle statistics

The create job endpoint returns the data about the job with a `data:` wrapper. This commit makes sure that, when the client is trying to process a job which has just been created, it looks inside the `data` wrapper.

## Revert alphagov/notifications-admin#918

Don’t need to catch the case where a job doesn’t have statistics any more.

## Depends on
- [x] https://github.com/alphagov/notifications-api/pull/650

### DIFF
--- a/app/notify_client/job_api_client.py
+++ b/app/notify_client/job_api_client.py
@@ -18,15 +18,14 @@ class JobApiClient(BaseAPIClient):
     @staticmethod
     def __convert_statistics(job):
         results = defaultdict(int)
-        if 'statistics' in job:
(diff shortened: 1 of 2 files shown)


leohemsted:
re-ran build cos it didn't trigger for https://github.com/alphagov/notifications-admin/pull/919/commits/be90cf2783cb49487d6641498ea3fd392b5fb029 for some reason - https://jenkins.notify.tools/job/notifications-admin/112/console

all passed, seems good

👍

PR #922, opened 2016-09-01
Let users cancel a job

quis:
image

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/656

Cancelled jobs:
- are hidden on the dashboard
- are hidden on the jobs page
- return a 404 if you try to go to their page


servingUpAces:
👍

PR #924, opened 2016-09-05
Fix bug where every service was ‘in’ trial mode

quis:
The message on the service settings page was being shown for every service, whether or not they were in trial mode.


servingUpAces:
👍

PR #926, opened 2016-09-06
Redirect and resend verification email when pending user attempts to login

imdadahad:
When a pending user attempts to create an account/login they will receive another verification email (incase the email was deleted/in spam)

@quis does the message on line 41 sound okay, can amend if required?

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -38,8 +38,8 @@ def sign_in():
         user = user_api_client.get_user_by_email_or_none(form.email_address.data)
         user = _get_and_verify_user(user, form.password.data)
         if user and user.state == 'pending':
-            flash("You haven't verified your email or mobile number yet.")
-            return redirect(url_for('main.sign_in'))
+            flash("You haven't verified your email or mobile number yet. Check your email for a verification link.")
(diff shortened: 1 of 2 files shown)


app/main/views/sign_in.py
```diff
@@ -38,8 +38,8 @@ def sign_in():
         user = user_api_client.get_user_by_email_or_none(form.email_address.data)
         user = _get_and_verify_user(user, form.password.data)
         if user and user.state == 'pending':
-            flash("You haven't verified your email or mobile number yet.")
-            return redirect(url_for('main.sign_in'))
+            flash("You haven't verified your email or mobile number yet. Check your email for a verification link.")
```
quis: I don’t think you need the flash message at all here. The page will look like:

>   **Check your email**
>
>   In order to verify your email address we've sent a new confirmation link to test@example.com

quis: And I’ve tweaked it a bit here: https://github.com/alphagov/notifications-admin/pull/927

imdadahad: Gotcha, just updated.

quis:
Tested locally, looks good 👍

PR #928, opened 2016-09-07
Simplified the config files for admin.

minglis:
Updated the config and readme to simply as per work in PR: 

https://github.com/alphagov/notifications-api/pull/664

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -242,21 +242,6 @@ def view_notifications(service_id, message_type):
     )
 
 
-@main.route("/services/<service_id>/jobs/<job_id>/notification/<string:notification_id>")
-@login_required
-@user_has_permissions('view_activity', admin_override=True)
-def view_notification(service_id, job_id, notification_id):
-    now = time.strftime('%H:%M')
-
-    return render_template(
-        'views/notification.html',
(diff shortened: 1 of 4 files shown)


app/main/views/jobs.py
```diff
@@ -250,9 +250,7 @@ def view_notification(service_id, job_id, notification_id):
 
     return render_template(
         'views/notification.html',
-        message=[
-            message for message in messages if message['id'] == notification_id
-            ][0],
+        message=[message for message in messages if message['id'] == notification_id][0],
```
leohemsted: i was originally going to write about how this code should use `next` rather than `[ list comprehension ][0]` but now i'm more concerned because messages isn't defined and i suspect this page doesn't exist

app/main/views/jobs.py
```diff
@@ -307,23 +305,23 @@ def _get_job_counts(job, help_argument):
...
             ],
             [
-              'failed', 'failed',
-              job.get('notifications_failed', 0)
+                'failed', 'failed',
+                job.get('notifications_failed', 0)
             ]
         ]
-    ]
+        ]
```
leohemsted: rogue indent

leohemsted:
👍

PR #929, opened 2016-09-07
Fix 500 on trial mode page

quis:
I was using the trial mode page to prototype the time picker as part of 225a61ddd31721296146d56421770ac0fb4c3b7d.

I didn’t clean up properly, and left an errant undefined variable, which caused the trial mode page to `500`.

This commit:
- removes the errant argument
- adds a test for all the static pages to make sure that they at least return a `200`


leohemsted:
👍

PR #930, opened 2016-09-07
Fix empty jobs table showing on dashbord

quis:
If you have scheduled and then cancelled jobs this would be enough to show the jobs block on the dashboard (in other words it wasn’t filtering out cancelled jobs). However the contents of the table _were_ filtering out cancelled jobs, so the table would be empty and look broken.

This commit changes the conditional to operate on the `list` of jobs with cancelled ones filtered out.


servingUpAces:
👍

PR #932, opened 2016-09-07
Visual tidy up

quis:
## Only show ‘Request to go live’ link to admin users

Only users with the manage service permission can access this page, so we should only show the link to users that have this permission.

## Remove ‘active’ title on team page

This section of the page is covered by the ‘Team members’ `<h1>`.

Only the invited users is a special case that needs its own title.

image

## Remove 2/3rds and 3/4 columns

These columns make the line length too short, and the text less readable.

image

## Tweak spacing on API keys page

Splits the clients section out a bit more clearly.

image

### DIFF
--- a/app/assets/stylesheets/views/users.scss
+++ b/app/assets/stylesheets/views/users.scss
@@ -1,6 +1,7 @@
 .user-list {
 
   @include core-16;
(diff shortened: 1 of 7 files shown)


minglis:
👍

PR #934, opened 2016-09-08
Update Python client

quis:
Just so that nobody else has to do it.

Implements:
- [x] https://github.com/alphagov/notifications-python-client/pull/29

Which is a breaking change requiring the renaming of method arguments.

### DIFF
--- a/app/notify_client/api_key_api_client.py
+++ b/app/notify_client/api_key_api_client.py
@@ -9,15 +9,15 @@ KEY_TYPE_TEST = 'test'
 
 
 class ApiKeyApiClient(BaseAPIClient):
-    def __init__(self, base_url=None, client_id=None, secret=None):
+    def __init__(self, base_url=None, service_id=None, api_key=None):
         super(self.__class__, self).__init__(base_url=base_url or 'base_url',
-                                             client_id=client_id or 'client_id',
-                                             secret=secret or 'secret')
+                                             service_id=service_id or 'service_id',
+                                             api_key=api_key or 'api_key')
 
     def init_app(self, app):
         self.base_url = app.config['API_HOST_NAME']
-        self.client_id = app.config['ADMIN_CLIENT_USER_NAME']
-        self.secret = app.config['ADMIN_CLIENT_SECRET']
+        self.service_id = app.config['ADMIN_CLIENT_USER_NAME']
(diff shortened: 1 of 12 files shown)


app/notify_client/api_key_api_client.py
```diff
@@ -9,15 +9,15 @@
 
 
 class ApiKeyApiClient(BaseAPIClient):
-    def __init__(self, base_url=None, client_id=None, secret=None):
+    def __init__(self, base_url=None, service_id=None, api_key=None):
         super(self.__class__, self).__init__(base_url=base_url or 'base_url',
-                                             client_id=client_id or 'client_id',
-                                             secret=secret or 'secret')
+                                             service_id=service_id or 'service_id',
+                                             api_key=api_key or 'api_key')
```
leohemsted: given that we don't invoke these clients with keys, does it make sense to remove the args entirely?

```python
class ApiKeyApiClient(BaseAPIClient):
    def __init__(self):
        super().__init__(base_url='foo', service_id='bar', api_key='baz')
```

(side note: i hate how we've got this init_app thing rather than using __init__ properly but that's definitely beyond the remit of this PR to change that)

quis: Good point, changed in 5fda35c.

And yeah, I’d never really thought about what those clients are doing and it’s nasty.

leohemsted:
👍

app/notify_client/api_key_api_client.py
```diff
@@ -9,15 +9,13 @@
 
 
 class ApiKeyApiClient(BaseAPIClient):
-    def __init__(self, base_url=None, client_id=None, secret=None):
-        super(self.__class__, self).__init__(base_url=base_url or 'base_url',
-                                             client_id=client_id or 'client_id',
-                                             secret=secret or 'secret')
+    def __init__(self):
+        super(self.__class__, self).__init__("a", "b", "c")
```
leohemsted: dont need `self.__class__, self` in super - that's a python 2.x crutch. `super()` is equivalent

quis: 😎   92aacc1a549b6a5be52c0cd40fece2bc9599eaf2

leohemsted:
👍

PR #935, opened 2016-09-09
Fix overlapping text on tables of notifications

quis:
This commit changes the tables of notifications from 3 columns to two columns. This is so the text has more room, so it doesn’t start overlapping.

It also makes sure that if the recipient gets really long that it will be cut off with an ellipsis, rather than overlapping…

I hypothesize that if a notification fails you probably don’t care when it failed, just that it failed.

## Before

image

## After

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -282,11 +282,11 @@ def format_notification_status(status, template_type):
     }.get(template_type).get(status, status)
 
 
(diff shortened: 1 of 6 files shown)


quis:
@leohemsted I’ve added date to failed notifications, and refactored a bit.

image

minglis:
👍

PR #936, opened 2016-09-09
Fix exception raised when a pending user attempts to complete 'forgotten password' flow

imdadahad:
1. If a pending user executes the 'forgotten password' flow (click link in email + two factor auth) an exception is raised as their account has not been activated. If a user was able to complete the 'forgotten password' flow successfully, their account should be activated as at this point we are sure they have access to the email account and phone. The corresponding test was also added.

2. Fixes a test ensuring that a user is activated after they verify. Previous to this an activated user was passed in which does not test the functionality correctly.

3. Ensure the DB write that activates the user is only carried out if necessary.


minglis:
👍 ship it

PR #942, opened 2016-09-14
Link to GOV.UK tech doc format for Python client

quis:
We want to try this out in research tomorrow, instead of the README style doc we’ve been using up to this point.


imdadahad:
👍

PR #948, opened 2016-09-20
[WIP] Add an index page for the API integration

quis:
Initial prototype:

screen shot 2016-09-14 at 13 09 40

***

Step 1:

image

PR #949, opened 2016-09-26
Add whitelist, trial mode, and log to API page

quis:
image

***

image

Cleaned up version of https://github.com/alphagov/notifications-admin/pull/948

## Depends on:

- [x] https://github.com/alphagov/notifications-api/pull/689


minglis:
👍

PR #950, opened 2016-09-26
DISCUSS: Reduce minimum password length to 8 characters

quis:
We see over and over in research that people are tripped up by the 10 character requirement because it’s longer than they are used to. Most sites require 6 or 8 characters for a password.

It goes against the CESG advice which is to not try increasing password strength by increasing the burden on the user:

> Traditionally, organisations impose rules on the length and complexity
> of passwords. However, people then tend to use predictable strategies
> to generate passwords, so the security benefit is marginal while the
> user burden is high.

https://www.cesg.gov.uk/guidance/password-guidance-simplifying-your-approach

Instead we should be relying on:

- [x] two factor authentication
- [x] blacklisting common passwords
- [x] locking out users after 10 failed attempts to log in


quis:
Turns out the blacklist wasn’t very thorough, so I made it better in 3da47f2:

- went to the OWASP repository of security lists: https://github.com/danielmiessler/SecLists

- downloaded `10k_most_common.txt`, `twitter-banned.txt` and `500-worst-passwords.txt`

- filtered out any under 8 characters:
  ```
  sed -r '/^.{,7}$/d' passwords-twitter.txt > passwords-combined.txt
  sed -r '/^.{,7}$/d' passwords-500.txt >> passwords-combined.txt
  sed -r '/^.{,7}$/d' passwords.txt >> passwords-combined.txt
  ```

- filtered out any duplicates:
  ```
  cat passwords-combined.txt | awk '!x[$0]++' > passwords-combined-deduped.txt
  ```

minglis:
👍

PR #952, opened 2016-09-27
AJAX the activity page

quis:
The activity page looks the same as the job page. So it should behave the same and not need manually refreshing to see the new stuff.

ajax-activity


minglis:
👍

PR #956, opened 2016-09-29
Give platform admin users visibilty of API pages

quis:
Platform admins should be able to see what’s going on with a service’s API integration, including:
- messages sent
- contents of whitelist
- names of keys

They should also be able to revoke keys in an emergency.

The only thing they _shouldn’t_ be able to do is create new keys (because then they’d be able to send messages as the service).


yahoopete:
👍 

Yep - this has been annoying me....

PR #961, opened 2016-10-03
Add monthly breakdown of usage

quis:
screen shot 2016-09-30 at 15 34 39

***

- shows all the months from start of given financial year to now or end of given financial year (whichever is earliest)
- shows a breakdown of free and paid text messages for each of these months

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/699

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -8,7 +8,9 @@ from flask import (
     url_for,
     session,
     jsonify,
-    current_app
+    current_app,
+    request,
+    abort
 )
 from flask_login import login_required
 
@@ -80,8 +82,16 @@ def template_history(service_id):
 @login_required
 @user_has_permissions('manage_settings', admin_override=True)
 def usage(service_id):
+    try:
+        year = int(request.args.get('year', 2016))
+    except ValueError:
+        abort(404)
     return render_template(
         'views/usage.html',
+        months=get_free_paid_breakdown_for_billable_units(
+            year, service_api_client.get_billable_units(service_id, year)
+        ),
+        year=year,
         **calculate_usage(service_api_client.get_service_usage(service_id)['data'])
     )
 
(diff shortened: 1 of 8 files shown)


app/main/views/dashboard.py
```diff
@@ -195,3 +205,56 @@ def format_weekly_stats_to_list(historical_stats):
...
+        if previous_cumulative < allowance:
+            return {
+                'paid': monthly_usage - (allowance - previous_cumulative),
+                'free': allowance - previous_cumulative
+            }
+        else:
+            return {
+                'paid': monthly_usage,
+                'free': 0
+            }
```
leohemsted: this block can be un-nested using 

```
if
elif
else
```

app/main/views/dashboard.py
```diff
@@ -195,3 +205,56 @@ def format_weekly_stats_to_list(historical_stats):
...
+                cumulative, previous_cumulative, monthly_usage
+            )
+        )
+
+
+def get_free_paid_breakdown_for_month(
+    cumulative,
+    previous_cumulative,
+    monthly_usage,
+    allowance=250000
```
leohemsted: this should be a local variable rather than a parameter since changing this isn't expected usage

leohemsted: this should be a local variable rather than a parameter since changing this isn't expected usage

app/main/views/dashboard.py
```diff
@@ -195,3 +205,56 @@ def format_weekly_stats_to_list(historical_stats):
...
+    for month in get_months_for_financial_year(year):
+        previous_cumulative = cumulative
+        monthly_usage = billable_units.get(month, 0)
+        cumulative += monthly_usage
+        yield dict(
+            name=month,
+            **get_free_paid_breakdown_for_month(
+                cumulative, previous_cumulative, monthly_usage
+            )
+        )
```
leohemsted: i'd rather we avoid `**` unpacking where possible, I think [personally] that it's a bit ugly and it's not immediately apparent - i'd probably rather `get_free_paid_breakdown_for_month` returned a tuple, and then we 
```
yield {
    'name': month,
    'paid': breakdown[0],
    'free': breakdown[1]
}

File: app/main/views/dashboard.py (hunk 1 above)
Comment (leohemsted) [inline app/main/views/dashboard.py]: this block can be un-nested using 

```
if
elif
else
```

app/main/views/dashboard.py
leohemsted: hmm maybe. i'd suggest a named tuple but they're silly. keep as is

quis:
@leohemsted This only looks at whatever’s in the billable units column, so whatever happens to that when you use research mode or a test key.

app/main/views/dashboard.py
quis: Fixed and squashed out

app/main/views/dashboard.py
quis: 👌

PR #962, opened 2016-10-05
Re-add Google Analytics

quis:
## Revert "Remove Google Analytics"

This reverts commit f31170f.

## Remove UUIDs from URLs sent to Google Analytics

If all our URLs are unique (because they contain service/job/template IDs) then it makes it hard to aggregate how users are behaving across a range of services/jobs/templates.

This commit replaces anything that looks like a UUID in a URL with `…`.

_Regex taken from http://stackoverflow.com/a/18516125/147318_

screen shot 2016-10-05 at 10 58 03


yahoopete:
:+1:

PR #965, opened 2016-10-06
Limit jobs on the dashboard

leohemsted:
The functional tests were running into problems due to the large amount of jobs that they create on live leading to browser slowdown. To combat this, we've added pagination to the jobs page. It's a pain to put this in the dashboard so we just limit the query to the first page (of 50)

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -1,8 +1,6 @@
 # -*- coding: utf-8 -*-
 import ago
-import time
 import dateutil
-import json
 from orderedset import OrderedSet
 from datetime import datetime, timedelta, timezone
 from itertools import chain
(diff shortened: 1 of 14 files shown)


leohemsted:
hmm should probably paginate the jobs page - cos that'll be limited to 50 and probably *should* show everything

quis:
Yeah, should follow the same next/previous pattern used on the activity page.

leohemsted:
@quis added pagination to the jobs page:

image

app/main/views/jobs.py
```diff
@@ -65,12 +65,33 @@ def _set_status_filters(filter_args):
...
 @user_has_permissions('view_activity', admin_override=True)
 def view_jobs(service_id):
+    page = int(request.args.get('page', 1))
+    # all but scheduled and cancelled
+    statuses_to_display = [
+        'pending',
+        'in progress',
+        'finished',
+        'sending limits exceeded',
+    ]
```
quis: Would be good to factor this out so they’re only hard coded once.

quis: And you could write the comment as code then:
`statuses_to_display = all_statuses - {'scheduled', 'cancelled'}`

PR #968, opened 2016-10-07
Present API key as a combination of secret and service ID

quis:
image

***

In research we’ve seen people mix up the service ID and API key because they’re both 36 character UUIDs. We can’t get rid of the service ID because it’s used to look up the API key.

Instead, we should change API key to be one long string, which contains both the service ID, API key and the name of the key. 

We still need to keep the old, separate, key and service ID for a while until people have updated their clients. But they’re now both on this page, rather than on two separate pages, which should make for less fussing anyway.

**This shouldn’t be rolled out until the new clients are available**

- [x] alphagov/notifications-python-client#36
- [x] alphagov/notifications-node-client#10
- [x] alphagov/notifications-ruby-client#15
- [x] alphagov/notifications-java-client#38
- [x] https://github.com/alphagov/notifications-php-client/pull/3

### DIFF
--- a/app/main/views/api_keys.py
+++ b/app/main/views/api_keys.py
@@ -3,7 +3,7 @@ from flask_login import login_required
 from app.main import main
 from app.main.forms import CreateKeyForm, Whitelist
 from app import api_key_api_client, service_api_client, notification_api_client, current_service
-from app.utils import user_has_permissions
+from app.utils import user_has_permissions, email_safe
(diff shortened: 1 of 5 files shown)


leohemsted:
this'll probably also break functional tests so make sure you run them [after we solve current team vs live api key thing]

leohemsted:
@quis pushed fix for broken test (though would probably have been solved just by re-runnign anyway)

PR #970, opened 2016-10-07
Make whitelist available to live services

quis:
image

Why would a live service need the whitelist? Because the team key will also let you send to members of your whitelist (so this commit relabels it to say so).

image

## Depends on

- [x] https://github.com/alphagov/notifications-api/pull/705


leohemsted:
are there any other docs about? theres `/docs/index.md` - is that actually used anywhere any more?

quis:
@leohemsted Good point, we should delete that file and `410 gone` `/documentation`

PR #974, opened 2016-10-10
notification-utils 9.0.5

leohemsted:
Uses more restrictive email regex to prevent people adding emails with leading quotes, semicolons, etc, that SES rejects and end up as a Technical Failure


quis:
https://github.com/alphagov/notifications-utils/compare/9.0.2...9.0.5 when the tag is there…

PR #978, opened 2016-10-11
remove filtering of test messages from jobs

leohemsted:
actually just code cleanup since this code is now duplicated on the api side

* [x] ~~depends on https://github.com/alphagov/notifications-api/pull/709~~


quis:
Rebased…

PR #979, opened 2016-10-12
Allow a job to be scheduled any time in next 4 days

quis:
4days-final
---

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/711

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -117,6 +117,7 @@ def create_app():
     application.add_template_filter(format_date)
     application.add_template_filter(format_date_normal)
     application.add_template_filter(format_date_short)
+    application.add_template_filter(format_datetime_relative)
     application.add_template_filter(format_delta)
     application.add_template_filter(format_notification_status)
     application.add_template_filter(format_notification_status_as_time)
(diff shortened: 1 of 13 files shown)


app/__init__.py
```diff
@@ -232,6 +233,23 @@ def format_datetime_short(date):
...
+def format_datetime_relative(date):
+    return '{} at {}'.format(
+        get_human_day(date),
+        format_time(date)
+    )
+
+
+def get_human_day(time):
+    #  Add 1 hour to get ‘midnight today’ instead of ‘midnight tomorrow’
+    time = (gmt_timezones(time) - timedelta(hours=1)).strftime('%A')
```
servingUpAces: Will this work after British Summer Time?

quis: @servingupaces Yeah, that's what the GMT timezones bit handles.

app/assets/javascripts/radioSelect.js
```diff
@@ -2,88 +2,144 @@
...
-    .filter(
-      (index, element) => $(element).not(':has(:checked)')
-    ).removeClass('selected');
-
-  var refocus = $element => setTimeout(
-    () => $element.blur().trigger('focus'),
-    10
-  );
+  let states = {
+    'initial': Hogan.compile(`
```
quis: Option 1: Load the templates via AJAX. This would be too slow.

Option 2: Inline the templates, eg `<script type='text/template'>…</script>`. Hogan doesn’t support this natively, so we’d have to write extra code, and it would also be slower.

So I think that while this is _slightly_ ugly, it’s the least worst solution.

PR #984, opened 2016-10-13
Send an email to the user when they change email address

servingUpAces:
This PR changes the flow to change an email address.
Once the user enter their password, they are told "Check your email".
An email has been sent to them containing a link to notify which contains an encrypted token.
The encrypted token contains the user id and new email address. Once the link is clicked the user's email address is updated to the new email address.
They are redirected to the /user-profile page.

Also in this commit is an update from flask.ext.login to flask_login.

### DIFF
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -1,19 +1,21 @@
+import json
+
 from flask import (
     render_template,
     redirect,
     url_for,
-    session
-)
+    session,
+    current_app)
 
-from flask.ext.login import current_user
-from flask_login import login_required
+from flask_login import login_required, current_user
+from notifications_utils.url_safe_token import check_token
+
 from app.main import main
 
 from app.main.forms import (
     ChangePasswordForm,
     ChangeNameForm,
     ChangeEmailForm,
-    ConfirmEmailForm,
     ChangeMobileNumberForm,
     ConfirmMobileNumberForm,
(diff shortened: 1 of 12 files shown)


app/main/views/user_profile.py
```diff
@@ -1,19 +1,21 @@
...
 
-from flask.ext.login import current_user
-from flask_login import login_required
 from app.main import main
 
 from app.main.forms import (
     ChangePasswordForm,
     ChangeNameForm,
     ChangeEmailForm,
-    ConfirmEmailForm,
```
leohemsted: can we remove this form then?

app/main/views/user_profile.py
```diff
@@ -95,6 +96,26 @@ def _check_password(pwd):
...
+                             current_app.config['DANGEROUS_SALT'],
+                             current_app.config['EMAIL_EXPIRY_SECONDS'])
+    token_data = json.loads(token_data)
+    user_id = token_data['user_id']
+    new_email = token_data['email']
+    user = user_api_client.get_user(user_id)
+    user.email_address = new_email
+    user_api_client.update_user(user)
+    if session.get(NEW_EMAIL, None):
+        del session[NEW_EMAIL]
```
leohemsted: `session.pop(NEW_EMAIL, None)` does the same without needing an if statement around it

PR #985, opened 2016-10-14
Error if a CSV file contains more than 50,000 rows

quis:
We want to limit the number of rows someone can have in a job, because it gets too slow to process the file otherwise.

This should be the first error that a user sees, because we can’t work out if there are other errors until they’ve got the file down to a processable size.

This also means adding a message to say that the file can’t be displayed if it doesn’t contain any processed rows.

**Story**

https://www.pivotaltracker.com/story/show/129830161

**Depends on**

- [x] https://github.com/alphagov/notifications-utils/pull/73

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -15,7 +15,22 @@
 
 {% block maincolumn_content %}
 
-  {% if not recipients.has_recipient_column %}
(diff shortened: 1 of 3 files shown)


imdadahad:
👍

PR #987, opened 2016-10-17
Make ‘too many rows’ error actually work

quis:
Was using the wrong attribute name on the instance of `RecipientCSV` (needed to match https://github.com/alphagov/notifications-utils/blob/19df1eb4ef061674708fe9807bfedaa507dd7186/notifications_utils/recipients.py#L156 )

The ‘can’t show file contents’ message wasn’t showing up, now it is.

image

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -15,7 +15,7 @@
 
 {% block maincolumn_content %}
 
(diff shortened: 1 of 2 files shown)


imdadahad:
Tested locally with both sms and email, works 👍

PR #992, opened 2016-10-19
Add error logging

leohemsted:
(we weren't previously logging anything on error!)

[also removed `monotonic` cos it's part of python stdlib now]


minglis:
👍

PR #995, opened 2016-10-21
Fix 500s

imdadahad:
If a user was on the notifications dashboard (email or sms) and had signed out of the admin app on another tab, 500 responses would be returned.

This changes so we check the user is authenticated first when grabbing their permission set and return the appropriate codes. Additionally further polling of notifications is stopped upon failure.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -42,7 +42,6 @@ def user_has_permissions(*permissions, admin_override=False, any_=False):
         @wraps(func)
         def wrap_func(*args, **kwargs):
             from flask_login import current_user
-
             if current_user and current_user.is_authenticated:
                 if current_user.has_permissions(
                     permissions=permissions,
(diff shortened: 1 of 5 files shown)


quis:
Seems like the dashboard AJAX endpoint returns a `302` (which isn’t technically a failure) if you’re not signed in.

app/utils.py
```diff
@@ -42,11 +42,17 @@ def wrap(func):
...
-                return func(*args, **kwargs)
+            if current_user and current_user.is_authenticated:
+                if current_user.has_permissions(
+                    permissions=permissions,
+                    admin_override=admin_override,
+                    any_=any_
+                ):
+                    return func(*args, **kwargs)
+                else:
+                    abort(403)
```
leohemsted: can you add some tests for this decorator

imdadahad: Just added 👍

tests/app/main/test_permissions.py
```diff
@@ -107,6 +108,17 @@ def test_platform_admin_user_can_not_access_page(app_,
         admin_override=False)
 
 
+def test_user_with_permissions_returns_401_unauthenticated_user(app_):
```
leohemsted: can you change name of test - cos there's not actually a user with permissions, there's no user at all. eg `test_no_user_returns_401_unauth(app_)` or something

PR #998, opened 2016-10-26
Invite nongov users

imdadahad:
At the moment, the only way for us to allow users without a *.gov.uk email address to use Notify is to whitelist their entire domain potentially disallowing access for gov users.

This changes makes it possible to *invite* users from any domain into a service.  However, they still cannot register without an invite if they are not on the whitelist.


app/templates/views/choose-service.html
```diff
@@ -12,12 +12,14 @@ <h1 class="heading-large">
   </h1>
 
   {{ browse_list(services) }}
+  {% if can_add_service %}
   {{ browse_list([
```
leohemsted: can you indent this section inside the `{% if can_add_service %}` block

app/utils.py
```diff
@@ -201,3 +201,9 @@ def from_file(cls, file_content, filename=''):
 
 def get_help_argument():
     return request.args.get('help') if request.args.get('help') in ('1', '2', '3') else None
+
+
+def user_in_whitelist(email_address):
+    valid_domains = current_app.config.get('EMAIL_DOMAIN_REGEXES', [])
+    email_regex = "[^\@^\s]+@([^@^\\.^\\s]+\.)*({})$".format("|".join(valid_domains))
+    return bool(re.match(email_regex, email_address))
```
leohemsted: can you refactor `validators.py::ValidEmailDomainRegex.__call__` to use this function so we don't have code repeated everywhere

also this is fucking gross and there must be a better way of doing it. we appear to already check use the regular email validator to check for the validity of the email - so hopefully you should be able to replace this with  `return any(email_address.endswith(domain) for domain in valid_domains)` and not worry about whether it's a valid email in this check .

[if that works, then can you rename ValidEmailDomainRegex to be `GovEmail` or something similar

leohemsted: also, `current_app.config.EMAIL_DOMAIN_REGEXES` is always defined so shouldn't need call `.get`

also, if it works without the regex, you'll need to update the `config.EMAIL_DOMAIN_REGEXES` to be non-regex escaped (ie `\.` -> `.`)

tests/app/main/views/test_add_service.py
```diff
@@ -101,3 +103,11 @@ def test_should_return_form_errors_with_duplicate_service_name_regardless_of_cas
...
             app.service_api_client.find_all_service_email_from.assert_called_once_with()
             assert not mock_create_service.called
+
+
+def test_non_whitelist_user_cannot_add_service(app_, mocker, client, fake_uuid):
+    non_whitelist_user = create_active_user(fake_uuid, 'someuser@notonwhitelist.com')
+    client.login(non_whitelist_user, mocker)
+    assert not user_in_whitelist(non_whitelist_user.email_address)
+    response = client.get(url_for('main.add_service'))
+    assert response.status_code == 403
```
leohemsted: can you also add tests for the following

* non-whitelisted users cant see the add service button from the list of services page
* non-whitelisted users get a 403 if they post to the add_service url (think this would currently work, but shouldn't)
* non-whitelisted users can't create an account
* non-whitelisted users can be invited by a whitelisted user

servingUpAces:
Invited non-gov users can not change their email address. However, I am not sure that they should be able to. Perhaps in users-profile they should not see edit button.

app/main/views/add_service.py
```diff
@@ -61,8 +73,11 @@ def add_service():
             help=1
         ))
     else:
-        return render_template(
-            'views/add-service.html',
-            form=form,
-            heading=heading
-        )
+        if not user_in_whitelist(current_user.email_address):
+            abort(403)
```
imdadahad: Silly me, moved and refactored.

app/templates/views/choose-service.html
imdadahad: Done

tests/app/main/views/test_add_service.py
imdadahad: Added 1 and 2 on the list. 

3 already exists in test_should_return_200_when_email_is_not_gov_uk and 4 was already added to test_invite_user

app/utils.py
```diff
@@ -201,3 +201,9 @@ def from_file(cls, file_content, filename=''):
 
 def get_help_argument():
     return request.args.get('help') if request.args.get('help') in ('1', '2', '3') else None
+
+
+def is_gov_user(email_address):
+    valid_domains = current_app.config['EMAIL_DOMAIN_REGEXES']
+    email_regex = "[.|@]({})$".format("|".join(valid_domains))
```
leohemsted: this should be an escaped `\.`

leohemsted: (and then make the string `r'...'` so that it doesn't moan about the escaping)

config.py
```diff
@@ -44,24 +44,25 @@ class Config(object):
...
+        "nhs.net",
+        "police.uk",
+        "kainos.com",
+        "salesforce.com",
+        "bitzesty.com",
+        "dclgdatamart.co.uk",
+        "valtech.co.uk",
+        "cgi.com",
+        "capita.co.uk",
+        "ucds.email"
```
leohemsted: since we're still using regex, these'll need to be `\.` again (and `r` prefixed). sorry

tests/app/main/views/test_add_service.py
```diff
@@ -101,3 +102,25 @@ def test_should_return_form_errors_with_duplicate_service_name_regardless_of_cas
...
+                                                              mock_login,
+                                                              mock_get_non_govuser,
+                                                              api_nongov_user_active):
+    client.login(api_nongov_user_active)
+    assert not is_gov_user(api_nongov_user_active.email_address)
+    response = client.get(url_for('main.add_service'))
+    assert response.status_code == 403
+
+
+def test_non_whitelist_user_cannot_create_service(app_,
```
leohemsted: unused fixture `app_` (also mock_get_non_govuser? but that might actually be used, if so feel free to keep it. these admin fixtures give me the heebie jeebies)

leohemsted: same comments apply to test above too

app/utils.py
```diff
@@ -201,3 +201,9 @@ def from_file(cls, file_content, filename=''):
 
 def get_help_argument():
     return request.args.get('help') if request.args.get('help') in ('1', '2', '3') else None
+
+
+def is_gov_user(email_address):
```
leohemsted: we should definitely have some fairly extensive separate tests to make sure this regex is robust. speak to me if you're not sure what test cases to have

imdadahad: test_validators.py already has an extensive set that tests against this

tests/app/main/views/test_add_service.py
imdadahad: I'll remove the _app, the mock_get_non_govuser is actually used otherwise it behaves like a normal user with _'non-gov'_ user permissions

app/utils.py
imdadahad: Updated 👍

config.py
imdadahad: Changed

config.py
```diff
@@ -44,23 +44,23 @@ class Config(object):
...
+        "nhs\.net",
+        "police\.uk",
+        "kainos\.com",
+        "salesforce\.com",
+        "bitzesty\.com",
+        "dclgdatamart\.co\.uk",
+        "valtech\.co\.uk",
+        "cgi\.com",
+        "capita\.co\.uk",
         "ucds.email"
```
leohemsted: missed escaping this `.`

leohemsted: also you should prefix all these strings with `r` to signify that they're raw strings, and python shouldn't treat the backslashes as escape characters itself

leohemsted: you can prove this works by adding a failing test case (`test@test.ucdszemail`) for example, which would currently pass i believe

imdadahad: Interesting that currently in master non of these use raw strings. Updated anyhow.

leohemsted: yeah - with `\.` it won't do anything bad but if it were, say, `\n` it would turn into a newline rather than a literal `\` followed by a `n` for the regex parser to pick up

imdadahad: yeah, would be surprised if a newline or another escape character appeared here but better to be safe 👍

PR #1001, opened 2016-10-27
ensure robustness of email_safe function

leohemsted:
* remove leading, trailing, or consecutive periods
* strip unicode accents, umlauts, diacritics etc

image

image


imdadahad:
👍

PR #1003, opened 2016-10-28
Use the host uid and gid for running the tests in the Docker container

bandesz:
Testing for now


bandesz:
Closing this for now, because this is imposing the same security risk as running as root

PR #1006, opened 2016-11-01
Make certain questions on request to go live radio buttons

quis:
Before | After
--- | ---
image image | image image


quis:
Have done some more work on this with Sheryll. Main change is making the MOU bit a question, rather than a banner:

image

yahoopete:
:+1:

PR #1008, opened 2016-11-02
Deactivate service

leohemsted:
# lets deactivate a service!
image

# no, really!
image

## it renames your service!
image

## it deletes your templates!
image

## it revokes your api keys!
image

## it's irreversible!
image

- [ ] https://github.com/alphagov/notifications-api/pull/723


quis:
As discussed yesterday:
- [ ] Make the button say ‘Archive service’
- [ ] Add a confirmation loop similar to when you delete a template

leohemsted:
@quis updated

PR #1011, opened 2016-11-03
Update user profile

imdadahad:
Previous to this we were passing the whole user object when making any update to their profile. This makes it so that we only pass in the changed attribute instead. 

This relies on:

- [x] https://github.com/alphagov/notifications-api/pull/725

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -3,6 +3,12 @@ from notifications_python_client.errors import HTTPError
 
 from app.notify_client.models import User
 
+ALLOWED_ATTRIBUTES = {
+    'name',
+    'email_address',
+    'mobile_number'
+}
+
 
 class UserApiClient(BaseAPIClient):
     def __init__(self):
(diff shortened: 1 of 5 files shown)


app/notify_client/user_api_client.py
```diff
@@ -47,9 +47,9 @@ def get_users(self):
             users.append(User(user, max_failed_login_count=self.max_failed_login_count))
         return users
 
-    def update_user(self, user):
-        data = user.serialize()
-        url = "/user/{}".format(user.id)
+    def update_user(self, user_id, **kwargs):
+        data = dict(**kwargs)
```
quis: This is fine, but you should probably guard against accidentally updating a user’s `id` or similar.

See https://github.com/alphagov/notifications-admin/blob/master/app/notify_client/service_api_client.py#L83-L100 for what we do with services.

servingUpAces:
This will not work with the latest version of the api, as it will fail the user_schema validation.

PR #1012, opened 2016-11-04
Make relationship between keys, whitelist, trial mode and team clearer

quis:
image

***

image

***

image

***

image

***

image

***

image


yahoopete:
:+1:

PR #1025, opened 2016-11-14
Show email and sms counts for each service on platform admin

imdadahad:
Before: 

screen shot 2016-11-14 at 14 51 45

After:

screen shot 2016-11-14 at 14 52 43

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -118,6 +118,10 @@
 
   }
 
+  &-noborder {
+    border: 0px;
+  }
+
   &-index {
     width: 15px;
   }
@@ -157,6 +161,12 @@
 .table-field-right-aligned {
   @extend %table-field;
   text-align: right;
+}
+
+
+.table-row-group {
+  border-top: 1px solid #BFC1C3;
+  border-bottom: 1px solid #BFC1C3;
 }
 
 .table-empty-message {
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -165,8 +165,8 @@
 
 
 .table-row-group {
-  border-top: 1px solid #BFC1C3;
-  border-bottom: 1px solid #BFC1C3;
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
 }
 
 .table-empty-message {
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -52,7 +52,7 @@
 {% macro row_group(id=None) %}
   <tbody class="table-row-group" {% if id %}id="{{id}}"{% endif %}>
     {{ caller() }}
-  </tr>
+  </tbody>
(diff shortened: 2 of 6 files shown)


app/assets/stylesheets/components/table.scss
```diff
@@ -159,6 +163,12 @@
   text-align: right;
 }
 
+
+.table-row-group {
+  border-top: 1px solid #BFC1C3;
+  border-bottom: 1px solid #BFC1C3;
```
quis: Should use SASS variables rather than hard-coded hex values. `$border-colour` is the one you’re looking for.

app/templates/components/table.html
```diff
@@ -49,10 +49,20 @@
...
 {%- endmacro %}
 
-{% macro field(align='left', status='') -%}
-  <td class="table-field{% if align == 'right' %}-right-aligned{% endif %}">
-    <span class="{{ 'table-field-status-' + status if status }}">{{ caller() }}</span>
-  </td>
+{% macro row_group(id=None) %}
+  <tbody class="table-row-group" {% if id %}id="{{id}}"{% endif %}>
+    {{ caller() }}
+  </tr>
```
leohemsted: opening tag `tbody`, closing tag `tr`

app/assets/stylesheets/components/table.scss
imdadahad: Just fixed

app/templates/views/platform-admin.html
```diff
@@ -2,49 +2,88 @@
...
+
+          {% call field(align='right', border=False) %}
+            {{ big_number(service['stats']['sms']['requested'], smaller=True) }}
+          {% endcall %}
+
+          {% call field(align='right', border=False) %}
+            {{ big_number(service['stats']['sms']['delivered'], smaller=True) }}
+          {% endcall %}
+
+          {% call field(align='right', border=False) %}
```
leohemsted: before this had a `status='error' if 0 else ''` which is complete garbage and always fails the check cos `if 0` - but i get that what it was meant to do is highlight any errors in red - do we want to do that?

`status='error' if service['stats']['sms']['failed'] else ''` probably

app/templates/views/platform-admin.html
```diff
@@ -2,49 +2,88 @@
...
+
+      {% call row_group() %}
+
+        {% call row() %}
+
+          {% call field(border=False) %}
+            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name'] }}</a>
+          {% endcall %}
+
+          {% call field(border=False) %}
```
leohemsted: i'd be inclined to put this call and the next 3 stats calls into a macro, which you pass in either 'sms' or 'email' to, to save repeating ourselves

app/templates/views/platform-admin.html
imdadahad: Yeah think we still want that, added

PR #1028, opened 2016-11-17
Render address block in letters

quis:
Depends on and implements: https://github.com/alphagov/notifications-utils/pull/84/files

Also adds a letter preview to a few places where it was missing.

***

image

image


quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/1037

PR #1035, opened 2016-11-25
New product page part 1

quis:
Still to do:

- [x] work through the content with Sheryll
- [x] get final versions of the illustrations from Stephen (the bottom two I just hacked together in Photoshop)
- [x] Look into using SVG instead of PNG for the illustrations
- [x] Make the beta badge less hacky

***

screencapture-localhost-6012-1480092344353

Main changes:

## Make the blue section of the page full-width and remove the phones

The phones image got us so far, but the rest of the product has evolved beyond it now.

The needs we identified for the top, blue section of the page are:
- what is Notify?
- can I use it?
- can I test it out?
- how do I create a log in

## Add a grid of features to the home page

This is the first step towards a fully-fledged ‘product’ page.

The needs for the next section of this page are:
- how will Notify help me work?
- will it work with my service?

This commit shows 4 features of Notify on the home page which address those needs. They are illustrated because:

1. We want to catch people’s attention – users are reluctant to scroll on this page because they just want to click the ‘create account button’. But we hypothesize that they will get on better with Notify if they look at some of this stuff first.

2. The concepts that they’re talking about are hard to explain with just words because they’re quite abstract. The illustrations help us be more specific.

3. Feedback we got from user research was that the product page didn’t gives users any sense of what it was like to actually use Notify.


quis:
Updated with:
- New `<h1>`
- More space between items in the blue panel
- The four features grouped more tightly
- ‘Sign in‘ instead of ‘Sign into your account’ in the header

***

screencapture-localhost-6012-1480345381489

quis:
Added an illustration to the ‘What is Notify’ section:

image

app/templates/views/signedout.html
```diff
@@ -1,66 +1,92 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <div class="product-page-section">
+    <div class="grid-row">
+      <div class="column-half">
+        <h2>
+          Get control of your content
```
pcraig3: Surely they had control at some point but feel as though they don't anymore.
Maybe "Take back control of your content"?

quis: Make your content great again?

leohemsted: 😑

quis:
Beta badge implemented properly:

image

quis:
Trying some more things with the wording in the feature area (making the sentences shorter and putting each one on its own line):

image

quis:
Final SVG illustrations from Stephen:

image

quis:
Latest changes from sitting down with Stephen, Sheryll and disembodied Gwen:

screencapture-localhost-6012-1481126052405

leohemsted:
oh man theres me making use of my famed pointing skills

PR #1036, opened 2016-11-25
Remove the red header meaning signed in to admin

quis:
> Make the header bar red
>
> Red for admin is a good reckon.

– 286fc308d9a64ab8effa3b9a575591b2f2f942da (part of https://github.com/alphagov/notifications-admin/pull/130)

Starting to think it’s not such a good reckon. Users could take a guess at what it meant, but they often guessed wrong or were mystified.

However, changing the colour of the header bar _is_ useful for us internally to see which environment we’re in. So this commit makes three changes:

1. On live, the header bar is always standard GOV.UK blue

2. On other environments, the header bar is some other colour (local is very different, staging and preview are related colours)

3. If an environment has a different header colour, it has it even when you’re not logged in.

***

**Local**

image

**Preview**

image

**Staging**

image

**Live**

image

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -5,11 +5,9 @@
   <!--[if gt IE 8]><!-->
   <link rel="stylesheet" media="screen" href="{{ asset_url('stylesheets/main.css') }}" />
   <!--<![endif]-->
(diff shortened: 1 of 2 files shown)


yahoopete:
:+1:

PR #1037, opened 2016-11-28
Render letter templates as PNGs and PDFs

quis:
Depends on and implements:
- [x] alphagov/notifications-utils#84
- [x] alphagov/notifications-utils#86
- [x] https://github.com/alphagov/notifications-utils/pull/94
- [x] https://github.com/alphagov/notifications-aws/pull/106
- [x] https://github.com/alphagov/notifications-utils/pull/100

---

image

image

## Add an endpoint for generating a PDF of a letter

Previewing a letter is different to previewing an email or text message because:

- a letter has a layout
- the layout is fixed, ie it doesn’t depend on the user’s device
- the ‘send yourself a test’ feature won’t be as useful because it has a lead time, so the feedback loop will be much longer

For these reasons a HTML-only preview of the letter won’t be enough (we don’t reckon). A PDF is more appropriate because:

- it can replicate the layout of the letter exactly
- it has a fixed size
- it is a print format, so the user could even print themselves a copy locally to get a feel for how it will look

This commit makes use of [Flask WeasyPrint](https://pythonhosted.org/Flask-WeasyPrint/) to take a HTML representation of the letter, convert it to a PDF and serve it back to the user.

The actual work to generate the HTML and specify the layout is done in utils,  same as we do for rendering other messages.

## Add endpoint for generating an image of a letter

The PDF preview is all good, but it’s hard, finickity, and feels dirty to embed a PDF in a web page. It’s a more natural thing to embed an image in a web page.

So this commit adds another endpoint to return an image of a letter template. It generates this image from the PDF preview, so the stack looks like:

1. `template.png` (generated in admin)
2. `template.pdf` (generated in admin)
3. HTML preview (generated by a `Renderer` in utils)
4. `Template` instance
5. serialised template from API
6. Template stored in database

The library used to convert the PDF to an image is [Wand](http://docs.wand-py.org/en/0.4.4/), which binds to ImageMagick underneath. So in order to get this working locally on a Mac you will probably need to do: `brew install imagemagick ghostscript cairo pango`.

It should work on Ubuntu/EC2 once this is merged: https://github.com/alphagov/notifications-aws/pull/106

## 	Use PDF link renderer to preview letters

Shows the image preview on the page, and links to the PDF.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,4 @@ wand==0.4.4
 
 git+https://github.com/alphagov/notifications-python-client.git@3.0.1#egg=notifications-python-client==3.0.1
 
-git+https://github.com/alphagov/notifications-utils.git@12.1.0#egg=notifications-utils==12.1.0
+git+https://github.com/alphagov/notifications-utils.git@12.1.1#egg=notifications-utils==12.1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,4 @@ wand==0.4.4
 
 git+https://github.com/alphagov/notifications-python-client.git@3.0.1#egg=notifications-python-client==3.0.1
 
-git+https://github.com/alphagov/notifications-utils.git@11.0.2#egg=notifications-utils==11.0.2
+git+https://github.com/alphagov/notifications-utils.git@11.1.0#egg=notifications-utils==11.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ Flask-WTF==0.11
 Flask-Login==0.3.2
 Flask-WeasyPrint==0.5
 git+https://github.com/quis/WeasyPrint.git@older-html5lib#egg=WeasyPrint==0.33
-html5lib==0.9999999
+html5lib==1.0b8
 credstash==1.8.0
 boto3==1.3.0
 Pygments==2.0.2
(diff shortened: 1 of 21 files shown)


requirements.txt
```diff
@@ -3,6 +3,9 @@ Flask==0.10.1
 Flask-Script==2.0.5
 Flask-WTF==0.11
 Flask-Login==0.3.2
+Flask-WeasyPrint==0.5
+git+https://github.com/quis/WeasyPrint.git@older-html5lib#egg=WeasyPrint==0.33
+html5lib==0.9999999
```
leohemsted: ah yes i forgot about how braindead this library was

leohemsted: https://github.com/html5lib/html5lib-python/issues/282

ah they've provided alternative versions. can you try: `html5lib==1.0b10`?

tests/app/main/views/test_templates.py
```diff
@@ -39,6 +40,39 @@ def test_should_show_page_for_one_template(
...
+    api_user_active,
+    mock_login,
+    mock_get_service,
+    mock_get_service_email_template,
+    mock_get_user,
+    mock_get_user_by_email,
+    mock_has_permissions,
+    fake_uuid
+):
+    client.login(api_user_active)
```
leohemsted: just fyi i made `logged_in_client` in my quest for better test fixtures to help remove some of these mocks :) it includes all the login decorators so you dont need to have them in your test (though it only currently works with api_user_active)

PR #1045, opened 2016-12-05
Add option to exclude test key usage in stats on platform admin page

jennyd:
Depends on https://github.com/alphagov/notifications-api/pull/758

This page currently includes all notifications for all services, including those sent using a test key. Stats on all other pages exclude test key usage, though, which can lead to confusion for admins comparing numbers between pages. Adding the option to toggle between including and excluding test key usage on the platform admin page gives context for this difference, and allows admins to see live usage of the platform as well as load associated with test key usage.

Also move "Today" heading and providers link to the left column and header respectively.

Thanks @quis for design help ✨ 

Before:
screen shot 2016-12-05 at 17 15 39

After:
screen shot 2016-12-05 at 17 16 06

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -1,6 +1,9 @@
 import itertools
 
-from flask import render_template
+from flask import (
+    render_template,
+    request
+)
 from flask_login import login_required
 
 from app import service_api_client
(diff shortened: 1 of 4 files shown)


jennyd:
After some discussion I've added another commit to move the day and test key messages to below the `h1` instead of having them in a new column:

screen shot 2016-12-06 at 12 48 13

PR #1047, opened 2016-12-07
Invitation expiry msg

servingUpAces:
Added a new 400.html error page, this error page will only be used when we get an HTTPError from the api.

This was done to solve the problem of invitations expiring and the user getting a 500 error message, which says to try again.

screen shot 2016-12-13 at 09 39 14


imdadahad:
👍

PR #1048, opened 2016-12-07
Fix a couple of things that got missed on the product page

quis:
Couple of things that got missed:
- `alt` text for the images
- redundant CSS


yahoopete:
:+1:

Unless you want to include the video with this commit?

quis:
Will leave the video for another time, need to figure out the accessible media player.

PR #1049, opened 2016-12-08
Run on Paas

bandesz:
Notes:
 - CloudFoundry injects two env variables:
   - VCAP_APPLICATION: app metadata in json format (it contains the space name which will be sandbox/preview/staging/production)
   - VCAP_SERVICES: contains all credentials for 3rd party services in json format (e.g. Postgres db, mmg, firetext, etc.)
 - the logs needs to be written to stdout/stderr

TODO:
 - [x] Sort out logging to stdout/stderr in json format (https://www.pivotaltracker.com/story/show/135951975)


app/config.py
```diff
@@ -111,10 +122,24 @@ class Live(Config):
...
 
 
+class CloudFoundryConfig(Config):
+    pass
+
+
+# CloudFoundry sandbox
+class Sandbox(CloudFoundryConfig):
+    HTTP_PROTOCOL = 'https'
+    HEADER_COLOUR = '#F499BE'  # $baby-pink
```
leohemsted: @quis do we want to change the colour to distinguish from preview? 

(note: this doesn't block this PR i think, it can be changed separately if we care)

PR #1051, opened 2016-12-09
users can't edit inactive services

leohemsted:
if a service gets archived, then we already block them from using any unrevoked api keys (https://github.com/alphagov/notifications-api/blob/master/app/authentication/auth.py#L66), but previously didn't stop them creating new templates, uploading CSVs, etc.

Now, block any calls to the API if the service is inactive and the user's not a platform admin

### DIFF
--- a/tests/app/notify_client/test_notify_admin_api_client.py
+++ b/tests/app/notify_client/test_notify_admin_api_client.py
@@ -1,0 +1,74 @@
+import uuid
+import pytest
+from unittest.mock import patch
+
+import werkzeug
+
+from tests import service_json
+from tests.conftest import api_user_active, platform_admin_user
+from app.notify_client import NotifyAdminAPIClient
+
+
+@pytest.mark.parametrize('method', [
+    'put',
+    'post',
+    'delete'
+])
+@pytest.mark.parametrize('user', [
+    api_user_active(str(uuid.uuid4())),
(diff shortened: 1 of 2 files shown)


tests/app/notify_client/test_notify_admin_api_client.py
```diff
@@ -0,0 +1,96 @@
...
+    platform_admin_user(str(uuid.uuid4()))
+], ids=['api_user', 'platform_admin'])
+@pytest.mark.parametrize('service', [
+    service_json(active=True),
+    None
+], ids=['active_service', 'no_service'])
+def test_active_service_can_be_modified(app_, method, user, service):
+    api_client = NotifyAdminAPIClient('api_key', 'base_url', 'service_id')
+
+    with app_.test_request_context() as request_context, app_.test_client() as client:
```
imdadahad: Can you use the client in conftest here?

leohemsted: unfortunately not, since I need to modify the request_context object :(

leohemsted: unless in conftest i attach request_context to client as a variable or similar, which i'm loathe to do cos it'll be undocumented

imdadahad: fair enough, looks good 👍

PR #1053, opened 2016-12-14
Switched off session protection

minglis:
Many departments are reporting being logged out during usage of the app
This may be because they present differing IP addresses to Notify on different requests due to outbound proxies and so on
Switching session protection to None means and IP change won't log a user out.

Note this was hard to test on my local build. I could replicate the error on production by switching from my mobile hotspot to wifi and back. However having to use the VPN to sign in (for SQS) from home maybe meant that the issue didn't occur as I couldn't replicate the problem locally (though I did in the office yesterday). So I can't verify this fixes it.

Whoever reviews this would be good to checkit in the office - using different wifi networks.

Thanks

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -99,6 +99,7 @@ def create_app():
     login_manager.init_app(application)
     login_manager.login_view = 'main.sign_in'
     login_manager.login_message_category = 'default'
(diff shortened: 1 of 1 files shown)


yahoopete:
:+1:

PR #1055, opened 2016-12-14
Fix misaligned sidebar on the tour

quis:
Got broken when we moved the service name out of the nav.

## Before

image

## After 

image


yahoopete:
:+1:

PR #1056, opened 2016-12-16
Fix button shadow colour on product page

quis:
The button was inheriting a green box shadow from the default GOV.UK button style, which is typically used on a white background. On the blue background, a darker blue shadow looks better I think.

The 15% darkening is the same as used by the default green button: https://github.com/alphagov/govuk_frontend_toolkit/blob/a3fe44e00ef3e364376a138fbffa8495af9906df/stylesheets/design-patterns/_buttons.scss#L36

Before | After
---|---
image | image

@nickcolley I have no idea now you spotted this, but thanks.

### DIFF
--- a/app/assets/stylesheets/views/product-page.scss
+++ b/app/assets/stylesheets/views/product-page.scss
@@ -36,6 +36,7 @@
 
     .button {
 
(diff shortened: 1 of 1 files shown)


nickcolley:
Nice! 🎉

PR #1060, opened 2016-12-19
Add triaging to the support and feedback flow

quis:
Adds more nuance to the process of raising a ticket, compared to the one-size-fits-all process we had before.

***

support

# Functional changes

TL;DR, as much as possible we should work out how to prioritise/categorise tickets and not put that burden on the user. However, there are some cases where we can’t. This PR adds UI for the cases where we can’t.

## Add a support index page

Our support process is about to get more fully fledged so we’ll need an index page to route people properly.

We reckon that users will also want to know what the support process is, so let’s explain it on this page.

## Rename support link in navigation to save space

Technically it’s _support, feedback and questions_ but that’s way too long. And it’s all support really.

## Split support into two pages

The kind of communications we’re getting at the moment can broadly be broken down into:
- problems
- questions and feedback

We will need to triage problems differently, because they could potentially be urgent/severe/emergency/P1/whatever language we use. Questions or feedback will never be P1.

Two reasons for making the user categorise their tickets themselves:

- Outside of hours we can’t get someone out of bed in order to decide if a ticket is a problem or just feedback

- We can tailor the subsequent pages to whether it’s a problem or feedback (eg showing a link to the status page if the user is having a problem)

This commit let’s users make the choice with a pair of radio buttons.

It also cleans up a bunch of the tests and parameterizes them so we’re testing the flow for both ticket types.

## Don’t ask for a user’s email address if we know it  …

If a user is logged in then we already know their name and email address. So there’s no need for them to fill them again on the support form.

One concern we might have about this is the user not realising we’re doing this, and the feedback form looking like a bit of a black hole. So we’re replaying their email address on this page to reassure them that:
- we know who they are
- and that they’ll get a reply

## Triage tickets based on time of day and services

In business hours all tickets are high priority, ie we will at least acknowledge them within 30 mins.

If we are not in business hours then we need to know if a ticket is serious enough to get someone out of bed. Only the user can tell us this, but we can give them some examples to help them decide.

In addition, out-of-hours tickets are only a priority if the user has live services. Normally we can determine this and do the priority-setting in the background.

If they can’t log in then we can’t determine what services they have. So in this case they will need to use the emergency email address, which only users with live services will have.

The logic for this gets fairly complex. It might be to easier to understand what’s going on by walking through the test cases, which are a bit more declarative.

N.B. Deskpro’s ‘urgency’ is descending, eg 10 is the most urgent and 1 is the least.

## Add a confirmation page

Generally I prefer confirmation pages to the flash message thing (they’re harder to miss). So this commit adds one.

It also adds some logic to this page, so that, depending what the user has told us about the thing they’ve submitted, we can tell them how quickly to expect a response.

## Preserve message in session if we go out of hours

This is a real edge case, but it seems worth handling.

How you’d get to this case:
- it’s 5:29pm and you start to describe the problem you’re having
- it’s 5:31pm and you click ‘submit’
- you’re redirected to the triage page because we’re now out of hours
- you click ‘this is a serious problem’

What would be bad thing to happen:
- you’re back on the message page and all the stuff you’ve written is gone

What would be a good thing to happen:
- we save the message in a session so that you can check it again before sending it


quis:
Final content for the support index page:

image

quis:
Final content for the triage page:

image

quis:
## Require email address for reporting problems

If you report a problem we want to be able to get back to you to find out more information, or to update you on the status of a fix. So it shouldn’t be possible to report a problem without providing an email address.

This commit makes `email_address` a required field when `ticket_type` is problem.

image

quis:
Updated support home page to:
- emphasise that you should subscribe to get updates
- not talk about the emergency email address (you don’t need to know about it at this stage – only if you can’t use the support process to raise a ticket)

***

image

quis:
Updated the out of hours emergencies page to remove the confusing bullet points.

***

image

app/main/views/feedback.py
```diff
@@ -1,26 +1,102 @@
...
+
+    if not form.feedback.data:
+        form.feedback.data = session.pop('feedback_message', '')
+
+    severe = request.args.get('severe')
+
+    urgent = any((
+        in_business_hours(),
+        (ticket_type == 'problem' and convert_to_boolean(severe))
+    ))
```
quis: Tidied up in a3b4b961f3b75a2f7dd6a0d9a89a7b67f9cc8c70

app/main/views/feedback.py
```diff
@@ -37,7 +113,99 @@ def feedback():
...
+    return render_template('views/support/bat-phone.html')
+
+
+@main.route('/support/thanks', methods=['GET', 'POST'])
+def thanks():
+    return render_template(
+        'views/support/thanks.html',
+        urgent=convert_to_boolean(request.args.get('urgent')),
+        anonymous=convert_to_boolean(request.args.get('anonymous')),
+        logged_in=(current_user and current_user.is_authenticated),
```
quis: Yeah, not sure why I needed to do this at one point. Don’t now so removed in 5a750ba0f231efe0f82c80064ed942d5ad749480

tests/app/main/views/test_feedback.py
```diff
@@ -1,94 +1,408 @@
...
-            assert resp.status_code == 200
+
+@freeze_time('2016-12-12 12:00:00.000000')
+@pytest.mark.parametrize('support_type, expected_h1', [
+    ('problem', 'Report a problem'),
+    ('question', 'Ask a question or give feedback'),
+])
+@pytest.mark.parametrize('logged_in, expected_form_field, expected_contact_details', [
+    (True, type(None), 'We’ll reply to test@user.gov.uk'),
+    (True, type(None), 'We’ll reply to test@user.gov.uk'),
```
quis: Removed in c7166216

app/main/views/feedback.py
```diff
@@ -37,7 +113,99 @@ def feedback():
...
+    )
+
+
+def in_business_hours():
+
+    now = datetime.now().replace(tzinfo=pytz.timezone('Europe/London'))
```
quis: Fixed in fb9e7c5d

app/main/views/feedback.py
```diff
@@ -37,7 +113,99 @@ def feedback():
...
+
+    if is_weekend(now) or is_bank_holiday(now):
+        return False
+
+    opening_time = now.replace(hour=9, minute=30, second=0, tzinfo=pytz.timezone('Europe/London'))
+    closing_time = now.replace(hour=17, minute=30, second=0, tzinfo=pytz.timezone('Europe/London'))
```
quis: Fixed in fb9e7c5d

app/main/views/feedback.py
```diff
@@ -37,7 +113,99 @@ def feedback():
...
+        "01/05/2017",
+        "29/05/2017",
+        "28/08/2017",
+        "25/12/2017",
+        "26/12/2017",
+    }
+
+
+def has_live_services(user_id):
+    return any(
```
quis: 😺

tests/app/main/views/test_feedback.py
```diff
@@ -1,94 +1,408 @@
...
+):
+    get_services_mock(mocker, fake_uuid)
+    assert has_live_services(12345) == expected_return_value
+
+
+@pytest.mark.parametrize('when, is_in_business_hours', [
+
+    ('2016-06-06 09:29:59', False),  # opening time, summer and winter
+    ('2016-12-12 09:29:59', False),
+    ('2016-06-06 09:30:00', True),
```
quis: Fixed in fb9e7c5d

app/main/views/feedback.py
```diff
@@ -1,26 +1,102 @@
...
+        in_business_hours(),
+        (ticket_type == 'problem' and convert_to_boolean(severe))
+    ))
+
+    anonymous = all((
+        (not form.email_address.data),
+        (not current_user.is_authenticated),
+    ))
+
+    if needs_triage(ticket_type, severe):
```
quis: Addressed in 02cd0ebf

app/main/views/feedback.py
```diff
@@ -1,26 +1,102 @@
...
+        session['feedback_message'] = form.feedback.data
+        return redirect(url_for('.triage'))
+
+    if needs_escalation(ticket_type, severe):
```
quis: Same fix in 02cd0ebf

app/main/views/feedback.py
```diff
@@ -37,7 +113,99 @@ def feedback():
...
+    return any(
+        service['restricted'] is False
+        for service in service_api_client.get_services({'user_id': user_id})['data']
+    )
+
+
+def needs_triage(ticket_type, severe):
+    return all((
+        ticket_type == 'problem',
+        severe is None,
```
quis: Consistency added in 02cd0ebf

PR #1063, opened 2016-12-20
Add remaining endpoints for PDFs and PNGs

quis:
Depends on:

- [x] https://github.com/alphagov/notifications-utils/pull/103
- [x] https://github.com/alphagov/notifications-utils/pull/105

Template | With CSV
---|---
image | image

***

Right now we can show what a letter template looks like as a PDF or PNG.

This PR completes the work so this is also possible when:

- showing a template with the placeholders replaced
- showing any version of a template

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -96,7 +96,7 @@ class UKMobileNumber(TelField):
         try:
             validate_phone_number(self.data)
         except InvalidPhoneError as e:
-            raise ValidationError(e.message)
+            raise ValidationError(str(e))
 
 
 def mobile_number():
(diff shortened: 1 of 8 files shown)


quis:
## Refactor PNG generating endpoint for code reuse and improve image quality

The PNGs we were generating looked a bit fuzzy, and the text rendering was a bit off.

This is because we weren’t rendering them at a high enough resolution. I’m guessing the default was 72dpi. This commit increases the resolution to 96dpi, which (locally) seems to be enough to give a good-looking preview.

The way we generate PNGs is the same for all the different endpoints, so the code to do this should be shared. This means we can set the resolution in one place and have it change everywhere.

PR #1065, opened 2016-12-22
Fix edge case bug in the invitation flow

servingUpAces:
After investigating a 500 in the admin app, I found an edge case in the invitation registration flow that can cause an error.

The flow is as follows:
If the invited user clicks on the /invitation/<token> link in the email (now on /register-from-invite),
then goes to another browser and registers for Notify.
Coming back to the other browser, submit the form for /register-from-invite.

This PR fixes that bug and adds a couple tests for register_from_invite

### DIFF
--- a/tests/app/main/views/test_register.py
+++ b/tests/app/main/views/test_register.py
@@ -1,9 +1,11 @@
+from datetime import datetime
+
 from flask import (
     url_for,
     session
 )
 
-from bs4 import BeautifulSoup
+from app.notify_client.models import InvitedUser
 
 
 def test_render_register_returns_template_with_form(app_):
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_register.py
```diff
@@ -139,3 +141,53 @@ def test_register_with_existing_email_sends_emails(app_,
...
+                               mock_send_verify_code,
+                               mock_accept_invite):
+
+    invited_user = InvitedUser(fake_uuid, fake_uuid, "",
+                               "invited@user.com",
+                               ["manage_users"],
+                               "pending",
+                               datetime.utcnow())
+    with app_.test_request_context():
+        with app_.test_client() as client:
```
leohemsted: just like in api, don't need `app_` here - there's a `client` fixture to get rid of these two `with` statements (though you'll still need the session_transaction one below)

servingUpAces: Ah I should have looked for that. Good one

PR #1066, opened 2016-12-28
Restrict letter previews to their folded height

quis:
folded

---

The letter previews take up too much space on the page (more than a typical email or text message). This means that users have to scroll too much.

When we send the real letters they will be folded in half to fit in the envelope.

So by showing the letter previews on the page cropped to half the letter’s height:
- it reduces the need for scrolling
- it gives an accurate preview of how the letter will be delivered

## Draw the folded half of the letter with CSS

Users might wonder why they can’t see the whole letter. We should make it obvious that it’s because the letter is folded. We can use CSS to draw the bottom half of the page behind the top half. With some transforms this makes it look like the letter is actually folded.

It’s a bit skeuomorphic but:
- I think it achieves the desired effect
- the way we show emails and text messages is also mildly skeuomorphic

## Make the letter unfold slightly on hover

This might be taking the skeuomorphism too far (especially with the animation). But it’s a way of indicating that the letter will ‘unfold’ if you click it. Might revert this, but let’s see how it feels.

### DIFF
--- a/app/assets/stylesheets/components/letter.scss
+++ b/app/assets/stylesheets/components/letter.scss
@@ -11,6 +11,8 @@ $outline-width: 5px;
   outline: $outline-width solid rgba($text-colour, 0.1);
   padding: 0;
   margin: $outline-width $outline-width $gutter;
+  height: 50%;
+  overflow: hidden;
 
   a {
     display: block;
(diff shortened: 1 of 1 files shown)


quis:
## Don’t skew bottom half of letter in folded state

Skewing the letter results in a bunch of slightly-off-vertical diagonal lines. This is a bit visually jarring when everything else on the page lines up vertically or horizontally.

This commit makes the bottom half of the letter straight, by offsetting it instead of transforming it. The ‘fold’ now has depth, so it has to be drawn in somehow, that’s what the `:before` is doing.

---

folded-2

quis:
## Thin the borders around the letter preview

1px borders match the keylines that we use on email template previews.

---

image

PR #1067, opened 2017-01-03
Add date range filter to platform admin

servingUpAces:
A date range has been added to the platform admin page. 

Be aware that when using this query it looks at the notification_history table, this may results in a slow running query. I will be testing the performance once I get it deployed to production with a large data set. 

screen shot 2017-01-03 at 11 39 07

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -60,7 +60,6 @@ class ServiceAPIClient(NotifyAdminAPIClient):
         """
         Retrieve a list of services.
         """
-        print(params_dict)
         return self.get('/service', params=params_dict)
 
     def get_active_services(self, params_dict=None):
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -60,6 +60,7 @@ class ServiceAPIClient(NotifyAdminAPIClient):
         """
         Retrieve a list of services.
         """
+        print(params_dict)
         return self.get('/service', params=params_dict)
 
     def get_active_services(self, params_dict=None):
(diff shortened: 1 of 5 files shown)


app/notify_client/service_api_client.py
```diff
@@ -60,6 +60,7 @@ def get_services(self, params_dict=None):
         """
         Retrieve a list of services.
         """
+        print(params_dict)
```
leohemsted: :printer:

tests/app/main/views/test_platform_admin.py
```diff
@@ -103,25 +99,36 @@ def test_platform_admin_toggle_including_from_test_key(
 ):
     with app_.test_request_context():
         with app_.test_client() as client:
+            print(include_from_test_key)
```
leohemsted: :printer:

imdadahad:
@leohemsted I personally try to avoid using date pickers. I've found in my experience that existing date picker libraries aren't that great (not easy to customise and not mobile friendly). I have seen developments by web browsers trying to push towards making them inbuilt but not sure how far that's come with browsers other than Chrome. 

In any case probably better to split the date out into separate fields, easier on the eye and easier for validation too. Checkout https://www.gov.uk/service-manual/design/dates

servingUpAces:
We definitely do not want to use date pickers. 
I don't like the look of these filters, but we are planning to do some more designing around this page and the other MI reports that we are going to need. 
As a first attempt at adding more filters to this page, I think it is ok to leave this as is. Remember this page is only seen by our team. 
I'd like to see how this page actually performs on production where we have more data to deal with.

leohemsted:
@servingUpAces running the load tests is super straightforward - i'd like us to test this on staging first to make sure it doesn't screw up if the system's under load when you request a large page

PR #1070, opened 2017-01-06
Fix timeout on download of large CSV reports

imdadahad:
This fixes the issue on admin where downloading a large CSV report (13,000+) times out.

# **Issue**

We load the contents of a CSV report into memory and attempt to return this to the user. If the CSV has a significant number of rows this can take a long time, or timeout (as discovered by one of our end users).

# **Fix**

Instead of attempting to load the entire contents of the CSV into memory and then return this to the user, we now stream it back in _chunks_ of 5000. This means that if a large CSV is requested for download, Chrome (for example) will take each of these and incrementally build the CSV in a streaming fashion. This prevents the timeout issue.

# **How to test**

Download a report (ideally with a large number of notifications) and keep an eye the bottom bar (on Chrome). The connection will initiate and the download in a streaming fashion.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -115,9 +115,14 @@ def get_errors_for_csv(recipients, template_type):
     return errors
 
 
-def generate_notifications_csv(*args, **kwargs):
+def generate_notifications_csv(**kwargs):
+    from app import notification_api_client
+
     csvfile = StringIO()
-    csvwriter = csv.writer(csvfile)
+    csvwriter = csv.DictWriter(
+        csvfile,
+        fieldnames=['Row number', 'Recipient', 'Template', 'Type', 'Job', 'Status', 'Time']
+    )
 
     def read_current_row():
         csvfile.seek(0)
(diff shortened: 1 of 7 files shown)


app/utils.py
```diff
@@ -108,24 +115,41 @@ def get_errors_for_csv(recipients, template_type):
...
+        csvfile.truncate()
+
+    # Initiate download quicker by returning the headers first
+    csvwriter.writerow(['Row number', 'Recipient', 'Template', 'Type', 'Job', 'Status', 'Time'])
+    line = read_current_row()
+    clear_file_buffer()
+    yield line
+
+    from app import format_datetime_24h, format_notification_status, notification_api_client
+    json_list = notification_api_client.get_notifications_for_service(**kwargs)['notifications']
```
leohemsted: do we need to look at making this a yielding call over on the API side as well?

app/utils.py
```diff
@@ -108,24 +115,41 @@ def get_errors_for_csv(recipients, template_type):
...
+    for x in json_list:
+        csvwriter.writerow([
+            int(x['job_row_number']) + 2 if 'job_row_number' in x and x['job_row_number'] else '',
```
leohemsted: `int(x['job_row_number']) + 2 if x.get('job_row_number') else ''`

app/utils.py
```diff
@@ -108,24 +115,41 @@ def get_errors_for_csv(recipients, template_type):
...
+            x['template']['name'],
+            x['template']['template_type'],
+            x['job']['original_file_name'] if x['job'] else '',
+            format_notification_status(x['status'], x['template']['template_type']),
+            format_datetime_24h(x['created_at'])
+        ])
+
+        line = read_current_row()
+        clear_file_buffer()
+        yield line
```
leohemsted: this stuff is super sketch  - we're maintaining a one line CSV, replacing that one line with each line as it comes, then yielding it? in this case, what does CSV writer give us - is it just quote escaping?

 is there any way we can do this without nasty seeking and truncating?

i'd almost be inclined to say we roll it ourself, and reduce the code to just:
```
yield ','.join(['Row number', 'Recipient', 'Template', 'Type', 'Job', 'Status', 'Time'])
for x in ...:
    yield ','.join([int(x['job_row_number') + 2, [...] )
```

we'd need to make sure we took care of new lines, quote escaping, essentially the stuff in [here](https://docs.python.org/3.4/library/csv.html#dialects-and-formatting-parameters), but it'd lead to much cleaner code, especially if we can hook into any internal formatting functions that the csv writer might use

tests/__init__.py
```diff
@@ -295,6 +298,24 @@ def single_notification_json(
     return data
 
 
+def csv_notifications(notifications_json):
```
leohemsted: think it's better to put this in `test_jobs.py` since we don't plan on using it anywhere else

tests/app/main/views/test_jobs.py
```diff
@@ -424,6 +445,7 @@ def test_should_show_notifications_for_a_service_with_next_previous(
 @freeze_time("2016-01-01 11:09:00.061258")
 def test_should_download_notifications_for_a_job(app_,
                                                  api_user_active,
+                                                 mocker,
```
leohemsted: looks like a good case for using the `logged_in_user` fixture, then you won't have to include most of these fixtures (only the ones that the actual function uses, not the logging in checks like get_service and mock_login)

app/utils.py
leohemsted: update: we looked into it and csvwriter just calls straight through to C code, so leave as is.

app/utils.py
leohemsted: update: we probably do. but this can be shipped independently. this PR still offers a serious tangible advantage, in that the download starts immediately so the user knows that the page is working

PR #1071, opened 2017-01-16
This PR allows the pentesters to operate as part of white listed root…

minglis:
… domains. Needed for full app testing.

### DIFF
--- a/config.py
+++ b/config.py
@@ -65,6 +65,7 @@ class Config(object):
         r"ucds\.email",
         r"naturalengland\.org\.uk",
         r"hmcts.net",
+        r"irmsecurity.com",  # temp addition for pen testers to be able to create account
     ]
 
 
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -65,6 +65,7 @@ class Config(object):
         r"ucds\.email",
         r"naturalengland\.org\.uk",
         r"hmcts.net",
+        r"irmsecurity.com",  # temp addition for pen testers to be able to create account
     ]
```
imdadahad: We need to add a `\` before the dot as `.` like the ones above so that something like someone@irmsecurityXcom" won't pass the regex match.

So change to:

```
r"hmcts\.net",
r"irmsecurity\.com"
```

minglis: Of course.

minglis: done

PR #1073, opened 2017-01-17
Update count of services on product page

quis:
Two new services, from a department who are already using Notify.


yahoopete:
15 years of web development leading to this moment.  Well done @quis.

:+1:

PR #1074, opened 2017-01-17
Add last updated column (formatted) in view providers table

imdadahad:
This adds a 'Last Updated' column in the provider list table found on Admin -> View Providers.

This will help to identify when a provider was last updated (e.g. priority was changed etc.) and will be especially useful when auto provider-switching is in place.

### DIFF
--- a/tests/app/main/views/test_providers.py
+++ b/tests/app/main/views/test_providers.py
@@ -33,7 +33,8 @@ stub_providers = {
             'priority': 1,
             'display_name': 'first_email_provider',
             'identifier': 'first_email',
-            'notification_type': 'email'
+            'notification_type': 'email',
+            'updated_at': None
         },
         {
             'active': True,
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_providers.py
```diff
@@ -36,7 +41,7 @@
             'display_name': 'second_email_provider',
             'identifier': 'second_email',
             'id': '0bd529cd-a0fd-43e5-80ee-b95ef6b0d51b',
-            'notification_type': 'email'
+            'notification_type': 'email',
```
leohemsted: for consistency, it's probably better for these other two (`first_email_provider` and `second_email_provider`) to have `'updated-at': None` as well too i guess

imdadahad: cool

imdadahad: Not sure why github isn't picking up the change...

PR #1075, opened 2017-01-18
Create or update templates with priority as platform admin

servingUpAces:
Add a selectField to edit and create templates that is only visible for platform admins that makes the template a `priority` template.

There is a check that the template can not be created as priority if the user is not a platform admin.
There is a check that the template can not change the `priority` unless they are a platform admin.

screen shot 2017-01-18 at 16 06 06

https://www.pivotaltracker.com/story/show/135429147

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -270,8 +270,8 @@ class SMSTemplateForm(Form):
             NoCommasInPlaceHolders()
         ]
     )
-    process_type = SelectField(u'Select priority or normal', choices=[('normal', 'normal'),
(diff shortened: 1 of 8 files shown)


app/main/forms.py
```diff
@@ -268,6 +270,9 @@ class SMSTemplateForm(Form):
             NoCommasInPlaceHolders()
         ]
     )
+    process_type = SelectField(u'Select priority or normal', choices=[('normal', 'normal'),
```
leohemsted: should be "process type"

PR #1078, opened 2017-01-19
Sanitize HTML in templates using utils

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/109

***

We can no longer trust that the content of templates stored in the database is safe.

Utils now has code to sanitise the content of templates.

This commit:
- updates utils to bring this code in
- modifies some integration tests to make sure everything is working (there are more extensive unit tests in utils)

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -479,7 +479,7 @@ def test_should_show_preview_letter_message(
     mock_get_service_letter_template.assert_called_with(service_id, template_id)
     assert mock_letter_preview.call_args[0][0]['message'] == (
         '<h2>Subject</h2>\n'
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_send.py
```diff
@@ -476,9 +476,10 @@ def test_should_show_preview_letter_message(
     assert response.status_code == 200
     assert response.content_type == expected_content_type
     mock_get_service_letter_template.assert_called_with(service_id, template_id)
+    print(mock_letter_preview.call_args)
```
leohemsted: 🖨

quis:
@leohemsted Good point, added those tests in 04966e4 (and squashed out the 🖨 )

PR #1079, opened 2017-01-19
when pressing back after sending CSV redirect properly

leohemsted:
previously we were issuing a flask redirect (302) from the function, which we then attempted to unpack as a dict further down the line.

raise a werkzeug.routing.RequestRedirect (301 MOVED PERMANENTLY) instead. note: only use this pattern when the URL they attempted to access will *NEVER* be valid, as 301s are cached by browsers.

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -1,15 +1,16 @@
+import uuid
 from io import BytesIO
 from os import path
 from glob import glob
 import re
 from itertools import repeat
 from functools import partial
-from unittest.mock import Mock, patch
+from unittest.mock import patch
+from urllib.parse import urlparse
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_send.py
```diff
@@ -882,3 +883,19 @@ def test_check_messages_shows_over_max_row_error(
...
+        'main.check_messages',
+        service_id=service_id,
+        template_type='sms',
+        upload_id=uuid.uuid4(),
+    ))
+
+    # permanent redirect
+    assert response.status_code == 301
+    path = urlparse(response.location).path
+    assert path == url_for('main.choose_template', service_id=service_id, template_type='sms')
```
quis: `_external=True` will mean not having to make the path relative yourself.

PR #1080, opened 2017-01-19
Add a breadcrumb to the product page

quis:
This copies what the other GaaP components will be doing for their product pages, and adds a link upwards to the Service Toolkit

# Before

screen shot 2017-01-19 at 14 59 40

# After

screen shot 2017-01-19 at 14 59 20

***

The SASS and HTML are taken from here:
https://github.com/alphagov/product-page-example/blob/f05ca1fb71f16728bb7a52462b11a850420f36f0/source/stylesheets/modules/_breadcrumbs.scss

Only changes I’ve made are:
- making the file paths work with our build pipeline

Changes to our code to accomodate this are:
- putting the padding on the product page `<h1>` not its container
- moving the hero image accordingly so that it lines up
- making the `<main>` element on the product page into an anchor so that the breadcrumb can link to it – screenreaders will then announce the link as “GOV.UK Notify, same page”

### DIFF
--- a/app/templates/fullwidth_template.html
+++ b/app/templates/fullwidth_template.html
@@ -1,7 +1,7 @@
 {% extends "admin_template.html" %}
 
 {% block fullwidth_content %}
-  <main role="main">
+  <main role="main" id="main">
     {% block maincolumn_content %}{% endblock %}
   </main>
 {% endblock %}
--- a/app/assets/stylesheets/components/vendor/breadcrumbs.scss
+++ b/app/assets/stylesheets/components/vendor/breadcrumbs.scss
@@ -1,0 +1,106 @@
+/**
+ * Breadcrumbs Component
+ *
+ * Based on the existing breadcrumbs from GOV.UK Elements, re-implemented
+ * using BEM, allowing for an `inverse` modifier (e.g. against a masthead), with
+ * structured data markup, and allowing for the 'current page' to be included
+ * as the last element in the breadcrumb.
+ *
+ * Notes:
+ *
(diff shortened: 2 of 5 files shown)


app/templates/fullwidth_template.html
```diff
@@ -1,7 +1,7 @@
 {% extends "admin_template.html" %}
 
 {% block fullwidth_content %}
-  <main role="main">
+  <main role="main" id="main">
```
36degrees: Worth noting – your skip link currently links to `#content`, but that doesn't appear to be an element on the page. I'd suggest either changing the skip link to `#main` or standardising on `#content` everywhere. 👍

quis: GOV.UK Template hardcodes the skiplink to point at `#content` (see [this line](https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L39)). But GOV.UK Elements adds styles to `#content` (see [this block](https://github.com/alphagov/govuk_elements/blob/d6226bd2c8c7e256f995c0f519daaa4035c6bd54/public/sass/elements/_layout.scss#L9-L20)). We don’t want those styles, especially not on the product page because it goes full-width.

Guessing this is why you’ve changed the skiplink to point at `#main` [in the example product page](https://github.com/alphagov/product-page-example/blob/f05ca1fb71f16728bb7a52462b11a850420f36f0/source/layouts/layout.erb#L21).

I’m going to raise an issue on GOV.UK Template.

quis: Issue raised: https://github.com/alphagov/govuk_template/issues/271

PR #1086, opened 2017-01-23
Fix: Platform admin showing incorrect sent totals

imdadahad:
This fixes an issue where the platform admin page shows incorrect sent total values.

# **Issue**

The total email and sms sent figures are set to the sum of delivered notifications. E.g. If we sent 100 sms messages and 80 of those were delivered while 20 failed, then the total would incorrectly show 80 only. 

# **Fix**

Changed the total to include the sum of delivered _and_ failed.

### DIFF
--- a/tests/app/main/views/test_platform_admin.py
+++ b/tests/app/main/views/test_platform_admin.py
@@ -147,7 +147,6 @@ def test_create_global_stats_sets_failure_rates(fake_uuid):
     )
 
     stats = create_global_stats(services)
-
     assert stats == {
         'email': {
             'delivered': 2,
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_platform_admin.py
```diff
@@ -323,3 +321,38 @@ def test_shows_archived_label_instead_of_live_or_research_mode_label(
     service_mode = table_body.find_all('tbody')[0].find_all('tr')[1].td.text.strip()
     # get second column, which contains flags as text.
     assert service_mode == 'archived'
+
+
+def test_should_show_correct_sent_totals_for_platform_admin(
+    app_,
```
leohemsted: use `client` rather than `app_` and you won't need the with statements

imdadahad: 👍

PR #1089, opened 2017-01-23
Add introducer video to the product page

quis:
It’s always been the plan to put this video on the product page. Just getting round to it now.

Watch the video here: https://www.youtube.com/watch?v=i2a2jiwYTd4

Uses the CSS developed by the service manual team, taken from here: https://github.com/alphagov/product-page-example/blob/a5c613f07bca47a15f67c529615bc83995aea3b5/source/stylesheets/modules/_responsive-embed.scss

---

image


quis:
Demo’d to @yahoopete

PR #1093, opened 2017-01-25
Remove pentesters

leohemsted:
are we going to delete their accounts on staging?


quis:
Oops, just did this: https://github.com/alphagov/notifications-admin/pull/1098

PR #1094, opened 2017-01-27
Filter and navigate usage by financial year

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/803

***

Right now we tell people that the usage page is for the current financial year. This is a lie – it’s for all time.

So this commit calls through to the API to get the stats for (by default) the current financial year.

We already do this for the monthly breakdown, this just does the same thing for the yearly totals.

It also adds navigation to show the data for other financial years:
- previous so you can go back and see your usage and verify that the bill you’re about to pay is correct
- next so that you can check what your SMS allowance is going to be before you actually get into it

***

usage-by-year

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -19,7 +19,7 @@ from app import (
     template_statistics_client
 )
 from app.statistics_utils import get_formatted_percentage, add_rate_to_job
-from app.utils import user_has_permissions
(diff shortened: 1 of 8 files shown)


quis:
Alternative which I considered but rejected. Don’t think it’s as clear.

screen shot 2017-01-30 at 11 45 01

yahoopete:
:+1:

This is good...

PR #1108, opened 2017-02-07
Provide better subject lines for Deskpro tickets

yahoopete:
They get threaded and become very confusing.... this will avoid that.


imdadahad:
👍

PR #1109, opened 2017-02-08
Break down usage by month, filter by year

quis:
Depends on:

- [x] https://github.com/alphagov/notifications-api/pull/820

***

The previous, weekly activity breakdown was what we reckoned might be useful. But now that we have people using the platform it feels like aggregating a service’s usage by month is better because:
- matches the timeframe users report on within their organisation
- is consistent with the usage page

And like the usage page this commit also limits the page to only show one financial year’s worth of data at once (rather than data for all time).

This commit also makes some changes to the jobs view code so that our aggregation of failure states is consistent between the dashboard pages and the jobs pages.

---

# Before 

screen shot 2017-02-08 at 11 04 37

# After

screen shot 2017-02-08 at 10 58 34

---

screen shot 2017-02-08 at 10 59 00

---

screen shot 2017-02-08 at 10 58 04


yahoopete:
👍

PR #1128, opened 2017-02-14
Fix permissions in tests

leohemsted:
Essentially, only use mock_has_permissions if you're really sure you don't care about permissions. Not in fixtures that are used by pretty much the entire test suite to log in.

a lot of the changes are replacing `logged_in_client` with the new fixture `logged_in_platform_admin_client` where applicable, and removing unneeded fixture invocations


leohemsted:
i'm going to go ahead and merge this, idk why jenkins hasn't run it - i triggered manually here  and it worked fine https://jenkins.notify.tools/job/notifications-admin/623/

PR #1129, opened 2017-02-14
Mark up ‘pill’ component as being tabs

quis:
Currently it’s not possible for a screen reader user to know which financial year they’re looking at. From the accessibility report:

> The financial year links are contained in a navigation region - tabbing or arrowing through only reads out the links, not the main information of "2016 to 2017 financial year" - that information is vital for understanding the page content.

This problem also applies to other pages which use the `pill` component, which is effectively tabbed navigation (that reloads the page rather than showing or hiding content on the page).

There are specific ARIA attributes that can be used to mark up a navigation as being tabbed. This commit:

- adds those attributes
- makes the selected ‘tab’ visible to screenreaders and keyboard focusable
- adds a visual focus indicator to the selected tab
- adds `id`s to the parts of the page that are controlled by the tabs so
  that they are labelled as such

This also means changing the pill component from being a `<nav>` to a `<ul>` because `tablist` is not a valid `role` for a `nav`.

Mostly follows the example here: http://accessibility.athena-ict.com/aria/examples/tabpanel2.shtml

## With Voiceover

aria-tabs

### DIFF
--- a/app/assets/stylesheets/components/pill.scss
+++ b/app/assets/stylesheets/components/pill.scss
@@ -9,7 +9,7 @@
   }
 
   a,
-  span {
+  &-selected-item {
     display: block;
     float: left;
     box-sizing: border-box;
@@ -45,7 +45,7 @@
     }
   }
 
-  span {
+  &-selected-item {
     border: 2px solid $black;
     outline: 1px solid rgba($white, 0.1);
     position: relative;
--- a/app/assets/stylesheets/components/pill.scss
+++ b/app/assets/stylesheets/components/pill.scss
@@ -2,24 +2,19 @@
 
   display: flex;
 
+  li {
+    width: 25%;
+    flex-grow: 1;
+    text-align: left;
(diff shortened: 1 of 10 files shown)


quis:
Rolling this up into the same PR because it would be hard to do both changes in parallel without merge conflicts:

##    Fix invalid nesting of HTML elements

In HTML you generally can’t nest an inline level element inside a block level one, if you want your HTML to validate.

 There were a couple of places where we were using a `<span>` as a containing element:
- inside every table cell (think we inherited this from Digital Marketplace)
- in the ‘pill’ navigation component for the selected tab

This meant that when we put components like big number inside these, the resulting HTML was invalid, because big number is built with a bunch of `<div>`s, which are block level.

This commit removes the use of a `<span>` tag in these places, and replaces it with a `<div>`. Nesting block level elements in fine in HTML.

yahoopete:
👍

PR #1131, opened 2017-02-15
Make template usage graph take up 100% of width

quis:
The bars were sitting in a table cell with some right padding, so they never extended all the way to the right. Making it right-aligned removes this padding, then setting the text to left aligned keeps things looking the same.

alignment


imdadahad:
👍

PR #1133, opened 2017-02-15
Don't let users add non-GSMish characters to sms templates

leohemsted:
- [x] https://github.com/alphagov/notifications-utils/pull/116
- [x] https://github.com/alphagov/notifications-admin/pull/1128

image

If the character can be downgraded safely (like the `“”` characters or non-breaking spaces etc) then that's fine, but if it can't, like emojis or non-latin characters, then we should error, because otherwise we'll be billed inconsistent amounts by our providers

Email templates are unaffected

The downgradeable characters are still persisted to the database

### DIFF
--- a/tests/app/main/test_validators.py
+++ b/tests/app/main/test_validators.py
@@ -1,6 +1,6 @@
 import pytest
 from app.main.forms import RegisterUserForm, ServiceSmsSender
-from app.main.validators import ValidGovEmail, NoCommasInPlaceHolders
+from app.main.validators import ValidGovEmail, NoCommasInPlaceHolders, OnlyGSMCharacters
 from wtforms import ValidationError
 from unittest.mock import Mock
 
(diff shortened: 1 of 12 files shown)


tests/app/main/test_validators.py
```diff
@@ -141,6 +141,21 @@ def test_for_commas_in_placeholders(
...
+def test_gsm_character_validation(client, msg):
+    OnlyGSMCharacters()(None, _gen_mock_field(msg))
+
+
+def test_non_gsm_character_validation(client):
+    with pytest.raises(ValidationError) as error:
+        OnlyGSMCharacters()(None, _gen_mock_field('∆ abc 📲 def 📵 ghi'))
+
+    assert str(error.value) == (
+        'You can’t use ‘∆’, ‘📲’ or ‘📵’ in text messages. '
```
quis: > You can’t use ∆, 📲 or 📵 in text messages.

I think for single characters it’s clearer without the surrounding quotation marks.

tests/app/main/test_validators.py
```diff
@@ -141,6 +141,21 @@ def test_for_commas_in_placeholders(
...
+        'They won’t show up properly on everyone’s phones.'
```
quis: Will this say ‘it won’t show up properly…’ if there’s only one errant character?

leohemsted: damn no it won't

quis:
Utils changes got brought in as part of https://github.com/alphagov/notifications-admin/pull/1137

Will resolve conflicts now…

quis:
Should be updated to include https://github.com/alphagov/notifications-utils/pull/119/files

PR #1135, opened 2017-02-16
Fix error message when recipient column missing

quis:
When your CSV file is missing the recipient column (eg ‘phone number’ or ‘email address’) we give you a helpful error message telling you that this is the case.

When we changed the recipient column to be columns, plural, we didn’t update the code that generated the error message. So you would get errors that looked this like this:

> Your file needs to have a column called ‘’

This commit fixes the error message.


yahoopete:
👍

PR #1136, opened 2017-02-16
Show a more useful message if you get signed out

quis:
image

> Users that allow their session to expire, or access a bookmarked link are told they need to "Sign in to access this page" - we should explain that it's because they've been away a while, so that they understand why they're being asked to log in again.

– https://www.pivotaltracker.com/story/show/140016919

The message we were showing before (Please log in to access this page) is the default message from Flask Login.

In order to stop this flash message from appearing, we need to override the default handler for when a user is unauthorised. We’re overriding it with the same behaviour, minus the flash message.

If you navigate deliberately to the sign in page it’s unchanged.

Content is Sheryll-approved.

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -16,6 +16,7 @@ from flask_login import (
 )
 
 from app import (
+    login_manager,
     user_api_client,
     service_api_client,
     invite_api_client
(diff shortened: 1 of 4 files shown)


yahoopete:
👍 +1

PR #1139, opened 2017-02-20
Make proposition image meet colour contrast

quis:
The message text in our previous illustration was white on light blue, which didn’t meet WCAG AA colour contrast. WCAG AA requires a contrast ratio of 4.5:1. The text in our image was only 3.8:1.

The text in this new image has a contrast ratio of 19.8:1, so easily passes WCAG AAA.

Required a slight tweak to the positioning of the image because it occupies slight more of the artboard horizontally.

image


yahoopete:
:+1:

PR #1142, opened 2017-02-20
Show template usage broken down by month

quis:
This follows on from:
- https://github.com/alphagov/notifications-admin/pull/1094
- https://github.com/alphagov/notifications-admin/pull/1109

It depends on:
- [x] https://github.com/alphagov/notifications-api/pull/829

A year is too long. Month-by-month is a better timeframe for making decisions or seeing patterns in your usage.

image

***
image

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-
+from functools import partial
 from flask import (
     render_template,
     url_for,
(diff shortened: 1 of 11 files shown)


app/main/views/dashboard.py
```diff
@@ -298,3 +310,32 @@ def requested_and_current_financial_year(request):
...
+        yield dict(
+            requested_count=sum(
+                template['counts'].get(status, 0)
+                for status in REQUESTED_STATUSES
+            ),
+            **template
+        )
+
+
+def get_tuples_of_finanical_years(
```
imdadahad: finanical_years -> financial_years 

Think this type is in other files too

quis: Good point. Changed and squashed out.

quis:
Bit of tidying up to make things clearer, now that all these pages are lined up:

# Relabel _show more_ links on the dashboard

- ‘messages sent per month’ is a better description of what will appear on the page than ‘activity breakdown’
- ‘templates used by month’ instead of ‘this year’ for consistency

I think ‘usage breakdown’ still works for the remaining link – it’s more than a monthly breakdown, it also breaks down the spend.

Doesn’t need to say ‘by month’ in the `<h1>`s themselves, because you can see from looking at the page that it’s broken down by month.

app/main/views/dashboard.py
```diff
@@ -65,16 +65,35 @@ def service_dashboard_updates(service_id):
 @login_required
 @user_has_permissions('view_activity', admin_override=True)
 def template_history(service_id):
-    template_statistics = aggregate_usage(
-        template_statistics_client.get_template_statistics_for_service(service_id)
-    )
```
servingUpAces: if template_statistics_client.get_template_statistics_for_service(service_id) is no longer used we should have a follow up PR to remove the old method and endpoint.

app/main/views/dashboard.py
```diff
@@ -298,3 +310,32 @@ def requested_and_current_financial_year(request):
         )
     except ValueError:
         abort(404)
+
+
+def format_template_stats_to_list(stats_dict):
+    if not stats_dict:
```
servingUpAces: do we have any unit tests for this method?

app/main/views/dashboard.py
```diff
@@ -298,3 +310,32 @@ def requested_and_current_financial_year(request):
...
+            requested_count=sum(
+                template['counts'].get(status, 0)
+                for status in REQUESTED_STATUSES
+            ),
+            **template
+        )
+
+
+def get_tuples_of_financial_years(
+    partial_url,
```
servingUpAces: Any unit tests for this method?

quis: Added a test in c122a5b

app/main/views/dashboard.py
quis: Added a test in c122a5b

app/main/views/dashboard.py
quis: Probably good to hold onto it a short while just in case we want to revert this. But generally agree.

PR #1143, opened 2017-02-20
Update two-factor to use new update password endpoint and refactor tests

imdadahad:
This makes the two-factor flow use the new method of updating passwords (new endpoint).

If the user is coming from the new password page, the password is set and the login count is reset. Instead of using the old `update_user` method, this uses the new `update_password` method. On the API, this resets the login count and changes the password.

### DIFF
--- a/app/main/views/two_factor.py
+++ b/app/main/views/two_factor.py
@@ -29,9 +29,7 @@ def two_factor():
             services = service_api_client.get_active_services({'user_id': str(user_id)}).get('data', [])
             # Check if coming from new password page
             if 'password' in session['user_details']:
-                user.set_password(session['user_details']['password'])
-                user.reset_failed_login_count()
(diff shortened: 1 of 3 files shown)


app/main/views/two_factor.py
```diff
@@ -29,9 +29,7 @@ def _check_code(code):
             services = service_api_client.get_active_services({'user_id': str(user_id)}).get('data', [])
             # Check if coming from new password page
             if 'password' in session['user_details']:
-                user.set_password(session['user_details']['password'])
-                user.reset_failed_login_count()
```
leohemsted: can we now get rid of that function completely? doesn't look like it's used anywhere else

imdadahad: yeah not used anywhere else in admin, removed 👍

PR #1145, opened 2017-02-22
Remove old API key and service ID combo

quis:
Once all our users have upgraded to the latest clients they won’t need this. The latest clients only use the combined key and service ID.

Discuss: when can we safely remove it?

### DIFF
--- a/app/templates/views/api/keys/show.html
+++ b/app/templates/views/api/keys/show.html
@@ -32,17 +32,4 @@
 
     </div>
 
-    <h2 class='heading-medium'>For older API clients</h2>
-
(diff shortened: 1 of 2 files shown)


leohemsted:
remind me to check the logs - we keep a record of the user agents and they should include client version info so we can see whos' actually using old keys

leohemsted:
if i've understood our logs correctly, then we're not ready for this - only 13 / 163 api hits in the last 15 minutes were using v2 endpoints

quis:
Good sleuthing. Will close for now and we can revisit.

PR #1153, opened 2017-02-23
Remove remember me cookie

leohemsted:
it's no longer needed cos the session now waits 20 hours so there's no point keeping the skip-2fa thing

* [x] merge after #1146


leohemsted:
@servingUpAces that was straightforward

removed

PR #1156, opened 2017-02-24
WIP: Performance info on the product page

quis:
image


quis:
Closing this because I don’t think it’s a working solution. Doesn’t give a rolling month.

PR #1157, opened 2017-02-27
Fix wrong error message if file is missing columns

quis:
Accidentally got broken here: https://github.com/alphagov/notifications-admin/commit/41fa1586353374d611f7b7e7690b03c13b830880#diff-bff3df90be0231a1e33e033fc51ba7f7L78

This commit changes it back to how it was before (but keeping the new macro for formatting the list).

Stops this from happening:

screen shot 2017-02-27 at 12 00 52

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -70,7 +70,7 @@
           ) }}.
         </p>
         <p>
-          It doesn’t have {{ recipients.column_headers | formatted_list(
+          It doesn’t have {{ recipients.missing_column_headers | formatted_list(
             conjunction='or',
             prefix='a column called ',
             prefix_plural='columns called '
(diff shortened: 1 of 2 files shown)


leohemsted:
we don't have a test for this?

app/templates/views/check.html
```diff
@@ -70,7 +70,7 @@ <h1 class='banner-title'>
           ) }}.
         </p>
         <p>
-          It doesn’t have {{ recipients.column_headers | formatted_list(
+          It doesn’t have {{ recipients.missing_column_headers | formatted_list(
```
leohemsted: whoops my bad 😰

quis:
@leohemsted Added a test and squashed it into e1f5376.

PR #1158, opened 2017-02-27
Don’t 500 when a CSV is missing rows

quis:
> When the CSV is missing the header row, we get an error and the user will see "Sorry, we are experiencing technical difficulties..."
>
> We should return a better error message for the user.

– https://www.pivotaltracker.com/story/show/140668615

This was caused by an attempt to access the `first_recipient` variable before it was assigned. It would only be assigned when there was at least one row in the file.

Fixing this means doing two things:
- defaulting `first_recipient` to be `None` before looking in the file
- adding an error message for when we can’t extract any rows out of the file (which is more nuanced than the file just being completely empty)

(There’s a nasty `sort` in the Jinja template because when there are no rows in the file the order of the required column headers is not deterministic.)

# After

image

### DIFF
--- a/app/templates/views/check.html
+++ b/app/templates/views/check.html
@@ -35,6 +35,23 @@
       {% endcall %}
     </div>
 
+  {% elif not count_of_recipients %}
+
+    <div class="bottom-gutter">
+      {% call banner_wrapper(type='dangerous') %}
+        <h1 class='banner-title'>
+          Your file is missing some rows
+        </h1>
+        <p>
+          It needs at least one row of data, and {{ recipients.missing_column_headers | sort() | formatted_list(
+            prefix='a column called',
+            prefix_plural='columns called'
+          ) }}.
+        </p>
+        {{ skip_to_file_contents() }}
+      {% endcall %}
+    </div>
+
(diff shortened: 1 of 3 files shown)


app/templates/views/check.html
```diff
@@ -35,6 +35,23 @@ <h1 class='banner-title'>
...
+
+    <div class="bottom-gutter">
+      {% call banner_wrapper(type='dangerous') %}
+        <h1 class='banner-title'>
+          Your file is missing some rows
+        </h1>
+        <p>
+          It needs at least one row of data, and {{ recipients.missing_column_headers | sort() | formatted_list(
+            prefix='a column called',
+            prefix_plural='columns called'
```
leohemsted: reckon it's obvious that this "a column called ..." means "a row of headers"?

quis: Hard to know without being able to see what people are uploading that was causing the previous error…

quis: Like what was the specific mistake they were making? Was it just an empty file?

leohemsted: yah fair enough 👍

PR #1168, opened 2017-03-02
add letter_contact_block edit fields

leohemsted:
no actual template functionality yet - just the ability for services that have letters enabled to edit a 10 line block that will in the future go on the top right hand side of their letters with contact information

text on the edit page probably needs a bit of TLC tho

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -496,8 +496,11 @@ class ServiceLetterContactBlock(Form):
     )
 
     def validate_letter_contact_block(form, field):
-        if field.data.strip().count('\n') >= 10:
-            raise ValidationError('Can only have 10 lines')
+        line_count = field.data.strip().count('\n')
+        if line_count >= 10:
+            raise ValidationError(
+                'Contains {} lines, maximum is 10'.format(line_count + 1)
+            )
 
 
 class ServiceBrandingOrg(Form):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -486,8 +486,18 @@ class ServiceSmsSender(Form):
     )
 
     def validate_sms_sender(form, field):
-        if field.data and not re.match('^[a-zA-Z0-9\s]+$', field.data):
+        if field.data and not re.match(r'^[a-zA-Z0-9\s]+$', field.data):
             raise ValidationError('Use letters and numbers only')
+
+
+class ServiceLetterContactBlock(Form):
+    letter_contact_block = TextAreaField(
+        'How should users contact you?'
+    )
+
+    def validate_letter_contact_block(form, field):
+        if field.data.strip().count('\n') >= 10:
+            raise ValidationError('Can only have 10 lines')
 
 
 class ServiceBrandingOrg(Form):
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -25,6 +25,7 @@ from app.main.forms import (
     RequestToGoLiveForm,
     ServiceReplyToEmailFrom,
     ServiceSmsSender,
+    ServiceLetterContactBlock,
     ServiceBrandingOrg
 )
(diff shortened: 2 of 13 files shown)


quis:
# Make letter contact textbox half-wide

This approximates the width of the contact block area in the letter template itself.

image

app/main/views/service_settings.py
```diff
@@ -259,6 +261,28 @@ def service_set_sms_sender(service_id):
         form=form)
 
 
+@main.route("/services/<service_id>/service-settings/set-letter-contact-block", methods=['GET', 'POST'])
+@login_required
+@user_has_permissions('manage_settings', admin_override=True)
+def service_set_letter_contact_block(service_id):
+    print(current_service)
+    print(current_service['can_send_letters'])
```
imdadahad: 🖨

app/main/forms.py
```diff
@@ -486,10 +486,20 @@ class ServiceSmsSender(Form):
...
 
     def validate_sms_sender(form, field):
-        if field.data and not re.match('^[a-zA-Z0-9\s]+$', field.data):
+        if field.data and not re.match(r'^[a-zA-Z0-9\s]+$', field.data):
             raise ValidationError('Use letters and numbers only')
 
 
+class ServiceLetterContactBlock(Form):
+    letter_contact_block = TextAreaField(
+        'Letter contact block'
```
imdadahad: Does this need a RequiredValidator?

leohemsted: it's not required - if you leave it empty, that's okay

app/main/views/service_settings.py
```diff
@@ -259,6 +261,28 @@ def service_set_sms_sender(service_id):
         form=form)
 
 
+@main.route("/services/<service_id>/service-settings/set-letter-contact-block", methods=['GET', 'POST'])
```
imdadahad: Slightly minor, but why is this endpoint prefixed with set? I know we do this with other endpoints but doesn't seem right given it's a `GET` too.

app/main/forms.py
imdadahad: 👍

tests/app/main/views/test_service_settings.py
```diff
@@ -538,6 +580,61 @@ def test_if_sms_sender_set_then_form_populated(
     assert page.find(id='sms_sender')['value'] == 'elevenchars'
 
 
+@pytest.mark.parametrize('method', ['get', 'post'])
+def test_cant_set_letter_contact_block_if_service_cant_send_letters(
```
imdadahad: Being pedantic here but `get` -> `GET` and with post?

app/main/views/service_settings.py
leohemsted: well, i guess the pages' purpose is to set the value of the block - it has a GET just cos the page has to be loaded and that's the WTForms pattern.

if you want to view the services' letter contact block, you'd just look at it on the service-settings page

or something idk /shrug

tests/app/main/views/test_service_settings.py
leohemsted: ah those are the actual functions, eg `logged_in_client.get(...)`. `logged_in_client.GET` will error

imdadahad: ah ofc, missed the `getattr`

app/main/views/service_settings.py
imdadahad: fair enough

imdadahad:
Happy to approve once those prints are removed 👍

leohemsted:
pending chris' work ofc

quis:
@imdadahad Can you review the commits I just added? 68113e7, 63f5239, 3dabd39, 92a51c3

Looks like this:
image

imdadahad:
@quis looks good 👍

PR #1170, opened 2017-03-03
Show who updated providers and provider history data

imdadahad:
This adds capability to see who updated a provider and to view update history for each provider on a separate page.

## Providers page

screen shot 2017-03-03 at 11 29 05

The above shows that the 'Notify service user' made an update to both providers and when the update was made. As an fyi, this indicates an automatic provider switch.

## View provider history for a given provider

screen shot 2017-03-03 at 11 30 33

A separate page that shows the update history for a given provider. The history is displayed in descending order by version and also shows who updated and when the updates were made.

# Additional notes

Grouped provider templates in their own folder and did endpoint refactoring to ensure good old RESTness :)


app/templates/views/providers/provider.html
```diff
@@ -0,0 +1,48 @@
...
+        {{ text_field('None') }}
+    {% endif %}
+
+    {{ text_field(item.priority) }}
+
+    {{ text_field(item.active) }}
+
+  {% endcall %}
+
+      <a class="page-footer-back-link" href="{{ url_for('main.view_providers') }}">Back to Providers</a>
```
servingUpAces: Is the button suppose to be grey?

quis: Should just be a link in this context.

```jinja
{{ page_footer(
  back_link=url_for(…),
  back_link_text="Back to providers"
) }}
```

imdadahad: 👍

quis: 👍  😎

PR #1183, opened 2017-03-13
Don’t halt Gulp watch task on JS/SASS errors

quis:
Problem
---
You make a minor typo, save the file and your Gulp process dies without you realising. You then spend 5 minutes trying to work out why your changes aren’t appearing, no matter what edits you make.

Solution
---

Catch errors raised in processing the JS/SASS files and log them, without killing the process.

How
---

Use a handy plugin called [Gulp PrettyError](https://www.npmjs.com/package/gulp-prettyerror).

What it looks like
---

Before | After
---|---
image | screen shot 2017-03-13 at 13 19 48

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -66,6 +66,7 @@ gulp.task('javascripts', () => gulp
     paths.src + 'javascripts/listEntry.js',
     paths.src + 'javascripts/main.js'
   ])
+  .pipe(plugins.prettyerror())
   .pipe(plugins.babel({
     presets: ['es2015']
   }))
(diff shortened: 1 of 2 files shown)


quis:
And while I’m here:

## Upgrade Gulp SASS

Gulp was spitting out a lot of noise about stuff being out of date. The cause seemed to be our outdated version of Gulp SASS. This commit upgrades Gulp SASS to the latest version.

### Changes

https://github.com/dlmanning/gulp-sass/compare/v2.3.1...v3.1.0

This is a breaking change because one of the underlying libraries includes a breaking change. But looking at their release notes, it doesn’t seem to affect any SASS features that we’re using:

https://github.com/sass/libsass/releases/tag/3.4.0

PR #1191, opened 2017-03-15
Pin six to a version that’s more likely to work

quis:
Everything is fine and happy locally and on Jenkins.

Getting a `cannot import name viewkeys` error on AWS.

Based on some Googling, it’s possible that this will fix it.


quis:
Gonna merge this to see if it works.

PR #1194, opened 2017-03-17
40 - 41 for GOV.UK Blogs

yahoopete:


minglis:
Hmm, this needs some automation......

yahoopete:
Correct, but it's slightly tricker than it seems as we have things like our test live services and Notify service etc.... also the organisation count is in the same file and that can't be done till we add org to the model.....

So yes, but not a triviality....

quis:
Could automate the count of services if not the count of depts.:

https://www.performance.service.gov.uk/data/govuk-notify/government-services?flatten=true&collect=count%3Asum&group_by=service

```json
{
  "data": [
    {
      "_count": 41.0, 
      "count:sum": 41.0, 
      "service": "govuk-notify"
    }
  ], 
  "warning": "Warning: This data-set is unpublished. Data may be subject to change or be inaccurate."
}
```

leohemsted:
```sql
SELECT count(*) FROM (
    SELECT s.name
    FROM services s
    JOIN user_to_service uts ON uts.service_id = s.id
    JOIN users u ON uts.user_id = u.id
    WHERE NOT s.restricted
    GROUP BY s.name
    HAVING NOT bool_or(u.platform_admin)
) as anon;
```

😎

PR #1196, opened 2017-03-20
Remove unwanted SASS

quis:
None of this SASS is used by the app any more. But we’re still shipping it to every user. Let’s save some filesize by not doing so.


quis:
Restarted the Jenkins job manually because it seemed stuck. Tests passed: https://jenkins.notify.tools/job/notifications-admin/731/console

PR #1197, opened 2017-03-20
Put template IDs back on the dashboard

quis:
Removed as part of refactoring the code to generate the graphs of template usage on the dashboard: https://github.com/alphagov/notifications-admin/commit/4a226a7a29d83a4187ae2df089098bbdcebf2202#diff-cf78cb5c29a2d3c4d45b61d8617824b7L29

Didn’t realise that they were used by the functional tests.

This commit puts them back while keeping the code reuse.

### DIFF
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -135,10 +135,11 @@
 
 {% macro spark_bar_field(
   count,
-  max_count
+  max_count,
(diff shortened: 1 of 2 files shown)


quis:
YOLOmerging because this is blocking my deploy.

PR #1198, opened 2017-03-20
Fix format_notification_status to handle the letter status.

servingUpAces:
Not sure what temp and perm failure should look like so I left them for now.


minglis:
👍

PR #1205, opened 2017-03-24
Handle lists as CSV field values

quis:
Brings in:
- [x] alphagov/notifications-utils#128

This means that `RecipientCSV` will sometimes return the value of a cell in a spreadsheet as a `list`, not a `str`. So we need to handle that, rather than putting a Python representation (`['one', 'two', 'three']`) on the page.

This commit handles it by putting a bulleted list on the page instead.

This breaks our model of showing the spreadsheet as it appears in Excel or whatever, because we’re showing the aggregation of the columns into a list. However:
- this is the easier thing to do for now
- it might actually be more usable because it keeps the table narrower

***

screen shot 2017-03-24 at 11 48 57


quis:
Test added in 2ad9753, merging…

PR #1211, opened 2017-04-03
Don’t use https for localhost

quis:
We set a URL for the app to use to access itself. On PaaS this is configurable, locally we default to `localhost`. `localhost` doesn’t (easily?) support HTTPS, so this default wasn’t working.


quis:
Merging but not deploying this because it only affects code running locally.

PR #1221, opened 2017-04-10
Increase width of edit letter textbox

quis:
# The problem

image

The textbox we use for editing letters is the same size as that for email and text messages.

This is problematic because:
- it feels quite cramped – letters will often be longer than emails or text messages
- it has a narrower line length than the printed letters (which is a constant, unlike for emails and text messages)

# Working out the solution

The printed letters have a line length of 137.5mm and a font size of 12.5pt.

137.5mm = 5.41 inches = 389.7pt line length

389.7pt/12.5pt = 31.8em

So we could make the box 31.8em wide, but then it wouldn’t align to our grid.

screen shot 2017-04-10 at 09 03 18

Our grid splits the page into quarters initially because this is how wide the navigation is. So this means that we can use grid units of 1/multiples of four, eg 1/4, 1/8, 1/12, 1/16, etc. But the smaller the denominator, the less effective the grid will be – it gets closer to no grid at all.

After having a play around, 5/8 of the page looks closest to 31.8em. Since the main column of the page is 3/4, we set a column of 5/6 width inside that, which equals 5/8 of the total page.

It’ll never break the lines at the exact same place as the preview, because the textbox uses a  different typeface (with different metrics) from the preview. But it’s close.

# After

image

image

### DIFF
--- a/app/assets/stylesheets/_grids.scss
+++ b/app/assets/stylesheets/_grids.scss
@@ -18,6 +18,10 @@
   @include grid-column(1/8);
 }
 
+.column-five-eighths {
+  @include grid-column(5/8);
+}
+
 .column-seven-eighths {
   @include grid-column(7/8);
 }
--- a/app/assets/stylesheets/components/sms-message.scss
+++ b/app/assets/stylesheets/components/sms-message.scss
@@ -2,7 +2,7 @@
 .sms-message-wrapper {
 
   width: 100%;
-  max-width: 450px;
+  max-width: 464px;
   box-sizing: border-box;
   padding: $gutter-half $gutter-half $gutter-half $gutter-half;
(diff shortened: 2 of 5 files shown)


quis:
# Also, fix alignment of edit letter links

Not sure how these got out of line (maybe when we brought the date into the left-hand text area). But this commit updates the percentages to match the comments.

This is so it’s clear from the position of the link which part of the letter you’re editing.
***

image

quis:
# Couple of other things that needed their widths tweaking

## Edit email

Edit email box gets wider to match more closely the width of the previewed and delivered emails


### Before 

image

### After

image


# Text message preview

SMS message preview gets slightly wider so it lines up with a 4/8 column

## Before

image

## After

image

PR #1223, opened 2017-04-11
Letter jobs page

kentsanggds:
## What 

Added letter jobs page to platform admin with supporting tests in order to allow platform admin users to manually send DVLA letter jobs to the ftp client via the API.

Initially sending a letter job will put the job status as `in progress`, the user will have to wait until it is `ready to send` before they can select the job for sending.

Attempting to click send without selecting a job will display the `No jobs to send` message.

There will be an indication of which jobs were sent, but this information will be lost when clicking on the `Letter jobs` link at the top.

## How to review

execute `pytest tests/app/main/views/test_letter_jobs.py`

### DIFF
--- a/app/main/views/letter_jobs.py
+++ b/app/main/views/letter_jobs.py
@@ -1,45 +1,31 @@
-from flask import (render_template, url_for, redirect, request, abort)
+import datetime
+
+from flask import render_template, request
+from flask_login import login_required
+
+from app import letter_jobs_client, format_datetime_24h
 from app.main import main
-from app import convert_to_boolean
-from flask_login import (login_required, current_user)
+from app.utils import user_has_permissions
 
 
 @main.route("/letter-jobs", methods=['GET', 'POST'])
 @login_required
+@user_has_permissions(admin_override=True)
 def letter_jobs():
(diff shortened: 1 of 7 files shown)


app/main/views/letter_jobs.py
```diff
@@ -0,0 +1,31 @@
...
+    letter_jobs_list = letter_jobs_client.get_letter_jobs()
+
+    if request.method == 'POST':
+        if len(request.form.getlist('job_id')) > 0:
+            job_ids = request.form.getlist('job_id')
+
+            response = letter_jobs_client.send_letter_jobs(job_ids)
+            msg = response['response']
+
+            for job_id in job_ids:
```
minglis: Is this to cover the fact that the job status is updated async? Cause at this point it I guess may fail? What happens if I come back to the page - will it give a consistent response ie on POST it fakes the status, whereas coming back to the page my show the older status? or have I misread it?

minglis: Agreed with Ken that this is OK as (a) platform admin and (b) we need to dig further into job lifecycle on a letter job

PR #1224, opened 2017-04-11
Letters from template-preview service

leohemsted:
This is to get letter previews working again. They now run from another service, so pango/imagemagick/ghostscript etc are no longer dependencies

### To run this service locally

Checkout https://github.com/alphagov/notifications-template-preview, and run it locally on port 6013.

### To run this service on PaaS

Ensure service has been pushed to PaaS, then make sure that notify-admin has the `notify-template-preview` cf service added to it - the config should pick up the credentials and URL from that

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1,6 +1,5 @@
 from datetime import datetime, timedelta
 from string import ascii_uppercase
-from io import StringIO
 
 from flask import (
     request,
@@ -9,21 +8,17 @@ from flask import (
     url_for,
     flash,
     abort,
-    send_file,
-    current_app
 )
 from flask_login import login_required, current_user
-from flask_weasyprint import HTML, render_pdf
 from dateutil.parser import parse
-import requests
 
 from notifications_utils.formatters import escape_html
-from notifications_utils.template import LetterPreviewTemplate
 from notifications_utils.recipients import first_column_headings
 from notifications_python_client.errors import HTTPError
 
 from app.main import main
-from app.utils import user_has_permissions, get_template, png_from_pdf
+from app.utils import user_has_permissions, get_template
+from app.template_previews import TemplatePreview
 from app.main.forms import (
     ChooseTemplateType,
     SMSTemplateForm,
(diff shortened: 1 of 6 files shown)


app/template_previews.py
```diff
@@ -0,0 +1,29 @@
...
+        data = {
+            "letter_contact_block": current_service['letter_contact_block'],
+            "admin_base_url": current_app.config['ADMIN_BASE_URL'],
+            "template": template,
+            "values": values
+        }
+        resp = requests.post(
+            '{}/preview.{}'.format(current_app.config['TEMPLATE_PREVIEW_SERVICE_URL'], filetype),
+            json=data,
+            headers={'Authorization': 'Token my-secret-key'}
```
imdadahad: Need to pull from cf config

quis:
Refactor here is good. I think we should run through it tomorrow with it all working on someone else’s machine before merging.

leohemsted:
yeah, agreed, both them running the service locally and also them pushing to sandbox and running against that so we can make sure creds are correct

app/main/views/templates.py
```diff
@@ -87,26 +84,12 @@ def view_template(service_id, template_id):
     )
 
 
-@main.route("/services/<service_id>/templates/<template_id>.pdf")
+@main.route("/services/<service_id>/templates/<template_id>.<filetype>")
```
imdadahad: We probably want to add some validation for the `filetype`

imdadahad: with tests too

leohemsted: done (and for send.py)

quis:
✅  Got it running locally

README.md
```diff
@@ -15,18 +15,10 @@ GOV.UK Notify admin application.
 
 ## First-time setup
 
-Brew is a package manager for OSX. The following command installs brew:
-```shell
-    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-```
-
```
quis: Brew is still needed to install Node.

PR #1227, opened 2017-04-11
Fix new checkboxes

quis:
Problems:
- WTForms expects the value of checkboxes to always be `y` (they don’t work like radio buttons, which is where I copied this code for)
- WTForms `BooleanField`s don’t have a checked attribute, they set their data attibute to `True` or `False`


quis:
Demo’d to @leohemsted. Going to merge now, but not follow the deploy through until tomorrow morning.

PR #1229, opened 2017-04-12
Redirect after letter-jobs post

kentsanggds:
## What

In order to prevent a letter job from being resubmitted a redirect was introduced after POST.

## How to review 

execute `pytest tests/app/main/views/test_letter_jobs.py`

### DIFF
--- a/app/main/views/letter_jobs.py
+++ b/app/main/views/letter_jobs.py
@@ -1,6 +1,6 @@
 import datetime
 
-from flask import render_template, request
+from flask import redirect, render_template, request, session, url_for
 from flask_login import login_required
 
 from app import letter_jobs_client, format_datetime_24h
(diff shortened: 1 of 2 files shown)


app/main/views/letter_jobs.py
```diff
@@ -12,20 +12,27 @@
...
+        session['msg'] = msg
+
+        return redirect(url_for('main.letter_jobs'))
+
+    msg = session.pop('msg', None)
+    job_ids = session.pop('job_ids', None)
+    if job_ids:
+        for job_id in job_ids:
+            job = [j for j in letter_jobs_list if job_id == j['id']][0]
+            job['sending'] = 'sending'
```
quis: This block is confusing to me, but I guess you’re not actually changing it as part of this PR.

PR #1234, opened 2017-04-19
Add platform admin setting for letter branding

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/906
- [ ] https://github.com/alphagov/notifications-template-preview/pull/8

***

Does two main things:
- defines what ‘brands’ we support, in terms of the ID that DVLA use
- adds a form to choose which branding a service uses (currently platform admin only, like email branding)

By doing this we will be able to (with some more work) preview and send letters with a variety of different branding.

Story: https://www.pivotaltracker.com/story/show/143506905

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -83,11 +83,6 @@ class Config(object):
         r"assembly\.wales",
     ]
 
-    LETTER_BRANDING = [
-        ('001', 'HM Government'),
-        ('500', 'Land Registry'),
-    ]
-
 
 class Development(Config):
     DEBUG = True
--- a/app/config.py
+++ b/app/config.py
@@ -83,6 +83,11 @@ class Config(object):
         r"assembly\.wales",
     ]
 
+    LETTER_BRANDING = [
+        ('001', 'HM Government'),
+        ('500', 'Land Registry'),
+    ]
+
 
 class Development(Config):
     DEBUG = True
--- a/app/notify_client/organisations_client.py
+++ b/app/notify_client/organisations_client.py
@@ -16,3 +16,6 @@ class OrganisationsClient(NotifyAdminAPIClient):
 
     def get_organisations(self):
         return self.get(url='/organisation')['organisations']
+
+    def get_letter_organisations(self):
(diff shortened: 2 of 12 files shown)


app/config.py
```diff
@@ -83,6 +83,11 @@ class Config(object):
         r"assembly\.wales",
     ]
 
+    LETTER_BRANDING = [
+        ('001', 'HM Government'),
+        ('500', 'Land Registry'),
+    ]
```
leohemsted: does it make sense getting this from an API endpoint to avoid duplication of DB data?

we can have a simple GET endpoint at the API that just returns `select * from dvla_organisation`

leohemsted:
i clicked approve but i meant to click comment /shrug

app/config.py
quis: Yes, does make sense, it’s how we do it for email branding IIRC.

For now or later?

leohemsted: eh might as well do it now, shouldn't take too long to add to the API

`GET /dvla_organisations` returns 
```
[
{"id": "001", "name": "HM Government"}, ...
]
```

sound alright?

quis: Yeah, seems legit.

quis:
@leohemsted Added code to call through to the API for the organisation list in 544c809. API should be returning the data as a dictionary, see mocked example here: https://github.com/alphagov/notifications-admin/pull/1234/commits/544c809ad12ef8430fd91bbfa06a3f3990d2da2d#diff-484462fced51d1a06b1d93b4a44dd535R1363

app/notify_client/organisations_client.py
```diff
@@ -16,3 +16,6 @@ def get_organisation(self, id):
 
     def get_organisations(self):
         return self.get(url='/organisation')['organisations']
+
+    def get_letter_organisations(self):
+        return self.get(url='/foo')
```
leohemsted: `/dvla_organisations`

quis: Fixed and squashed out.

PR #1236, opened 2017-04-20
Fix CSV download timeout issue

imdadahad:
This fixes the issue where downloading a large report (10,000+) would timeout after 30s.

## Summary

There were a number of issues identified in Admin:

1. The **API call to retrieve 5000 notifications was slow**, clocking in about 3-4s. This was due to serialisation on the API taking pretty long. This has been reduced down to 2ish seconds in [1]

2. **Buffering**: We were storing each notification inside a buffer and then using `csv` module to prepare in the CSV format. This buffering added an extra 1-2 seconds. 

3. **Formatting**: We were doing multiple operations on the `created_at` for each notification which slowed us down significantly. In particular we were converting back into a `datetime` with `dateutil.parser.parse` and then called `datetime.strftime` twice to finally get it into a friendly format. This was a big bottleneck and cocked in at 1-3s per 5000. This has been moved to the API and refactored down in [1]

## Results

A 50k csv locally running both apps with gunicon (important) now takes 15ish seconds. We previously timed out at 30s for 10k notifications.

## Dependencies 
- [x] https://github.com/alphagov/notifications-api/pull/908 [1]

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -150,7 +150,8 @@ def view_job_csv(service_id, job_id):
                 job_id=job_id,
                 status=filter_args.get('status'),
                 page=request.args.get('page', 1),
-                page_size=5000
+                page_size=5000,
+                format_for_csv=True
             )
         ),
         mimetype='text/csv',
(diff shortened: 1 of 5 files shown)


minglis:
👍

PR #1237, opened 2017-04-20
Show all pages of a letter in the app

quis:
In research we’ve seen two problems with the click-to-see-PDF thing:

- it’s not very intuitive that the letter is clickable, or what you can expect when clicking the letter
- people get lost of stuck in the PDF view because it opens in the same tab, or they open it in a new tab and then get find their way back, or …

So this commit changes the show template page to show the entire contents of the letter, same as we do for emails and text messages.

Right now it only does it on the view template page. I think we’ll have to work out a way of showing some kind of truncated version on the _Send yourself a test_ and _Preview_ pages. But that’s for later.

***

Depends on:

- [x] https://github.com/alphagov/notifications-template-preview/pull/7

***

screencapture-localhost-6012-services-42a9d4f2-1444-4e22-9133-52d9e406213f-templates-8b7efa5e-8fbd-4f1a-8f0f-2a5d6b1656d2-1492700163132

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -8,6 +8,7 @@ from flask import (
     url_for,
     flash,
     abort,
+    json,
 )
 from flask_login import login_required, current_user
 from dateutil.parser import parse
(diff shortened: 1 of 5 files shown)


quis:
Should probs find a way to fix the alignment of those edit buttons 🤔

minglis:
Think this is OK 👍

We should talk about caching some of the PNG/PDF generation as this is slow for a UI to be backed onto.

PR #1240, opened 2017-04-24
Add international sms feature flag

kentsanggds:
Added feature flag to allow sending of international sms


tests/app/main/views/test_service_settings.py
```diff
@@ -845,6 +847,39 @@ def test_switch_service_disable_letters(
...
+
+def test_switch_service_disable_international_sms(
+    logged_in_platform_admin_client,
+    service_one,
+    mocker,
+):
+    service_one['can_send_international_sms'] = True
+    mocked_fn = mocker.patch('app.service_api_client.update_service_with_properties', return_value=service_one)
+
+    response = logged_in_platform_admin_client.get(
```
minglis: Should this be a POST - state is changing

app/main/views/service_settings.py
```diff
@@ -192,6 +192,17 @@ def service_switch_can_send_letters(service_id):
     return redirect(url_for('.service_settings', service_id=service_id))
 
 
+@main.route("/services/<service_id>/service-settings/can-send-international-sms")
```
minglis: Should be explicity a POST I think as state changes.

PR #1244, opened 2017-04-25
Sort platform admin page by requested messages.

minglis:
- puts services that are doing things at the top.

### DIFF
--- a/tests/app/main/views/test_platform_admin.py
+++ b/tests/app/main/views/test_platform_admin.py
@@ -7,7 +7,7 @@ from bs4 import BeautifulSoup
 from tests.conftest import mock_get_user
 from tests import service_json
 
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_platform_admin.py
```diff
@@ -347,3 +347,92 @@ def test_should_show_correct_sent_totals_for_platform_admin(
...
+        emails_delivered=50,
+        emails_failed=50,
+        sms_requested=200,
+        sms_delivered=50,
+        sms_failed=50
+    )
+
+    mock_get_detailed_services.return_value = {'data': services}
+    mock_get_user(mocker, user=platform_admin_user)
+    client.login(platform_admin_user)
```
imdadahad: Think we have a fixture for this now `logged_in_platform_admin_client`

PR #1247, opened 2017-04-26
Add a back button to the scheduler

quis:
We’ve seen in research a user getting stuck playing with the scheduler. They picked a day, but then didn’t want to choose one of the options for that day. There’s no way to do this except pick a day and then un-pick it.

What they ended up doing was clicking the grey back button, which took them back to the previous page, making them upload their file again.

This commit adds a ‘back’ link for the scheduler. ‘Back’ seems like sensible naming because that’s the thing that the user tried to click, and the UI of a link matches the thing they clicked to get into this situation.

***

add-back-link-scheduler

### DIFF
--- a/app/assets/javascripts/radioSelect.js
+++ b/app/assets/javascripts/radioSelect.js
@@ -56,7 +56,7 @@
   let focusSelected = function() {
     setTimeout(
       () => $('[type=radio]:checked').next('label').blur().trigger('focus').addClass('selected'),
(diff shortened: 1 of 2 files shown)


quis:
Similarity, proximity and good continuation 😉 : http://graphicdesign.spokanefalls.edu/tutorials/process/gestaltprinciples/gestaltprinc.htm

PR #1254, opened 2017-04-27
add sent as a new status

leohemsted:
functions as delivered - shows up in delivered column, contributes
to delivered stats etc.

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -314,7 +314,8 @@ def format_notification_status(status, template_type):
             'permanent-failure': 'Email address doesn’t exist',
             'delivered': 'Delivered',
             'sending': 'Sending',
-            'created': 'Sending'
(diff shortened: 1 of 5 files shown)


app/__init__.py
```diff
@@ -314,7 +314,8 @@ def format_notification_status(status, template_type):
             'permanent-failure': 'Email address doesn’t exist',
             'delivered': 'Delivered',
             'sending': 'Sending',
-            'created': 'Sending'
+            'created': 'Sending',
+            'sent': 'Sent internationally'
```
quis: Appreciate that this is to stop the app crashing if we do something unexpected, but should probably map to ‘Delivered’ for `email` and `letter`.

leohemsted: squashed into cf125d0

PR #1264, opened 2017-05-08
Only show back button if you’ve come from settings

quis:
You can also reach this page from the individual template view.


quis:
This doesn't work as intended – do not merge.

PR #1270, opened 2017-05-11
Show all pages of a letter everywhere

quis:
Right now showing all the pages in full is the only way we have of showing a letter that makes sense to our users. Maybe in the future we show some kind of truncated version, but the end of the first page is not a good place to truncate the letter.

This commit just extracts the code for showing multiple pages from the `template` views, refactors it for reuse, and includes it in the `send` views.

### DIFF
--- a/app/template_previews.py
+++ b/app/template_previews.py
@@ -1,4 +1,4 @@
-from flask import current_app
+from flask import current_app, json
 import requests
 
 from app import current_service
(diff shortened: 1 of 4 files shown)


app/template_previews.py
```diff
@@ -32,3 +32,14 @@ def from_utils_template(cls, template, filetype, page=None):
             template.values,
             page=page,
         )
+
+
+def get_page_count_for_letter(template, values=None):
```
leohemsted: should have a couple o' unit tests for this - make sure it handles non letter templates fine, and that it gets the count out of the json blob

quis: Tests added for those two things and squashed in.

quis:
Going to merge this since I’ve addressed the one comment and @leohemsted is away.

PR #1277, opened 2017-05-17
Show one field at a time on send yourself a test

quis:
# Screenshots

## Before

image

## After

tour-opp

***

tour-opp

# Background on the _Send yourself a test_ feature

We built it because it’s useful for:
- constructing an email/text message/letter without uploading a CSV file
- seeing what the thing your going to send will look like (either by getting it in your inbox or downloading the PDF)
- learning the concept of placeholders, ie understanding they’re thing that gets populated with _stuff_

# The problem with it

The current UI breaks when a template has a lot of placeholders. This is especially apparent with letter templates, which have a minimum of 7 placeholders by virtue of the address.

# Why we have this problem

The idea behind having the form fields side-by-side was to help people understand the relationship between their spreadsheet columns and the placeholders. But this means that the page was doing a lot of work, trying to teach both:
- replacement of placeholders
- link between placeholders and spreadsheet columns

The latter is better explained by the example spreadsheet shown on the upload page. So it can safely be removed from the send yourself a test page – in other words the fields don’t need to be shown side by side.

# Why this solution

Showing them one-at-a-time works well because:
- it’s really obvious, even on first use, what the page is asking you to do
- as your step through each placeholder, you see the message build up with the data you’ve entered – you’re learning how replacement of placeholders works by repetition

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,7 +1,6 @@
 import re
 import pytz
 
-from flask_login import current_user
 from flask_wtf import Form
 from datetime import datetime, timedelta
 from notifications_utils.recipients import (
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,6 +1,6 @@
 import re
-
 import pytz
+
 from flask_login import current_user
 from flask_wtf import Form
 from datetime import datetime, timedelta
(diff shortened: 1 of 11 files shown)


app/main/forms.py
```diff
@@ -1,6 +1,7 @@
+import copy
```
quis: Removed the `copy` and squashed into 8c03feb3.

Removed other unused import in 5da24288.

tests/app/main/views/test_send.py
```diff
@@ -336,7 +335,33 @@ def test_send_test_sms_message(
...
+    logged_in_client,
+    mocker,
+    api_user_active,
+    mock_login,
+    mock_get_service,
+    mock_get_service_template,
+    mock_has_permissions,
+    mock_get_users_by_service,
+    mock_get_detailed_service_for_today,
+    fake_uuid
```
quis: Fixtures addressed in 4072658e

app/main/views/send.py
```diff
@@ -147,62 +157,130 @@ def get_example_csv(service_id, template_id):
...
+    return redirect(url_for(
+        '.send_test_step',
+        service_id=service_id,
+        template_id=template_id,
+        step_index=0,
+        help=get_help_argument(),
+    ))
+
+
+@main.route("/services/<service_id>/send/<template_id>/test/step-<int:step_index>", methods=['GET', 'POST'])
```
quis: Think I’ve addressed all of those in 2f8e3e12

PR #1283, opened 2017-05-19
Set default text message sender to GOVUK

quis:


quis:
Closing this because we should just use whatever’s in the DB.

PR #1284, opened 2017-05-19
Stop letting people enter empty SMS senders

quis:
This shouldn’t be deployed until we’ve migrated the database to default to GOVUK.


leohemsted:
- [ ] https://github.com/alphagov/notifications-api/pull/973

PR #1286, opened 2017-05-23
Remove fallback for empty SMS sender

quis:
The API will **(not yet)** always return a value for this now (defaults to GOV.UK)


leohemsted:
- [  ] https://github.com/alphagov/notifications-api/pull/977

quis:
- [ ] https://github.com/alphagov/notifications-api/pull/988

PR #1291, opened 2017-05-24
Add page to show two-way conversation

quis:
two-way-scroll

### DIFF
--- a/app/main/views/conversation.py
+++ b/app/main/views/conversation.py
@@ -1,0 +1,55 @@
+from flask import (
+    render_template,
+    url_for,
+)
+from flask_login import login_required
+from notifications_utils.recipients import format_phone_number_human_readable
+from notifications_utils.template import SMSPreviewTemplate
+from app.main import main
+from app.utils import user_has_permissions
+from app import notification_api_client, service_api_client
+
+
+@main.route("/services/<service_id>/conversation/<notification_id>")
+@login_required
+@user_has_permissions('view_activity', admin_override=True)
+def conversation(service_id, notification_id):
+
+    user_number = get_user_number(service_id, notification_id)
+
+    return render_template(
+        'views/conversations/conversation.html',
+        conversation=get_sms_thread(service_id, user_number=user_number),
+        user_number=user_number,
+    )
+
+
+def get_user_number(service_id, notification_id):
+    try:
+        user_number = service_api_client.get_inbound_sms_by_id(service_id, notification_id)['user_number']
+    except Exception:
+        user_number = notification_api_client.get_notification(service_id, notification_id)['to']
+    return format_phone_number_human_readable(user_number)
+
+
+def get_sms_thread(service_id, user_number):
+    return [
+        {
+            'inbound': ('notify_number' in notification),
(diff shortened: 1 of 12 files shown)


app/main/views/conversation.py
```diff
@@ -0,0 +1,55 @@
...
+        user_number = service_api_client.get_inbound_sms_by_id(service_id, notification_id)['user_number']
+    except Exception:
+        user_number = notification_api_client.get_notification(service_id, notification_id)['to']
+    return format_phone_number_human_readable(user_number)
+
+
+def get_sms_thread(service_id, user_number):
+
+    for notification in sorted((
+        notification_api_client.get_notifications_for_service(service_id, to=user_number)['notifications'] +
```
minglis: How does/will pagination work? (guess it's possible - maybe for the future?)

app/notify_client/service_api_client.py
```diff
@@ -238,11 +239,22 @@ def get_billable_units(self, service_id, year):
             params=dict(year=year)
         )
 
-    def get_inbound_sms(self, service_id):
+    def get_inbound_sms(self, service_id, user_number=''):
         return self.get(
-            '/service/{}/inbound-sms'.format(service_id)
+            '/service/{}/inbound-sms?user_number={}'.format(
```
minglis: Guess we need to figure out how to kill this.

app/main/views/conversation.py
quis: I think only showing the `X` most recent messages is probably fine.

app/notify_client/service_api_client.py
quis: Yup. Shouldn’t block this PR though.

app/main/views/conversation.py
```diff
@@ -0,0 +1,55 @@
...
+        user_number=user_number,
+    )
+
+
+def get_user_number(service_id, notification_id):
+    try:
+        user_number = service_api_client.get_inbound_sms_by_id(service_id, notification_id)['user_number']
+    except Exception:
+        user_number = notification_api_client.get_notification(service_id, notification_id)['to']
+    return format_phone_number_human_readable(user_number)
```
leohemsted: need a test for this fn

tests/app/main/views/test_conversation.py
```diff
@@ -0,0 +1,95 @@
...
+    ),
+    (
+        'template content',
+        'yesterday at midnight',
+    ),
+    (
+        'template content',
+        'yesterday at midnight',
+    ),
+]))
```
leohemsted: i'd prefer you moved all these parametrised values to a separate test that just calls `get_human_day`, and then just have one test that actually calls test_view_conversation

app/main/views/conversation.py
quis: Tests added in eabd3f5

tests/app/main/views/test_conversation.py
quis: Removed the parametrizing in fd3f6f9

Still need this as an integration test because it’s testing the ordering and threading of the messages.

PR #1293, opened 2017-05-25
Make ‘send a one-off message’ a thing

quis:
one-off

*** 

This pull request started off as making an extra feature to send individual messages, which would be invite-only for service teams that want it.

I think we should replace _Send yourself a test_ with this new feature for everyone. Reasons being:
- the service permissions stuff isn’t ready
- it keeps the proposition/feature set/surface of the app smaller and easier to understand

It adds:
- another page in the flow, for you to enter an email address phone number
- a link on that page to prefill the message with your own email address/phone number, so you still don’t ever have to type in your own phone number/email address

### DIFF
--- a/app/assets/stylesheets/_grids.scss
+++ b/app/assets/stylesheets/_grids.scss
@@ -26,6 +26,19 @@
   @include grid-column(7/8);
 }
 
+%top-gutter,
+.top-gutter {
+  @extend %contain-floats;
+  display: block;
+  margin-top: $gutter;
+  clear: both;
+}
+
+.top-gutter-4-3 {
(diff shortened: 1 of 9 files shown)


quis:
# Rename the feature to ‘send to one recipient’ 

‘One-off’ is a bit wooly. Feels like our name for the thing. ‘Send to one recipient’ matches ‘Upload recipients’.

This also means making the `<h1>` on job page ‘Report’ for one-off messages. It doesn’t make sense to call the feature ‘send to one recipient’ when we’re not using the language of one-off any more.

Flow looks like:

image

***

image

***

image

***

image

***

image

PR #1297, opened 2017-05-30
Add search box to filter notifications by recipient

quis:
_Should not be merged until Friday 2nd June, ie until https://github.com/alphagov/notifications-api/pull/981 has fully propagated._

***

image

***

> Service teams that use the admin interface often need to know the outcome of a message... at the moment they have to page through all the results in the activity stream. They should be able to find notifications by email address or phone number.

– https://www.pivotaltracker.com/n/projects/1443052

# The filter funnel

The notifications are filtered down like this:

> _all notifications for service_ → _only failed_ → _only to this phone number_

For this reason, if the user navigates to a different ‘bucket’ of notifications (eg delivered, failed) then the search term is reset, because they’ve changed the filter which is at a level above the search term.

This also means that the counts aren’t updated based on the search term (in reality each service will only send a handful of notifications to one person in any 7 day period, so counting doesn’t feel like a requirement).

# Validation

For this pass at the story it doesn’t do any validation – the user will just get no results if they search by something which isn’t a phone number or email address.

This means that the API needs to accept anything as a search string, so this PR depends on:

- [x] https://github.com/alphagov/notifications-api/pull/997

### DIFF
--- a/tests/app/main/views/test_jobs.py
+++ b/tests/app/main/views/test_jobs.py
@@ -392,6 +392,14 @@ def test_can_show_notifications(
     ),
     (
         {
+            'message_type': 'sms',
+            'to': '+33(0)5-12-34-56-78',
+        },
+        'sending,delivered,failed',
+        '+33(0)5-12-34-56-78',
+    ),
+    (
+        {
             'status': 'failed',
             'message_type': 'email',
             'page': '99',
--- a/tests/app/main/views/test_jobs.py
+++ b/tests/app/main/views/test_jobs.py
@@ -8,6 +8,7 @@ from bs4 import BeautifulSoup
 
 from app.main.views.jobs import get_time_left, get_status_filters
 from tests import notification_json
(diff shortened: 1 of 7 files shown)


tests/app/main/views/test_jobs.py
```diff
@@ -367,6 +382,54 @@ def test_can_show_notifications(
...
+            'status': 'failed',
+            'message_type': 'email',
+            'page': '99',
+            'to': 'test@example.com',
+        },
+        'failed',
+        'test@example.com',
+    ),
+])
+def test_search_recipient_form(
```
imdadahad: Worth adding a test for an intl number?

quis: Added in 311732ff

imdadahad: 👍

quis:
I’m gonna merge this now. It’s not quite Friday yet, but it’s getting to the end of the working day so should be fine.

PR #1300, opened 2017-06-02
Service setting for inbound sms

servingUpAces:
Add a platform admin button to service-settings page to turn the inbound_sms messaging on and off.

If clicked you will be prompted to enter a sms sender number, when setting the permission on or off.
Team members will always be able to see the number, but will only be able to change it if the inbound_sms permission is off.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -49,6 +49,7 @@ def service_settings(service_id):
         letter_branding=letter_branding_organisations.get(
             current_service.get('dvla_organisation', '001')
         ),
+        can_receive_inbound=('inbound_sms' in current_service['permissions']),
         letter_contact_block=Field(current_service['letter_contact_block'], html='escape')
     )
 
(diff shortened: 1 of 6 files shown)


app/main/views/service_settings.py
```diff
@@ -49,6 +49,7 @@ def service_settings(service_id):
         letter_branding=letter_branding_organisations.get(
             current_service.get('dvla_organisation', '001')
         ),
+        can_receive_inbound=('inbound_sms' in current_service['permissions']),
```
quis: We should make something nicer than this for checking service permissions. Not for this PR though.

app/main/views/service_settings.py
```diff
@@ -267,14 +268,27 @@ def service_set_reply_to_email(service_id):
...
     form = ServiceSmsSender()
-    if request.method == 'GET':
-        form.sms_sender.data = current_service.get('sms_sender')
     if form.validate_on_submit():
-        service_api_client.update_service(
-            current_service['id'],
-            sms_sender=form.sms_sender.data or None
-        )
+        set_inbound_sms = request.args.get('set_inbound_sms', False)
+        if set_inbound_sms == 'True':
```
imdadahad: `if set_inbound_sms:` instead?

servingUpAces: set_inbound_sms is not a bool, it was easier to check it as a string then try to turn it into a boolean.

PR #1302, opened 2017-06-05
Add Care Quality Commission to email whitelist

quis:
CQC is an executive non-departmental public body, sponsored by the Department of Health.

They have asked to be allowed to register for Notify using the `cqc.org.uk` and `digital.cqc.org.uk` domains. We know that this really is there domain because it’s linked to from here:

https://www.gov.uk/government/organisations/care-quality-commission


servingUpAces:
It would be nice to be able to set these domains without doing a deploy.

PR #1324, opened 2017-06-16
One off status page

leohemsted:
We want to make one-off messages not need a job.

We want to be able to see the results of sending these messages.

For one-off messages, they'll have a new page, but for the help tutorial we still want to show the messages in the old job style.

This PR duplicates the job status page, but referring to a single notification.

- [ ] https://github.com/alphagov/notifications-api/pull/1036


tests/conftest.py
```diff
@@ -1671,3 +1700,28 @@ def os_environ():
...
     os.environ = old_env
+
+
+@pytest.fixture
+@pytest.fixture
+def client_request(logged_in_client):
+    class ClientRequest:
+
+        @staticmethod
+        def get(endpoint, endpoint_kwargs=None, expected_status=200, follow_redirects=False):
```
quis: Do you think the underscore pattern that `url_for` uses would be better/more consistent here? eg

```python
def get(endpoint, **kwargs, _expected_status=200, _follow_redirects=False)
```

quis: Also `follow_redirects=True` won’t do anything.

tests/conftest.py
```diff
@@ -1671,3 +1700,28 @@ def os_environ():
     os.environ = {}
     yield
     os.environ = old_env
+
+
+@pytest.fixture
+@pytest.fixture
```
quis: Duplicated lines?

tests/conftest.py
leohemsted: yeah i think that might look nice, makes it an easier transition from having previously called url_for using kwargs if we don't need to change them into a dict too 👍

app/config.py
```diff
@@ -96,6 +96,7 @@ class Development(Config):
 
 class Test(Development):
     DEBUG = True
+    TESTING = True
```
quis: Did you leave this in by accident?

leohemsted: http://flask.pocoo.org/docs/0.12/testing/

> disable the error catching during request handling so that you get better error reports when performing test requests against the application.

tests/conftest.py
```diff
@@ -1630,6 +1633,33 @@ def mock_reset_failed_login_count(mocker):
     return mocker.patch('app.user_api_client.reset_failed_login_count')
 
 
+@pytest.fixture
+def mock_get_notification(mocker, fake_uuid, notification_status='delivered'):
```
quis: This mock is already defined on line 1100. Can you remove the existing one to avoid confusion for anyone doing ⌘ + F on this file?

PR #1327, opened 2017-06-21
Create inbound api

kentsanggds:
## What 

Added service settings to manage inbound api which are shown when `inbound sms` permissions are on.

## How to review

Execute `pytest tests/app/main/views/test_service_settings.py`

Also you can manually check it by logging into admin and navigating to the service settings under a platform admin log in to ensure that -

- the API endpoint is shown when inbound sms are allowed and hidden when not allowed. 
- check that you can set the API endpoint, and can also see updates to it.
- check that only the domain is shown on the service settings screen and not the full url.
- check that you can update url / bearer token or both

### DIFF
--- a/app/templates/views/service-settings/set-inbound-api.html
+++ b/app/templates/views/service-settings/set-inbound-api.html
@@ -1,0 +1,35 @@
+{% extends "withnav_template.html" %}
+{% from "components/textbox.html" import textbox %}
+{% from "components/page-footer.html" import page_footer %}
+
+{% block service_page_title %}
+  Inbound api
+{% endblock %}
+
+{% block maincolumn_content %}
+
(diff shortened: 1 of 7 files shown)


app/templates/views/service-settings/set-inbound-api.html
```diff
@@ -0,0 +1,38 @@
...
+      <form method="post">
+        {{ textbox(
+          form.url,
+          width='2-3',
+          hint='Valid https url'
+        ) }}
+        {{ textbox(
+          form.bearer_token,
+          width='1-4',
+          hint='At least 10 characters'
```
servingUpAces: The bearer_token is always empty. But if I want to update either or both of the url and the bearer_token, it would be good to show **** for the bearer_token, so we know it has been set.

PR #1335, opened 2017-06-23
Use new service permissions

kentsanggds:
Remove references to old style of service flags to use new service permissions.


app/main/views/service_settings.py
```diff
@@ -208,25 +208,32 @@ def service_switch_research_mode(service_id):
     return redirect(url_for('.service_settings', service_id=service_id))
 
 
+def switch_service_permissions(service_id, permission):
```
servingUpAces: Can this new method be used in the service_set_sms_sender controller?

kentsanggds: ok

PR #1340, opened 2017-06-26
Send one off messages without jobs

leohemsted:
Lots of code change here (sorry!), but this PR changes the one off / send a test flow to, instead of creating a one-line CSV file and submitting a job, instead post a notification to a new endpoint. This is complicated. Three main reasons:

* Letters should still go via CSV because they aren't actually submitted as notifications
* In CSV files, the recipient is just another column. When sending as an API request, recipient is another field altogether, so we need to unpick that.
* The send_test_step function is already pretty long and complicated

image

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -638,14 +638,23 @@ def _check_notification(service_id, template_id, exception=None):
 
 
 def get_template_error_dict(exception):
+    # TODO: Make API return some computer-friendly identifier as well as the end user error messages
     if 'service is in trial mode' in exception.message:
         error = 'not-allowed-to-send-to'
+    elif 'Exceeded send limits' in exception.message:
+        error = 'too-many-messages'
+    elif 'Content for template has a character count greater than the limit of' in exception.message:
+        error = 'message-too-long'
     else:
         raise exception
 
     return {
         'error': error,
-        'SMS_CHAR_COUNT_LIMIT': 0
+        'SMS_CHAR_COUNT_LIMIT': current_app.config['SMS_CHAR_COUNT_LIMIT'],
+        'current_service': current_service,
+
+        # used to trigger CSV specific err msg content, so not needed for single notification errors.
+        'original_file_name': False
     }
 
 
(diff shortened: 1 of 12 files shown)


app/templates/views/notifications/check.html
```diff
@@ -0,0 +1,34 @@
+{% extends "withnav_template.html" %}
+<!-- {% from "components/banner.html" import banner_wrapper %} -->
+<!-- {% from "components/table.html" import list_table, field, text_field, index_field, hidden_field_heading %} -->
+<!-- {% from "components/file-upload.html" import file_upload %} -->
+<!-- {% from "components/page-footer.html" import page_footer %} -->
```
leohemsted: TODO: remove these lines

app/main/views/send.py
```diff
@@ -568,3 +577,83 @@ def get_send_test_page_title(template_type, help_argument):
...
+    if (
+        (
+            not session.get('recipient') or
+            not all_placeholders_in_session(template.placeholders)
+        )
+        and back_link
+    ):
+        return redirect(back_link)
+
+    template.values = get_receipient_and_placeholders_from_session(template.template_type)
```
quis: Recipient misspelled here.

quis: (and elsewhere)

app/templates/views/notifications/check.html
```diff
@@ -0,0 +1,30 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  <h1 class="heading-large">
+    Preview of {{ template.name }}
+  </h1>
+
+  {{ template|string }}
+
```
quis: This page needs to handle errors. The main one I can think of is sending to a phone number that’s not yours when in trial mode (right now it raises an exception). There may be others.

leohemsted: known errors:

* External phone number in trial mode
* Text message too long
* Service exceeded daily limit

leohemsted:
Error messages:

image
image
image


Error messages only come up after you hit submit, which isn't ideal, but we don't have a separate validate notification endpoint yet.

PR #1341, opened 2017-06-27
Add feature to mark a template as redacted

quis:
Works similarly to the delete template flow, because it’s a destructive, one-way action.

Not on the edit template page, because it’s not something you want to be considering every time you’re editing a template. And we saw that people couldn’t find the delete button when it was on this page.

Adds a bit more CSS for the `dangerous` banner type, because the content here is quite complicated. Breaking it into a list helps, but the spacing didn’t look right, so needed some tweaking.

Can ship independently of the code that shows the redaction, but needs the API first.

***

**Wording needs work:**

image

image

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -383,26 +383,10 @@ def delete_service_template(service_id, template_id):
     )
 
 
-@main.route("/services/<service_id>/templates/<template_id>/redact", methods=['GET', 'POST'])
+@main.route("/services/<service_id>/templates/<template_id>/redact", methods=['GET'])
 @login_required
 @user_has_permissions('manage_templates', admin_override=True)
-def redact_template(service_id, template_id):
-
-    if request.method == 'POST':
-
-        service_api_client.redact_service_template(service_id, template_id)
-
-        flash(
-            'Personalised content will be hidden for messages sent with this template',
-            'default_with_tick'
-        )
-
-        return redirect(url_for(
-            '.view_template',
-            service_id=service_id,
-            template_id=template_id,
-        ))
-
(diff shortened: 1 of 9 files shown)


quis:
Updated wording:

image

image

image

app/main/views/templates.py
```diff
@@ -383,6 +383,44 @@ def delete_service_template(service_id, template_id):
     )
 
 
+@main.route("/services/<service_id>/templates/<template_id>/redact", methods=['GET', 'POST'])
+@login_required
+@user_has_permissions('manage_templates', admin_override=True)
+def redact_template(service_id, template_id):
+
+    if request.method == 'POST':
```
leohemsted: i'm leaning towards separating out `GET` and `POST` into two separate functions, especially in situations like this where they don't have any shared form setup code.

flask handles all the routing just fine.

quis: Good point, changed in 23e665cc

quis:
Also it's weird that the API doesn't 400 for this, even though it's not built yet.

app/notify_client/service_api_client.py
```diff
@@ -170,6 +170,12 @@ def update_service_template(self, id_, name, type_, content, service_id, subject
         endpoint = "/service/{0}/template/{1}".format(service_id, id_)
         return self.post(endpoint, data)
 
+    def redact_service_template(self, service_id, id_):
+        return self.post(
+            "/service/{}/template/{}".format(service_id, id_),
+            {'redact_personalisation': True}
```
leohemsted: you'll also need to include an `'updated_by_id'`

quis:
Added confirmation:

image

PR #1360, opened 2017-07-07
Make conversation and inbox pages update using AJAX

quis:
Fairly self-explanatory. Uses the same pattern of breaking things up into functions as the jobs page.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -143,6 +143,23 @@ def monthly(service_id):
 @user_has_permissions('view_activity', admin_override=True)
 def inbox(service_id):
 
+    return render_template(
+        'views/dashboard/inbox.html',
+        partials=get_inbox_partials(service_id),
+        updates_url=url_for('.inbox_updates', service_id=service_id),
+    )
+
+
+@main.route("/services/<service_id>/inbox.json")
+@login_required
+@user_has_permissions('view_activity', admin_override=True)
+def inbox_updates(service_id):
+
+    return jsonify(get_inbox_partials(service_id))
+
+
+def get_inbox_partials(service_id):
+
     if 'inbound_sms' not in current_service['permissions']:
         abort(403)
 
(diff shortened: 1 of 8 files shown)


app/main/views/dashboard.py
```diff
@@ -143,6 +143,23 @@ def monthly(service_id):
...
+    return render_template(
+        'views/dashboard/inbox.html',
+        partials=get_inbox_partials(service_id),
+        updates_url=url_for('.inbox_updates', service_id=service_id),
+    )
+
+
+@main.route("/services/<service_id>/inbox.json")
+@login_required
+@user_has_permissions('view_activity', admin_override=True)
```
leohemsted: should also add tests for using the right permissions to access `inbox` and `inbox.json`

tests/app/main/views/test_dashboard.py
```diff
@@ -198,6 +199,30 @@ def test_inbox_not_accessible_to_service_without_permissions(
...
 
 
+def test_view_inbox_updates(
+    logged_in_client,
+    service_one,
+    mocker,
+    mock_get_inbound_sms_with_no_messages,
+):
+
+    service_one['permissions'] = ['inbound_sms']
```
leohemsted: `service_one` is just a json blob, so won't actually update the values in `current_service` i think?

app/main/views/dashboard.py
quis: Permissions fix is in this PR: https://github.com/alphagov/notifications-admin/pull/1361

tests/app/main/views/test_dashboard.py
quis: giphy-1

leohemsted: essentially, if you put a pdb inside the actual function, does `current_service['permissions'] == ['inbound_sms']`

quis: Yes. Somehow. 😬

app/main/views/dashboard.py
quis: Permissions test for the json endpoint added in 813c348c

app/main/views/dashboard.py
```diff
@@ -143,6 +143,25 @@ def monthly(service_id):
 @user_has_permissions('view_activity', admin_override=True)
 def inbox(service_id):
 
+    print(current_service)
```
leohemsted: 🖨

quis: It’s gone.

PR #1361, opened 2017-07-07
Fix wrong permissions on dashboard pages

quis:
Anyone with access to a service should be able to see:
- monthly breakdown (non financial)
- incoming messages

Not sure how we test this better to stop it happening again?

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -122,7 +122,7 @@ def usage(service_id):
 
 @main.route("/services/<service_id>/monthly")
 @login_required
-@user_has_permissions('manage_settings', admin_override=True)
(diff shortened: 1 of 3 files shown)


quis:
I found a way to add tests for this, but it uses a pretty old fixture `validate_route_permissions`. Probably a way we could roll this up into `client_request` and make testing this stuff less verbose. But works for now.

leohemsted:
yeah i think we need to look at the permissions stuff - ideally we have a single test file that hits every route on the admin app, and can raise an exception if we haven't added permissions data to that test file for a new endpoint. ideally not using that old `validate_route_permissions` fn which i really don't like

but it's fine for now 🚢

PR #1362, opened 2017-07-12
Add function to estimate letter delivery date

quis:
TL;DR, turns this picture into Python:

image

***

Letter delivery depends on:
- how long it takes to print
- how long it takes to post

Both of these process are impacted by weekends, because people don’t work on weekends.

It also depends on if you submit your letter before or after 5pm, because that’s the cut off time for getting a letter printed on a given day – ie after 5pm on Monday is effectively the same as Tuesday and so on.

But I reckon all our users need to know is roughly how long it will take until the letter turns up on the user’s doorstep. So this commit adds a function to calculate this. Doesn’t surface it on the front end _yet_.

### DIFF
--- a/tests/app/test_utils.py
+++ b/tests/app/test_utils.py
@@ -5,6 +5,7 @@ from csv import DictReader
 import pytest
 
 from collections import OrderedDict
+from freezegun import freeze_time
 
 from app.utils import (
     email_safe,
@@ -12,7 +13,7 @@ from app.utils import (
     generate_previous_dict,
     generate_next_dict,
     Spreadsheet,
-    get_estimated_delivery_date_for_letters,
+    get_letter_timings,
 )
 
 from tests import notification_json, single_notification_json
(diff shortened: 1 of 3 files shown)


quis:
# Add print timings to letter timings

It will also be useful to know (especially for the API):
- when a letter was printed
- if it’s been printed or not

This commit:
- adds code to calculate these two pieces of information
- refactors the function to return a `namedtuple` – a tuple of two items was manageable, but with four items it was getting hard to know what each one meant – this lets us label each piece of information that is being returned

***

This code should probably go into utils at the moment, but it’s only going to be used in admin for now.

tests/app/test_utils.py
```diff
@@ -154,3 +156,137 @@ def test_generate_notifications_csv_calls_twice_if_next_link(mocker):
...
+    expected_print_time,
+    is_printed,
+    expected_earliest,
+    expected_latest,
+):
+    timings = get_letter_timings(upload_time)
+    assert timings.printed_by.strftime('%A %H:%M') == expected_print_time
+    assert timings.is_printed == is_printed
+    assert timings.earliest_delivery.strftime('%A %Y-%m-%d') == expected_earliest
+    assert timings.latest_delivery.strftime('%A %Y-%m-%d') == expected_latest
```
leohemsted: good use of strftime to make tests readable 💯

PR #1367, opened 2017-07-13
Letter notification page

quis:
_Branched off https://github.com/alphagov/notifications-admin/pull/1362_

# Before

image
***
image

# After 

image
***
image

_ignore bug with the contact block not getting populated_

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -19,6 +19,7 @@ from app.utils import (
     get_help_argument,
     get_template,
     get_time_left,
+    get_letter_timings,
     REQUESTED_STATUSES,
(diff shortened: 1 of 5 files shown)


app/main/views/notifications.py
```diff
@@ -63,6 +63,8 @@ def view_notification(service_id, notification_id):
         job = job_api_client.get_job(service_id, notification['job']['id'])['data']
     else:
         job = None
+
+    print(get_letter_timings(notification['created_at']))
```
leohemsted: 🖨

quis: 🚫 🖨

PR #1378, opened 2017-07-21
Revert vendoring dependencies for cloundfoundry

imdadahad:
Despite following the instructions described here for vendoring dependencies: http://docs.cloudfoundry.org/buildpacks/python/#vendoring, it doesn't seem to work and instead runs `pip install -r requirements.txt` as opposed to with `--find-links --no-index` (https://github.com/cloudfoundry/python-buildpack/blob/1588bd4099b9b2f75448c62bcea80e38dd5795a8/bin/steps/pipenv). 

Reverting for now until we have some time to look later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,6 @@ generate-version-file: ## Generates the app version file
 build: dependencies generate-version-file ## Build project
 	npm run build
 	. venv/bin/activate && PIP_ACCEL_CACHE=${PIP_ACCEL_CACHE} pip-accel install -r requirements.txt
-	mkdir -p vendor/
-	cp ${PIP_ACCEL_CACHE}/sources/* vendor/
 
 .PHONY: cf-build
 cf-build: dependencies generate-version-file ## Build project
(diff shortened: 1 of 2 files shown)


leohemsted:
do you want to raise an issue on the python-buildpack project?

imdadahad:
I've sent a message on their slack channel so hoping they'll reply soon. 

If that doesn't lead anywhere I'll do that 👍 .

Makefile
```diff
@@ -86,8 +86,7 @@ generate-version-file: ## Generates the app version file
 build: dependencies generate-version-file ## Build project
 	npm run build
 	. venv/bin/activate && PIP_ACCEL_CACHE=${PIP_ACCEL_CACHE} pip-accel install -r requirements.txt
-	rm -rf vendor
-	mkdir -p vendor
+	mkdir -p vendor/
 	cp ${PIP_ACCEL_CACHE}/sources/* vendor/
```
leohemsted: you haven't actually reverted everything tho

imdadahad: Yeah just doing that now - will push up commits in a sec. Sorry should've mentioned that in the summary.

imdadahad: Also going to remove the wheel stuff anyway as that was _fairly_ independent of this.

leohemsted: 👍

PR #1388, opened 2017-07-26
Don’t set combined API on Notify python client

quis:
Because we’re setting the API key and service ID after calling the `__init__` method of the client it wasn’t doing the thing where it splits the combined key into the two individual UUIDs. So we still need to set them directly, individually on the client.


quis:
Tested this by:
- replicating bug on master locally, confirming fix in this branch
- deploying this branch to the prototype environment

Going to merge now to unblock the pipe.

PR #1389, opened 2017-07-26
Update last used message when deleting a template

imdadahad:
This will now show a generic message in the case that a given template to be deleted has been used more than seven days ago.

This can be merged before the API change that will ensure we no longer query the `NotificationHistory` table get the last usage of a template.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -428,12 +428,13 @@ def delete_service_template(service_id, template_id):
         last_used_notification = template_statistics_client.get_template_statistics_for_template(
             service_id, template['id']
         )
-        message = 'It was last used {} ago.'.format(
-            get_human_readable_delta(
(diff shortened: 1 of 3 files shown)


quis:
Think you’ve managed to revert some changes in the rebase. These lines in the tests should stay as is, and the code should be modified to make the tests pass.

```python
assert page.h1.text == 'Are you sure you want to delete Two week reminder?'
assert normalize_spaces(page.select('.banner-dangerous p')[0].text) == (
    'It was last used 10 minutes ago.'
)
```
– https://github.com/alphagov/notifications-admin/pull/1389/files#diff-f2eb304b93cc383727c0ab7fc8fbd464L701

imdadahad:
@quis yeah rebase troubles, updated now think it should all be okay.

PR #1392, opened 2017-07-27
Update org logos to use logo CDN

kentsanggds:
## What -

Updated `get_branding_as_dict` function used when setting organisation logo to get the logo from the logo CDN and not the static CDN which is pointing at the static web app folder.

Step 4/5 - https://www.pivotaltracker.com/story/show/148289053

## How to review -

When changing email branding on platform admin setting - 

- you should be able to see the logos
- right clicking on the logo should reveal the source to be coming from the logo CDN
>- `static-logos.[environment]\[image filename]`

### DIFF
--- a/tests/app/test_utils.py
+++ b/tests/app/test_utils.py
@@ -13,7 +13,11 @@ from app.utils import (
     generate_previous_dict,
     generate_next_dict,
     Spreadsheet,
+<<<<<<< HEAD
     get_letter_timings,
+=======
+    get_cdn_domain
+>>>>>>> Update emails to use logos cdn
 )
 
 from tests import notification_json, single_notification_json
@@ -158,6 +162,7 @@ def test_generate_notifications_csv_calls_twice_if_next_link(mocker):
     assert mock_get_notifications.mock_calls[1][2]['page'] == 2
 
 
+<<<<<<< HEAD
 @freeze_time('2017-07-14 14:59:59')  # Friday, before print deadline
 @pytest.mark.parametrize('upload_time, expected_print_time, is_printed, expected_earliest, expected_latest', [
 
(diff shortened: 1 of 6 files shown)


tests/app/test_utils.py
```diff
@@ -290,3 +288,15 @@ def test_get_estimated_delivery_date_for_letter(
     assert timings.is_printed == is_printed
     assert timings.earliest_delivery.strftime('%A %Y-%m-%d') == expected_earliest
     assert timings.latest_delivery.strftime('%A %Y-%m-%d') == expected_latest
+
+
+def test_get_cdn_domain_on_localhost(client, mocker):
+    mocker.patch.dict('app.current_app.config', values={'ADMIN_BASE_URL': 'http://localhost:6012'})
```
imdadahad: Think there's a `set_config` fixture you can use for this

tests/app/test_utils.py
```diff
@@ -290,3 +288,15 @@ def test_get_estimated_delivery_date_for_letter(
...
+    domain = get_cdn_domain()
+    assert domain == 'static-logos.notify.tools'
+
+
+def test_get_cdn_domain_without_logo_base_domain_env_returns_admin_base_domain(client, mocker):
+    mocker.patch.dict('app.current_app.config', values={'ADMIN_BASE_URL': 'https://some.admintest.com'})
```
imdadahad: Same as above comment

tests/app/test_utils.py
```diff
@@ -290,3 +288,15 @@ def test_get_estimated_delivery_date_for_letter(
...
     assert timings.latest_delivery.strftime('%A %Y-%m-%d') == expected_latest
+
+
+def test_get_cdn_domain_on_localhost(client, mocker):
+    mocker.patch.dict('app.current_app.config', values={'ADMIN_BASE_URL': 'http://localhost:6012'})
+    domain = get_cdn_domain()
+    assert domain == 'static-logos.notify.tools'
+
+
+def test_get_cdn_domain_without_logo_base_domain_env_returns_admin_base_domain(client, mocker):
```
imdadahad: This is essentially getting the domain name for a non `localhost` domain?

kentsanggds: yes, because at the moment the logos will be uploaded to the s3 bucket on tools, so need to ensure that it looks for the CDN at that location when the app is run for real.

kentsanggds: ok, can rename to `test_get_cdn_domain_for_non_localhost`

PR #1403, opened 2017-08-02
Add pages to create and manage organisation branding for Emails

kentsanggds:
## What - 

This is the final step, 5/5, for allowing platform admin users to create / update existing organisation branding - https://www.pivotaltracker.com/story/show/148289053.

- `Organisations` link in platform admin settings area
- admin can select an org to manage or create a new one
- upload a new logo, a file with a unique name will be uploaded to the s3 bucket, existing files will remain in case they are referenced in existing emails
- temporary files in s3 will be purged automatically once a logo is saved, otherwise they can be easily identified and manually purged as they start with `temp-`
- `uploader.py` has been renamed to `s3_client.py` as this feels more appropriate

## How to review -

Log into an account with platform admin rights, click on Organisations at the top
- select an organisation to edit and make changes
>- save the changes, changes should appear on the select orgs page
>- do not save the changes, there should be no changes to the organisation

- create a new organisation, save the changes and see that the new organisation is available 
>- Save button will be initially disabled as a logo is required for an organisation
>- valid colour entries should update a colour bar next, eg `red`, `#888`
>- save the changes, changes should appear on the select orgs page
>- do not save the changes, there should be no changes to the organisation

- send yourself an email associated with the updated / new organisation

- additionally check that no temp files remaining in the s3 bucket if changes have been saved
>- any temp files remaining would be due to a user uploading logos but not saving the change, 
>>- these will clear when the user next performs a save


quis:
Don’t have time now to look at the whole thing, but the _Organisations_ link should go in the new platform admin navigation (see https://github.com/alphagov/notifications-admin/pull/1382) not in the top bar like this:

image

kentsanggds:
OK, sure will move it to new area

PR #1407, opened 2017-08-08
Don’t let users in trial mode send letters

quis:
image

## Why we’re making this change

Users in trial mode haven’t signed the MOU. This means that they haven’t agreed to pay for any costs they incur.

Unlike text messages and emails, we don’t give you any free allowance of letters. Sending _any_ letters will cost the user money.

Therefore we shouldn’t let users who haven’t agreed that they will pay for the service to incur costs by sending letters.

Getting this in place means we can turn letters on for users in trial mode without worrying that they’ll accidentally send real letters, which would result in:
- us having to absorb those costs
- or some awkward conversations

### The pattern

The pattern used for this is roughly the same as other trial mode errors that we have already, ie a red box that says you’re not allowed. Not sure if this is exactly right because it’s not exactly an error so the pattern might feel too heavy-handed.

### 50 recipients limit

We should also make sure that users trying to send letters don’t get told about the 50 recipients limit. The ‘can’t send to more than 50 recipients in trial mode’ error doesn’t apply for letters (they can’t send to _any_ recipients).  So we should make sure that the error message about not being able to send to any recipients always comes up instead of the 50 recipients one, whether you’re trying to upload a file with 1 or 111 rows.

### The trial mode page

The meaning of trial mode has evolved now that we’re adding letters. So this PR also updates the trial mode content page to make it clear:
- that the existing conditions only apply to texts and emails
- we need to explain what trial mode means for letters

### DIFF
--- a/app/templates/views/check/column-errors.html
+++ b/app/templates/views/check/column-errors.html
@@ -83,10 +83,6 @@
           {% include "partials/check/not-allowed-to-send-to.html" %}
         {% endwith %}
 
-      {% elif recipients.more_rows_than_can_send %}
-
-        {% include "partials/check/too-many-messages.html" %}
-
       {% elif trying_to_send_letters_in_trial_mode %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="Trying to send letters in trial mode" data-error-label="{{ upload_id }}">
@@ -97,6 +93,10 @@
           In <a href="{{ url_for('.trial_mode') }}">trial mode</a> you
           can only preview how your letters will look
         </p>
+
+      {% elif recipients.more_rows_than_can_send %}
+
+        {% include "partials/check/too-many-messages.html" %}
 
       {% endif %}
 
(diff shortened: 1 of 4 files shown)


app/templates/views/check/column-errors.html
```diff
@@ -75,14 +75,29 @@ <h1 class='banner-title' data-module="track-error" data-error-type="Missing plac
...
       {% elif not recipients.allowed_to_send_to %}
+
         {% with
           count_of_recipients=count_of_recipients,
           template_type_label=recipients.recipient_column_headers[0]
         %}
           {% include "partials/check/not-allowed-to-send-to.html" %}
         {% endwith %}
+
+      {% elif trying_to_send_letters_in_trial_mode %}
```
leohemsted: should this not be at the top of the list? i think this should come before column mismatches or file too large messages

quis:
@servingUpAces Yes.

app/templates/views/check/column-errors.html
quis: @leohemsted This is consistent with what we do for emails and text messages. It’s better that users can find out how the columns and accepted file types work even when they’re in trial mode.

PR #1409, opened 2017-08-08
Let api check unique email

servingUpAces:
The service name and email_from address needs to be unique, the AddServiceForm was doing this check in a very inefficient way. This PR lets the create_service call to the api through the integrity error. We should see a vast improvement to the performance of creating a new service. 

Next to come is to fix the update service name.


app/main/views/add_service.py
```diff
@@ -39,14 +39,30 @@ def _add_invited_user_to_service(invited_user):
...
+                                                       message_limit=current_app.config['DEFAULT_SERVICE_LIMIT'],
+                                                       restricted=True,
+                                                       user_id=session['user_id'],
+                                                       email_from=email_from)
+        session['service_id'] = service_id
+        return service_id, None
+    except HTTPError as e:
+        if e.status_code == 400 and e.message['name']:
+            form.name.errors.append("This service name is already in use")
+            return None, e
```
leohemsted: we should probably have an `else: raise` here, or we'll swallow up legit errors

app/main/views/add_service.py
```diff
@@ -60,24 +76,20 @@ def add_service():
     if not is_gov_user(current_user.email_address):
         abort(403)
 
-    form = AddServiceForm(service_api_client.find_all_service_email_from)
```
leohemsted: i think we should delete this `find_all_service_email_from` function now

app/main/views/add_service.py
```diff
@@ -60,24 +76,20 @@ def add_service():
...
 
     if form.validate_on_submit():
         email_from = email_safe(form.name.data)
         service_name = form.name.data
-        service_id = _create_service(service_name, email_from)
 
-        if (len(service_api_client.get_active_services({'user_id': session['user_id']}).get('data', [])) > 1):
+        service_id, error = _create_service(service_name, email_from, form)
+        if error:
+            return render_template('views/add-service.html', form=form, heading=heading)
```
leohemsted: is there any way to put the creation step into the `form.validate_on_submit` function? that way, we wouldn't need this extra if check, since if there's an error that the Form object can detect the `validate_on_submit` function returns false and we render the same thing with the same args anyway

app/templates/views/signedout.html
```diff
@@ -107,7 +107,7 @@ <h2 class="with-keyline">
     </h2>
     <div class="responsive-embed responsive-embed--16by9 responsive-embed--bordered bottom-gutter-2">
       <div class="responsive-embed__wrapper">
-        <iframe width="560" height="315" src="https://www.youtube.com/embed/i2a2jiwYTd4" frameborder="0" allowfullscreen></iframe>
+        <iframe width="560" height="315" src="https://www.youtube.com/embed/_90cv1YgQo4" frameborder="0" allowfullscreen></iframe>
```
leohemsted: this commit appears to already be in master 🤔

app/main/views/add_service.py
servingUpAces: I can't - update service name is still using it.

app/templates/views/signedout.html
servingUpAces: rebase :(

PR #1423, opened 2017-08-14
Add domains of all organisations on GOV.UK

quis:
We keep having to manually add domains as users from obscure bits of government try to sign up for Notify.

The way we check the validity of these domains is often by seeing if they’re linked to from that organisation’s page on GOV.UK. This commit adds all those organisations that have a separate domain linked to from gov.uk. Which should mean we get fewer support tickets about this stuff.

It feels a bit questionable as to whether some of them should have access. We should review.

There is no clean API or register of these domains, so the full detail of how I scraped them from GOV.UK is here: https://gist.github.com/quis/9c2625225b7e381da2b0d523ae54b3b7


quis:
Going to close this because a lot of these domains belong to organisations (charities, museums, etc)  that we don’t want to give access to yet.

PR #1426, opened 2017-08-14
use iget_array to reduce memory footprint. and removed useless import…

chfw:
I felt iget_array would be better suited for the use case. The benefit is that when dealing with large excel files, the memory footprint of pyexcel is reduced. Especially with csv, tsv, its memory consumption is constant. And if you would only read ods and if your ods file could be large, I would recommend that you choose pyexcel-odsr, which consumes less memory because it does not load everything into memory in one go.

Here is the documentation on iget_array:

http://pyexcel.readthedocs.io/en/latest/generated/pyexcel.iget_array.html#pyexcel.iget_array

Since version pyexcel 0.2.2, no longer the imports of pyexcel.ext.xls, pyexcel.ext.ods etc. are required. And here is the [plugin loading methodology](http://lml.readthedocs.io/en/latest/design.html).

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -22,10 +22,6 @@ from flask import (
 )
 from flask_login import current_user
 import pyexcel
-import pyexcel_io
-import pyexcel_xls
-import pyexcel_xlsx
-import pyexcel_ods3
 
 from notifications_utils.template import (
     SMSPreviewTemplate,
@@ -246,12 +242,16 @@ class Spreadsheet():
             return cls(Spreadsheet.normalise_newlines(file_content), filename)
 
         if extension == 'tsv':
-            file_content = StringIO(Spreadsheet.normalise_newlines(file_content))
+            file_content = StringIO(
+                Spreadsheet.normalise_newlines(file_content))
 
-        return cls.from_rows(pyexcel.get_sheet(
-            file_type=extension,
-            file_content=file_content.read()
-        ).to_array(), filename)
+        instance = cls.from_rows(
+            pyexcel.iget_array(
+                file_type=extension,
+                file_stream=file_content),
+            filename)
+        pyexcel.free_resources()
+        return instance
 
 
 def get_help_argument():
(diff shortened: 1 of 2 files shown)


quis:
Hey @chfw thanks very much for contributing. These all seem like sensible changes and we’d be happy to accept them.

However, it seems to have broken our ability to handle `.ods` files. When running the tests I get the following exception:
```
TypeError: invalid file: <FileStorage: 'tests/spreadsheet_files/open document spreadsheet.ods' ('application/vnd.oasis.opendocument.spreadsheet')>
```

I’ve copied the [full stack trace here](https://gist.github.com/quis/18d6d7bb51ed8e7a2b4c25512d9f3f0b).

You should be able to get the tests running on your local machine by following the instructions in [the README](https://github.com/alphagov/notifications-admin#readme). But if you need any help, or would like one of the team here to have a look at why it’s not working, do let us know.

chfw:
Got too many errors: 

> tests/app/main/views/test_send.py ..FFFFFFFFFFFFFFFFFFFFFFF...............................FF....FFF..FFF...F.....F........................FFFFFFFFF

> =================================== FAILURES ===================================
>  test_upload_files_in_different_formats[tests/spreadsheet_files/excel_97.xls-True] 
> 
> self = <flask_wtf.csrf.CSRFProtect object at 0x7ff0fbbf6160>
> 
>     def protect(self):
>         if request.method not in current_app.config['WTF_CSRF_METHODS']:
>             return
>     
>         try:
> >           validate_csrf(self._get_csrf_token())
> 
> ../nadminenv/lib/python3.5/site-packages/flask_wtf/csrf.py:249: 
> _ _ _ _ _ _ _ _ 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

quis:
1. Make sure you’re working inside a Virtualenv
2. Run `pip install -r requirements_for_test.txt`
2. Run `source environment_test.sh` (I think this is the undocumented step that you’re missing)
3. In the same shell, run `py.test tests/app/main/views/test_send.py`

chfw:
All passed.
```
collected 113 items                                                                                          

tests/app/main/views/test_send.py .................................................................................................................

```

chfw:
@quis , yep. I also learnt that the sister library, pyexcel-ods(depends on odfpy) has no problem in handling `FileStorage`.

PR #1429, opened 2017-08-15
Inbound number admin

servingUpAces:
This PR will allow the platform admin user to set an inbound number for a service.
There is a new table to store the available inbound sms numbers. When a platform admin user now turns the `inbound_sms`permission on, the service will be allocated the next available sms number. 

There is a new view to show all inbound SMS numbers and if a service is allocated to that number,  /inbound-sms-admin. 

Left outstanding:
 - Make the inbound_sms_admin.html beautiful


app/templates/views/service-settings/confirm-inbound-number.html
```diff
@@ -0,0 +1,33 @@
+{% extends "withnav_template.html" %}
```
kentsanggds: I don't think this template is being used so could be removed

app/main/views/service_settings.py
```diff
@@ -329,24 +335,41 @@ def service_set_reply_to_email(service_id):
...
         form=form)
 
 
+@main.route("/services/<service_id>/service-settings/set-inbound-number", methods=['GET'])
+@login_required
+@user_has_permissions('manage_settings', admin_override=True)
+def service_set_inbound_number(service_id):
+    switch_service_permissions(current_service['id'], 'inbound_sms')
+    set_inbound_sms = request.args.get('set_inbound_sms', False)
+    try:
```
kentsanggds: Just a thought - perhaps the logic for allocating the inbound_number should be in the API whenever `inbound_sms` permissions is updated?

app/templates/views/service-settings/confirm-inbound-number.html
```diff
@@ -0,0 +1,33 @@
...
+{% from "components/table.html" import text_field, edit_field %}
+{% from "components/page-footer.html" import page_footer %}
+
+{% block service_page_title %}
+  Inbound SMS Assigned Number
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  <h1 class="heading-large">Inbound SMS Assigned Nubmer</h1>
```
klssmith: Just a minor typo in this heading

PR #1491, opened 2017-09-15
Platform admin Timeout problem

venusbb:
When viewing platform admin services, we do an sql queries and whether it's trial mode or live services is filtered at the admin views.

This PR pass the filtered live/trial mode service to the query and reduce query time.

This is in relation to PR https://github.com/alphagov/notifications-api/pull/1257


quis:
This is going to break the platform admin homepage, which counts up notifications sent by both live and trial mode services.

PR #1501, opened 2017-09-20
show letters as accepted, not created or sending, on the api page

leohemsted:
- [ ] https://github.com/alphagov/notifications-api/pull/1241

Letters are shown as `accepted`, emails and sms are still shown as `created` or `sending`

### DIFF
--- a/app/main/views/api_keys.py
+++ b/app/main/views/api_keys.py
@@ -13,13 +13,18 @@ from app.notify_client.api_key_api_client import KEY_TYPE_NORMAL, KEY_TYPE_TEST,
 def api_integration(service_id):
     return render_template(
         'views/api/index.html',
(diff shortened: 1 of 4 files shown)


app/main/views/api_keys.py
```diff
@@ -13,13 +13,18 @@
...
-        api_notifications=notification_api_client.get_notifications_for_service(
+        api_notifications=map_letters_to_accepted(notification_api_client.get_notifications_for_service(
             service_id=service_id,
             include_jobs=False,
             include_from_test_key=True
-        )
+        ))
     )
 
+def map_letters_to_accepted(notifications):
```
quis: Feel like this is a great example of where the logic could live in the client, and the view could be kept cleaner.

leohemsted: great point. also adding a test now.

PR #1509, opened 2017-09-25
Allow user to add multiple reply-to addresses

gov-cjwaszczuk:
Added new routes for email reply-to / edit / add
User can now create multiple reply to addresses

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -27,7 +27,7 @@ from app.main.forms import (
     ConfirmPasswordForm,
     ServiceNameForm,
     RequestToGoLiveForm,
-    ServiceReplyToEmailFrom,
+    ServiceReplyToEmailForm,
     ServiceSmsSender,
     ServiceLetterContactBlock,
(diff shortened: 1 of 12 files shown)


app/main/views/service_settings.py
```diff
@@ -308,23 +314,65 @@ def service_set_email(service_id):
...
+@login_required
+@user_has_permissions('manage_settings', admin_override=True)
+def service_edit_email_reply_to(service_id, reply_to_email_id):
+    form = ServiceReplyToEmailForm()
+    reply_to_email_address = service_api_client.get_reply_to_email_address(service_id, reply_to_email_id)
+    if request.method == 'GET':
+        form.email_address.data = reply_to_email_address['email_address']
+        form.is_default.data = reply_to_email_address['is_default']
+    if form.validate_on_submit():
+        print(form.is_default.data)
```
servingUpAces: Do we want this print?

gov-cjwaszczuk: what print? ;)

PR #1510, opened 2017-09-26
Allow services to turn international text messaging on and off

quis:
We didn’t make this self-service before because the pricing information wasn’t published (ie we had to send it to services that asked for it). 

Now that we publish pricing information in the app, there’s no reason why services can’t make an informed decision about whether they want international SMS or not.

So this commit:
- removes the platform admin button
- adds some radio buttons that our users can click with their mice

***

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -658,6 +658,16 @@ class ServiceInboundApiForm(Form):
                                                            Length(min=10, message='Must be at least 10 characters')])
 
 
+class InternationalSMSForm(Form):
+    enabled = RadioField(
+        'Send text messages to international phone numbers',
+        choices=[
+            ('on', 'On'),
+            ('off', 'Off'),
+        ],
+    )
+
+
 def get_placeholder_form_instance(
     placeholder_name,
     dict_to_populate_from,
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -224,26 +224,36 @@ def service_switch_research_mode(service_id):
     return redirect(url_for('.service_settings', service_id=service_id))
 
 
-def switch_service_permissions(service_id, permission, sms_sender=None, force=None):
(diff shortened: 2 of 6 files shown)


app/main/views/service_settings.py
```diff
@@ -222,13 +224,22 @@ def service_switch_research_mode(service_id):
...
+        else:
+            permissions.append(permission)
     else:
-        permissions.append(permission)
-    current_service['permissions'] = permissions
+        if force is True:
+            permissions.append(permission)
+        elif force is False and permission in permissions:
+            permissions.remove(permission)
+
```
imdadahad: Not sure about this, would it cleaner to have another method to force set a permission?

app/main/forms.py
```diff
@@ -658,6 +658,16 @@ class ServiceInboundApiForm(Form):
                                                            Length(min=10, message='Must be at least 10 characters')])
 
 
+class InternationalSMSForm(Form):
+    enabled = RadioField(
+        'Send text messages to international phone numbers',
+        choices=[
+            ('on', 'On'),
+            ('off', 'Off'),
+        ],
```
imdadahad: Being picky here any reason why we don't have:

```
    (True, 'On'),
    (False, 'Off'),
```

quis: `True` and `False` are Python, you can’t put them as the value of a HTML attribute (ie this gets turned into `<input value='on'>`)

app/main/views/service_settings.py
quis: I dislike this too. Will have a think about how it could be better.

app/main/forms.py
imdadahad: Ah I see, thanks for clarifying.

app/main/views/service_settings.py
quis: @imdadahad How about a3bd596 ?

PR #1512, opened 2017-09-26
Tweak email reply to pages, add IDs

quis:
This pull request makes a few small design tweaks to the email reply to pages, and adds one major new thing, showing the IDs of the reply to addresses on these pages.

***

The first users of multiple email reply to addresses will be using the API. This means that the need to be able to specify the ID of the reply to address they want.

We chose to implement it like this instead of by passing the address in directly because that means deploying code. For some teams deploying code can take weeks, and we’d like to let teams have the flexibility to make changes faster than this.

Same as for templates, you shouldn’t have to go to the _edit_ page in order to get the ID. This means listing them on the page where you see all the reply to addresses.

Listing the IDs like this means that it’s not really a table any more, because the information isn’t organised in columns. So I think it makes sense to reuse the pattern from the manage team page, which has a similar relationship between the information.

***

screen shot 2017-09-26 at 15 59 42

screen shot 2017-09-26 at 15 56 18

### DIFF
--- a/app/assets/stylesheets/components/tick-cross.scss
+++ b/app/assets/stylesheets/components/tick-cross.scss
@@ -42,12 +42,12 @@
   &-list {
 
     @extend %grid-row;
-    margin-top: 5px;
     position: relative;
 
     &-permissions {
 
       @include grid-column(3/4);
+      margin-top: 5px;
 
       li {
         display: block;
@@ -59,7 +59,7 @@
     &-edit-link {
       text-align: right;
       position: absolute;
-      top: -1.6em;
+      top: -25px;
       right: -135px;
     }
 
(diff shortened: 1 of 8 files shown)


quis:
# Hide ID when only one reply to address is shown

Most user will only have one reply to address. Which means they should never have to worry about IDs. And if you only have one then you never need its ID, because the last remaining address will always be the default.

So IDs should only be shown when a service has created more than one reply to address.

This required a bit of visual tweaking of the _user list_ pattern, because its spacing wasn’t defined in a way that worked when only the name of the thing, and not its details were shown on the page.

## Before 

screen shot 2017-09-28 at 11 38 20

## After

screen shot 2017-09-28 at 11 38 38

## Team list still looks unchanged

image

PR #1513, opened 2017-09-27
Add letters to the request to go live page

quis:
It’s not either text messages, or emails, or both now – it’s any combination of the three channels.

This commit adds ‘letters’ as an option on the request to go live page by changing the radio buttons to a group of checkboxes, so the user can choose as many or as few as they want.

This commit also does a bunch of housekeeping stuff around the tests for this page, because they haven’t been touched in quite some time.

## Before

image

## After

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -431,15 +431,9 @@ class RequestToGoLiveForm(Form):
         ],
         validators=[DataRequired()]
     )
-    channel = RadioField(
-        'What kind of messages will you be sending?',
-        choices=[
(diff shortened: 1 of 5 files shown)


imdadahad:
Do we have a test that checks at least one is selected?

quis:
Since there is no native WTForms control for a group of checkboxes, we’d have to write some custom validation to make sure that at least one is checked. Don’t think it’s worth it just for a form which sends an email – we can always follow up if they don’t tick any.

PR #1516, opened 2017-09-27
make sure inbound sms page doesn't break if you receive a blank inbou…

leohemsted:
…nd message

### DIFF
--- a/app/main/views/conversation.py
+++ b/app/main/views/conversation.py
@@ -73,7 +73,10 @@ def get_sms_thread(service_id, user_number):
             'inbound': is_inbound,
             'content': SMSPreviewTemplate(
                 {
-                    'content': notification.get('content') or notification['template']['content']
+                    'content': (
+                        notification['content'] if is_inbound else
+                        notification['template']['content']
+                    )
                 },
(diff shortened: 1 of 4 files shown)


app/main/views/conversation.py
```diff
@@ -59,8 +59,14 @@ def get_user_number(service_id, notification_id):
...
+            dict(type='outbound', **x) for x in
+            notification_api_client.get_notifications_for_service(service_id, to=user_number)['notifications']
+        ] +
+        [
+            dict(type='inbound', **x) for x in
+            service_api_client.get_inbound_sms(service_id, user_number=user_number)
+        ]
     ), key=lambda notification: notification['created_at']):
 
         is_inbound = ('notify_number' in notification)
```
quis: This should go now if `notification['type']` means the same thing. Or should be `is_inbound = notification['type'] == 'inbound'` if there’s a good reason to keep it.

quis:
Also, what does the empty message look like on the screen?

leohemsted:
it looks: pretty bad

image


image

quis:
it looks OK

PR #1520, opened 2017-10-02
Added letters info to service / admin dashboard & activity page

gov-cjwaszczuk:
#  Summary 

This pull request adds a new third column onto the service / platform admin dashboard pages. It also adds a letter activity page to go alongside the email and sms activity pages.

The activity page does not show the sending statuses for letters, this is because they aren't applicable to the letter process, and therefore would not make a whole lot of sense. It also does not contain the search bar.

The sizing of numbers within platform admin has been decreased, this is due to the large numbers overflowing the columns, and was a more appropriate solution than implementing flex into the columns.

*** 
## Before:

image

image

## After:

image

image

image

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -42,7 +42,8 @@
   position: relative;
   margin-bottom: $gutter-two-thirds;
 
-  .big-number {
+  .big-number,
+  .big-number-smaller {
     padding: $gutter-half;
     position: relative;
     background: $black;
--- a/tests/app/main/views/test_jobs.py
+++ b/tests/app/main/views/test_jobs.py
@@ -8,7 +8,7 @@ from bs4 import BeautifulSoup
 
 from app.main.views.jobs import get_time_left, get_status_filters
 from tests import notification_json
-from tests.conftest import SERVICE_ONE_ID, normalize_spaces
+from tests.conftest import SERVICE_ONE_ID, normalize_spaces, mock_get_notifications
 from freezegun import freeze_time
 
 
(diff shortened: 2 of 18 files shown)


quis:
Can you do some screenshots of the relevant pages?

app/assets/stylesheets/components/big-number.scss
```diff
@@ -42,7 +42,7 @@
   position: relative;
   margin-bottom: $gutter-two-thirds;
 
-  .big-number {
+  .big-number, .big-number-smaller {
```
quis: Prefer one line per selector, eg
```CSS
.big-number, 
.big-number-smaller {
```

tests/app/main/views/test_jobs.py
```diff
@@ -172,14 +172,13 @@ def test_should_show_letter_job(
...
     )
-
     assert normalize_spaces(page.h1.text) == 'thisisatest.csv'
     assert normalize_spaces(page.select('p.bottom-gutter')[0].text) == (
         'Sent by Test User on 1 January at 11:09am'
     )
     assert page.select('.banner-default-with-tick') == []
     assert normalize_spaces(page.select('tbody tr')[0].text) == (
-        '07123456789 template content'
+        '07123456789 template content Delivered 1 January at 11:10am'
```
quis: Think it’s the fixtures that are set up wrong here (phone number for letter job) but letter jobs shouldn’t show a status in the table.

tests/conftest.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 from notifications_python_client.errors import HTTPError
-from flask import url_for
+from flask import url_for, Markup
```
quis: Unused import.

PR #1522, opened 2017-10-03
Updated runtime python version

gov-cjwaszczuk:
CloudFoundry pulled python 3.5.2.


sakisv:
Putting this here as a reference:
https://github.com/cloudfoundry/python-buildpack/commit/40995d807b5cab59a840ed2278a904af13aab3c7

PR #1526, opened 2017-10-03
Rename count from ‘departments’ to ’organisations’

quis:
They’re not all departments.

Could also say ‘departments and agencies’ as per the Performance Platform page. Which is more accurate?


quis:
image

Needs the link text changing as well. And still not sure if ‘organisations’ is right.

quis:
Updated the link. Think ‘organisations’ is OK after all:

image

PR #1543, opened 2017-10-16
Allow one-off email sending to select the reply-to address

gov-cjwaszczuk:


app/main/views/send.py
```diff
@@ -166,6 +167,76 @@ def get_example_csv(service_id, template_id):
...
+    }[template_type]
+
+
+def get_sender_details(service_id, template_type):
+    api_call = {
+        'email': 'get_reply_to_email_addresses',
+        'letter': 'get_letter_contacts',
+        'sms': 'get_sms_senders'
+    }[template_type]
+    return getattr(service_api_client, api_call)(service_id)
```
leohemsted:
```python
api_call = {
    'email': service_api_client.get_reply_to_email_addresses,
    'letter': service_api_client.get_letter_contacts,
}[template_type]
return api_call(service_id)
```

app/main/views/send.py
```diff
@@ -166,6 +167,69 @@ def get_example_csv(service_id, template_id):
...
+            url_for('.send_one_off', service_id=service_id, template_id=template_id)
+        )
+
+    sender_context = get_sender_context(sender_details, template['template_type'])
+
+    form = SetSenderForm(sender=sender_context['default_id'])
+    form.sender.choices = sender_context['value_and_label']
+    form.sender.label.text = sender_context['description']
+    option_hints = {}
+    option_hints[sender_context['default_id']] = 'Default'
```
leohemsted: `option_hints = {sender_context['default_id']: 'Default'}`

app/main/views/send.py
```diff
@@ -166,6 +167,69 @@ def get_example_csv(service_id, template_id):
...
+            'default_id': default_sender,
+            'title': "Choose where to send replies",
+            'description': "Select an email address that recipients can reply to"
+        },
+        'letter': {
+            "value_and_label": [(sender['id'], sender[sender_format]) for sender in sender_details],
+            "default_id": default_sender,
+            "title": "Choose sender address",
+            "description": "Select an address that recipients can reply to"
+        }
```
leohemsted: ~is it worth moving these descriptions to the sender form class, and chosing between two classes based on options? so having `SmsSenderForm` and `EmailSenderForm` or something~

e: never mind, refactored a bit and its nicer now. i think. maybe.

app/main/views/templates.py
```diff
@@ -57,6 +57,7 @@
 )
 def view_template(service_id, template_id):
     template = service_api_client.get_service_template(service_id, str(template_id))['data']
+    one_off_endpoint = '.set_sender' if template["template_type"] == "email" else '.send_one_off'
```
leohemsted: pop this in set-sender rather than in the view imo

tests/app/main/views/test_send.py
```diff
@@ -35,6 +41,98 @@
...
+    mocker
+):
+    template_mock(mocker)
+    sender_data(mocker)
+
+    page = client_request.get(
+        '.set_sender',
+        service_id=service_one['id'],
+        template_id=fake_uuid
+    )
```
leohemsted: not doing any asserts here

PR #1544, opened 2017-10-16
Remove HTML5 validation

quis:
This commit:
- uses WTForms email fields wherever we collect email addresses
- makes sure we don’t let the default HTML5 form validation get trigged (using `novalidate`[1])

We don’t want to use the default validation because:
- it looks out of place
- it’s not well supported by assistive technology

1. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-novalidate


quis:
@leohemsted this one’s for you…

PR #1547, opened 2017-10-17
 Let users download a CSV of inbound messages

quis:
# Before 

screen shot 2017-10-17 at 12 24 39

# After 

screen shot 2017-10-17 at 12 23 56

# Download

screen shot 2017-10-17 at 11 26 11
screen shot 2017-10-17 at 11 26 07

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -6,7 +6,8 @@ from flask import (
     session,
     jsonify,
     request,
-    abort
+    abort,
+    Response,
 )
 from flask_login import login_required
 
@@ -20,6 +21,9 @@ from app import (
     service_api_client,
     template_statistics_client,
     inbound_number_client,
+    format_datetime_short,
+    format_date_numeric,
+    format_datetime_numeric,
 )
 from app.statistics_utils import get_formatted_percentage, add_rate_to_job
 from app.utils import (
@@ -27,6 +31,7 @@ from app.utils import (
     get_current_financial_year,
     FAILURE_STATUSES,
     REQUESTED_STATUSES,
+    Spreadsheet,
 )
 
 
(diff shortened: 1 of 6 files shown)


app/main/views/dashboard.py
```diff
@@ -161,6 +166,31 @@ def inbox_updates(service_id):
...
+def inbox_download(service_id):
+    return Response(
+        Spreadsheet.from_rows(
+            [[
+                'Phone number',
+                'Message',
+                'Received',
+            ]] + [[
+                message['user_number'],
+                message['content'].lstrip(('=+-@')),
```
leohemsted: any reason we're doing this strip here?

leohemsted: ah i see from tests, to prevent excel formula injection. https://www.contextis.com/blog/comma-separated-vulnerabilities. Is it worth following that blog's advice and pre-pending a single `'` apostrophe? Not sure how I feel about remove content from messages.

quis: I was following https://www.contextis.com/blog/comma-separated-vulnerabilities which suggests prepending a tab character.

I don’t feel too bad about removing content here because:
- can’t see a valid use case for starting a message with one of these characters
- inserting characters means that those characters are visible for someone using a non-Excel spreadsheet application, eg loading the CSV and looking for messages starting with the word ‘STOP’
- hacks like inserting tabs/apostrophes are undocumented, ie they could stop working at any time, leaving users vulnerable again

Alternatively, we could take the advice from the post I linked to:
> swear off the world, get yourself a cabin with the woods and maybe try being friends with squirrels for a while.

leohemsted: my only concern is that this is inbound content coming from end users - it could be any content, of any kind, eg

> Gov: "reply to this message with a phone number and we'll ring you when stuff happens"
> User: "+1 (555) 867-5309"

 agree with your second and third bullets point completely tho.

quis: Discussed in person, agreed that stripping leading formula-initialising characters is the least-worst option.

PR #1556, opened 2017-10-19
Add syslog drain service to preview and staging

sakisv:
This should and will be moved back to manifest-base.yml when
it is configured for production usage.

### DIFF
--- a/manifest-preview.yml
+++ b/manifest-preview.yml
@@ -2,6 +2,14 @@
 
 inherit: manifest-base.yml
 
+services:
+  - notify-aws
+  - notify-config
+  - notify-template-preview
+  - hosted-graphite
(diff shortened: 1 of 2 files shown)


manifest-preview.yml
```diff
@@ -2,6 +2,14 @@
 
 inherit: manifest-base.yml
 
+services:
+  - notify-aws
```
richardc0: You've added other services too?

sakisv: No, I'm just overriding the services from the [base file](https://github.com/alphagov/notifications-admin/blob/master/manifest-base.yml#L5)

Had I just added the new service, that would be the only one it would be bound to the app.

PR #1568, opened 2017-10-23
Improve the performance of /platform-admin

servingUpAces:
Using a new endpoint in API to get the aggregate platform admin stats.

- [x]  https://github.com/alphagov/notifications-api/pull/1332

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -123,36 +123,6 @@ def create_global_stats(services):
     return stats
 
 
-def platform_failure_percentages(services):
-    stats = {
-        'email': {
-            'delivered': 0,
-            'failed': 0,
-            'requested': 0
-        },
-        'sms': {
-            'delivered': 0,
-            'failed': 0,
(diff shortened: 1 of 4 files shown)


app/main/views/platform_admin.py
```diff
@@ -120,6 +123,36 @@ def create_global_stats(services):
...
+            'delivered': 0,
+            'failed': 0,
+            'requested': 0
+        }
+    }
+
+    for service in services:
+        for msg_type, status in itertools.product(('sms', 'email', 'letter'), ('delivered', 'failed', 'requested')):
+            import pdb
+            pdb.set_trace()
```
leohemsted: pdb

app/main/views/platform_admin.py
```diff
@@ -120,6 +123,36 @@ def create_global_stats(services):
     return stats
 
 
+def platform_failure_percentages(services):
```
leohemsted: this function isn't used anywhere?

PR #1580, opened 2017-10-24
Added Multiple SMS sender functionality

gov-cjwaszczuk:
Added multiple SMS sender for service settings. Before you could only amend a single SMS sender. In the new format, you can add multiple SMS sender; the default will be selected from the senders when a notification is to be sent. 

Also changes to the functionality of the inbound SMS number. When you choose to set an inbound sms number, you will be shown a list of the current available inbound numbers. If there aren't any, you will be given a message saying so. If the service already has an inbound number attached to it but somehow ends up on the set inbound number page again, they will be given a message telling them so. If the user only has a single sms sender, when they enable inbound SMS, the sender will be updated. If they had more than one sender, the new inbound number will be a separate sms sender. 

With regards to editing an inbound SMS number, the user only has the ability to switch it away from being default. The record can't be changed like other sms senders. 

## :o: Service Settings
image

## :o: Show all sms senders
image
## :o: Add sms sender
image
## :o: Edit sms sender
image
## :o: Edit inbound number
image

# Inbound Number screenshots
## :o: Number already set
image
## :o: No available numbers
image
## :o: Available numbers
image

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -28,6 +28,7 @@ from app.main.forms import (
     RenameServiceForm,
     RequestToGoLiveForm,
     ServiceReplyToEmailForm,
+    ServiceInboundNumberForm,
     ServiceSmsSender,
     ServiceLetterContactBlockForm,
     ServiceBrandingOrg,
@@ -82,6 +83,11 @@ def service_settings(service_id):
     default_letter_contact_block = next(
         (Field(x['contact_block'], html='escape') for x in letter_contact_details if x['is_default']), "Not set"
     )
+    sms_senders = service_api_client.get_sms_senders(service_id)
+    sms_sender_count = len(sms_senders)
+    default_sms_sender = next(
+        (Field(x['sms_sender'], html='escape') for x in sms_senders if x['is_default']), "None"
+    )
     return render_template(
         'views/service-settings.html',
         organisation=organisation,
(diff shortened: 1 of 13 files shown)


app/main/views/service_settings.py
```diff
@@ -430,22 +438,35 @@ def service_set_sms_sender(service_id):
...
-    try:
-        if set_inbound_sms == 'True':
-            inbound_number_client.activate_inbound_sms_service(service_id)
-            return redirect(url_for('.service_settings', service_id=service_id))
-        else:
-            inbound_number_client.deactivate_inbound_sms_permission(service_id=service_id)
-            return redirect(url_for('.service_set_sms_sender', service_id=service_id))
-    except HTTPError as e:
+    available_inbound_numbers = service_api_client.get_available_inbound_sms_numbers()
+    print(available_inbound_numbers)
```
servingUpAces: The print stmt can go

app/notify_client/service_api_client.py
```diff
@@ -249,6 +249,9 @@ def get_whitelist(self, service_id):
     def update_whitelist(self, service_id, data):
         return self.put(url='/service/{}/whitelist'.format(service_id), data=data)
 
+    def get_available_inbound_sms_numbers(self):
```
servingUpAces: Should this one be in the inbound_number_client?

PR #1589, opened 2017-10-30
Standardise analytics code on failwhale

quis:
Missed this page when doing 62fcc2429fe01f81937d06ddfd513766542f4e78


leohemsted:
there's also https://github.com/alphagov/notifications-aws/blob/master/terraform/modules/admin-failwhale/files/custom_503.html

not sure if that's used?

PR #1594, opened 2017-10-31
Use annual_billing tables at backend for getting and updating free_sm…

venusbb:
This PR is related to these 2 pages: 
image

and in service settings:
image

**What:** 
Created billing_api_client to get and update 
* new service will create new entry in the table using default organisation type. 
* usage page will use data from the data to display and calculate monthly usage
* free_sms_fragment_limit setting page will use and update data to the table. 

**Some hidden logic already implemented in API**
* when updating the free-sms-fragment-limit in service-setting page, the current year and any  free_sms_fragment_limit years entry in annual_billing table will be updated. 
* when displaying the free-sms-fragement-limit in the usage page, if it is a past year and the row doesn't exist, the oldest allowance limit will be displayed. If it is a current or future year, the newest entry of that service will be used to create an entry for that requested year in API and be displayed. 
* We assume at least one year entry exist in the annual-billing table. Otherwise, 503 error code will be returned from the API

Related API pull requests:
alphagov/notifications-api#1342, alphagov/notifications-api#1355

Pivotal Story
#152067121


quis:
I think you meant https://github.com/alphagov/notifications-api/pull/1342 not #1342 and https://github.com/alphagov/notifications-api/pull/1355 not #1355.

app/main/views/dashboard.py
```diff
@@ -287,9 +290,9 @@ def get_dashboard_totals(statistics):
     return statistics
 
 
-def calculate_usage(usage):
+def calculate_usage(usage, free_sms_fragment_limit):
     # TODO: Don't hardcode these - get em from the API
```
quis: This comment can be deleted now 🎉

app/main/views/dashboard.py
```diff
@@ -287,9 +290,9 @@ def get_dashboard_totals(statistics):
...
-    sms_free_allowance = 250000
+    sms_free_allowance = free_sms_fragment_limit         # VB-Progress
```
quis: What does this comment mean?

app/main/views/dashboard.py
```diff
@@ -115,15 +115,18 @@ def usage(service_id):
...
         )),
         selected_year=year,
         years=get_tuples_of_financial_years(
             partial(url_for, '.usage', service_id=service_id),
             start=current_financial_year - 1,
             end=current_financial_year + 1,
         ),
-        **calculate_usage(billing_api_client.get_service_usage(service_id, year))
+        **calculate_usage(billing_api_client.get_service_usage(service_id, year),
+                          billing_api_client.get_free_sms_fragment_limit_for_year(service_id, year))
```
quis: You’re making the same API call twice in one endpoint here. It would be better to store the result in a variable.

app/main/views/service_settings.py
```diff
@@ -704,13 +709,23 @@ def set_organisation_type(service_id):
...
     if form.validate_on_submit():
         service_api_client.update_service(
             service_id,
+            # TODO: Retire this after new end points are added.
             free_sms_fragment_limit=form.free_sms_allowance.data,
         )
+        # get a list of all the free sms allowance entries for this service
+        sms_list = billing_api_client.get_free_sms_fragment_limit_for_all_years(service_id)
+
+        for item in range(0, len(sms_list)):
```
quis: You can iterate over `sms_list` directly, eg `for year in sms_list:`

app/main/views/service_settings.py
```diff
@@ -704,13 +709,23 @@ def set_organisation_type(service_id):
...
+            if sms_list[item]['financial_year_start'] >= get_current_financial_year():
+                form.set_free_sms_allowance = \
+                    billing_api_client.create_or_update_free_sms_fragment_limit_for_year(service_id,
+                                                                                         form.free_sms_allowance.data,
+                                                                                         sms_list[item][
+                                                                                             'financial_year_start'])
```
quis: I don’t understand why we’re doing this. It means that different years will get updated depending on whether the user has looked at the page for next year.

Really, this update should only ever be called with the current year (so no need for the `for` loop and `if` statement) and the API should handle updating the current and any future years that exist.

app/notify_client/billing_api_client.py
```diff
@@ -23,3 +25,47 @@ def get_service_usage(self, service_id, year=None):
...
+                result = self.get(
+                    '/service/{0}/billing/free-sms-fragment-limit'.format(service_id),
+                    params=dict(financial_year_start=year)
+                )
+            return result['free_sms_fragment_limit']
+        except HTTPError:
+            current_app.logger.info(
+                'Requested free_sms_fragment_limit entry for service {0} and year {1} does not exist'
+                .format(service_id, year))
+            return -1
```
quis: Why would we ever want to tell someone that their allowance is -1?

app/notify_client/billing_api_client.py
```diff
@@ -23,3 +25,47 @@ def get_service_usage(self, service_id, year=None):
...
+    def get_free_sms_fragment_limit_for_all_years(self, service_id, year=None):
+        try:
+            return self.get(
+                '/service/{0}/billing/free-sms-fragment-limit'.format(service_id),
+            )
+        except HTTPError:
+            current_app.logger.info(
+                'No free_sms_fragment_limit entry exists for service {0} '
+                .format(service_id, year))
+            return []
```
quis: Again, this should raise a `500` right? Assuming every service should have at least one row in this table.

app/notify_client/billing_api_client.py
```diff
@@ -23,3 +25,47 @@ def get_service_usage(self, service_id, year=None):
...
+
+    def create_or_update_free_sms_fragment_limit_for_year(self, service_id, free_sms_fragment_limit, year=None):
+        if year is None:
```
quis: If it’s allowable for `year` to be `None` then this method shouldn’t be called `_for_year`.

app/notify_client/billing_api_client.py
venusbb: I have changed this to return whatever api is returning, which should be a 500.

app/notify_client/billing_api_client.py
venusbb: OK - I will change and pass API 503 errors instead.

app/main/views/dashboard.py
venusbb: Changed

app/main/views/dashboard.py
venusbb: Done

app/main/views/dashboard.py
venusbb: Removed

app/main/views/service_settings.py
venusbb: No longer relevant as we don't use the end-point to return a list anymore.

app/main/views/service_settings.py
venusbb: The logic has been pushed back to API as discussed.

app/notify_client/billing_api_client.py
venusbb: Refractored

app/main/views/add_service.py
```diff
@@ -40,17 +41,21 @@ def _add_invited_user_to_service(invited_user):
...
             message_limit=current_app.config['DEFAULT_SERVICE_LIMIT'],
-            free_sms_fragment_limit=current_app.config['DEFAULT_FREE_SMS_FRAGMENT_LIMITS'].get(organisation_type),
+            free_sms_fragment_limit=free_sms_fragment_limit,
             restricted=True,
             user_id=session['user_id'],
             email_from=email_from,
         )
         session['service_id'] = service_id
+
+        billing_api_client.create_or_update_free_sms_fragment_limit(service_id, free_sms_fragment_limit)
```
quis: Is there an edge case here where the `create_service` API call succeeds but the `create_or_update_free_sms_fragment_limit` returns `4xx` or `5xx` for whatever reason and we end up with mismatched data?

tests/app/main/views/test_service_settings.py
```diff
@@ -29,6 +29,16 @@
...
+from freezegun import freeze_time
+
+
+@pytest.fixture
+def mock_get_service_settings_page_common(
+    mock_get_letter_organisations,
+    mock_get_inbound_number_for_service,
+    mock_get_free_sms_fragment_limit,
+):
+    return
```
quis: Nice.

PR #1598, opened 2017-11-01
Email auth for inviting members and editing permissions

gov-cjwaszczuk:
Added email auth functionality for user. If the service has email auth on, they can begin to invite users with email authentication as their primary login authentication. They can also edit user's permissions so that they log in using email authentication. Users who have no mobile number cannot be changed to email auth.

## :o: Managing users
image
## :o: Inviting users
image
## :o:  Editing user permissions
image

### DIFF
--- a/tests/app/main/views/test_manage_users.py
+++ b/tests/app/main/views/test_manage_users.py
@@ -10,6 +10,7 @@ from tests.conftest import (
     SERVICE_ONE_ID,
     active_user_with_permissions,
     active_user_view_permissions,
+    active_user_no_mobile,
     active_user_manage_template_permission,
 )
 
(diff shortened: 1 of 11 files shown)


gov-cjwaszczuk:
edited` test_edit_user_permissions_including_authentication_with_email_auth_service` if you'd like to check it out, also added test_invite_user_with_email_auth_service (refactoring invite_user test made it overly confusing)

tests/app/main/views/test_manage_users.py
```diff
@@ -220,7 +363,60 @@ def test_invite_user(
...
+        follow_redirects=True
+    )
+
+    assert response.status_code == 200
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+    assert page.h1.string.strip() == 'Team members'
+    flash_banner = page.find('div', class_='banner-default-with-tick').string.strip()
+    assert flash_banner == 'Invite sent to test@example.gov.uk'
+
+    expected_permissions = 'manage_api_keys,manage_settings,manage_templates,manage_users,send_emails,send_letters,send_texts,view_activity'  # noqa
```
quis: Doesn’t matter for this PR, but a useful thing to know is that you can split a string across multiple lines with brackets, eg:

```python
p = (
    'one'
    'two'
    '…'
)
```

PR #1599, opened 2017-11-01
New account verification emails no longer call API

leohemsted:
This frees up the API call to be used for 2fa email auth

Also, catch malformed tokens, and show a little error page

 image


leohemsted:
@quis feel free to change wording - this is a generic page for all email tokens

* invites
* password reset
* change email verification
* new account email verification

PR #1603, opened 2017-11-02
Select text message sender for one off notifications

servingUpAces:
This PR adds the ability to select a text message sender if more than one exist for the service.

If the service has more than one text message sender set then when `send to one recipient` will allow the user to select which text message sender to use for that notification.

screen shot 2017-11-02 at 12 10 26


quis:
Can we tell which one (if any) is the inbound number?

servingUpAces:
@quis I have now added that

PR #1605, opened 2017-11-02
make sure old invites get the proper err message

leohemsted:
we were accidentally covering up the expiry message with a more
generic one

### DIFF
--- a/app/main/views/invites.py
+++ b/app/main/views/invites.py
@@ -7,6 +7,7 @@ from flask import (
     abort,
     current_app
 )
+from itsdangerous import SignatureExpired
 from markupsafe import Markup
 from notifications_utils.url_safe_token import check_token
 from flask_login import current_user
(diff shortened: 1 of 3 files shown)


app/main/views/invites.py
```diff
@@ -21,12 +22,20 @@
...
+            token,
+            current_app.config['SECRET_KEY'],
+            current_app.config['DANGEROUS_SALT'],
+            current_app.config['INVITATION_EXPIRY_SECONDS']
+        )
+    except SignatureExpired:
+        errors = [
+            'Your invitation to GOV.UK Notify has expired. '
+            'Please ask the person that invited you to send you another one'
+        ]
```
leohemsted: stolen from https://github.com/alphagov/notifications-api/blob/master/app/accept_invite/rest.py#L37-L38

PR #1606, opened 2017-11-02
Fix a typo: generates -> generate

tuzz:


tuzz:
I don't have permissions to merge so someone else will have to.

quis:
@tuzz No worries, someone will merge it when they’re next doing a deploy.

PR #1610, opened 2017-11-03
Allow updates to SMS prefixing setting

quis:
We’re extracting this from being determined based on what the sender name is to its own setting.

This commit will let users set it independently.

Until the explicitly set it, it will still be determined based on whether their default sender name matches the default for the platform.

---

image

---

image

### DIFF
--- a/tests/app/main/views/test_service_settings.py
+++ b/tests/app/main/views/test_service_settings.py
@@ -44,7 +44,7 @@ from tests.conftest import (
         'Label Value Action',
         'Send text messages On Change',
         'Text message sender GOVUK Manage',
-        'Start all text messages with service name On Change',
+        'Text messages start with service name On Change',
         'International text messages Off Change',
         'Receive text messages Off Change',
 
@@ -64,7 +64,7 @@ from tests.conftest import (
         'Label Value Action',
         'Send text messages On Change',
         'Text message sender GOVUK Manage',
-        'Start all text messages with service name On Change',
+        'Text messages start with service name On Change',
         'International text messages Off Change',
         'Receive text messages Off Change',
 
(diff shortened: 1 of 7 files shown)


quis:
Wording changes based on feedback from Thom:

image

image

tests/app/main/views/test_service_settings.py
```diff
@@ -2063,3 +2067,46 @@ def test_empty_letter_contact_block_returns_error(
...
+
+def test_show_sms_prefixing_setting_page(
+    client_request,
+    mock_update_service,
+):
+    page = client_request.get(
+        'main.service_set_sms_prefix', service_id=SERVICE_ONE_ID
+    )
+    assert normalize_spaces(page.select_one('legend').text) == (
+        'Start all text messages with ‘service one:’'
```
venusbb: This has not been updated to the new text

tests/app/main/views/test_service_settings.py
```diff
@@ -2063,3 +2067,46 @@ def test_empty_letter_contact_block_returns_error(
...
+    )
+    radios = page.select('input[type=radio]')
+    assert len(radios) == 2
+    assert radios[0]['value'] == 'on'
+    assert radios[0]['checked'] == ''
```
venusbb: Default is off?

venusbb:
Do you also need to test posting to the end point? - changing the setting?

tests/app/main/views/test_service_settings.py
quis: Sorry – push failed – this is updated now

tests/app/main/views/test_service_settings.py
quis: The HTML looks like

1. `<input checked='' … >`
2. `<input … >`

So `radio['checked']` on the checked one returns an empty string, on the non-checked one it will raise a `KeyError`.

quis:
# POST to the correct endpoint when updating

`prefix_sms_with_service_name` is a computed attribute on the service model. It’s where we get the value from, and the API does some work to get it from the database, or derive it from the default SMS sender. It can’t be updated, because it’s not itself a database column.

`prefix_sms` is the name of the actual database column. This is the thing that we need to update.

This will go away eventually.

PR #1617, opened 2017-11-06
remove flask-script

leohemsted:
- [x] make sure logging is okay - it's different now? flask appears to be logging other stuff? https://stackoverflow.com/questions/14888799/disable-console-messages-in-flask-server
- [x] make sure gunicorn is okay - i had to change some stuff in wsgi.py
- [x] make sure paas runs and logs all proper and stuff.

It's non trivial!

flask-script has been deprecated by the internal flask.cli module, but making this carries a few changes with it

* you should add FLASK_APP=application.py and FLASK_DEBUG=1 to your environment.sh.
* instead of using `python app.py runserver`, now you must run `flask run -p 6012`. The -p command is important - the port must be set before the config is loaded, so that it can live reload nicely. (https://github.com/pallets/flask/issues/2113#issuecomment-268014481)
* find available commands by just running `flask`.
  * run them using flask. eg `flask list_routes`
  * define new tasks by giving them the decorator `@app.cli.command('task-name')`. Task name isn't needed if it's just the same as the function name. Alternatively, if app isn't available in the current scope, you can invoke the decorator directly, as seen in app/commands.py


app/__init__.py
```diff
@@ -519,3 +496,25 @@ def setup_event_handlers():
     from app.event_handlers import on_user_logged_in
 
     user_logged_in.connect(on_user_logged_in)
+
+
+def add_template_filters(application):
```
quis: `for template_filter in (…):`?

app/commands.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def list_routes():
+    """List URLs of all application routes."""
+    for rule in sorted(current_app.url_map.iter_rules(), key=lambda r: r.rule):
+        print("{:10} {}".format(", ".join(rule.methods - set(['OPTIONS', 'HEAD'])), rule.rule))
+
+
+def setup_commands(application):
+    application.cli.command('list-routes')(list_routes)
```
quis: Do we ever use this stuff? Could we just kill it?

leohemsted: i'd prefer to keep it just so that there's an example of adding a command to flask

PR #1626, opened 2017-11-08
Stop escaping special characters in inbound messages, version 2

quis:
This PR reimplements https://github.com/alphagov/notifications-admin/pull/1614, but with a bug around handling unicode characters fixed (see ed3f6fb for just the new changes).

---

Our previous, reverted, attempt at handling escaped strings (https://github.com/alphagov/notifications-admin/pull/1614) was too naïve and didn’t play well with characters outside the normal ASCII range. We have examples of real inbound messages containing `👍` and `’`, so we should definitely handle them.

It’s a bit tricky, because the strings we get from our providers are a mixture of escape sequences (eg `\n`) and unicode bytes (eg `😨`). So we have to first convert the unicode bytes `😨` into an escape sequence, for example `\U0001f628`. We do this using the `raw_unicode_escape` encoding:

> Latin-1 encoding with \uXXXX and \UXXXXXXXX for other code points. Existing backslashes are not escaped in any way.

– https://docs.python.org/3/library/codecs.html#text-encodings

Then, as before, we turn this back into a string using the `unicode_escape` codec, which transforms all escape sequences into their literal representations (eg `\U0001f628` becomes `😨` and `\n` becomes a newline).

### DIFF
--- a/app/main/views/conversation.py
+++ b/app/main/views/conversation.py
@@ -10,7 +10,7 @@ from notifications_utils.recipients import format_phone_number_human_readable
 from notifications_utils.template import SMSPreviewTemplate
 from app.main import main
 from app.main.forms import SearchTemplatesForm
-from app.utils import user_has_permissions
+from app.utils import user_has_permissions, unescape_string
 from app import notification_api_client, service_api_client
 from notifications_python_client.errors import HTTPError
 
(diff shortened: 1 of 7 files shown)


quis:
Closing in favour of https://github.com/alphagov/notifications-api/pull/1378

PR #1627, opened 2017-11-09
replace user PUT with POSTs

leohemsted:
the update_user fn was used in two places, for things that are handled fine by update_user_attribute. Reduce complexity in the API by killing the PUT, which is more dangerous (might silently overwrite things that shouldn't be, like "last_logged_in_at" etc).

Had to change the code not received mobile number form, and the activate user function.

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -54,12 +54,6 @@ class UserApiClient(NotifyAdminAPIClient):
             users.append(User(user, max_failed_login_count=self.max_failed_login_count))
         return users
 
-    def update_user(self, user):
-        data = user.serialize()
-        url = "/user/{}".format(user.id)
-        user_data = self.put(url, data=data)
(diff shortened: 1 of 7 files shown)


app/notify_client/user_api_client.py
```diff
@@ -155,7 +149,9 @@ def is_email_unique(self, email_address):
     def activate_user(self, user):
         if user.state == 'pending':
             user.state = 'active'
-            return self.update_user(user)
+            url = "/user/{}".format(user.id)
+            user_data = self.post(url, data={'state': 'active'})
+            return User(user_data['data'], max_failed_login_count=self.max_failed_login_count)
```
leohemsted: couldn't use `update_user_attribute` here because state is a disallowed attribute to update

leohemsted:
@servingUpAces yup, will remove that in my mobile number nullable pr 🎉

PR #1634, opened 2017-11-09
Make dashboard totals smaller if numbers are big

quis:
Numbers over a billion overflow the two column layout. Numbers over one hundred thousand overflow the three column layout.

This commit makes the type size smaller in these cases, so that the numbers still fit in the boxes.

image

image

screen shot 2017-11-09 at 17 31 05

screen shot 2017-11-09 at 17 30 48

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -10,7 +10,6 @@ from flask import (
     Response,
 )
 from flask_login import login_required
-from math import pow
 from notifications_utils.recipients import format_phone_number_human_readable
 
 from app.main import main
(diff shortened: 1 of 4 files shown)


app/main/views/dashboard.py
```diff
@@ -244,7 +243,16 @@ def get_dashboard_partials(service_id):
         for job in job_api_client.get_jobs(service_id, limit_days=7, statuses=statuses_to_display)['data']
     ]
     service = service_api_client.get_detailed_service(service_id)
-    column_width = 'column-third' if 'letter' in current_service['permissions'] else 'column-half'
+    column_width, max_notifiction_count = get_column_properties(
+        3 if 'letter' in current_service['permissions'] else 2
```
richardc0: What are three and 2? Consider making them file consts COLUMN_HALF?

app/main/views/dashboard.py
```diff
@@ -437,3 +448,10 @@ def get_tuples_of_financial_years(
         )
         for year in range(start, end + 1)
     )
+
+
+def get_column_properties(number_of_columns):
+    return {
+        2: ('column-half', 999999999),
```
richardc0: seconded time they are used so I'd go for a file const

app/main/views/dashboard.py
quis: `3` and `2` are the number of columns. In other words, if there are three columns then each of them is a third of the total width. I think this is more clearly explained with a named argument, which is what I’ve done in d6fec9f.

PR #1639, opened 2017-11-14
Allow delete email reply to address, SMS senders

quis:
For both SMS senders and email reply to addresses this commit adds:
- a delete link
- a confirmation loop

It doesn’t let users delete:
- default SMS senders or reply to addresses (they always have to have at least one)
- inbound numbers

It assumes that the API will allow updating of an attribute named `active` on the respective database rows. It could work in a different way. We can’t do complete deletion though because these will still be keyed to notifications.

---

image

image


quis:
This will probably need changing to send a `DELETE` request to the API instead (once we’ve stopped foreign-keying notifications to senders).

leohemsted:
@quis what happened with this

quis:
@leohemsted Still don’t think there’s an API side to this. But should be easy enough now that we’re persisting the text of reply to addresses/senders not foreign keying them.

app/notify_client/service_api_client.py
```diff
@@ -391,13 +409,21 @@ def add_sms_sender(self, service_id, sms_sender, is_default=False, inbound_numbe
             data["inbound_number_id"] = inbound_number_id
         return self.post("/service/{}/sms-sender".format(service_id), data=data)
 
-    def update_sms_sender(self, service_id, sms_sender_id, sms_sender, is_default=False):
+    def update_sms_sender(self, service_id, sms_sender_id, sms_sender=None, active=None, is_default=False):
+        data = {
+            "is_default": is_default,
+        }
+        if sms_sender is not None:
```
richardc0: Do you need another check here? i.e. sms_sender and active = False then error?

Does this make sense? (Guess its the implementation on API?)

data = {
 "is_default": false,
 "sms_sender"" "1234567890"
 "active": false,
}

app/notify_client/service_api_client.py
```diff
@@ -391,13 +409,21 @@ def add_sms_sender(self, service_id, sms_sender, is_default=False, inbound_numbe
...
+            data.update({
+                "sms_sender": sms_sender,
+            })
+        if active is not None:
+            data.update({
+                'active': active,
```
richardc0: For consistency " instead of '

app/templates/views/service-settings/email-reply-to/edit.html
```diff
@@ -22,16 +34,23 @@ <h1 class="heading-large">
...
+      ) }}
     {% else %}
       <div class="form-group">
         {{ checkbox(form.is_default) }}
       </div>
+      {{ page_footer(
+        'Save',
+        back_link=url_for('.service_email_reply_to', service_id=current_service.id),
+        back_link_text='Back',
+        delete_link=url_for('.service_confirm_delete_email_reply_to', service_id=current_service.id, reply_to_email_id=reply_to_email_address_id),
```
richardc0: Would it be more consistent to have a delete button?

Or 

move the Link to the list on the page before so you have 'Change' and 'Delete'

if Not I'd say the screen before should be 'manage' and not 'change'

app/templates/views/service-settings/sms-sender/edit.html
```diff
@@ -9,9 +10,20 @@
 
 {% block maincolumn_content %}
 
-  <h1 class="heading-large">
-    Edit text message sender
-  </h1>
+  {% if confirm_delete %}
+    <div class="bottom-gutter">
+      {% call banner_wrapper(type='dangerous', subhead="Are you sure you want to delete this text message sender?") %}
```
richardc0: Again maybe one for the backend BUT...
can I add another email or SMS which is the same as I have just deleted?
e.g. I have added test@test.com and then delete it can I add it again?

quis:
@richardc0 story written here: https://www.pivotaltracker.com/story/show/156476018

I think this PR will have to change a bit to make a delete request, not an update request.

klssmith:
I've not made any changes to the first commit in this PR (just rebased it). 

The API side of this story is complete - it deletes reply to addresses by changing their `archived` attribute to `True` and has separate endpoints for deleting (see https://github.com/alphagov/notifications-api/pull/1847). The second commit changes the code to use the new endpoints.

app/templates/views/service-settings/sms-sender/edit.html
klssmith: Email reply-to addresses and SMS senders are not unique, so if you delete test@test.com then try to add test@test.com again it will create a new row in the database. You would have two rows with the email address of test@test.com, one where `archived` is `True` and one where `archived` is `False`

servingUpAces:
This has been tested and approved.

PR #1640, opened 2017-11-14
Email invites!

leohemsted:
### 1. invite a user with email auth selected

image

### 2. Receive a totally normal email invite

image

### 3. Fill out a form - there's no mobile number form here

image

### Other things work - for example:

Can't click if you're logged in:
image

You don't have a phone number in your profile:
image

(but you can add one if you want to by editing that field)

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -171,13 +171,11 @@ class RegisterUserForm(Form):
     email_address = email_address()
     mobile_number = international_phone_number()
     password = password()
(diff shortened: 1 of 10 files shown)


quis:
For the mobile number on the user’s profile page, can you use the `optional_text_field` macro?

leohemsted:
@quis 
image

quis:
giphy-1

PR #1642, opened 2017-11-14
Add a meta description tag to the product page

quis:
Google tries to auto-generate a snippet of a site’s content to show in search results. Currently it’s not doing a great job of this for Notify.

There’s a chance that if we give it better content in the site’s meta description then it will use that instead. Worth a go…

The content is adapted from the blue box on the product page.

It’s 145 characters, which is within the 160 characters recommended[1].

It matches the content in the page, and contains words that users are likely to be searching for (GOV.UK Notify, emails, text messages).

It’s only on the homepage, because it shouldn’t be duplicated across multiple pages.

1. https://yoast.com/meta-descriptions/

# Before 

screen shot 2017-11-14 at 17 20 13

# After 

Let’s see what Google does with it…


quis:
# Some time later…

image

PR #1643, opened 2017-11-14
Add proxy header check

sakisv:
This adds a check for all incoming requests to ensure they are coming from our proxies.

Counterpart of https://github.com/alphagov/notifications-api/pull/1391
Depends on https://github.com/alphagov/notifications-utils/pull/250


sakisv:
Depends on https://github.com/alphagov/notifications-admin/pull/1619 because of the utils version bump to 23.0.1

PR #1649, opened 2017-11-15
Existing user email auth invite

leohemsted:
make sure existing users accepting invites get their auth types set properly, EXCEPT FOR:
* if the service issuing the invite does not have permission to edit auth types, don't let them do anything. This will stop them turning existing email_auth users back to sms auth
* if the user hasn't got a mobile number, but the invite is for sms login, don't do anything either. They won't have a phone number if they signed up via an email_auth invite previously.

in these cases, we accept the invite and add the user to the service as normal, however, just don't update the user's auth type.

### DIFF
--- a/app/main/views/invites.py
+++ b/app/main/views/invites.py
@@ -72,6 +72,8 @@ def accept_invite(token):
         if existing_user in service_users:
             return redirect(url_for('main.service_dashboard', service_id=invited_user.service))
         else:
+            if invited_user.auth_type != existing_user.auth_type:
+                user_api_client.update_user_attribute(existing_user.id, auth_type=invited_user.auth_type)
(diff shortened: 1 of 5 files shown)


kentsanggds:
Should I be allowed to sign in on different browsers using the same link? 
As I don't think I was able to do this when signing in via SMS but have managed to do this using Chrome, incognito Chrome and Safari and have managed to continue to use the different sessions on the different instances. 

Trying again I wasn't able to replicate it time, and I'm not sure if the code in this PR is handling that but perhaps worth adding extra tests or testing manually as it's quite easy to copy and paste the link around.

kentsanggds:
Not sure if this is a concern as it's an edge case but if you turn off the edit user auth, there is no indication of what type of user auth they log in with?

PR #1650, opened 2017-11-16
Added new template usage page which will replace template-activity

richardc0:
The current template-activity page is slow as it is using the end point which uses notification_history  and hence is timing out. This adds a new pages (so that they can be compared side by side) which will be hidden until is is approved with the larger data set and tested.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -106,6 +106,61 @@ def template_history(service_id):
     )
 
 
+@main.route("/services/<service_id>/template-usage")
+@login_required
+@user_has_permissions('view_activity', admin_override=True)
+def template_usage(service_id):
+
+    year, current_financial_year = requested_and_current_financial_year(request)
+    stats = template_statistics_client.get_monthly_template_usage_for_service(service_id, year)
+
+    stats = sorted(stats, key=lambda x: (x['count']), reverse=True)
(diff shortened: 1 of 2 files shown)


app/main/views/dashboard.py
```diff
@@ -106,6 +106,61 @@ def template_history(service_id):
...
 
+@main.route("/services/<service_id>/template-usage")
+@login_required
+@user_has_permissions('view_activity', admin_override=True)
+def template_usage(service_id):
+
+    year, current_financial_year = requested_and_current_financial_year(request)
+    stats = template_statistics_client.get_monthly_template_usage_for_service(service_id, year)
+
+    stats = sorted(stats, key=lambda x: (x['count']), reverse=True)
```
leohemsted: i think with a defaultdict you could make this whole fn into a pretty short fn.

```python
import calendar
from collections import defaultdict

templates_by_month = defaultdict(list)
for stat in stats:
    templates_by_month[stat['month'].append({
        'name': stat['name'],
        'type': stat['type'],
        'requested_count': stat['count']
    })
months = [{
        'name': calendar.month_name[k],
        'templates_used': v
    } for k, v in templates_by_month
]
most_used_template_count = stats.get(0, {'count': 0})['count']
```

PR #1653, opened 2017-11-20
Improve page <h1>s in send one off flow

quis:
# Description of changes

## Fix misleading `<h1>` in one-off flow

Changing the `<h1>` in #1638 turned out to be quite confusing. The combination of the word
"recipient" and a selection of email addresses on the page was confusing.

This commit changes the page title to be much more explicit about what is expected from the page, rather than what is consistent with the text of the link that the user clicked.

## Hide form legend on choose reply page

The `<h1>` on this page says all the user needs to know. In research we saw that users didn’t even read the legend, even when prompted to!

## Improve page titles later on in the one off flow

Now that the page title for picking a sender/reply to has been improved, I think these pages are also less clear than they could be.

This commit changes the page titles to (I hope) be clearer about what is needed from the user on these pages.

# Changes to the email flow 

screen shot 2017-11-20 at 16 49 03
screen shot 2017-11-20 at 16 49 09
screen shot 2017-11-20 at 16 49 16
screen shot 2017-11-20 at 16 49 25

## Changes to the text message flow 

screen shot 2017-11-20 at 16 49 36
screen shot 2017-11-20 at 16 49 44
screen shot 2017-11-20 at 16 49 49
screen shot 2017-11-20 at 16 49 56

# Pages that don’t change 

## The tour

screen shot 2017-11-20 at 16 51 26

## Printing a test letter 

image

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -421,7 +421,11 @@ def send_test_step(service_id, template_id, step_index):
         skip_link = None
     return render_template(
         'views/send-test.html',
-        page_title=get_send_test_page_title(template.template_type, get_help_argument()),
+        page_title=get_send_test_page_title(
+            template.template_type,
+            get_help_argument(),
+            entering_recipient=not session['recipient']
+        ),
         template=template,
         form=form,
         skip_link=skip_link,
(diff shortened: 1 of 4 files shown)


app/main/views/send.py
```diff
@@ -421,7 +421,11 @@ def send_test_step(service_id, template_id, step_index):
         skip_link = None
     return render_template(
         'views/send-test.html',
-        page_title=get_send_test_page_title(template.template_type, get_help_argument()),
+        page_title=get_send_test_page_title(
+            template.template_type,
+            get_help_argument(),
+            entering_recipient=not session['recipient']
```
leohemsted: elsewhere we've used `step_index == 0` to check, rather than looking at the session. Not sure if it's big enough a distinction for me to care

quis: A `step_index` of `0` means different things in the send test/send one off routes 😱

quis:
```
>       assert page.h1.text.strip() == expected_h1
E       AssertionError: assert 'Who should t...e be sent to?' == 'Personalise this message'
E         - Who should this message be sent to?
E         + Personalise this message
```

leohemsted: oh good 😩

PR #1654, opened 2017-11-21
Updated link to for browsing to financial years

richardc0:
- The link which when clicked allows the user to view different financial years was pointing to the template_activity page
- Added Template id so that the link through to the template is maintained
- Handled empty data sets from API, so empty months are displayed to the user

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -116,23 +116,25 @@ def template_usage(service_id):
 
     stats = sorted(stats, key=lambda x: (x['count']), reverse=True)
 
-    months = [
-        {
-            'name': yyyy_mm_to_datetime(month).strftime('%B'),
-            'templates_used': [],
-        }
-        for month in get_months_for_financial_year(year, time_format='%Y-%m')
-    ]
-
-    for i, d in enumerate(months):
-        for stat in stats:
-            if d['name'] == calendar.month_name[int(stat['month'])]:
-                d['templates_used'].append({
+    def get_monthly_template_stats(month_name, stats):
+        return {
+            'name': month_name,
+            'templates_used': [
+                {
                     'id': stat['template_id'],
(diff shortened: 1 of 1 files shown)


leohemsted:
We've used `RequestRedirect` elsewhere to raise a `301 PERMANENTLY MOVED` (ie: your browser will cache the redirect)

app/main/views/dashboard.py
```diff
@@ -117,9 +117,18 @@ def template_usage(service_id):
 
     stats = sorted(stats, key=lambda x: (x['count']), reverse=True)
 
+    months = [
+        {
+            'name': yyyy_mm_to_datetime(month).strftime('%B'),
+            'templates_used': {},
+        }
+        for month in get_months_for_financial_year(year, time_format='%Y-%m')
+    ]
```
quis: Where is this being used? It’s redefined on [line 136](https://github.com/alphagov/notifications-admin/pull/1654/files#diff-9354168bac7d329e54c40f0505e3d278R136).

richardc0: There is also the other issue that months where are are no stats are not printed out. re-organised to use months and add to that.

quis:
I think you can do this without the nested `for` loops and mutating data, like this (hard to explain without actually writing the code, sorry):

```python
def get_monthly_template_stats(month_name, stats):
  return {
      'name': month_name,
      'templates_used': [
        {
          'id': stat['template_id'],
          'name': stat['name'],
          'type': stat['type'],
          'requested_count': stat['count']
        }
        for stat in stats
        if calendar.month_name[int(stat['month'])] == month_name
      ],
  }

months = [
    get_monthly_template_stats(month, stats)
    for month in get_months_for_financial_year(year, time_format='%B')
]
```

richardc0:
@quis thanks. Updated and tested.

PR #1662, opened 2017-11-24
Disabled the template_history endpoint

richardc0:
- Removed the code for the template_history endpoint and replaced with a redirect to the new page so that anyone is forwarded on
- Updated the template to point to the new template_usage page
- Updated tests and added a new mock_get_monthly_template_usage
- Deleted get_monthly_template_statistics_for_service
- Added new test to test the redirection of the old endpoint

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -81,7 +81,7 @@ def template_history(service_id):
 
 @main.route("/services/<service_id>/template-usage")
 @login_required
-@user_has_permissions(admin_override=True)
+@user_has_permissions('view_activity', admin_override=True)
 def template_usage(service_id):
 
     year, current_financial_year = requested_and_current_financial_year(request)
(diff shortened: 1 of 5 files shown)


app/main/views/dashboard.py
```diff
@@ -75,40 +76,12 @@ def service_dashboard_updates(service_id):
...
-            ), default=0)
-            for month in months
-        ),
-        years=get_tuples_of_financial_years(
-            partial(url_for, '.template_history', service_id=service_id),
-            end=current_financial_year,
-        ),
-        selected_year=year,
-    )
+    raise RequestRedirect("/services/{}/template-usage".format(service_id))
```
quis: We tend to use Flask’s built in `redirect` for this, eg https://github.com/alphagov/notifications-admin/blob/f02bcd3338d15e0dce0825583157600e840ce911/app/main/views/sign_out.py#L16

And `url_for` is preferable to formatting the path by hand.

PR #1665, opened 2017-11-28
Change the CSRF error from 400 to 403

sakisv:
All 400s are handled by a generic errorhandler which logs them and
converts them to 500.

We don't need to be alerted for CSRF problems.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -92,7 +92,7 @@ def create_app(application):
     init_app(application)
     statsd_client.init_app(application)
     logging.init_app(application, statsd_client)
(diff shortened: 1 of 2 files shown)


leohemsted:
dont merge i want to push a test to this branch

leohemsted:
@quis wanna have a look at the wording? And/or the general error handling? I'm really not sure how to make it obvious that a user just needs to go back to the form page, refresh, and fill it in again. Not sure if we can easily do the first half of that flow for them?

but i think we can merge this for now just to get rid of unhandled exceptions

PR #1673, opened 2017-11-30
Rationalise top navigation and footer 

quis:
# Signed out

## Before

image

## After 

image

# Signed in

## Before 

image

## After 

image

# Footer

## Before 

image

## After 

image

# Individual changes 

## Make features and documentation always visible

These links should always appear in the header, because they’re useful to both signed in and signed out users.

## Remove performance link from features nav

The features nav is supposed to navigate your between pages in the app. It’s very unexpected to have it open an external link.

Performance isn’t strictly a part of Support, but it’s worked having it there for long enough that it’s probably not a bother.

Before | After
--- | ---
image | image

## Make order of About links match features sidebar

The links in the footer should be in the same order as those in the sidebar of the features page. That’s the order that’s been thought about more.

## Make pricing its own column in footer

This makes the footer match the order and grouping of links in the top navigation.

## Remove headings from footer

It doesn’t make sense to have a section called ‘Support’ which has a link called ‘Support’ in it.

And by splitting up, and reducing the number of links in the footer, they don’t _need_ headers – hopefully they’re self explanatory.

## Rename Slack link

It was quite long, and Thom wasn’t a fan of the wording.

### DIFF
--- a/app/main/views/sub_navigation_dictionaries.py
+++ b/app/main/views/sub_navigation_dictionaries.py
@@ -13,11 +13,6 @@ def features_nav():
             "link": "main.security",
         },
         {
-            "name": "Performance",
-            "link": "https://www.gov.uk/performance/govuk-notify",
-            "external_link": True,
-        },
-        {
             "name": "Terms of use",
             "link": "main.terms",
         },
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -87,11 +87,11 @@
         <div class="column-one-third">
           <h2>About</h2>
           <ul>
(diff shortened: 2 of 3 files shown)


gov-cjwaszczuk:
It all seems to look good, my only concern is whether the sign in nav doesn't drop to a new line on smaller laptops. Is it also worth removing the external link functionality from the sub_navigation, or will it potentially play a part in the future?

quis:
Looks like this on smaller screens, which is… not totally broken:

image

quis:
Good point about the external nav, removed it in 1817d3f0

gov-cjwaszczuk:
Looks good to me now, menu seems to hold fine down to 1000px which should cover most small desktops.

PR #1681, opened 2017-12-04
Allow service to set callback url for notifications

gov-cjwaszczuk:
## Summary
Adds an option to allow a user to specify a callback url for their service to receive delivery receipts from. Refactors previous inbound api implementation also, removing it from the service settings page. Both now sit under a button called callbacks within the api integration page. If inbound sms is not on, the user will immediately go to the delivery receipts callback page, rather than the table page. 
### :small_orange_diamond: :small_blue_diamond: Api integration page :small_blue_diamond: :small_orange_diamond:
image
### :small_orange_diamond: :small_blue_diamond: API callbacks page :small_blue_diamond: :small_orange_diamond:
image
### :small_orange_diamond: :small_blue_diamond: Delivery receipts page :small_blue_diamond: :small_orange_diamond:
image
### :small_orange_diamond: :small_blue_diamond: Received messages page :small_blue_diamond: :small_orange_diamond:
image
### :small_blue_diamond: :small_orange_diamond: Service settings page :small_blue_diamond: :small_orange_diamond:
image
### :small_blue_diamond: :small_orange_diamond: Documentation page :small_blue_diamond: :small_orange_diamond:
image

Pivotal: https://www.pivotaltracker.com/story/show/153087419


app/main/views/api_keys.py
```diff
@@ -113,3 +115,46 @@ def revoke_api_key(service_id, key_id):
...
+            current_service['id'],
+            current_service.get('service_callback_api')[0]
+        )
+
+
+@main.route("/services/<service_id>/set-callback-api", methods=['GET', 'POST'])
+@login_required
+def set_callback_api(service_id):
+    callback_api = get_callback_api()
+    print(callback_api)
```
quis: 🖨

tests/app/main/views/test_api_keys.py
```diff
@@ -391,3 +391,117 @@ def test_should_validate_whitelist_items(
...
     mock_update_whitelist.assert_not_called()
+
+
+@pytest.mark.parametrize('url, bearer_token, expected_errors', [
+    ("", "", "Can’t be empty Can’t be empty"),
+    ("http://not_https.com", "1234567890", "Must be a valid https URL"),
+    ("https://test.com", "123456789", "Must be at least 10 characters"),
+])
+def test_set_callback_api_validation(
+    logged_in_client,
```
quis: Should be using the `client_request` fixture instead of `logged_in_client` wherever possible.

app/templates/views/api/set-callback-api.html
```diff
@@ -0,0 +1,38 @@
...
+
+      <form method="post">
+        {{ textbox(
+          form.url,
+          width='2-3',
+          hint='Valid https URL'
+        ) }}
+        {{ textbox(
+          form.bearer_token,
+          width='2-3',
```
quis: Can you make both of these textboxes `1-1` wide please? The URLs can be fairly long, from looking at what’s in the DB.

quis:
Also, I presume this depends on https://github.com/alphagov/notifications-api/pull/1459 before it can be shipped?

venusbb:
Chrome remembers the previous URL and Bearer_token as it recognises it as a username/password field. Can we somehow disable the autocomplete.

app/templates/views/callbacks.html
```diff
@@ -60,4 +62,32 @@ <h2 class="heading-medium">Format of the callback</h2>
...
+        field_headings_visible=True,
+        caption_visible=False
+      ) %}
+        {% for key, description, format in [
+          ('id', 'Notify’s id for the status receipts', 'UUID'),
+          ('reference', 'The reference sent by the service', '12345678'),
+          ('to', 'The email address of the recipient', 'hello@gov.uk'),
+          ('status', 'The status of the notification', 'delivered | permanent-failure | temporary-failure | technical-failure'),
+          ('created_at', 'The time the service sent the request', '2017-05-14T12:15:30.000000Z'),
+          ('updated_at', 'The last time the status was updated', '2017-05-14T12:15:30.000000Z'),
```
leohemsted: when querying data through the API, we don't return `updated_at`, since it's a bit unhelpfully vague. Instead, we return `completed_at` (since that's what this represents in this case).

I know the gov.uk integration callbacks doc says `updated_at`, but that's probably a mistake - we should really be using `completed_at`, especially since all of these are specifically about something that reached a final state.

venusbb:
Continue in PR #1700 because of permission problem pushing to Fork.

venusbb:
Continue in PR #1700 because of permission problem pushing to Fork.

PR #1682, opened 2017-12-04
update free sms fragment limit via billing client instead of service

leohemsted:


tests/app/main/views/test_dashboard.py
```diff
@@ -631,7 +631,7 @@ def test_usage_page_with_year_argument(
     mock_get_usage,
     mock_get_billable_units,
     mock_get_free_sms_fragment_limit,
-    mock_create_or_update_free_sms_fragment_limit
+    # mock_create_or_update_free_sms_fragment_limit
 ):
```
venusbb: Do we still need this line?

leohemsted: good spot, nope

PR #1685, opened 2017-12-05
Update error message about commas in placeholders

quis:
We call the yellow things ‘double brackets’ on the frontend, not fields or placeholders. This error message was a bit out of date.

Also refactored it to use the `Field` class; this code was probably written before `Field` was factored out of `Template`.

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -45,7 +45,7 @@ class ValidGovEmail:
 
 class NoCommasInPlaceHolders:
 
-    def __init__(self, message='You can’t put commas inbetween double brackets'):
+    def __init__(self, message='You can’t put commas between double brackets'):
         self.message = message
 
     def __call__(self, form, field):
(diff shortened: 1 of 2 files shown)


app/main/validators.py
```diff
@@ -45,11 +45,11 @@ def __call__(self, form, field):
 
 class NoCommasInPlaceHolders:
 
-    def __init__(self, message='You can’t have commas in your fields'):
+    def __init__(self, message='You can’t put commas inbetween double brackets'):
```
leohemsted: is `inbetween` a word? Should this be "in-between" or just "between"?

gov-cjwaszczuk: probably should just be "between"

quis: Thom pls halp.

quis: Changed to ‘between’ in 19c7b1a5718e89716eefb283011f8809f04169a6

PR #1686, opened 2017-12-05
Allow horizontal scrolling for big spreadsheets

quis:
_Not sure if we should just roll this out, but we don’t really have capacity to usability test it at the moment. I’ve shown it to a few other designers and they seem to think it makes sense. Doesn’t make sense to just have it growing stale on my computer._

_There’s more to come on this, ie the pages you see for column errors and a successful upload, but think this can go out independently._

***

_Update: see [later comments](https://github.com/alphagov/notifications-admin/pull/1686#issuecomment-352016823) for how I’ve made this less of a drastic change._

***

fullscreen

***

Two bits of context:

1. As we start dealing with letters, which have more columns, it’s more  likely that people’s spreadsheets won’t fit in our current layout.

2. We already removed the view of the template from the page that shows row-level errors (eg bad phone number or missing personalisation) in spreadsheets because you don’t need to know about the content of the message in order to fix the errors.

This commit goes further by removing anything that isn’t to do with the errors, including the normal GOV.UK header and the service’s navigation.

This means the content can go the width of the page, which means it can be allowed to scroll horizontally without being a usability car crash. Which means that the layout doesn’t break with a spreadsheet that has lots of columns.

***

Getting the scrolling to play nice with the fixed position elements is a bit wild. Here’s a picture of what the Javascript is doing (the table overflows and scrolls horizontally inside the blue box).

img_7479


quis:
Worth noting that row-level errors are the most common we see (numbers are for 3 months to date): 

image

quis:
Tested without JS; tested in IE9:

fullscreen-ie9

quis:
# Change tables to scroll in-page, not full screen

There were three problems with showing tables fullscreen:
- it was over-optimised for very big spreadsheets, whereas most users will only have a few columns in their files
- it was jarring to go from full screen and back to the normal layout
- it was a bit change for existing users, where we prefer incremental changes that make things better without disrupting people’s work (where possible)

So this commit changes the big table to scroll horizontally in the page, not take up the full width of the page. This is an enhancement to the way things work currently, rather than change.

From the fullscreen table it keeps:
- the shimming method to keep the horizontal scrollbar at the bottom of the screen at all times

It introduces some more refinements to make it nicer to use:
- fixing the first column, so you always know what row you’re on
- adding shadows indicate where there is content that’s scrolled outside the edges of the container

---

Looks like this:

scrolly-thing

quis:
And with Windows scrollbars:

windows-scroll

PR #1689, opened 2017-12-06
Updated service manual link on using-notify page

gov-cjwaszczuk:
Changes the link on the using-notify page to be on the 'Service Manual' text, rather than the bullet point.

### DIFF
--- a/app/templates/views/using-notify.html
+++ b/app/templates/views/using-notify.html
@@ -16,10 +16,10 @@
     <h1 class="heading-large">Using Notify</h1>
 
     <h2 class="heading-medium">Writing and sending text messages, emails and letters</h2>
-    <p>Check the Service Manual for guidance on how to:</p>
+    <p><a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">Check the Service Manual</a> for guidance on how to:</p>
     <ul class="list list-bullet">
-      <li><a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">plan and write text messages and emails</a></li>
-      <li><a href=""><a href="https://www.gov.uk/service-manual/technology/how-to-email-your-users"></a>send emails in a way that protects users from spam and phishing</a></li>
+      <li>plan and write text messages and emails</li>
+      <li><a href="https://www.gov.uk/service-manual/technology/how-to-email-your-users">send emails</a> in a way that protects users from spam and phishing</li>
     </ul>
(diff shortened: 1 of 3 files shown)


app/templates/views/using-notify.html
```diff
@@ -16,9 +16,9 @@
     <h1 class="heading-large">Using Notify</h1>
 
     <h2 class="heading-medium">Writing and sending text messages, emails and letters</h2>
-    <p>Check the Service Manual for guidance on how to:</p>
+    <p>Check the <a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">Service Manual</a> for guidance on how to:</p>
     <ul class="list list-bullet">
-      <li><a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">plan and write text messages and emails</a></li>
+      <li>plan and write text messages and emails</li>
       <li><a href=""><a href="https://www.gov.uk/service-manual/technology/how-to-email-your-users"></a>send emails in a way that protects users from spam and phishing</a></li>
```
quis: There’s something weird going on here. One `<a>` inside another. Should sort this out before merging.

app/templates/views/using-notify.html
```diff
@@ -16,10 +16,10 @@
     <h1 class="heading-large">Using Notify</h1>
 
     <h2 class="heading-medium">Writing and sending text messages, emails and letters</h2>
-    <p>Check the Service Manual for guidance on how to:</p>
+    <p>Check the <a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">Service Manual</a> for guidance on how to:</p>
```
quis: Doesn’t this link belong on the first bullet point?

ie:
> Check the Service Manual for guidance on how to:
>
> - [plan and write](https://www.gov.uk/service-manual/design/sending-emails-and-text-messages)  text messages and emails
> - [send emails](https://www.gov.uk/service-manual/technology/how-to-email-your-users) in a way that protects users from spam and phishing

gov-cjwaszczuk: wait no this actually is correct, it was a request from Thom to change it: 

image

gov-cjwaszczuk: Even though it's actually still wrong ..

gov-cjwaszczuk:
If we leave this open, I'll make the other changes that Thom request today aswell and alter the PR naming

PR #1696, opened 2017-12-08
Add URLs for previewing all rows of a spreadsheet

quis:
We’ve heard from some users, especially those sending letters, that they’d like to check that a spreadsheet they’ve uploaded has populated the template correctly.

My reckon is that seeing just one row of the spreadsheet populate the template isn’t enough to give people confidence that everything’s working properly. So this commit is the first step towards being able to preview an arbitrary row of a template, by extending the URL structure to optionally accept a row number for pages or files (ie PNG) that preview successfully uploaded spreadsheets.

What this commit doesn’t do is link to these pages; that will come as part of a subsequent commit.


app/main/views/send.py
```diff
@@ -522,19 +521,16 @@ def _check_messages(service_id, template_type, upload_id, letters_as_pdf=False):
...
 
-    with suppress(StopIteration):
-        first_recipient = None
-        template.values = next(recipients.rows)
-        first_recipient = template.values.get(
-            Columns.make_key(recipients.recipient_column_headers[0]),
-            ''
-        )
+    try:
+        template.values = recipients[preview_row]
```
richardc0: You could be defensive here and check rather than exception-ing as exceptions are expensive performance-wise e.g. if recipients.len() == preview_row then template.values = recipients[preview_row] otherwise abort

quis: Good idea, rewrote in 5f5dd3a

PR #1722, opened 2017-12-20
Letters on usage page

servingUpAces:
This PR add the letter data on the usage page. 
Monthly billing now contains letter data. 
If the service has the letter permission the usage page will show the letter usage data. Otherwise the letter data will not show up on the page.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -137,8 +137,12 @@ def usage(service_id):
     free_sms_allowance = billing_api_client.get_free_sms_fragment_limit_for_year(service_id, year)
     units = billing_api_client.get_billable_units(service_id, year)
     yearly_usage = billing_api_client.get_service_usage(service_id, year)
+
+    usage_template = 'views/usage.html'
+    if 'letter' in current_service['permissions']:
+        usage_template = 'views/usage-with-letters.html'
     return render_template(
-        'views/usage.html',
+        usage_template,
         months=list(get_free_paid_breakdown_for_billable_units(
             year,
             free_sms_allowance,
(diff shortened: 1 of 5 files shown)


app/main/views/dashboard.py
```diff
@@ -135,20 +135,26 @@ def usage(service_id):
     year, current_financial_year = requested_and_current_financial_year(request)
 
     free_sms_allowance = billing_api_client.get_free_sms_fragment_limit_for_year(service_id, year)
+    units = billing_api_client.get_billable_units(service_id, year)
+    yearly_usage = billing_api_client.get_service_usage(service_id, year)
+
+    usage_template = 'views/usage.html'
+    if 'letter' in current_service['permissions']:
+        usage_template = 'views/usage-with-letters.html'
```
leohemsted: can we remove `usage-with-letters` and instead just update the existing `usage.html`.


I'd probably add a `show_letters='letter' in current_service['permissions']` variable.

Then at the top of the top of the jinja file you can write `{% set column_size = 'column-one-third' if show_letters else 'column-half' %}`, and then we can just use that in the `<div class='{{column_size}}>`.

servingUpAces: I talked to chs about it. Because you have to change from using 2 to 3 columns to add the complexity in the template is not pretty. We thought it would actually be easier to deal with 2 templates than one.

servingUpAces:
Ok - chs is going to look at the template next year.

app/templates/views/usage-with-letters.html
```diff
@@ -0,0 +1,141 @@
...
+              {% endif %}
+              {% if month.paid %}
+                <li class="tabular-numbers">{{ "{:,}".format(month.paid) }} text messages at
+                {{- ' {:.2f}p'.format(sms_rate * 100) }}</li>
+              {% endif %}
+              {% if not (month.free or month.paid) %}
+                <li aria-hidden="true">–</li>
+              {% endif %}
+              </ul>
+             <ul>
```
leohemsted: i don't want it to stop this merge, but @gov-cjwaszczuk i think this `</ul><ul>` probs shouldn't be here, should it?

PR #1726, opened 2017-12-21
Make email addresses case insensitive when inviting users to services

klssmith:
Email addresses in invites should be case insensitive. This is to stop
the bug where a user creates their account using a lower case email
address (e.g. user1@gov.uk), but is then invited to a service using
their email address in a different case (e.g. USER1.gov.uk) and sees
an error message telling them that they can't accept an invite for a
different email address.

[Pivotal story](https://www.pivotaltracker.com/n/projects/1443052/stories/153825749)

### DIFF
--- a/app/main/views/invites.py
+++ b/app/main/views/invites.py
@@ -38,7 +38,7 @@ def accept_invite(token):
 
     invited_user = invite_api_client.check_token(token)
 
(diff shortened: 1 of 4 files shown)


klssmith:
@richardc0 I double checked about api and how we store email addresses there - the plan is to keep storing the email addresses as the user enters them, but whenever we compare email addresses we should always ignore the case.

PR #1740, opened 2018-01-02
Force example CSV to download, not open in browser

quis:
We wouldn’t wan’t anyone just seeing the raw CSV data in their browser – it isn’t clear what a user would do with it at that point. And we wouldn’t want it navigating them away from the send page, because this might cause them to lose their place.

This commit forces the file to download using the HTML5 `download` attribute:

> This attribute instructs browsers to download a URL instead of navigating to it, so the user will be prompted to save it as a local file. 

– https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download

### DIFF
--- a/app/templates/partials/jobs/notifications.html
+++ b/app/templates/partials/jobs/notifications.html
@@ -31,7 +31,7 @@
           </p>
         {% elif notifications %}
           <p class="bottom-gutter">
-            <a href="{{ download_link }}" download="download" class="heading-small">Download this report</a>
+            <a href="{{ download_link }}" download class="heading-small">Download this report</a>
             &emsp;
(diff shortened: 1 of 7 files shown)


quis:
Discussed offline with @gov-cjwaszczuk:

We should standardise on `<a download>` rather than `<a download="download">` everywhere. The value of the download attribute tells the browser what filename to use, but is overridden by the [`Content-Disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) HTTP header. Since it’s not being used, we should remove it for the sake of disambiguation.

quis:
Updated in: c2a0c39

PR #1744, opened 2018-01-03
Add make targets to enable and disable failwhale

sakisv:
Also add a couple of convenience targets and bring back the DNS_NAME
variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,25 +39,30 @@ venv/bin/activate:
 .PHONY: check-env-vars
 check-env-vars: ## Check mandatory environment variables
 	$(if ${DEPLOY_ENV},,$(error Must specify DEPLOY_ENV))
+	$(if ${DNS_NAME},,$(error Must specify DNS_NAME))
 
 .PHONY: sandbox
 sandbox: ## Set environment to sandbox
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -188,3 +192,23 @@ cf-rollback: ## Rollbacks the app to the previous release
 .PHONY: cf-push
 cf-push:
 	cf push -f manifest-${CF_SPACE}.yml
+
+.PHONY: cf-target
+cf-target: check-env-vars
```
richardc0: I like the target space being supplied as part of the command so that you don't by mistake deploy it to the wrong  space? What do you think?

sakisv: Agree 100%. That's why the target `enable-failwhale` has two "dependencies": `enable-failwhale: cf-target cf-failwhale-deployed`. 
`cf-target` is responsible for actually targeting the space, thus checking (via its own dependency, `check-env-vars`) if the `DEPLOY_ENV` has been defined, so if you just call `make enable-failwhale`, you will get an error.

Example:
```bash
$ make enable-failwhale
Makefile:41: *** Must specify DEPLOY_ENV.  Stop.
```

PR #1765, opened 2018-01-10
Download link activity page

servingUpAces:
We removed the download link on the activity page (click the big numbers) because it would time out for large requests. 
When we fix the download functionality we didn't put the link back on the page. This PR adds the link back.

Some refactoring is required as there is some replication of code, but I wanted to get this out there and tested. We may also want to iterate on the columns returned in the CSV.


app/main/views/jobs.py
```diff
@@ -35,39 +32,11 @@
     user_has_permissions,
     generate_notifications_csv,
     get_time_left,
-    REQUESTED_STATUSES,
-    FAILURE_STATUSES,
-    SENDING_STATUSES,
-    DELIVERED_STATUSES,
     get_letter_timings,
-)
+    parse_filter_args, set_status_filters)
```
quis: I’d prefer bracket on a new line and trailing comma like this:
```python
    parse_filter_args,
    set_status_filters,
)
```
makes for cleaner diffs in the future

app/utils.py
```diff
@@ -132,21 +135,36 @@ def generate_notifications_csv(**kwargs):
...
+                    notification['status'],
+                    notification['created_at']
+                ]
+                line = ','.join(str(i) for i in values) + '\n'
+                yield line
+        else:
+            # Change here
+            for notification in notifications:
+                values = [
+                    notification.get('row_number', None),
```
quis: I think we should lose this column, doesn’t make too much sense in the all notifications report.

app/utils.py
```diff
@@ -132,21 +135,36 @@ def generate_notifications_csv(**kwargs):
...
+                    notification['to'],
+                    notification['template']['name'],
+                    notification['template']['template_type'],
+                    notification.get('job_name', None),
+                    notification['status'],
+                    notification['created_at']
```
quis: Should have `updated_at` as a column here too – need to think of a sensible column name for it.

servingUpAces: We use "completed_at" for get_notifications

PR #1766, opened 2018-01-11
Fix spacing of table footers

quis:
When implementing the horizontally scrollable tables I messed up the spacing above the text that says “Only showing the first 50 rows”, where it’s displayed below a table that doesn’t have horizontal scrolling.

This PR makes the spacing correct in all these cases:

# Example spreadsheet

image

# Uploaded file with no horizontal overflow

image

# Uploaded file with horizontal overflow

image


quis:
And this one too:

image

PR #1768, opened 2018-01-11
Include letters on monthly messages sent

gov-cjwaszczuk:
Adds letters into the monthly messages sent page. Also added a new option into the mapping table; **equal length**.  This will take the amount of columns in a table and create an equal width to accommodate them all.

## Screenshots 
image


quis:
This PR could do with some tests and screenshots.

I’m going to think about the grid and column spacing stuff.

quis:
I’ve done some thinking about the grids stuff. Think we’ll have to iterate this, but it’s a future thing, not a now thing.

PR #1775, opened 2018-01-12
Add links to navigate to/from callbacks page

quis:
# Link to callbacks from inbound page

If you’ve got an inbound number and are using the API you’ll probably be interested in this new feature we’ve got. And if you do know about it, you might not be able to find it because we’ve moved it.

# Add link back to API integration from callbacks 

Matches what we do on the API keys and whitelist pages.


gov-cjwaszczuk:
May be a little pedantic, but it would be harmless to through a test in to ensure the link only shows up when manage api keys is enabled.

PR #1788, opened 2018-01-16
Replace Deskpro ad-hoc requests with a client from utils

allait:
Moves the duplicated calls to Deskpro to notification_utils and rewrites feedback and service go-live requests to use the shared client.


quis:
Will need to incorporate the changes from this PR: https://github.com/alphagov/notifications-admin/pull/1776

allait:
@quis updated

PR #1764, opened 2018-01-10
Updated email preview to mimic sent emails

gov-cjwaszczuk:
image
@quis

PR #1797, opened 2018-01-17
Update utils to fix Deskpro empty user_email error

allait:
Deskpro user_email will now use the default address from the app
configuration if `user_email` argument was set to an empty string.


quis:
This is going to introduce a bug in the display of email templates until https://github.com/alphagov/notifications-admin/pull/1764 is merged. But this is bug is more serious so 👍

PR #1803, opened 2018-01-19
Allows services to choose if they can send letters

quis:
Our support ticket analysis shows that the most common action request after going live is turning on letters:

image

We just do this for any team that requests it – there’s no gatekeeping. So we should just allow people to make the change themselves.

This will be a better experience for our users, and less work for us. The design of the page replicates roughly what we have for international text messaging.

# Before 

image

# After 

screen shot 2018-01-19 at 14 05 00


quis:
This takes letters from 3 to 4:

image

PR #1811, opened 2018-01-24
Updated roadmap 23 Jan 2018

yahoopete:
Likely to change a bit, but don't they always. We can update following a run through, keen to get something updated asap though.


quis:
localhost-6012-features-roadmap ipad pro

yahoopete:
Not bad... coded that in the Matrix ;)

PR #1822, opened 2018-01-26
Remove syslog drain from staging

sakisv:


quis:
It’s not in production: https://github.com/alphagov/notifications-admin/blob/363c27f409dbfcef23bef872b0a32b05d232b1e0/manifest-production.yml

It’s still in preview: https://github.com/alphagov/notifications-admin/blob/363c27f409dbfcef23bef872b0a32b05d232b1e0/manifest-preview.yml#L5-L6

sakisv:
Yes, it's intentional. We can take a look at preview if we want to see how the data from syslog drain looks like in kibana. Having it on staging too was an overkill.

PR #1830, opened 2018-01-31
Allow full stops in SMS senders

quis:
_First three commits in this PR are just refactoring. Substantive change detailed below._

---

We have a team who want their (short) web address as the text message sender. This pull request updates the validation of text message senders to allow `.` as a valid character, which is currently blocking them from doing this.

We can be fairly confident this works because:

- the team are sending large volumes of messages already with their existing provider
- we’ve tested it with all combinations of
  - both our text message providers
  - an Android phone and n iPhone

### DIFF
--- a/tests/app/main/views/test_service_settings.py
+++ b/tests/app/main/views/test_service_settings.py
@@ -810,6 +810,7 @@ def test_incorrect_letter_contact_block_input(
 
 @pytest.mark.parametrize('sms_sender_input, expected_error', [
     ('elevenchars', None),
+    ('11 chars', None),
     ('', 'Can’t be empty'),
     ('abcdefghijkhgkg', 'Enter 11 characters or fewer'),
     (' ¯\_(ツ)_/¯ ', 'Use letters and numbers only'),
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_service_settings.py
```diff
@@ -809,23 +809,36 @@ def test_incorrect_letter_contact_block_input(
...
-    no_sms_senders
+    no_sms_senders,
+    mock_add_sms_sender,
 ):
     page = client_request.post(
         'main.service_add_sms_sender',
         service_id=SERVICE_ONE_ID,
         _data={'sms_sender': sms_sender_input},
-        _expected_status=200
+        _expected_status=(200 if expected_error else 302)
```
sakisv: I think this is a bit confusing:
Why 200 if we expect an error and why 302 if we don't expect error (so, in a success scenario?) ?

quis: `200` is no redirect, ie you stay on the same page with an error message

`302` means you get redirected; generally we do [post/redirect/get](https://en.wikipedia.org/wiki/Post/Redirect/Get) for form submissions.

PR #1833, opened 2018-02-01
Inbox page performance fix

servingUpAces:
The performance of the inbox page had degraded to the point of failing to load.

We pin pointed the problem to a bad loop that was calling the format_phone_number_human_readable 216 for 25 rows, yikes.

This PR fixes that performance problem.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -237,7 +237,7 @@ def get_inbox_partials(service_id):
             messages_to_show[human_readable] = message
 
     count_of_users = len(messages_to_show)
-    messages_to_show = reversed(sorted(messages_to_show.values(), key=lambda x: x['created_at']))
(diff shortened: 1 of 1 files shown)


app/main/views/dashboard.py
```diff
@@ -227,15 +227,17 @@ def get_inbox_partials(service_id):
...
-            format_phone_number_human_readable(message['user_number'])
-            for message in messages_to_show
-        }:
-            messages_to_show.append(message)
+        human_readable = format_phone_number_human_readable(message['user_number'])
+        if human_readable not in messages_to_show:
+            messages_to_show[human_readable] = message
+
+    count_of_users = len(messages_to_show)
+    messages_to_show = reversed(sorted(messages_to_show.values(), key=lambda x: x['created_at']))
```
richardc0: Would be better than just the reverse? 

Something like? sorted(messages_to_show.values(), key=lambda x: x['created_at'], reverse=True)

PR #1837, opened 2018-02-05
Removed reference to MOU now that we have contracts and MOUs.

yahoopete:


quis:
I think without the quotation marks is better – check with Thom?

quis:
I’m asking @thomchambers

quis:
Gotta get the details right.

giphy-6

PR #1843, opened 2018-02-07
Change organisations to email branding

gov-cjwaszczuk:
Pull request to adjust all previous organisation bits inside admin to their correct name, email branding. In order for us to implement the organisations, we need to change the email branding to no longer use specific URLS.

- Admin API calls have all been changed to now point to 'email-branding' rather than 'organisation'
- Tests were, for the most part, re-written to follow conventions across the board
- Javascript fixed on email branding page for colour preview
- Validation added to colour textbox
- New colour preview component

## Screenshots
image
image
image


app/main/views/email_branding.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@main.route("/email-branding/create", methods=['GET', 'POST'])
+@main.route("/email-branding/create/<logo>", methods=['GET', 'POST'])
+@login_required
+@user_has_permissions(admin_override=True)
+def create_email_branding(logo=None):
+    form = ServiceCreateEmailBranding()
+
+    logo = logo if logo else None
```
leohemsted: don't need this line

leohemsted: oh wait i guess it captures empty strings? Is that a thing we see? I would have thought flask would strip that, can we check. if not, `logo = logo or None` will do

app/templates/components/textbox.html
```diff
@@ -57,3 +57,65 @@
...
+  highlight_tags=False,
+  autofocus=False,
+  help_link=None,
+  help_link_text=None,
+  width='2-3',
+  suffix=None,
+  safe_error_message=False,
+  colour='#FFFFFF',
+  rows=8
+) %}
```
leohemsted: given that this is only used in one place, can we make this tailored specifically? so we can get rid of every option and conditional that we don't specifically use in the email branding update flow

app/templates/views/email-branding/manage-branding.html
```diff
@@ -0,0 +1,42 @@
+{% extends "views/platform-admin/_base_template.html" %}
+{% from "components/file-upload.html" import file_upload %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/textbox.html" import textbox, colour_textbox %}
+
+{% block service_page_title %}
+  {{ '{} an organisations email branding'.format('Update' if email_branding else 'Create')}}
```
leohemsted: needs an apostrophe `organisation’s` right

PR #1846, opened 2018-02-08
Add organisations

kentsanggds:
## What -

Added a platform admin page to show list of organisations and create an organisation.

Clicking on the organisation will take you to a organisation stub page which will eventually show a list of services once the service has been linked to an organisation which will be done in another story - https://www.pivotaltracker.com/story/show/153989801

## Reference

https://www.pivotaltracker.com/story/show/153989748

## Dependencies 

- [x] https://github.com/alphagov/notifications-api/tree/create-new-organisation-model Needs to be deployed before this PR

## Note 

Added additional functionality to edit organisation name which is not required as part of this story. It might not remain in the current location but the code could be useful for another story on editing an organisation name - https://www.pivotaltracker.com/story/show/153990200

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -56,10 +56,10 @@ def update_organisation(org_id):
     )
 
 
-@main.route("/organisations/create", methods=['GET', 'POST'])
+@main.route("/organisations/add", methods=['GET', 'POST'])
 @login_required
 @user_has_permissions(admin_override=True)
-def create_organisation():
+def add_organisation():
     form = CreateOrUpdateOrganisation()
 
     if form.validate_on_submit():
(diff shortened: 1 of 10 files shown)


app/main/views/organisations.py
```diff
@@ -0,0 +1,75 @@
...
+    form.name.data = org['name']
+
+    return render_template(
+        'views/organisations/manage-organisation.html',
+        form=form,
+        organisation=org
+    )
+
+
+@main.route("/organisations/create", methods=['GET', 'POST'])
```
servingUpAces: The naming convention in the admin app would be /organisations/add

kentsanggds: OK, can change, grabbed the `/create` from the `email_branding.py` but will go with majority

kentsanggds:
I've taken the format from `manage-branding.html`, should this page be updated or is it slightly different?

PR #1851, opened 2018-02-09
Add route secret key header to the API requests

allait:
Currently requests to the API made from the admin app are going from PaaS admin app to the nginx router ELB, which then routes them back to the api app on PaaS.

This makes sense for external requests, but for requests made from the admin app we could skip nginx and go directly to the api PaaS host, which should reduce load on the nginx instances and potentially reduce latency of the api requests.

API apps on PaaS are checking the X-Custom-Forwarder header (which is set by nginx on proxy_pass requests) to only allow requests going through the proxy.

This adds the custom header to the API client requests, so that they can pass that header check without going through nginx.

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -16,9 +16,8 @@ class UserApiClient(NotifyAdminAPIClient):
         super().__init__("a" * 73, "b")
 
     def init_app(self, app):
-        self.base_url = app.config['API_HOST_NAME']
-        self.service_id = app.config['ADMIN_CLIENT_USER_NAME']
-        self.api_key = app.config['ADMIN_CLIENT_SECRET']
+        super().init_app(app)
+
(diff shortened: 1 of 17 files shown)


app/notify_client/user_api_client.py
```diff
@@ -16,9 +16,8 @@ def __init__(self):
         super().__init__("a" * 73, "b")
 
     def init_app(self, app):
-        self.base_url = app.config['API_HOST_NAME']
-        self.service_id = app.config['ADMIN_CLIENT_USER_NAME']
-        self.api_key = app.config['ADMIN_CLIENT_SECRET']
+        super().init_app(app)
+
```
quis: Rebase needed to include this change: https://github.com/alphagov/notifications-admin/pull/1850/files#diff-ac0194de3598c438e547a9b8e15b690eR23

quis:
Going to rebase and force push this…

PR #1852, opened 2018-02-09
Add another prototyping app on PaaS

quis:
Just a copy/paste of the existing prototype. 

Bit of a gotcha with how we run the deploy command though:

---

Some time between version 6.32 and 6.34 of the Cloudfoundry CLI the ability to redirect the output of a command into `cf push -f` was broken.

The only alternative we can think of is writing the file to disk, doing the deploy, and then deleting it.

We’re careful to write to a directory outside the current repo to avoid:
- including secrets in the deployed package
- accidentally checking the secrets into source control

Right now this only applies to people deploying from their local machines. At some point it will affect Jenkins too, but isn’t now. So this commit only fixes the problem for the commands that developers run locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,6 +194,11 @@ cf-deploy-prototype: cf-target ## Deploys the app to Cloud Foundry
 	$(if ${CF_SPACE},,$(error Must specify CF_SPACE))
 	cf push -f <(make -s CF_MANIFEST_FILE=manifest-prototype-${CF_SPACE}.yml generate-manifest)
 
+.PHONY: cf-deploy-prototype-2
+cf-deploy-prototype-2: cf-target ## Deploys the app to Cloud Foundry
+	$(if ${CF_SPACE},,$(error Must specify CF_SPACE))
+	cf push -f <(make -s CF_MANIFEST_FILE=manifest-prototype-2-${CF_SPACE}.yml generate-manifest)
+
 .PHONY: cf-rollback
 cf-rollback: ## Rollbacks the app to the previous release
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -190,9 +190,18 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	cf delete -f notify-admin-rollback
 
 .PHONY: cf-deploy-prototype
-cf-deploy-prototype: cf-target ## Deploys the app to Cloud Foundry
+cf-deploy-prototype: ## Deploys the app to Cloud Foundry
 	$(if ${CF_SPACE},,$(error Must specify CF_SPACE))
-	cf push -f <(make -s CF_MANIFEST_FILE=manifest-prototype-${CF_SPACE}.yml generate-manifest)
+	(make -s CF_MANIFEST_FILE=manifest-prototype-${CF_SPACE}.yml generate-manifest) > ../temp_manifest.yml
```
leohemsted: i'd suggest putting the file in `/tmp/`

quis: Good idea. Squashed into 6f77e9e

Makefile
```diff
@@ -190,9 +190,18 @@ cf-deploy: ## Deploys the app to Cloud Foundry
 	cf delete -f notify-admin-rollback
 
 .PHONY: cf-deploy-prototype
-cf-deploy-prototype: cf-target ## Deploys the app to Cloud Foundry
```
leohemsted: assume this still needs cf-target?

PR #1857, opened 2018-02-12
don't log 404s as ERROR

leohemsted:


app/__init__.py
```diff
@@ -430,7 +430,6 @@ def useful_headers_after_request(response):
 
 def register_errorhandlers(application):  # noqa (C901 too complex)
     def _error_response(error_code):
-        application.logger.exception('Admin app errored with %s', error_code)
         resp = make_response(render_template("error/{0}.html".format(error_code)), error_code)
         return useful_headers_after_request(resp)
```
venusbb: In ```def render_http_error(error):```, do we also want to just log 500 related errors? rather than logging all of them?

leohemsted: any api error that gets here is unexpected (views should handle expected errors eg form validation), so we should log all of them

PR #1865, opened 2018-02-14
Remove dashboard from dashboard URL

quis:
It’s weird that `/services/<service_id>` returns `404`. The home page for every service is the dashboard – should be possible to get there from any other page just by stemming the URL.

Also makes it consistent with organisations, which will have `/organisations/<org_id>`.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -51,6 +51,13 @@ def temp_service_history(service_id):
 @main.route("/services/<service_id>/dashboard")
 @login_required
 @user_has_permissions('view_activity', admin_override=True)
+def old_service_dashboard(service_id):
+    return redirect(url_for('.service_dashboard', service_id=service_id))
+
+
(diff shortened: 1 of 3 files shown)


app/main/views/dashboard.py
```diff
@@ -51,6 +51,13 @@ def temp_service_history(service_id):
 @main.route("/services/<service_id>/dashboard")
 @login_required
 @user_has_permissions('view_activity', admin_override=True)
+def old_service_dashboard(service_id):
```
leohemsted: you can just stick two route decorators on the same function

leohemsted: although if you do that i'm not sure what `url_for` returns 🤔

quis: Yeah, I couldn’t find a way of specifying which one to use as the default

quis: And I think it will serve at both URLs; we want a redirect.

PR #1868, opened 2018-02-14
Organisation dashboard

gov-cjwaszczuk:
## Summary
PR to add an organisation's services to it's frontpage / dashboard, also adds team members to the sidebar with a static page for the time being. Also has a few changes to link services within it. If there is no organisations, the page will convey this, and not allow the user to try and post. If the organisation isn't set on the service, the 'Not Set' will now be grey rather than black.
## Screenshots 
image
image


app/templates/org_nav.html
```diff
@@ -0,0 +1,7 @@
+<nav class="navigation">
+  <ul>
+    {% if current_user.has_permissions(admin_override=True) %}
```
quis: Can you add a ‘Dashboard’ link here are well? Should link to the list of services for that org.

app/templates/org_template.html
```diff
@@ -0,0 +1,33 @@
...
+{% block fullwidth_content %}
+  <div id="content">
+    <div class="navigation-service">
+      <div class="navigation-service-name">
+        {{ current_org.name }}
+      </div>
+      <a href="{{ url_for('main.choose_service') }}" class="navigation-service-switch">Switch service</a>
+    </div>
+    <div class="grid-row">
+      {% if help %}
```
quis: Don’t need the `help` stuff here – it’s only for new users creating their first service.

app/main/__init__.py
```diff
@@ -3,34 +3,34 @@
...
+    register,
+    send,
+    service_settings,
+    sign_in,
+    sign_out,
+    styleguide,
+    templates,
+    two_factor,
+    user_profile,
+    verify
```
quis: Ideally this kind of rearranging things would be better done as a separate commit or PR. It makes it harder to see what’s going on when it’s mixed into functional changes.

If alphabetical is what we want then it would be better to change it everywhere, maybe using a tool like [isort](https://pypi.python.org/pypi/isort), and then linting for violations.

app/templates/views/organisations/organisation/index.html
```diff
@@ -0,0 +1,24 @@
+{% extends "org_template.html" %}
+
+{% block per_page_title %}
+  {{ current_org.name }}
+{% endblock %}
+
+{% block service_page_title %}
+  {{ current_org.name }}
+{% endblock %}
+
```
quis: You shouldn’t have `per_page_title` and `service_page_title`. One overrides the other.

app/templates/views/organisations/organisation/update-organisation.html
```diff
@@ -3,16 +3,16 @@
...
 
 {% block per_page_title %}
-  {{ '{} an organisation'.format('Update' if organisation else 'Create')}}
+  Update an organisation
 {% endblock %}
 
 {% block service_page_title %}
-  {{ '{} an organisation'.format('Update' if organisation else 'Create')}}
+  Update an organisation
 {% endblock %}
```
quis: Same thing with `service_page_title` and `per_page_title` here.

app/templates/views/organisations/organisation/users/index.html
```diff
@@ -0,0 +1,17 @@
+{% extends "org_template.html" %}
+
+{% block service_page_title %}
+  Team members
+{% endblock %}
```
quis: Actually is it worth having a separate thing called `org_page_title`, since this is not a page for a service?

app/templates/views/service-settings.html
```diff
@@ -257,7 +256,9 @@ <h2 class="heading-medium">Platform admin settings</h2>
       ) %}
         {% call row() %}
           {{ text_field('Organisation')}}
-          {{ optional_text_field(organisation or 'Not Set') }}
+          {% call field(status='default' if not organisation else '') %}
+              {{ organisation or 'Not Set' }}
+          {% endcall %}
```
quis: Why doesn’t the `optional_text_macro` work here? Either way it should be _Not set_ not _Not Set_.

tests/app/main/views/test_service_settings.py
```diff
@@ -2037,6 +2037,25 @@ def test_select_organisation(
...
+        url_for('.link_service_to_organisation', service_id=service_one['id']),
+    )
+
+    assert response.status_code == 200
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+
+    assert normalize_spaces(page.select_one('main p').text) == "No organisations"
+    assert not page.select_one('main button')
+
+
```
quis: Need a test to check that the organisations are shown if the API returns them.

gov-cjwaszczuk: Is this not covered in test_should_show_overview?

quis: Yeah, my bad, didn’t realise this code was already in.

PR #1874, opened 2018-02-15
Added a validation for a minimum of 3 character.

servingUpAces:
If the SMS sender is less than 3 characters one of our providers will return an error.


tests/app/main/test_validators.py
```diff
@@ -206,3 +206,7 @@ def test_sms_sender_form_validation(
     form.sms_sender.data = '###########'
     form.validate()
     assert 'Use letters and numbers only' == form.errors['sms_sender'][0]
+
+    form.sms_sender.data = '0'
```
quis: Can you make the test explicit that 3 is not OK but that 4 is?

servingUpAces: 👍

PR #1913, opened 2018-02-28
Update Features page with new content

thomchambers:
Edited some things in here based on user support tickets and new features we've added recently.


quis:

PR #1914, opened 2018-02-28
Update SMS sender validation to reject senders starting with 00

klssmith:
Having SMS senders that start with 00 can cause issues with Firetext due
to Firetext's validation rules, so we shouldn't allow SMS senders to start
with 00.

Firetext treats a double 00 at the start of the senderID as an international
prefix, so removes them. A sender of 00447876574016 would become 447876574016.

Under Firetext's validation rules, an SMS sender of five 0s (00000) would
become  4400. This is because the first 00 are removed (as the international
prefix). The third 0 is seen as the start of a phone number, and becomes 44,
leaving the final 00 = 4400.

[Pivotal story](https://www.pivotaltracker.com/story/show/155584129)

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -32,6 +32,7 @@ from wtforms.validators import DataRequired, Length, Optional, Regexp
 from app.main.validators import (
     Blacklist,
     CsvFileValidator,
+    DoesNotStartWithDoubleZero,
     LettersNumbersAndFullStopsOnly,
     NoCommasInPlaceHolders,
     OnlyGSMCharacters,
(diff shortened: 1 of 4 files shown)


quis:
Do we still have a problem if someone puts in `4400` as their sender then?

klssmith:
@quis It looks like having a sender of `4400` is ok. With senders starting with `00` you can see the error in Kibana, but a sender of `4400` doesn't cause any errors to show up for the notification.

quis:
Cool, 🚢 it!

PR #1919, opened 2018-03-01
Update Pricing content

thomchambers:
Similar to the other changes on Features and Security yesterday - just updating stuff based on user feedback/new features.

@quis I know you had some concerns over the word 'account'. Do say if you want to strike that word out/find an alternative before publishing - happy to discuss if it's not fit for purpose.


quis:

PR #1921, opened 2018-03-02
Catch duplicate recipient columns in spreadsheets

quis:
If someone has duplicate recipient columns in their file we don’t know which one to use. This commit adds an error message which should help them fix the duplication.

This commit doesn’t go to the extra effort to actually show the duplication in the preview. Don’t think it’s worth the effort/complexity for how infrequently we’ve seen this error.

---

image

---

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/376


quis:
Update: went to the extra effort of fixing the displayed column headers (but not the values):

image

PR #1922, opened 2018-03-02
Update the free allowance when the organisation type is changed

servingUpAces:
The free allowance affect the number of free text messages a services get per year, as seen on the usage page.
This allowance is being set properly when an organisation is created by not updated.
This PR updates the free allowance when the organisation type is updated.
The free allowance can also be changed if service has an exceptional free allowance.

### DIFF
--- a/tests/app/main/views/test_service_settings.py
+++ b/tests/app/main/views/test_service_settings.py
@@ -1554,7 +1554,7 @@ def test_should_show_page_to_set_organisation_type(
     ('central', 250000),
     ('local', 25000),
     ('nhs', 25000),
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_service_settings.py
```diff
@@ -1560,6 +1560,7 @@ def test_should_set_organisation_type(
     logged_in_platform_admin_client,
     mock_update_service,
     organisation_type,
+    mock_create_or_update_free_sms_fragment_limit
```
quis: Should check that this mock is being called with the expected value.

PR #1937, opened 2018-03-08
Make terms page smarter about the agreement

quis:
People are emailing us asking if their organisation has signed the agreement. In some cases they have, so this is a waste of their and our time.

This commit adds a bit of logic to the terms of use page to tell users when their organisation has already signed the agreement. This only applies in cases where we can know for sure from their email address that they have.

If we don’t know | Now we know
---|---
image | image

The wording here needs reviewing by @thomchambers, and we should check that it’s correct legally-speaking.

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -134,6 +134,9 @@ def feedback(ticket_type):
             abort(500, "Feedback submission failed")
         return redirect(url_for('.thanks', urgent=urgent, anonymous=anonymous))
 
+    if not form.feedback.data:
+        form.feedback.data = request.args.get('body', '')
+
     return render_template(
         'views/support/{}.html'.format(ticket_type),
         form=form,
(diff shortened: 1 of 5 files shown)


quis:
We need to say specifically ‘has accepted the GOV.UK Notify data sharing and financial agreement’ or ‘we already have a data sharing and financial agreement in place with your organisation’.

gov-cjwaszczuk:
I'm leant towards the first, it implies their involvement in the agreement more than the second.

thomchambers:
Suggested wording changes are in the second half of this doc. 'If we don't know' first, then 'Now we know'.
https://docs.google.com/document/d/1jnLN_JQjYbJUb39-1fzaBlUoabe2_OAO9X9SYgp9aO8/edit#

quis:

quis:
Updated wording to better differentiate between the agreement and the terms:

If we don’t know | Now we know
---|---
image | image

quis:
And when you contact us, this gets prefilled:

image

quis:
Complete wording of terms page:

localhost-6012-features-terms ipad

yahoopete:
:+1:

PR #1940, opened 2018-03-08
Add catch for http error on invite token check

gov-cjwaszczuk:
### Summary
When a user tried to accept an invite with a malformed token, the API would return a 500. Now that the API no longer returns a 500, we can handle the 400 that is returned instead. Will redirect to the sign-in page with the error provided from the API.

### Screenshot
image

### DIFF
--- a/tests/app/main/views/test_accept_invite.py
+++ b/tests/app/main/views/test_accept_invite.py
@@ -1,8 +1,10 @@
-from unittest.mock import ANY
+from unittest.mock import ANY, Mock
 
 from bs4 import BeautifulSoup
 from flask import url_for
+from notifications_python_client.errors import HTTPError
 from tests.conftest import mock_check_invite_token as mock_check_token_invite
+from tests.conftest import normalize_spaces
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_accept_invite.py
```diff
@@ -215,6 +217,36 @@ def test_cancelled_invited_user_accepts_invited_redirect_to_cancelled_invitation
...
+    ))
+
+    response = client.get(url_for('main.accept_invite', token='thisisnotarealtoken'), follow_redirects=True)
+
+    assert response.status_code == 200
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+
+    assert normalize_spaces(
+        page.select_one('.banner-dangerous').text
+    ) == 'Something’s wrong with this link. Make sure you’ve copied the whole thing.'
```
leohemsted: have we settled on one out of:

`There’s something wrong with the link you’ve used`

vs

`Something’s wrong with this link`

PR #1941, opened 2018-03-08
Add what we know of agreements to go live request

quis:
When we process a go live request it’s a pain to have to go fishing in the Google Drive to see if there’s a signed agreement in there. This should make it easier when we know there’s already an agreement.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -462,6 +462,14 @@ class GovernmentDomain:
             self.agreement_signed
         ) = self._get_details_of_domain()
 
+    @classmethod
+    def from_user(cls, user):
+        return cls(user.email_address if user.is_authenticated else '')
+
+    @classmethod
+    def from_current_user(cls):
+        return cls.from_user(current_user)
+
     @staticmethod
     def get_matching_function(email_address_or_domain):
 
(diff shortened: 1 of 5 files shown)


app/utils.py
```diff
@@ -462,6 +462,34 @@ def __init__(self, email_address_or_domain):
...
+    @classmethod
+    def from_user(cls, user):
+        return cls(user.email_address if user.is_authenticated else '')
+
+    @classmethod
+    def from_current_user(cls):
+        return cls.from_user(current_user)
+
+    @property
+    def as_human_readable(self):
```
leohemsted: can we rename this function to make it clear that this primarily returns MOU information? 

GovernmentDomain.as_human_readable, to me, would just return something `digital.cabinet-office.gov.uk (GDS) (Crown)` or something.

`GovernmentDomain.get_human_readable_agreement_status()` or something? 🤔

quis: Yeah, I planned to do that as part of a separate PR. Was leaning towards something like `AgreementInfo`. Then:
- `AgreementInfo.owner`
- `AgreementInfo.agreement_signed`
- `AgreementInfo.crown_status`
- `AgreementInfo.as_human_readable`

leohemsted: 👍 cool, this is good to go then

PR #1942, opened 2018-03-08
Precompiled page count for preview

richardc0:
Updated so that a precompiled pdf page count can be passed to the rendering template. Currently it uses the template from the API to calculate this which for a precompiled template is always 1.

Gets the PDF and then uses the utils method to get the page count.

* Added logic for precompiled letters
* Added test to test the new path
* Updated existing tests now the path has changed

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -46,9 +46,8 @@ def view_notification(service_id, notification_id):
     notification = notification_api_client.get_notification(service_id, str(notification_id))
     notification['template'].update({'reply_to_text': notification['reply_to_text']})
 
-    file_contents = view_letter_notification_as_preview(service_id, notification_id, "pdf")
-
     if notification['template']['is_precompiled_letter']:
+        file_contents = view_letter_notification_as_preview(service_id, notification_id, "pdf")
         page_count = pdf_page_count(io.BytesIO(file_contents))
     else:
         page_count = get_page_count_for_letter(notification['template'])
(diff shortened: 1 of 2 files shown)


app/main/views/notifications.py
```diff
@@ -44,6 +46,13 @@ def view_notification(service_id, notification_id):
     notification = notification_api_client.get_notification(service_id, str(notification_id))
     notification['template'].update({'reply_to_text': notification['reply_to_text']})
 
+    file_contents = view_letter_notification_as_preview(service_id, notification_id, "pdf")
```
leohemsted: should this be inside the `if precompiled letter` block - so that it doesn't make a request for other letters?

tests/app/main/views/test_notifications.py
```diff
@@ -339,3 +350,35 @@ def test_notification_page_has_expected_template_link_for_letter(
...
+
+    mock_get_notification(mocker, fake_uuid, template_type='letter', is_precompiled_letter=True)
+
+    mock_pdf_page_count = mocker.patch(
+        'app.main.views.notifications.pdf_page_count',
+        return_value=1
+    )
+
+    mocker.patch(
+        'app.notify_client.notification_api_client.NotificationApiClient.get',
```
leohemsted: is there not a more targeted function to mock here?
`api.main.views.notifications.notification_api_client.get_notification_letter_preview` probably

I don't feel comfortable mocking every single get request to the api in case we mock something that we didn't know we were calling.

PR #1953, opened 2018-03-12
Allow platform admin to get back to services they don’t own

quis:
Otherwise we have a link which doesn’t do what it says it does.


quis:
@leohemsted 👍

leohemsted:
https://github.com/alphagov/notifications-admin/pull/1954/commits/334234960f8017229041e2f640b2b119eff5d571

PR #1954, opened 2018-03-13
Choose accounts

leohemsted:
choose account page now shows you all your organisations, all the services you know about within those organisations, and then any other services you know about (that you don't know about the organisation of). Also, redirects to organisations if they're in the session, 

- [ ] https://github.com/alphagov/notifications-api/pull/1766

image


app/templates/views/choose-account.html
```diff
@@ -0,0 +1,42 @@
+{% extends "withoutnav_template.html" %}
+
+{% block per_page_title %}
+  Choose account
```
servingUpAces: Please leave this as Choose service, until we are introducing the users to the concept of organisations. We would have more time to consider the flow and wording on the page. It's ok to leave the url as /accounts.

app/templates/views/choose-account.html
```diff
@@ -0,0 +1,42 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  <h1 class="heading-large">
+    Choose account
```
servingUpAces: Please leave this as Choose service, until we are introducing the users to the concept of organisations. We would have more time to consider the flow and wording on the page.It's ok to leave the url as /accounts.

leohemsted:
TODO: redirect old endpoints to new endpoints

PR #1955, opened 2018-03-13
Clean up request to go live page

quis:
Since we’ve changed the flow that leads into this page, there are a couple of things that we can clean up or need to change:

# 1. Remove the MoU question from request to go live

This question was designed to make people feel like it was OK to submit their request without getting the MoU signed. We reckoned that this was the fastest way of getting their service live (because the MoU is the bit that’s most likely to slow them down).

We now have a better way of telling people:
- if they’ve signed the MoU already
- or to contact us if they haven’t (which is what the majority of teams seem to do now)

We were never actually using the answer to this question – we were still checking for every service whether they had it signed.

So this commit removes this now-redundant question.

# 2. Retitle request to go live submission page

`<h1>`s should be unique across the site. This page’s `<h1>` matches that of the previous page (the one with the checklist).

This commit re-titles it to:
- be unique
- more accurately describe the content of the page

---

Before | After 
---|---
localhost-6012-services-905e65d1-a816-4f48-b595-d37eabfe19d9-service-settings-submit-request-to-go-live ipad 1 | localhost-6012-services-905e65d1-a816-4f48-b595-d37eabfe19d9-service-settings-submit-request-to-go-live ipad

### DIFF
--- a/app/templates/views/service-settings/submit-request-to-go-live.html
+++ b/app/templates/views/service-settings/submit-request-to-go-live.html
@@ -13,7 +13,7 @@
 
       <h1 class="heading-large">Request to go live</h1>
 
-      <form method="post">
+      <form method="post" class="top-gutter">
         {{ checkbox_group('What kind of messages will you be sending?', [
           form.channel_email,
           form.channel_sms,
@@ -29,7 +29,7 @@
           form.method_upload,
           form.method_api
         ]) }}
-        <p>
+        <p class="bottom-gutter">
           By requesting to go live you’re agreeing to our <a href="{{ url_for('.terms') }}">terms of use</a>.
         </p>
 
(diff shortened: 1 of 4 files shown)


quis:
@thomchambers Thoughts on the page title?

app/templates/views/service-settings/submit-request-to-go-live.html
```diff
@@ -6,21 +6,14 @@
 {% from "components/banner.html" import banner_wrapper %}
 
 {% block service_page_title %}
-  Request to go live
+  How do you plan to use Notify?
```
leohemsted: i'm not convinced by this personally (since you click on the "request to go live" link to get here, and the action at the bottom is still "request to go live" but i'll concede to you and thom on this

quis:

PR #1961, opened 2018-03-16
Add client reference to API Integration / message log

kentsanggds:
## What

`client_reference` added to api message log

- makes it clearer for users and easier to copy
- also updated text in preview message


quis:
Is it worth adding tests for these changes?

PR #1963, opened 2018-03-19
fixed sessions overriding while uploading CSVs

CJWaszczuk:
If a user had multiple tabs open with different CSVs, the session would be overridden for the initial upload. PR to ensure that the session can no longer be overridden through the use of maps.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -126,10 +126,14 @@ def send_messages(service_id, template_id):
                 Spreadsheet.from_file(form.file.data, filename=form.file.data.filename).as_dict,
                 current_app.config['AWS_REGION']
             )
-            session['upload_data'] = {
-                "template_id": template_id,
-                "original_file_name": form.file.data.filename
-            }
+            if 'file_uploads' not in session:
+                session['file_uploads'] = {}
+            session['file_uploads'].update({
+                upload_id: {
+                    "template_id": template_id,
+                    "original_file_name": form.file.data.filename
+                }
+            })
             return redirect(url_for('.check_messages',
                                     service_id=service_id,
                                     upload_id=upload_id,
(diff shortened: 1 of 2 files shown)


app/main/views/send.py
```diff
@@ -468,7 +472,7 @@ def send_test_preview(service_id, template_id, filetype):
 
 def _check_messages(service_id, template_type, upload_id, preview_row, letters_as_pdf=False):
 
-    if not session.get('upload_data'):
+    if not session.get('file_uploads'):
```
leohemsted: think this should be `if upload_id not in session.get('file_uploads', {}):`

in case they have other upload ids but not this one

app/main/views/send.py
```diff
@@ -610,24 +614,23 @@ def check_messages_preview(service_id, template_type, upload_id, filetype, row_i
 @login_required
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def recheck_messages(service_id, template_type, upload_id, row_index=0):
-
-    if not session.get('upload_data'):
+    if not session.get('file_uploads'):
```
leohemsted: think this should be `if upload_id not in session.get('file_uploads', {}):`

in case they have other upload ids but not this one

app/main/views/send.py
```diff
@@ -610,24 +614,23 @@ def check_messages_preview(service_id, template_type, upload_id, filetype, row_i
...
-    return send_messages(service_id, session['upload_data'].get('template_id'))
+    return send_messages(service_id, session['file_uploads'][upload_id].get('template_id'))
 
 
 @main.route("/services/<service_id>/start-job/<upload_id>", methods=['POST'])
 @login_required
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def start_job(service_id, upload_id):
-    upload_data = session['upload_data']
+    upload_data = session['file_uploads'][upload_id]
```
leohemsted: this might key error too. all of these endpoints should have checks to make sure the data is in the session

PR #1964, opened 2018-03-19
Let users download the data sharing and financial agreement 

quis:
Rather than making users contact us to get the agreement, we should just let them download it, when we know which version to send them.

This PR adds two endpoints:
- one to serve a page which links to the agreement
- one to serve the agreement itself

This PR also adds a link from the terms of use page to download the agreement, when we can know which agreement the user should be signing:

***

image

***

image

***

image

_(ignore the .docx in this example)_

***

The dependencies for deploying this are:

1. Add the buckets for each environment (done in https://github.com/alphagov/notifications-aws/pull/336)
1. Upload the files to the buckets in each environment
    - [x] crown
    - [ ] ~~non-crown~~

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -69,6 +69,7 @@ class Config(object):
     STATSD_PORT = 8125
     NOTIFY_ENVIRONMENT = 'development'
     LOGO_UPLOAD_BUCKET_NAME = 'public-logos-local'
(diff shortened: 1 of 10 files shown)


quis:
## Only offer agreement download to non-crown bodies for now

We don’t have the crown agreement in a nice downloadable format at the moment.

quis:

PR #1968, opened 2018-03-19
Fixed sessions overriding while uploading csvs

gov-cjwaszczuk:
If a user had multiple tabs open with different CSVs, the session would be overridden for the initial upload. PR to ensure that the session can no longer be overridden through the use of maps.

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -262,12 +262,8 @@ def test_should_not_allow_files_to_be_uploaded_without_the_correct_permission(
 
 @pytest.mark.parametrize(
     "filename, acceptable_file",
-    list(zip(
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_send.py
```diff
@@ -2697,3 +2767,72 @@ def test_sms_sender_is_previewed(
...
+            **extra_args,
+            _expected_status=301
+        )
+    else:
+        client_request.post(
+            endpoint,
+            service_id=SERVICE_ONE_ID,
+            **extra_args,
+            _expected_status=301
+        )
```
leohemsted: looks good - do they all redirect to the same location? can we check that URL? (or it can be a param in the first set of params)

PR #1969, opened 2018-03-20
Don’t populate support form with arbitrary text

quis:
I don’t think it’s a massive risk (we’re certainly mitigating against any XSS), but having a page on a GOV.UK domain where you can pre-fill text on the page from a query string probably isn’t great.

So this commit restricts pre-filling the support form to a set of named questions.


quis:
Query string has only been used 4 times since we shipped it yesterday:

image

PR #1974, opened 2018-03-20
Org links don't show if user doesn't have permissions

gov-cjwaszczuk:
Currently, users see all the services in an organisation with link. If they click on one they don't have permissions to see, they get a 403. This PR removes the link so this flow is no longer possible.

image

### DIFF
--- a/tests/app/main/views/organisations/test_organisation_invites.py
+++ b/tests/app/main/views/organisations/test_organisation_invites.py
@@ -99,14 +99,17 @@ def test_organisation_services_show(
 
     assert len(page.select('.browse-list-item')) == 3
 
-    for i in range(0, 3):
+    for i in range(0, 2):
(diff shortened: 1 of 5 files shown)


tests/app/main/views/organisations/test_organisation_invites.py
```diff
@@ -99,14 +99,15 @@ def test_organisation_services_show(
...
         service_id = mock_get_organisation_services(mock_get_organisation['id'])[i]['id']
 
         assert normalize_spaces(page.select('.browse-list-item')[i].text) == service_name
-        assert normalize_spaces(
-            page.select('.browse-list-item a')[i]['href']
-        ) == '/services/{}'.format(service_id)
+        if i > 1:
+            assert normalize_spaces(
+                page.select('.browse-list-item a')[i]['href']
+            ) == '/services/{}'.format(service_id)
```
leohemsted: if i == 1:
    assert not a link

PR #1986, opened 2018-03-26
Redirect to job if no session on send message

gov-cjwaszczuk:
There have been issues with people being redirected to the choose templates page after they have pressed the send button. The only way this is possible is because the session id for their upload no longer exists, and we think this may be because the job has already been created and the session popped. To solve this issue, we will now redirect them straight to the job page, so as to not cause any confusion.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -628,7 +628,15 @@ def start_job(service_id, upload_id):
     try:
         upload_data = session['file_uploads'][upload_id]
     except KeyError:
-        return redirect(url_for('main.choose_template', service_id=service_id), code=301)
+        return redirect(
+            url_for(
+                'main.view_job',
+                job_id=upload_id,
+                service_id=service_id,
+                help=request.form.get('help'),
+                just_sent='yes',
+            )
+        )
 
     if request.files or not upload_data.get('valid'):
         # The csv was invalid, validate the csv again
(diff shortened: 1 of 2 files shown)


quis:
We should pop the session after creating the job, otherwise you could end up in a situation where there’s:
- no upload info in session
- no job with that UUID

Which means that redirecting to the job page would `404`. 

I think this is good to go otherwise though.

quis:
Also worth noting that we already debounce form submissions (https://github.com/alphagov/notifications-admin/pull/1490) so the only way this could be happening is if they are deliberately clicking the button again (because it’s taking a long time, for example).

app/main/views/send.py
```diff
@@ -628,7 +628,15 @@ def start_job(service_id, upload_id):
     try:
         upload_data = session['file_uploads'][upload_id]
     except KeyError:
-        return redirect(url_for('main.choose_template', service_id=service_id), code=301)
+        return redirect(
+            url_for(
+                'main.view_job',
+                job_id=upload_id,
+                service_id=service_id,
+                help=request.form.get('help'),
```
quis: Shouldn’t have the `help` argument at all here.

quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/1998

PR #1989, opened 2018-03-27
Add link to request to go live from ‘Using Notify’

quis:
We used to have this link. It went away at some point. This should reduce the number of users raising support tickets asking ‘how do I get a live API key’ and similar.


quis:

PR #2001, opened 2018-03-29
Stop storing `template_id` and `original_file_name` in session

quis:
At the moment you can’t press refresh on the check page if there’s errors. This is because the session gets cleared when there’s errors. This is a bad user experience.

The data that this page is relying on (from the session) is:
- template ID
- original file name

Neither of these things need to be in the session because:
- they are not secret
- the user can modify them already (by choosing a different template or renaming their file locally)

So this PR moves them to being stores in the URL.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -134,10 +134,13 @@ def send_messages(service_id, template_id):
                     "original_file_name": form.file.data.filename
                 }
             })
-            return redirect(url_for('.check_messages',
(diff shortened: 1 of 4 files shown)


gov-cjwaszczuk:
Is letter still meant to be storing the template_id and filename in the session via `make_and_upload_csv_file` ?

quis:
@gov-cjwaszczuk Good point – removed it now.

PR #2024, opened 2018-04-18
Fix one off session bug

gov-cjwaszczuk:
### Summary
In the current implementation, when a user starts two one off flows, the original will be overwritten by the the second flow. This PR solves this issue by mapping the one off flows to UUIDs to allow for unique session flows.
#### Fixed overriding sessions 
The app was only setup to allow for one session at a time, to solve this, we put a simple map in to allow for sessions to be unique to each flow. This was then stored in the url throughout the one off flow so that the correct session param could be selected. The new structure of the one off sender is: 
```
Session: {
   'other_stuff': 'example'
   'one_off': {
      $one_off_id: {
         'created_at': current time of creation
         'recipient': who the message is being sent to
         'placeholders': dynamic data added to message
         'sender': {
            'sender_id': id of sender
            'sender_name' name of sender
         }
      }
   }
}
```
The use of `created_at` is to ensure that the one_off session will be cleared after a certain amount of time, this stops the amount of one_off sessions building up too quickly within the user's browser.
#### Fixed sender visual issues bug
Currently on live, when going through the one off flow you never see the sender_name for the service. This was because the code was trying to get the sender_id from the session which would always have been none, on top of this, it was getting the id instead of the name, so a UUID would just end up printed. The solution to this was to select the default on the service for CSV uploads, and then in the case of one offs, they would get the default if the service only had 1 sender or be sent to the set sender page to select one. From here it would be stored in the one_off session under 
```
'one_off': {
   $one_off_id: {
      'sender': {'sender_id': $ID, 'sender_name': $Name}
   }
}
```
where the ID is used to correctly send the message, and the name is used as a visual cue in the flow.


quis:
Closing this because of merge conflicts, and because we want to avoid overloading the cookie storage.

The next person who picks up this story can use this branch as a starting point or cherry-pick code from it.

PR #2026, opened 2018-04-19
Cache `GET /template…` methods in Redis 

quis:
A lot of the frequently-used pages in the admin app rely on the API to get templates. So we can cache them the same way we’re doing for service and user.

This commit adds three new caches:
- a single template version (including `None`, which is the current version)
- all the templates for a service
- all versions of a template

The first will be the most crucial for performance, but there’s not much cost to adding the other two.

---

There’s a bit of refactoring involved to keep the code easy to read, see individual commits for details.

### DIFF
--- a/app/notify_client/cache.py
+++ b/app/notify_client/cache.py
@@ -7,13 +7,14 @@ from inspect import signature
 TTL = int(timedelta(hours=24).total_seconds())
 
 
-def _get_argument(argument_name, args, kwargs, client_method):
+def _get_argument(argument_name, client_method, args, kwargs):
 
     with suppress(KeyError):
         return kwargs[argument_name]
 
     with suppress(ValueError, IndexError):
-        return args[list(signature(client_method).parameters).index(argument_name) - 1]
+        argument_index = list(signature(client_method).parameters).index(argument_name)
+        return args[argument_index - 1]  # -1 because `args` doesn’t include `self`
 
     with suppress(KeyError):
         return signature(client_method).parameters[argument_name].default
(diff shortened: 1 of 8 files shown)


app/notify_client/cache.py
```diff
@@ -1,27 +1,51 @@
...
 
-def set(prefix, key_from_args=None):
+def set(prefix, *key_from_args):
 
     def _set(client_method):
 
         @wraps(client_method)
         def new_client_method(client_instance, *args, **kwargs):
-            redis_key = _make_key(prefix, args, key_from_args)
+            redis_key = _make_key(prefix, key_from_args, locals())
```
leohemsted: i'm not a huge fan of passing around `locals()` - i think five params isn't a ridiculous amount to just define in the `_make_key` definition

app/notify_client/cache.py
```diff
@@ -1,27 +1,51 @@
...
+
+def list_of_strings(list_of_stuff):
+    return list(map(str, filter(None, list_of_stuff)))
+
+
+def _make_key(prefix, key_from_args, local_variables):
     return '-'.join(
-        [prefix] + [args[index] for index in key_from_args]
+        [
+            local_variables['prefix']
```
leohemsted: already passing in prefix

app/notify_client/service_api_client.py
```diff
@@ -139,6 +139,7 @@ def remove_user_from_service(self, service_id, user_id):
         data = _attach_current_user({})
         return self.delete(endpoint, data)
 
+    @cache.delete('templates', 'service_id')
```
leohemsted: does the key for this look like `templates-00000000-0000-0000-0000-000000000000`? do we want to make this look like `services-00000000-0000-0000-0000-000000000000-templates` or something?

alternatively if we're doing introspection, how do you feel about making the delete param take in a format string (so this might be `templates-{service_id}` and then it calls .format on that) - might help to make the resultant key a little more explicit

leohemsted: how do you feel about `service-{service_id}-templates`?

leohemsted: in fact i really like that, it can make things much more flexible and clear, especially with regard to params that aren't named with 100% clarity. eg:

```
@cache.set('template-{id_}-v{version}')
def get_template_version(id_, version):
    ...
```

quis:
@leohemsted All comments addressed in 06de94f

PR #2029, opened 2018-04-19
Tweak to make page GDPR relevant

yahoopete:


app/templates/views/terms-of-use.html
```diff
@@ -38,7 +38,7 @@ <h2 class="heading-medium">When using Notify</h2>
       <li>
         send messages that meet the GOV.UK Service Manual standards for <a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages">writing text messages and emails</a></li>
       <li>not send messages containing any personally or commercially sensitive information</li>
-      <li>check that the data you add to Notify is accurate and complies with Data Protection Act principles</li>
+      <li>check that the data you add to Notify is accurate and complies with data protection legislation</li>
```
sakisv: I'm not sure if that's correct, but I would treat the word "data" as plural here.

Doing some digging however, shows that there doesn't seem to be a consensus on how the word data is supposed to be treated [1], [2], [3] - 🤷‍♂️ 

[1]: https://www.theguardian.com/news/datablog/2010/jul/16/data-plural-singular
[2]: https://blog.oxforddictionaries.com/2014/04/29/data-media-two-tricky-latin-plurals-handle/
[3]: https://en.wikipedia.org/wiki/Data_(word)

PR #2034, opened 2018-04-24
send zendesk rather than deskpro tickets

leohemsted:
- [x] https://github.com/alphagov/notifications-utils/pull/467
- [x] https://github.com/alphagov/notifications-credentials/pull/46


app/main/views/service_settings.py
```diff
@@ -191,7 +191,7 @@ def submit_request_to_go_live(service_id):
 
     if form.validate_on_submit():
         try:
-            deskpro_client.create_ticket(
+            zendesk_client.create_ticket(
```
richardc0: Do we raise Zendesk from API automatically?

Basically is this the right place to put this or should this communicate with API?

app/__init__.py
```diff
@@ -78,7 +78,7 @@
 org_invite_api_client = OrgInviteApiClient()
 asset_fingerprinter = AssetFingerprinter()
 statsd_client = StatsdClient()
-deskpro_client = DeskproClient()
+zendesk_client = ZendeskClient()
```
richardc0: I know you have done the utils side but it would have been good to use an interface or something to abstract the implementation of the actual client. i.e. HelpDeskClient. A lot of changes here are related to the name change and hence didn't necessarily need changing - still hopefully we won't be changing again anytime soon.

tests/app/test_utils.py
```diff
@@ -18,6 +17,7 @@
     get_cdn_domain,
     get_letter_timings,
 )
+from tests.conftest import fake_uuid
```
richardc0: is this needed? Its the only change in the file?

leohemsted: all the import changes are required by the isort changes i added in https://github.com/alphagov/notifications-admin/pull/2034/commits/be038e345d9f9059042b06c7d78e734a866838af

app/main/views/service_settings.py
leohemsted: i think it should be here - the API might be having issues that they're trying to report for example, and also the API can't add any additional context - we don't gain much from moving this to the api I think

PR #2047, opened 2018-04-27
Store info about files as S3 metadata

quis:
Storing things in the session is proving buggy – we still have one user (that we know about) where the session data isn’t getting written, so they’re blocked from uploading a file.

Since all the info we’re storing in the session is about the file, it makes sense to store it with the file.

This commit only does the writing of the metadata, once we’re sure this is working we can do subsequent work to read it back, and remove reliance on the session.

### DIFF
--- a/app/main/s3_client.py
+++ b/app/main/s3_client.py
@@ -15,6 +15,17 @@ def get_s3_object(bucket_name, filename):
     return s3.Object(bucket_name, filename)
 
 
+def get_csv_location(service_id, upload_id):
+    return (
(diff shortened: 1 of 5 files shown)


app/main/s3_client.py
```diff
@@ -15,6 +15,17 @@ def get_s3_object(bucket_name, filename):
     return s3.Object(bucket_name, filename)
 
 
+def get_csv_location(service_id, upload_id):
```
allait: I feel like this should return the `"{}/{}"` formatted string

quis: Can’t because the `upload` method needs the two things separate.

PR #2048, opened 2018-04-27
Fix import order

quis:


quis:
closing in favour of https://github.com/alphagov/notifications-admin/pull/2049

PR #2054, opened 2018-04-30
Store original filename as S3 metadata

quis:
By doing this we no longer have to store it in the request args. This is the last piece of metadata that’s currently posted to the API, so removing it means that all file metadata is passed by storing it in S3.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -558,7 +558,7 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
             valid=True,
             original_file_name=unicode_truncate(
                 request.args.get('original_file_name', ''),
-                1872
+                1600,
             ),
         )
     else:
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -46,6 +46,7 @@ from app.utils import (
     get_errors_for_csv,
     get_help_argument,
     get_template,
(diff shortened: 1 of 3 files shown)


quis:
S3 has a limit of 2kb for metadata.

This means we have a limit of 1870 bytes for the filename:
```python
encoded = 'notification_count50000template_id665d26e7-ceac-4cc5-82ed-63d773d21561validTrueoriginal_file_name'.encode('utf-8')
sys.getsizeof(b)
>>> 130
2000-130
>>> 1870
```

Or, in other words, 918 characters:
```python
sys.getsizeof(('ü'*918).encode('utf-8'))
>>> 1869
```

app/main/views/send.py
```diff
@@ -555,7 +556,10 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
             notification_count=len(recipients),
             template_id=str(template_id),
             valid=True,
-            original_file_name=request.args.get('original_file_name'),
+            original_file_name=unicode_truncate(
+                request.args.get('original_file_name', ''),
+                1872
```
allait: Should we do something like 1600? I think the limit is on the length of the headers, so it would include the `x-amz-` prefixes etc. It's a pretty large limit either way.

PR #2056, opened 2018-04-30
Remove use of session storage from file upload journey

quis:
giphy-11

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -641,17 +641,6 @@ def check_messages_preview(service_id, template_id, upload_id, filetype, row_ind
 @login_required
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def start_job(service_id, upload_id):
-    try:
-        upload_data = session['file_uploads'][upload_id]
-    except KeyError:
-        current_app.logger.exception('upload_id not in session')
-        return redirect(url_for('main.choose_template', service_id=service_id), code=301)
-
-    if request.files or not upload_data.get('valid'):
-        # The csv was invalid, validate the csv again
-        return send_messages(service_id, upload_data.get('template_id'))
-
-    session['file_uploads'].pop(upload_id)
 
     job_api_client.create_job(
         upload_id,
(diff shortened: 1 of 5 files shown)


app/main/views/send.py
```diff
@@ -477,6 +479,17 @@ def send_test_preview(service_id, template_id, filetype):
...
+    with suppress(HTTPError):
+        job_api_client.get_job(service_id, upload_id)
+        # If we just return a `redirect` (302) object here, we'll get
+        # errors when we try and unpack in the check_messages route.
+        # Rasing a werkzeug.routing redirect means that doesn't happen.
+        raise RequestRedirect(url_for(
+            '.send_messages',
+            service_id=service_id,
+            template_id=template_id
+        ))
```
leohemsted: so if the job already exists, redirect to the choose template page - if it doesn't exist it trips the suppress and flow continues? that's a little confusing because of the similarity between a 301 RequestRedirect and an API HTTPError - would a try catch be more legible?

```
try:
    get_job()
except HTTPError:
    raise RequestRedirect(...)
```

quis: Your example does the opposite – redirects if the job _doesn’t_ exist.

The `try`/`except` alternative would be:
```python
try:
    get_job()
    raise RequestRedirect(...)
except HTTPError:
    pass
```

quis: I feel like the cleaner option would be `if request.status != 404: raise Redirect` but I’m not sure how to do that with our API client.

quis: Or what do you think about:
```python
    try:
        job_api_client.get_job(service_id, upload_id)
    except HTTPError:
        pass
    else:
        # If we just return a `redirect` (302) object here, we'll get
        # errors when we try and unpack in the check_messages route.
        # Rasing a werkzeug.routing redirect means that doesn't happen.
        raise RequestRedirect(url_for(
            '.send_messages',
            service_id=service_id,
            template_id=template_id
        ))
```

quis: Decided the clearest thing was to add a comment.

PR #2072, opened 2018-05-08
Check that all domains in our lists are valid

quis:
Since we have too many to add individual tests every time we add a new one now.

Easiest way I could figure to do this using our existing code was to validate the domain as part of an email address.


quis:
I think it’s still worth testing that stuff like `examplegov.uk` doesn’t get through.

PR #2074, opened 2018-05-09
Let people choose which agreement to download if signed in

quis:
If we don’t know whether people belong to a crown organisation we should give them the option of self-selecting, because they might themselves know.

This commit adds a new version of the ‘agreement’ page which gives people exactly that choice. We can’t just offer this page to anyone, but we can offer to it to anyone who’s signed in.

It also adds a new version of the agreement page for people who have already signed the agreement, because they don’t need the signing instructions.

---

This should be better because:
- it’ll mean fewer tickets for us to deal with
- people will get the agreement without having to wait for us to send it to them

# Variants

When we know if you’re crown/non-crown | When we don’t know | When you’ve already signed
---|---|---
image | image | image

### DIFF
--- a/app/main/views/agreement.py
+++ b/app/main/views/agreement.py
@@ -13,22 +13,8 @@ def agreement():
 
     agreement_info = AgreementInfo.from_current_user()
 
-    if agreement_info.crown_status is None:
-        return render_template(
-            'views/agreement-choose.html',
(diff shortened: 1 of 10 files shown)


quis:
# Changes based on Thom’s feedback

image

---

image

PR #2088, opened 2018-05-18
Add page where users can say they want branding

quis:
image

---

At the moment branding is an undocumented feature. We get a bunch of support tickets from teams asking its possible.

This commit:
- lets people know it’s possible, and what the options are
- is the first step towards making this process as self-service as possible

In some cases we will be able to infer a user’s organisation from there email address, and Google image search their logo. So the experience for them is that they press a button and government just sorts it out for you (also known as "the dream").

In other cases we will have to get back to people asking for a copy of their logo, or to find out about their service, but this is what we have to do at the moment anyway.

--- 

image


quis:
# Added GOV.UK as one of the options 

Otherwise we might be implying it’s something you shouldn’t have.

image

PR #2095, opened 2018-05-25
Bump utils and sanitise header values

klssmith:
Updated `notifications-utils`. This brings in
- the renamed character sanitization classes
- the change to allow unicode in letter addresses (this lets us delete
a test that is no longer relevant)

Also replaced non-ascii characters in headers. This fixes a bug where
non-ascii characters in a CSV filename were causing errors since the
filename is also used in the header.


klssmith:
This needs https://github.com/alphagov/notifications-utils/pull/485 to be merged first.

PR #2110, opened 2018-06-07
Show contact link on service settings page

klssmith:
The contact link for a service is now displayed on the service settings page if the service has the `upload_document` permission. When adding the `upload_document` permission, the user will be taken to the form to add a contact link if the service doesn't already have one. If the service does already have a contact link, the platform admin button will toggle the permission on or off.

#### Settings page with the changes - 
screen shot 2018-06-07 at 11 29 38

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -265,3 +265,9 @@ a.table-show-more-link {
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+.truncate-text {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -78,7 +78,7 @@
   </td>
 {%- endmacro %}
 
-{% macro text_field(text, status='') -%}
+{% macro text_field(text, status='', truncate=false) -%}
   {% call field(status=status) %}
     {% if text is iterable and text is not string %}
       <ul class="list list-bullet">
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/components/table.scss
```diff
@@ -265,3 +265,9 @@ a.table-show-more-link {
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+.truncate-text {
```
richardc0: Check with CHS on this? Maybe we should wrap because with https://www we'll only ever see the first couple of characters

app/templates/components/table.html
```diff
@@ -89,15 +89,20 @@
         {% endfor %}
       </ul>
     {% else %}
-      {{ text }}
+      {% if truncate %}
```
richardc0: Do you need the Python doesn't the css just make this work?

app/assets/stylesheets/components/table.scss
klssmith: The CSS styles here are the ones that Chris suggested - I had the link wrapped originally, but it didn't look great because the link was taking up loads of lines

app/templates/components/table.html
klssmith: We don't always want to add the `.truncate-text` class. The macro has been updated to take `truncate` as a boolean argument and the `if` statement is used to determine whether or not to add the class.

app/assets/stylesheets/components/table.scss
quis: You could add the `title` attribute to the `<div>` – then when someone hovers over the item they’ll see the whole, untruncated text as a tooltip.

klssmith: 👍 I've added the title attribute in https://github.com/alphagov/notifications-admin/blob/6fafe37dfbd12957c240e3094c1c437c2e2bd7c2/app/templates/components/table.html#L93

PR #2113, opened 2018-06-13
Let users invite people as or change people to be ‘caseworkers’

quis:
This pull request modifies the form that the user sees when inviting or editing another user, if the service has the ‘caseworking’ permission set.

This will allow creating a new type of user, one who only has the `send_messages` permission, without the `view_activity` permission.

We are doing this because we think there are a number of services with a lot of users who don’t need to see the dashboard, or the other team members, and that we can make a simpler, faster interface for these users.

--- 

Adding the pages that this new kind of user sees will come in subsequent PRs.

cwp


quis:
# Remove show/hide behaviour from permissions form 

In research we found that:
- people didn’t initially realise that the permissions expanded when the ‘admin’ option was selected
- not having all the options visible at once makes it hard to know what permissions you are (and more importantly aren’t) giving to people

This commit makes it so that:
- the options within the ‘admin’ option are always visible
- a bit of Javascript logic makes it so you can pick ‘caseworker’ and ‘manage service’, for example (by deselecting one when you pick the other)

permissions-expanded

quis:
# Style conditional radios to match the GOV.UK Design System  

Updated thickness of border and spacing to match
https://design-system.service.gov.uk/components/radios/#conditionally-revealing-content

## Notify 

screen shot 2018-06-27 at 13 05 18

## Design system

screen shot 2018-06-27 at 13 05 23

quis:
Merged in #2114

PR #2114, opened 2018-06-14
Let caseworkers send one off messages and see sent messages

quis:
_This is branched off https://github.com/alphagov/notifications-admin/pull/2113_

***

image

image

***

We think ‘caseworkers’ do two main tasks:
- send messages one at a time
- (less frequently) go back and investigate whether individual messages have been sent

So this pull request:
- makes sure users who don’t have the `view_activity` permission (ie not ‘admin’ users) can still send messages
- gives these users permission to see the page of all notifications (and makes it work across all channels)
- adds new navigation options so that users can get to the start of these two tasks

***


quis:
# Older version

This had:
- different labels for the navigation items
- some ‘quick links’ that show up after sending a message 

---

quis:
# Even older version

This was a more radically different version, with a 2 column layout.

PR #2130, opened 2018-06-26
Add new platform admin stats page

klssmith:
Added a new platform admin page, at `/plaform-admin-new` which shows different data. This no longer offers the option to filter by test-key, only by date, and also gives a more detailed break-down of the notifications and failures sent with a normal / research key. Email complaints are also included on this page.

The existing platform admin stats page `/platform-admin` has not been
deleted yet so that both pages can be compared.

Warning thresholds are currently set at 
- Greater than 0.02% for email complaints
- Greater than 3% for notifications in the state of `permanent-failure` or `temporary-failure`
- 1 or more notification with the status of `technical-failure` or `virus-scan-failed`

### New platform admin page
screen shot 2018-06-28 at 12 08 30


klssmith:
We've changed api to add in a new platform-stats blueprint, so I've updated this PR to create a platform-stats client. Also added in a commit to validate the date format.

app/assets/stylesheets/app.scss
```diff
@@ -251,3 +251,9 @@ details .arrow {
     cursor: pointer;
   }
 }
+
+.bordered-text-box {
```
tombye: Still learning the conventions here but looks like this could be a variant of big number (`.big-number-bordered`?). It would mean the padding would have to be a bit less to allow for the border-width.

tombye: The above assumes big number is just a Sass pattern. It's not, it's part of a component, along with Jinja and (I assume any JavaScript needed) so ignore this as it doesn't consider the impact of using the Sass outside of this component.

PR #2136, opened 2018-07-02
Guess people’s names when they’re invited

quis:
Most people’s names, especially in government are in the format firstname.lastname@dept.gov.uk. This means that you can pretty reliably guess that their name is ‘Firstname Lastname’.

When users are invited to Notify we know their email address already.

So this commit pre-populates the registration form based on this guess.

This is a nice little detail, but it should also stop the browser pre-filling the name field with someone’s email address (which I think happens because the browser assumes a registration form will have an
email field).

Before | After 
---|---
image | image

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -25,6 +25,7 @@ from flask import (
     url_for,
 )
 from flask_login import current_user
+from notifications_utils.formatters import make_quotes_smart
 from notifications_utils.recipients import RecipientCSV
 from notifications_utils.template import (
     EmailPreviewTemplate,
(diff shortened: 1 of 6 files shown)


app/utils.py
```diff
@@ -613,3 +614,13 @@ def __init__(self, email_address_or_domain):
...
+
+
+def guess_name_from_email_address(email_address):
+
+    possible_name = re.split(r'[\@\+]', email_address)[0]
+
+    if '.' not in possible_name:
+        return ''
+
+    return make_quotes_smart(possible_name.replace('.', ' ').title())
```
leohemsted: we should strip numbers

leohemsted: maybe strip initials if they're not first or last, else leave blank.

```
fred.e.smith -> fred smith
f.smith -> ''
f.e.smith -> ''
```

quis: @leohemsted see 660fadbce7771f2844187ffe1fbb1917a03f73ea

***

# Make the guessing a bit more sophisticated

Things we’ve noticed from looking at real data that we could handle in a smarter way:
- removing numbers (there might be a tom.smith2@dept.gov.uk if tom.smith is already taken)
- removing middle initials (again, these tend to be used for disambiguation and aren’t included when we ask people for their names)
- ignoring email addresses which only have someone’s initial, not their first name (because we can’t make a decent guess in this case)

PR #2137, opened 2018-07-02
Don’t error if a user keeps same service name

quis:
Sometimes a user will hit the green button without actually changing the service name (we saw this in a usability testing session).

Currently this gives an error saying ‘service name already in use’, which is technically true (it’s in use by the current service) but practically just annoying.

This commit adds some logic to silently ignore such requests.


quis:
Interesting that this is the 5th most common error message:

image

PR #2144, opened 2018-07-05
Pin wtforms==2.1

leohemsted:
Tests fail with `wtforms==2.2.1`.  We're not sure of the reason but on production this version is used and locally it's not, because we only require flask-wtforms, which doesn't pin its requirements at all. We should probably pin all requirements from jenkins onwards to prevent this kind of thing happening again.


leohemsted:
the snyk vuln is: https://github.com/wtforms/wtforms/issues/315

PR #2158, opened 2018-07-11
Admin can find users by email and see information about them

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/158497234
Related API pull request: https://github.com/alphagov/notifications-api/pull/1943

### Screenshots: 
screen shot 2018-07-13 at 12 09 21
  
_____
  

screen shot 2018-07-13 at 12 09 40
   
____
   
screen shot 2018-07-13 at 12 10 48


app/templates/views/find-users/user-information.html
```diff
@@ -0,0 +1,45 @@
...
+  User information for {{ user.name }}
+{% endblock %}
+
+{% block org_page_title %}
+  User information for {{ user.name }}
+{% endblock %}
+
+{% block platform_admin_content %}
+  <div class="grid-row bottom-gutter">
+    <div class="column-two-thirds">
```
quis: Can this be `column-whole` so they email address doesn’t wrap onto two lines?

app/templates/views/find-users/user-information.html
```diff
@@ -0,0 +1,45 @@
...
+      {% if not user.logged_in_at %}
+      <p>This person has never logged in</p>
+      {% else %}
+      <p>Last logged in
+        <time class="timeago" datetime="{{ user.logged_in_at }}">
+          {{ user.logged_in_at|format_delta }}
+        </time>
+      </p>
+      {% endif %}
+      <p{% if user.failed_login_count > 0 %} style="color:red;"{% endif %}>
```
quis: Can you change:
- `red` to `#b10e1e` (from https://design-system.service.gov.uk/styles/colour/#main-colours)
- the `if` statement to wrap the whole paragraph (so this line of text isn’t shown when `user.failed_login_count` is `0`)

app/templates/views/find-users/find-users-by-email.html
```diff
@@ -0,0 +1,55 @@
...
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+  </form>
+
+  {% if users_found %}
+  <nav class="browse-list">
+    <ul>
+    {% for user in users_found %}
+      <li class="browse-list-item">
+        <a href="{{url_for('.user_information', user_id=user.id)}}" class="browse-list-link">{{ user.email_address }}</a>
+        <p class="browse-list-sub-item">{{ user.name }}</p>
```
quis: `browse-list-hint` will look better here.

quis:
Oh, and it should be Find users by ‘email address’, not ‘e-mail’

app/templates/views/find-users/user-information.html
CrystalPea: 👍

app/templates/views/find-users/user-information.html
CrystalPea: 👍

app/templates/views/find-users/find-users-by-email.html
CrystalPea: It looks like this with the hint class, is it how we want it?

screen shot 2018-07-13 at 11 02 55

CrystalPea:
@quis I applied requested changes, let me know if it's alright now :)

tests/app/main/views/test_find_users.py
```diff
@@ -0,0 +1,140 @@
...
+    mocker
+):
+    mock_get_user(mocker, user=platform_admin_user)
+    client.login(platform_admin_user)
+    response = client.get(url_for('main.find_users_by_email'))
+    assert response.status_code == 200
+
+    document = html.fromstring(response.get_data(as_text=True))
+    assert document.xpath("//h1/text()[normalize-space()='Find users by email']")
+    assert len(document.xpath("//input[@type='search']")) > 0
```
leohemsted: i won't make you change all this stuff now, but for future reference, we added the `client_request` pytest fixture to get rid of some of the busywork.

It handles login, etc, asserts that the status code is 200, and returns a beautifulsoup object (which allows you to navigate html in a nicer way than lxml)
eg this test would become


```
def test_find_users_by_email_page_loads_correctly(client_request, platform_admin_user):
    document = client_request.get('main.find_users_by_email')

    assert document.h1.text.strip() == 'Find users by email'
    assert len(document.find_all('input', {'type': 'search})) > 0
```

CrystalPea: Ahh, that's neat :3

CrystalPea: I applied it to most of the tests, but I left two bottom tests as they were, because client_request's login function interfered with later mocking `user_api_client.get_user`'s response value 😬

PR #2161, opened 2018-07-12
[DON’T MERGE YET] Launch basic view

quis:
This commit makes basic view available to anyone who wants it, by linking to the page where they can switch it on and off.

***

image


quis:
We found a non-intrusive way to make ‘basic view’ available to every service, so there’s no need to have a page where services can turn it on and off any more.

PR #2162, opened 2018-07-12
Sanitise csv filenames before saving to S3

klssmith:
The CSV filename gets used as part of the metadata when uploading the file to S3. Since the S3 metadata can only contain ASCII characters, we need to sanitise the filename before uploading it to S3.


quis:
I feel like it’s happened infrequently-enough that the encoding isn’t worth it.

PR #2166, opened 2018-07-16
Revise how we talk about what basic view is

quis:
_Changes and writeup from @karlchillmaid_

# The page where you switch on the feature

Before | After 
---|---
image | image

This content aims to describe:
- the benefit of basic view – ‘make Notify quicker and simpler’
- who it benefits – ‘team members who only need to send messages’
- how it does it – ‘by hiding…’
- what it prevents users from being able to do or see –  ‘everything except…’
- what it allows users to do – ‘send messages’, [see] ‘templates, a list of sent messages’

I’m still keen to mention sent messages here, as it feels weird not to mention it at all when it’s 1 of only 2 options in Basic view. I don’t think it’s as important to mention it on the Edit team member screen.

I’ve specifically used ‘a list of sent messages’ rather than just ‘sent messages’, to make it seem less like a noun (new feature).

# The page where you choose whether someone has basic view

Before | After
---|---
image | image

Switches the focus from what you can see to what you can’t.

Aims to be consistent with both:
- the description of permissions in admin view
- the language used to describe basic view in settings

### DIFF
--- a/app/templates/views/manage-users/permissions.html
+++ b/app/templates/views/manage-users/permissions.html
@@ -8,8 +8,8 @@
 
         <div class="bottom-gutter-1-3">
           {{ radio(option, option_hints={
-              'admin': 'Show dashboard, templates and team members',
-              'caseworker': 'Hide dashboard and other options, send messages from existing templates'
+              'admin': 'Show dashboard and other options',
+              'caseworker': 'Send messages, hide dashboard and other options'
           }) }}
         </div>
         {% if option.data == 'admin' %}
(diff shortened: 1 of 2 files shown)


quis:
We need to think about the wording on this some more…

karlchillmaid:
@quis I'll be working on this on Monday.

Some interesting observations in the pop-up testing this week.

quis:
image

image

karlchillmaid:
@quis how do you feel about the hint text wrapping?

quis:
image

image

PR #2167, opened 2018-07-16
Fix displaying of services on User Information page

CrystalPea:
Services for the user did not show up earlier, because `user_api_client.get_user` only fetches service id, not the name. Hence, I added a second API call to fetch service objects for that user.

After Chris' comment I also made service name into a link and updated the screenshot.

Below a screenshot of the page with services displayed:
screen shot 2018-07-16 at 17 30 46

### DIFF
--- a/app/templates/views/find-users/user-information.html
+++ b/app/templates/views/find-users/user-information.html
@@ -18,7 +18,7 @@
         <ul>
         {% for service in services %}
           <li class="browse-list-item">
-            <p class="browse-list-hint">{{ service.name }}</p>
+            <a class="browse-list-hint" href={{url_for('.service_dashboard', service_id=service.id)}}>{{ service.name }}</a>
           </li>
         {% endfor %}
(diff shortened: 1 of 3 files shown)


app/templates/views/find-users/user-information.html
```diff
@@ -16,9 +16,9 @@ <h1 class="heading-large">
       <h2 class="heading-medium">Services</h2>
       <nav class="browse-list">
         <ul>
-        {% for service in user.services %}
+        {% for service in services %}
           <li class="browse-list-item">
-            <p class="browse-list-sub-item">{{ service.name }}</p>
+            <p class="browse-list-hint">{{ service.name }}</p>
```
quis: We should link to the service’s dashboard here (to make it easy for a platform admin user to go and see what the service is all about).

CrystalPea: Done 👍

PR #2171, opened 2018-07-17
Flexible data retention

servingUpAces:
A service can have flexible data retention for each notification type. This PR adds the pages to add, view and update.
The days of retention must be between 3 and 90 days. 
The data retention is only viewable and maintained by a platform admin user. 

- [x] https://github.com/alphagov/notifications-api/pull/1945

screen shot 2018-07-19 at 11 30 58


tests/app/main/views/test_service_settings.py
```diff
@@ -2790,3 +2793,60 @@ def test_submit_email_branding_request(
         'Thanks for your branding request. We’ll get back to you '
         'within one working day.'
     )
+
+
+def test_show_service_data_retention(
```
leohemsted: do we need test for making sure it doesn't show if you're logged in as a normal user? 

do we already have something we can easily modify in `test_service_setting_permissions.py` perhaps?

tests/app/main/views/test_service_settings.py
```diff
@@ -2790,3 +2793,60 @@ def test_submit_email_branding_request(
...
+                                                          }
+                                                    )
+    assert response.status_code == 302
+    settings_url = url_for(
+        'main.data_retention', service_id=service_one['id'], _external=True)
+    assert settings_url == response.location
+    assert mock_create_service_data_retention.called
+
+
+def test_update_service_data_retention(
```
leohemsted: can we also have a test where we give some bad data (eg a negative days_of_retention) to make sure the error path works

leohemsted: and also a test that `GET`s `main.edit_data_retention` and ensures that it populates with existing data from api

quis:
This would look better as a table with 3 columns, eg

-|-|-
---|---|---
Email | 4 days | Change
Text messages | 7 days | Change
Letters | 30 days | Change

tests/app/main/views/test_service_settings.py
servingUpAces: I believe that what tests.app.main.views.test_service_settings.test_should_show_overview does.

tests/app/main/views/test_service_settings.py
```diff
@@ -2790,3 +2793,98 @@ def test_submit_email_branding_request(
...
+    response = logged_in_platform_admin_client.post(url_for('main.edit_data_retention',
+                                                            service_id=service_one['id'],
+                                                            data_retention_id=fake_uuid),
+                                                    data={'days_of_retention': -5}
+                                                    )
+    assert response.status_code == 200
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+    error_message = page.find('span', class_='error-message').text.strip()
+    assert error_message == 'Must be between 3 and 90'
+    mock_get_service_data_retention_by_id.called
```
leohemsted: missing an `assert`

PR #2178, opened 2018-07-20
Check if any jobs exist before querying jobs

quis:
At the moment the dashboard does two API calls to find out if a service has:

1. Scheduled jobs
2. Normal jobs

API calls are slow because they are synchronous, go over the network and touch the database. We can’t cache these API calls because:
- a scheduled job could become a normal job at any time
- the statistics on a normal job are constantly updating

However there are plenty of services which don’t have any jobs, and probably never will. And finding out if a service has any jobs is reliably cacheable (because as soon as a service creates its first job
it has some jobs).

So this commit:
- refactors the way we get scheduled/normal jobs into the job_api_client to make the view a bit slimmer
- makes an additional, Redis-wrapped call to find out if any jobs exist before trying to get the jobs

This should result in a speedup on the dashboard, and can be used in the future if there’s anywhere else we want to show or hide something depending on whether a service has created any jobs (I have some ideas).

### DIFF
--- a/app/notify_client/job_api_client.py
+++ b/app/notify_client/job_api_client.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 
-from app.notify_client import NotifyAdminAPIClient, _attach_current_user
+from app.notify_client import NotifyAdminAPIClient, _attach_current_user, cache
 
 
 class JobApiClient(NotifyAdminAPIClient):
@@ -15,6 +15,8 @@ class JobApiClient(NotifyAdminAPIClient):
         'ready to send',
         'sent to dvla'
     }
+
+    NORMAL_JOB_STATUSES = JOB_STATUSES - {'scheduled', 'cancelled'}
 
     def __init__(self):
         super().__init__("a" * 73, "b")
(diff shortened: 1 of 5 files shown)


app/notify_client/job_api_client.py
```diff
@@ -16,6 +16,8 @@ class JobApiClient(NotifyAdminAPIClient):
         'sent to dvla'
     }
 
+    NORMAL_JOB_STATUSES = JOB_STATUSES - {'scheduled', 'cancelled'}
```
CrystalPea: "normal" is a word that does not really tell much, could we maybe name this constans better?

tests/app/notify_client/test_job_client.py
```diff
@@ -300,3 +308,59 @@ def test_cancel_job(mocker):
...
+):
+    mock_redis_get = mocker.patch(
+        'app.notify_client.RedisClient.get',
+        return_value=cache_value,
+    )
+
+    assert JobApiClient().has_jobs(fake_uuid) is return_value
+    mock_redis_get.assert_called_once_with(
+        'has_jobs-{}'.format(fake_uuid)
+    )
```
CrystalPea: Here I would also check that variable `mock_get`, similar to one in previous test, is not called.

app/notify_client/job_api_client.py
quis: Good point, see eb8e433

tests/app/notify_client/test_job_client.py
quis: Added in 0906eb0

CrystalPea: Sweet 👍

quis:
# Before

image

# After 

image

# Improvement

Roughly 25% faster (134ms)

CrystalPea:
niiice 👍💯

PR #2179, opened 2018-07-20
Updated the service and org counts

yahoopete:


yahoopete:
https://media.giphy.com/media/3o7aDfuOPVroMTRjig/200w_d.gif

PR #2182, opened 2018-07-24
Remove view displaying data form monthly_billing table

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/159164066
connected API pull request: https://github.com/alphagov/notifications-api/pull/1981

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -141,37 +141,6 @@ def template_usage(service_id):
     )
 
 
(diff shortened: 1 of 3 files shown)


klssmith:
I think you can also delete `get_billable_units` and `get_service_usage` from the `BillingAPIClient` - it doesn't look like either of those functions get used anywhere apart from in the code that's being deleted.

PR #2192, opened 2018-07-31
[WIP] Show uploaded files in basic view 

quis:
We have teams with caseworkers uploading files on a regular basic. Since this is the only interaction they have with Notify, basic view would be suitable for them, if only it showed the jobs they’ve sent.

# A service with no uploaded files

image

# With some historic files 

image

# With a scheduled file

image


yahoopete:
:+1:

PR #2209, opened 2018-08-07
Preview the email branding before setting it

tombye:
Adds a preview to the 'set email branding' page for platform admins.

https://www.pivotaltracker.com/story/show/159361200

Note: this adds an extra `<source>` of `'self'` to the `frame-src` directive in our `Content-Security-Policy` header to allow the /_email page to be iframe'd.

With JS it shows a sample email in an iframe above the selection made:

set_email_branding_with_preview

Without JS it adds an extra preview step to the process, showing the branding in the same iframe:

**Set email branding**

set_email_branding_no_js

...which goes to:

**Preview email branding**

preview_email_branding


quis:
Design-wise this works great 👍 

For some reason the `iframe` doesn’t span the full width of the page for me locally in Chrome though:
image

tombye:
@quis looks like the CSS for the iframe got lost (probably in a rebase). Now added, in https://github.com/alphagov/notifications-admin/pull/2209/commits/5afb355ed7e024b1c740ef3f353d68ef1bf80573

PR #2215, opened 2018-08-09
bump utils to fix slow notification rendering

leohemsted:


quis:
https://github.com/alphagov/notifications-utils/pull/511

Makefile
```diff
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL := help
 SHELL := /bin/bash
-DATE = $(shell date +%Y-%m-%d:%H:%M:%S)
+DATE = $(shell date +%Y-%m-%dT%H:%M:%S)
```
tombye: Can this be split off into its own commit?

PR #2219, opened 2018-08-10
Allow service contact details to be a phone number, email or url

klssmith:
Service contact details are needed if the upload document permission is enabled - this used to be a link but services can now choose to use a link, email address or phone number. The form to add or change service contact details now gives these options and validates the data according to the type of contact details provided.

When validating phone numbers we can't use the existing validation because we want to allow landlines too, so there is a basic check that the phone number is the right length and doesn't include certain characters.

[Pivotal story](https://www.pivotaltracker.com/story/show/159490824)

### Updated settings page
screen shot 2018-08-10 at 14 14 10

### Updated form
screen shot 2018-08-10 at 14 14 30

### DIFF
--- a/app/templates/views/service-settings/contact_link.html
+++ b/app/templates/views/service-settings/contact_link.html
@@ -1,24 +1,46 @@
 {% extends "withnav_template.html" %}
 {% from "components/textbox.html" import textbox %}
 {% from "components/page-footer.html" import page_footer %}
+{% from "components/radios.html" import radio, radios_wrapper %}
 
 {% block service_page_title %}
-  {{ 'Change link on' if 'upload_document' in current_service.permissions else 'Add link for' }} ‘Download your document’ page
+  {{ 'Change' if 'upload_document' in current_service.permissions else 'Add' }} contact details for ‘Download your document’ page
 {% endblock %}
 
 {% block maincolumn_content %}
   <div class="grid-row">
     <div class="column-five-sixths">
       <h1 class="heading-large">
-        {{ 'Change link on' if 'upload_document' in current_service.permissions else 'Add link for' }} ‘Download your document’ page
+        {{ 'Change' if 'upload_document' in current_service.permissions else 'Add' }} contact details for ‘Download your document’ page
       </h1>
       <p>
-        When you send users a document to download, you need to include a link to your service
+        When you send users a document to download, you need to include the contact details for your service
         on the download page. This is so users can contact you if there’s a problem (for example,
         if the link to download the document has expired).
       </p>
-      <form method="post">
-        {{ textbox(form.url, width='1-1') }}
+      <form method="post" novalidate>
+
+        {% call radios_wrapper(form.contact_details_type, hide_legend=true) %}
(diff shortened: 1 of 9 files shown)


app/templates/views/service-settings/contact_link.html
```diff
@@ -1,24 +1,46 @@
...
         on the download page. This is so users can contact you if there’s a problem (for example,
         if the link to download the document has expired).
       </p>
-      <form method="post">
-        {{ textbox(form.url, width='1-1') }}
+      <form method="post" novalidate>
+
+        {% call radios_wrapper(form.contact_details_type, hide_legend=true) %}
+          {% for option in form.contact_details_type %}
+            {% if option.data == 'url' %}
```
CrystalPea: This could optionally be refactored to:
```
              {% info_type = option.data.replace('_', '-') %}
              {{ radio(option, data_target="{}-type".format(info_type)) }}
              <div class="panel panel-border-narrow js-hidden" id="{}-type".format(info_type)>
                {{ textbox(form.url, label=' ', width='1-1') }}
              </div>
```

app/templates/views/service-settings/contact_link.html
```diff
@@ -1,24 +1,46 @@
...
+              {{ radio(option, data_target="url-type") }}
+              <div class="panel panel-border-narrow js-hidden" id="url-type">
+                {{ textbox(form.url, label=' ', width='1-1') }}
```
leohemsted: `textbox(getattr(form, option.data), label=' ', width='1-1')`

app/templates/views/service-settings/contact_link.html
leohemsted: `textbox(getattr(form, option.data), label=' ', width='1-1')`

klssmith: It would be nice to refactor this, but when I try the JavaScript stops working - when you select a radio button the relevant text box should appear but it doesn't.

This is what I've tried (you can't use `getattr` in jinja, but from reading the docs the jinja `attr` filter does the same thing) - 
```python
{% set details_type = option.data.replace('_', '-') %}
{{ radio(option, data_target="{}-type".format(details_type)) }}
<div class="panel panel-border-narrow js-hidden" id="{}-type".format(details_type)>
  {{ textbox(form|attr(option.data), label=' ', width='1-1') }}
</div>
```

CrystalPea: Wow, that's curious 😮

leohemsted: that's weird - did you have a look at the resulting HTML? Because the templates are all serverside and calculated well before any js is run, so if the HTML is the same, the js will (_should_) function exactly the same

klssmith: Turns out I was forgetting that you can't use Python functions such as `format` in jinja - now refactored 😅

CrystalPea: Woo, go you! 🎉  
And good to know about `.format()` not working in jinja, thanks ☺️

PR #2222, opened 2018-08-13
Allow service contact_link or contact_details naming

klssmith:
The Service `contact_link` property in Notifications API will be renamed
`contact_details`. To ensure that there is no downtime or errors while the
name is being changed, this commit takes both the old and new name into
account by
- checking for both names where we were previously just checking for
`contact_link`
- sending values for both `contact_link` and `contact_details` to the API

Once the name has been changed in API, we can replace `contact_link` with
`contact_details` instead of keeping both names.


klssmith:
Closing since we've decided to keep the original name for now.

PR #2223, opened 2018-08-13
 Replace all references to contact_link with contact_details

klssmith:
Now that `contact_link` has been renamed to `contact_details` in Notifications API, we can update the name everywhere in the code.

_NOTE_ - branched off a branch - only the latest commit has been added in this PR


klssmith:
Closing since we've decided to keep the original name for now.

PR #2230, opened 2018-08-14
Show the 'name' of the email branding in the adding and editing screens

tombye:
The API was updated to give email_branding entries an extra field of 'name' (see [related story](https://www.pivotaltracker.com/story/show/159300417)).

This plugs that work into the admin interface so you can select email_branding by a name rather than the logo or banner text.

https://www.pivotaltracker.com/story/show/159624563

## Service settings email branding

Before | After
---|---
service-settings-set-email-branding_before | service-settings-set-email-branding_after

## Platform admin select email branding

Before | After
---|---
platform-admin-select-email-branding_before | platform-admin-select-email-branding_after

## Platform admin create new branding/edit existing branding

Before | After
---|---
platform-admin-create-email-branding_before | platform-admin-create-email-branding_after

## Things to note

The platform admin page for creating/editing brands has the 'name' field above the logo field. This is because the logo has its own form so can't* be embedded into the one the other fields share.

*well it could be, with some JS magic but that would add work and make it harder to understand.

The radios overlap on some of the pages when the brands they represent don't have anything in their 'name' fields. This will go away once they do.

### DIFF
--- a/tests/app/main/views/test_email_branding.py
+++ b/tests/app/main/views/test_email_branding.py
@@ -45,7 +45,7 @@ def test_edit_email_branding_shows_the_correct_branding_info(
 
     assert page.select_one('#logo-img > img')['src'].endswith('/example.png')
     assert page.select_one('#name').attrs.get('value') == 'Organisation name'
-    assert page.select_one('#text').attrs.get('value') == 'Organisation name'
+    assert page.select_one('#text').attrs.get('value') == 'Organisation text'
     assert page.select_one('#colour').attrs.get('value') == '#f00'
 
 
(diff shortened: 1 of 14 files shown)


tests/app/main/views/test_email_branding.py
```diff
@@ -47,6 +45,7 @@ def test_edit_email_branding_shows_the_correct_branding_info(
 
     assert page.select_one('#logo-img > img')['src'].endswith('/example.png')
     assert page.select_one('#name').attrs.get('value') == 'Organisation name'
+    assert page.select_one('#text').attrs.get('value') == 'Organisation name'
```
servingUpAces: How do you know you aren't get name for text? They are the same value

PR #2235, opened 2018-08-15
Sort branding radios alphabetically

tombye:
On both pages where you can choose a branding style from a list of radios.

https://www.pivotaltracker.com/story/show/159623862

### DIFF
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -29,7 +29,8 @@ def email_branding():
     brandings = email_branding_client.get_all_email_branding()
 
     form = ServiceSelectEmailBranding()
-    form.email_branding.choices = get_branding_as_value_and_label(brandings) + [('None', 'Create a new email branding')]
+    email_brandings = sorted(get_branding_as_value_and_label(brandings), key=lambda tup: tup[1].lower())
+    form.email_branding.choices = email_brandings + [('None', 'Create a new email branding')]
 
     if form.validate_on_submit():
         if form.email_branding.data != 'None':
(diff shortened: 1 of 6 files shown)


app/main/views/email_branding.py
```diff
@@ -29,7 +29,8 @@ def email_branding():
     brandings = email_branding_client.get_all_email_branding()
 
     form = ServiceSelectEmailBranding()
-    form.email_branding.choices = get_branding_as_value_and_label(brandings) + [('None', 'Create a new email branding')]
+    email_brandings = sorted(get_branding_as_value_and_label(brandings), key=lambda tup: tup[1].lower())
+    form.email_branding.choices = email_brandings + [('None', 'Create a new email branding')]
```
leohemsted: i'd be inclined to just update the `get_all_email_branding` function to put the sort there.

app/main/views/email_branding.py
```diff
@@ -29,7 +29,8 @@ def email_branding():
     brandings = email_branding_client.get_all_email_branding()
 
     form = ServiceSelectEmailBranding()
-    form.email_branding.choices = get_branding_as_value_and_label(brandings) + [('None', 'Create a new email branding')]
+    email_brandings = sorted(get_branding_as_value_and_label(brandings), key=lambda tup: tup[1].lower())
```
leohemsted: i'd be inclined to just update the `get_all_email_branding` function to put the sort there.

tombye: Makes sense. How's it look now I fixed it (in f85a9306)?

PR #2241, opened 2018-08-20
New columns for email branding

servingUpAces:
Added banner_colour, single_id_colour and domain to email branding view. Now able to set the fields.

However, the new fields are not being used yet.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -731,8 +731,23 @@ class ServiceUpdateEmailBranding(StripWhitespaceForm):
 
     name = StringField('Name of brand')
     text = StringField('Text')
+    domain = StringField('Domain')
     colour = StringField(
         'Colour',
+        render_kw={'onchange': 'update_colour(this)'},
+        validators=[
+            Regexp(regex="^$|^#(?:[0-9a-fA-F]{3}){1,2}$", message='Must be a valid color hex code')
+        ]
+    )
+    banner_colour = StringField(
+        'Banner colour',
+        render_kw={'onchange': 'update_colour(this)'},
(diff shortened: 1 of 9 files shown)


app/main/forms.py
```diff
@@ -731,22 +731,23 @@ class ServiceUpdateEmailBranding(StripWhitespaceForm):
...
-
-class ServiceCreateEmailBranding(StripWhitespaceForm):
-
-    name = StringField('Name of brand')
-    text = StringField('Text')
-    colour = StringField(
-        'Colour',
+    banner_colour = StringField(
+        'Banner colour',
+        render_kw={'onchange': 'update_colour(this)'},
```
CrystalPea: `render_kw={'onchange': 'update_colour(this)'},`

Wow, what's happening here? Is this like some javascripty thing that will update branding colour on the page when it is changed by the user?

Sorry if it's a silly question, I just never saw anything like this 😳

app/main/views/index.py
```diff
@@ -78,6 +78,13 @@ def design_content():
     return render_template('views/design-patterns-content-guidance.html')
 
 
+def _set_colour(branding_style, email_branding):
+    if branding_style in ['both', 'org']:
```
CrystalPea: is `'both'` like both logo and banner? 

And if it is both logo and banner, then why do we choose `''single_id_colour''` over `'banner_colour'`?

app/main/forms.py
quis: This icky Javascript will go away in https://github.com/alphagov/notifications-admin/pull/2242

PR #2243, opened 2018-08-20
Rename 'reply-to address'

karlchillmaid:
We currently use the label `reply to address` (no hyphen) in Settings.

Suggest we use `reply-to address` (with a hyphen) instead. 

Reasons:

- this is consistent with the way Gmail and other email providers usually label it
- 'reply to address' (no hyphen) looks confusing when it appears within a sentence

We may need to update our documentation as a result of this change.

### DIFF
--- a/app/templates/views/service-settings/data-retention.html
+++ b/app/templates/views/service-settings/data-retention.html
@@ -4,7 +4,7 @@
 {% from "components/table.html" import mapping_table, row_group, row, text_field, edit_field, field, boolean_field, list_table with context %}
 
 {% block service_page_title %}
-  Email reply to addresses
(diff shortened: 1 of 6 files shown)


app/templates/views/service-settings/data-retention.html
```diff
@@ -4,7 +4,7 @@
 {% from "components/table.html" import mapping_table, row_group, row, text_field, edit_field, field, boolean_field, list_table with context %}
 
 {% block service_page_title %}
-  Email reply to addresses
+  Email reply-to addresses
```
quis: This one should say ‘Data retention’ instead (copy/paste error I suspect).

PR #2244, opened 2018-08-21
Tell people they can have their logo on letters

quis:
image

***

image

***

image

***

At the moment there’s nothing in Notify that says the logo on letter templates can be changed from ‘HM Government’. While some people guess that it’s possible, or are motivated enough to enquire, others might be assuming that the branding can’t be changed (and thus deciding Notify letters aren’t for them).

In the future we should make this process slicker (eg with a file upload) but the easiest, quickest improvement to make for now is:
- let people know that the branding can be changed
- direct them to support if they do want to change it


app/templates/views/service-settings/request-letter-branding.html
```diff
@@ -0,0 +1,29 @@
...
+  <h1 class="heading-large">Letter branding</h1>
+  <div class="grid-row">
+    <div class="column-three-quarters">
+      <p>
+        Your letters have the {{ letter_branding }} logo.
+      </p>
+      <p>
+        If you want a different logo then
+        <a href="{{ url_for('main.feedback', ticket_type='ask-question-give-feedback', body='letter-branding') }}">contact support</a>
+        and we’ll help you set one up.
```
karlchillmaid: We could lose "and we'll help you set one up"?

Alternative suggestion:
`If you want to use a different logo then contact support.`

quis: image

PR #2248, opened 2018-08-22
Add filter to branding selection

tombye:
We've textboxes for filtering lists of content in our pages that are too long to view on most screens. This adds some for the pages that list the brands used for emails.

https://www.pivotaltracker.com/story/show/159623907

|Before|After|
|---|---|
|email_branding_select_to_edit_before|email_branding_select_to_edit_after|

I also put one in the page for choosing the combination of brand type and style because it has the same need:

|Before|After|
|---|---|
|email_branding_choose_branding_before|email_branding_choose_branding_after|

As with the existing ones, these only show if you have over a set number of items onscreen.

### Other changes

The textboxes were already in several other pages so I made it into a component to help ensure we use it consistently.

I also added tests for whether it shows in the pages where it already exists.


quis:
Hope you don’t mind, tacked on a few tweaks to make it a bit more usable:
image

app/main/views/email_branding.py
```diff
@@ -30,13 +30,10 @@ def email_branding():
...
     email_brandings = get_branding_as_value_and_label(brandings)
-    form.email_branding.choices = email_brandings + [('None', 'Create a new email branding')]
+    form.email_branding.choices = email_brandings
 
     if form.validate_on_submit():
-        if form.email_branding.data != 'None':
-            return redirect(url_for('.update_email_branding', branding_id=form.email_branding.data))
-        else:
-            return redirect(url_for('.create_email_branding'))
+        return redirect(url_for('.update_email_branding', branding_id=form.email_branding.data))
```
tombye: Can you explain about removing the check for `None` here? I'm not up-to-speed on the history of having it so would be useful to know why it's not needed any more.

quis: It’s checking for the `None` that’s appended to the list of brandings on this line: https://github.com/alphagov/notifications-admin/pull/2248/files#diff-d4aa3b36c51ad33984973afe6051632eL32

PR #2249, opened 2018-08-23
Add brand type to email branding.

servingUpAces:
Removed banner_colour and single_id_colour.
We only really need one colour now.


tests/app/main/views/test_email_branding.py
```diff
@@ -294,7 +284,8 @@ def test_temp_logo_is_shown_after_uploading_logo(
 
     response = logged_in_platform_admin_client.post(
         url_for('main.create_email_branding'),
-        data={'file': (BytesIO(''.encode('utf-8')), 'test.png')},
+        data={'file': (BytesIO(''.encode('utf-8')), 'test.png'),
```
tombye: I don't think the `brand_type` param is needed here. The controller will issue a redirect if `file` is in the form data so it never gets to the stage of needing/using any other params:

https://github.com/alphagov/notifications-admin/blob/master/app/main/views/email_branding.py#L99

servingUpAces: Can you take another look @tombye

PR #2250, opened 2018-08-23
Try to get more users to do the things we require before going live

quis:
Before | After 
---|---
image | image 

# Use the task list pattern on the request to go live page

We’ve found a significant property of users (about 25%) who request to go live aren’t completing all the items on the checklist.

In 1 of 6 (17%) of the usability testing sessions we did on this process we saw someone skip straight past the checklist page because of big green button syndrome. While 1 in 6 people would normally be a small number<sup>1</sup> in the context of a usability testing session, it’s enough to cause a big workload for our team (assuming it is the sole cause of people not completing the items on the checklist).

The initial reason for using the tick cross pattern for the checklist was:
- it was coherent with the rest of Notify
- the task list pattern didn’t have a way of showing that something still needed doing – it put more visual emphasis on the things the user had already done

There’s been some interesting discussion on the GOV.UK Design System backlog about users failing to complete items in the task list. A few people have tried different patterns for communicating that items in the task list still need ‘completing’.

So this commit:
- adds a task list pattern
- uses the task list pattern for the request to go live checklist

The task list is adapted from the one in the design system in that:
- the ‘completed’ label has a black, not blue background (because Notify often uses blocks of blue to indicate something that’s clickable)
- it adds an explicit ‘not complete’ label which is visually not filled in (sort of how ticked/unticket radio buttons work)

# Check text message sender before going live

We often check that a service has an appropriate text message sender as a condition of them going live. We don’t mention this anywhere.

The services for whom GOVUK is definitely not an appropriate sender are those in local government. As we have more of these teams starting to use Notify, we should streamline the process by making this check automated.

This commit adds that check, for teams who:
- have text message templates
- have self-declared as NHS or local government

***

1. With the caveat that looking only at task completion and quantifying qualitative research can both be problematic and the intention here is to show that the numbers are close enough to say that they could be symptoms of the same problem. [Leisa Reichelt’s Mind the Product talk](https://vimeo.com/284015765) is good on this stuff.

### DIFF
--- a/app/assets/stylesheets/components/task-list.scss
+++ b/app/assets/stylesheets/components/task-list.scss
@@ -1,0 +1,41 @@
+$indicator-colour: $black;
+
+%task-list-indicator {
+  @include bold-16;
+  display: inline-block;
+  padding: 3px 8px 1px 8px;
+  position: absolute;
+  right: 0;
+  top: 50%;
+  margin-top: -15px;
+  border: 2px solid $indicator-colour;
+}
+
+.task-list {
+
+  border-bottom: 1px solid $border-colour;
+  margin: $gutter 0;
+
+  &-item {
+    border-top: 1px solid $border-colour;
+    padding: 15px 0;
+    padding-right: 20%;
+    position: relative;
+  }
+
+  &-indicator-completed {
+    @extend %task-list-indicator;
+    background-color: $indicator-colour;
+    color: $grey-4;
+    // Just a pinch of letter spacing to make reversed-out text a bit
+    // easier to read
+    letter-spacing: 0.02em;
+  }
+
+  &-indicator-not-completed {
+    @extend %task-list-indicator;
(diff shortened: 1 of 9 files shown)


karlchillmaid:
**Content suggestions:**

H1: `Before you request to go live`

We probably don't need to use a lead-in for a task list / table.

Tasks:
`Add a team member who can manage settings, team and usage`
`Add templates that show the kind of messages you’ll send`
`Add an email reply-to address`
`Change your text message sender name`

☝️ I've tried to keep these short so they don't wrap, but we can add to or edit them if necessary.

I also suggest we move the line about making the service live within one day to the [next screen](https://user-images.githubusercontent.com/28294225/44736342-dfe00280-aae6-11e8-81ad-f619f4d36c19.png).

quis:
Updated first page:
image

…and second page: 
image

PR #2262, opened 2018-08-29
Remove branding type from set email branding page

tombye:
Updating the 'Set email branding' page, from service settings, so it uses the new email_branding model added in https://github.com/alphagov/notifications-admin/pull/2249.

For a task in this story: https://www.pivotaltracker.com/story/show/159986276

### Set email branding page

|Before|After|
|---|---|
|old_set_email_branding|newer_set_email_branding|

Includes changes to the email preview page (iframe'd into the 'Set email branding' page) and the brand request page. They both needed changes so they accessed the brand type from the new email_branding model.

### DIFF
--- a/app/assets/javascripts/emailPreviewPane.js
+++ b/app/assets/javascripts/emailPreviewPane.js
@@ -5,18 +5,16 @@
   const root = this,
       $ = this.jQuery;
 
-  let branding_type = $('.multiple-choice input[name="branding_type"]:checked');
   let branding_style = $('.multiple-choice input[name="branding_style"]:checked');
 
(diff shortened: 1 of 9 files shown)


yahoopete:
Does none mean GOV.UK only branding?  If so we should change the label to be clear about that....

tombye:
@yahoopete we had a good look into that and none means it removes email branding from your service. That being the case, me and @quis reckon we show GOV.UK as what you get when you don't have a brand yet elsewhere (the branding request page) so it makes sense to do it here.

TLDR: Yep, now updated with a label of 'GOV.UK' (see new screenshot in the description).

quis:
I get this error when I try to set a branding:

image

PR #2268, opened 2018-08-30
Make the data we get from the go live requests more useful

quis:
# Request to go live page

Before | After 
---|---
image | image

# Ticket we receive 

Before | After 
---|---
image | image

# The magic

copyp


app/templates/views/service-settings/submit-request-to-go-live.html
```diff
@@ -13,32 +13,22 @@
...
-          form.channel_email,
-          form.channel_sms,
-          form.channel_letter
-        ]) }}
+      <form method="post">
         <div class="form-group">
-          {{ textbox(form.start_date, width='1-1') }}
-          {{ textbox(form.start_volume, width='1-1', hint='For example, ‘1,000 per month’.') }}
-          {{ textbox(form.peak_volume, width='1-1', hint='For example, ‘Messages will increase to 20,000 per month in January’.') }}
+          {{ textbox(form.volume_email, width='1-2', hint='For example, 1,000,000') }}
```
karlchillmaid: Are the numbers in these hints realistic? If it doesn't really matter we should keep them all the same.

quis: Pete’s happy with those examples. Can always tweak later if we think of something better.

app/main/views/service_settings.py
```diff
@@ -214,32 +216,42 @@ def submit_request_to_go_live(service_id):
         zendesk_client.create_ticket(
             subject='Request to go live - {}'.format(current_service.name),
             message=(
-                'Service: {}\n'
-                '{}\n'
+                'Service: {service_name}\n'
```
CrystalPea: I really like how you used named arguments here, less chance of mixing something up <3

feels_good_man

PR #2282, opened 2018-09-03
Look up email branding from user’s domain

quis:
Uses this:

image

To show this:
image

Then when a user creates a service, this happens automatically:

image

So when they send an email:

hmr

---

This also works (differently in the background but the same way for the user) for users who choose ‘NHS’ as the answer to the ‘Who runs this service’ question. In other words those services will automatically get NHS branding.

---

Stories:
- General https://www.pivotaltracker.com/story/show/159623304
- NHS-specific https://www.pivotaltracker.com/story/show/159623188

### DIFF
--- a/app/domains.yml
+++ b/app/domains.yml
@@ -65,7 +65,8 @@ innovateuk.gov.uk:
   owner: Innovate UK
   agreement_signed: true
 ncsc.gov.uk: gchq.gov.uk
-nhs.net:
+nhs.net: nhs.uk
+nhs.uk:
   owner: NHS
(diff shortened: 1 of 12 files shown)


quis:
**Sector** trumps **domain** trumps **GOV.UK**.

So:

Sector| Email address| Branding
---|---|---
NHS | user@nhs.uk | NHS
Central | user@nhs.uk | NHS
NHS | user@gov.uk | NHS
Central | user@gov.uk | GOV.UK

yahoopete:
Oh nice!

PR #2287, opened 2018-09-05
Remove old API key and service ID combo

quis:
This PR attempts to resurrect https://github.com/alphagov/notifications-admin/pull/1145

---

Once all our users have upgraded to the latest clients they won’t need this. The latest clients only use the combined key and service ID.

Can we safely remove it?


servingUpAces:
@quis Should we give this a go? I think it's ok

PR #2311, opened 2018-09-18
Add letter class row to settings page, visible to platform admin only

klssmith:
Added a new row to the settings table, 'Post class', which shows the
default postage of a service and is only visible to Platform Admin.

Also added a new page to enable Platform Admin users to change the
default postage for a service - this only has two options at the
moment, 1st class only and 2nd class only.

[Pivotal story](https://www.pivotaltracker.com/story/show/160231362)

### Settings page as seen by platform admin user
screen shot 2018-09-19 at 09 19 52

### Change postage class page
screen shot 2018-09-19 at 09 19 37


quis:
Use ‘to’ in time ranges per https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

So _Monday to Saturday_ not _Monday - Saturday_.

app/templates/views/service-settings/set-postage.html
```diff
@@ -0,0 +1,39 @@
+{% extends "withnav_template.html" %}
+{% from "components/radios.html" import radios %}
+{% from "components/page-footer.html" import page_footer %}
+
+{% block service_page_title %}
+  Postage classes
```
tombye: This should be the same as the h1 so either change the title or the h1. Not sure it matters which.

app/templates/views/service-settings/set-postage.html
```diff
@@ -0,0 +1,39 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  <div class="grid-row">
+    <div class="column-five-sixths">
+      <h1 class="heading-large">Postage</h1>
+      <p>You can send letters by first or second class post.</p>
+      <p>
+        See a list of <a href="{{ url_for(".pricing", _anchor="letters") }}">postage prices</a>.
```
tombye: Think the double quotes used in `url_for` should be single to match style used in other templates.

app/templates/views/service-settings/set-postage.html
klssmith: 👍 I've changed the title to 'Postage'.

app/templates/views/service-settings/set-postage.html
klssmith: Now fixed. We were also using the wrong type of single quotes in the text (the straight type not the curvy type) so I've changed that too.

PR #2314, opened 2018-09-19
Don’t trigger HTML5 validation on any form

quis:
~~We should probably have a reusable thing for this…~~ I made this into a reusable component in d9da219


quis:
Really went for it and made this change across all forms in d9da219

PR #2324, opened 2018-09-24
Display postage on letter notification page

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/story/show/160232054

### Screenshot: 
screen shot 2018-09-25 at 11 09 20


karlchillmaid:
@quis We should always write it as 'first class' or 'second class', 'post' is optional depending on the context.

This is the content I had for this page:

`Sent by second class post on 14 August at 1:58pm`
`Estimated delivery date: 17 August`

@CrystalPea I'll share the document with both of you.

PR #2327, opened 2018-09-25
Add line about the data sharing and financial agreement to the checklist

quis:
Before | After 
---|---
image | image

2/3 of our incomplete requests to go live are incomplete because the Data Sharing and Financial Agreement isn’t signed.

We reckon we can be pushier about this by saying it’s ‘incomplete’ where we know the agreement is signed.

Where the agreement is signed we should confirm this, rather than make the line disappear. This is so it makes more sense to someone who sees it as ‘incomplete’, signs it, then comes back to the page.

If we don’t know whether or not the agreement is signed we should wait until someone has got in touch with us by requesting to go live to figure it out. So that’s why we’re not showing that line at all.

***

Needs @karlchillmaid’s input on the content before merging.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -67,8 +67,7 @@ from app.notify_client.billing_api_client import BillingAPIClient
 from app.notify_client.complaint_api_client import ComplaintApiClient
 from app.notify_client.platform_stats_api_client import PlatformStatsAPIClient
 from app.commands import setup_commands
-from app.utils import get_cdn_domain
-from app.utils import gmt_timezones
+from app.utils import get_cdn_domain, gmt_timezones, id_safe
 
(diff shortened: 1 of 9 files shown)


quis:
# Make wording around agreement more straightforward 

image

We think that the next page explains the nuance around service vs organisation sufficiently well:

image

quis:
# Remove ‘Download the’ call to action 

Users come to this page from various places, including the new 'Sign the…' link on the request to go live page.

Of these users:
- some won't have signed it
- some will have signed it
- some will see that it's complete and wonder why, as they haven’t  actually done anything

So it’s more appropriate for the title of this page to be descriptive,
rather than an action.

Before | After 
---|---
image | image

PR #2349, opened 2018-10-03
Display postage on usage page

klssmith:
The usage page shows the monthly letter spending broken down by price, but this change also adds postage details to the page.

Also updated some of mocks being used in the tests, since these were not mocking the correct fields from notifications-api.

[Pivotal story](https://www.pivotaltracker.com/story/show/160660570)

#### Usage page before
screen shot 2018-10-03 at 12 05 09

#### Usage page after
screen shot 2018-10-03 at 15 21 22


klssmith:
Needs https://github.com/alphagov/notifications-api/pull/2130 to be merged first

quis:
Summary of discussion:
- sort should be type, class, price ascending (this is already done but above screenshot is out of date)
- ‘1 letters’ shouldn’t be pluralised

PR #2351, opened 2018-10-04
Preview the validation of precompiled letters

CrystalPea:
### Work done:
- add validation preview view to platform admin
- allow admin users to upload a pdf file to check if the contents for within printable areas
- do antivirus check on the files
- send a POST request to template preview to do the boundaries check

Pivotal ticket: https://www.pivotaltracker.com/story/show/159697071

### Validation passed:
screen shot 2018-10-05 at 11 54 01

### Validation failed:
screen shot 2018-10-05 at 11 54 27


app/main/views/platform_admin.py
```diff
@@ -240,6 +244,42 @@ def platform_admin_returned_letters():
...
+
+        if not virus_free:
+            flash("Document didn't pass the virus scan")
+            abort(400)
+
+        try:
+            response = validate_letter(pdf_file)
+            if response.status_code == 200:
+                pages, message, result = response.json()["pages"], response.json()["message"], response.json()["result"]
+        except HTTPError as error:
```
leohemsted: this is catching a `notifications_python_client.errors.HTTPError` - however, that won't be thrown, because we're using requests, rather than our api clients.

You should be calling `response.raise_for_status()`, and then catching a `requests.RequestException`

app/main/views/platform_admin.py
```diff
@@ -240,6 +244,42 @@ def platform_admin_returned_letters():
...
+    message, pages, result = None, [], None
+    form = PDFUploadForm()
+
+    if form.validate_on_submit():
+        pdf_file = form.file.data
+
+        try:
+            virus_free = antivirus_client.scan(pdf_file)
+        except AntivirusError:
+            flash("Antivirus API error")
```
leohemsted: The user doesn't need to know this detail - I wouldn't bother with this `flash`. They'll still see the "something went wrong" page.

app/main/views/platform_admin.py
```diff
@@ -240,6 +244,42 @@ def platform_admin_returned_letters():
...
+            abort(503)
```
leohemsted: `abort(503)` won't work - it then tries to find a `503.html` error file, which doesn't exist. This'll obfuscate the original exception when we come to look at the logs - we can just re-raise the `AntivirusError`, which'll be caught by the `handle_bad_request` error handler in `app/__init__.py` (it'll return a 500 and the something went wrong page).

In combination with the line above about not needing the flash, we don't need the try catch at all.

app/main/views/platform_admin.py
```diff
@@ -240,6 +244,42 @@ def platform_admin_returned_letters():
...
+
+        if not virus_free:
+            flash("Document didn't pass the virus scan")
+            abort(400)
```
leohemsted: this `abort(400)` (and the one below for the HTTPError) will return a generic error page - with no side nav, way to get back easily, etc.

I think we can call the `render_template` function, but with `pages = [], message = "Document didn't pass the virus scan", result = "error"`.

then we can return a tuple as in `return render_template(...), status_code` to return a non-200.

I think this'll look nicer and give users a chance to upload the correct file etc.

app/main/views/platform_admin.py
```diff
@@ -240,6 +244,42 @@ def platform_admin_returned_letters():
...
+
+        try:
+            response = validate_letter(pdf_file)
+            if response.status_code == 200:
+                pages, message, result = response.json()["pages"], response.json()["message"], response.json()["result"]
+        except HTTPError as error:
+            if error.status_code == 400:
+                flash("Something was wrong with the file you tried to upload. Please upload a valid PDF file.")
```
leohemsted: should we be returning a 400 in this case, same as the virus scan error?

PR #2364, opened 2018-10-09
Update pricing for 1 October onwards - adding first class rates

yahoopete:


yahoopete:
I hear a distant LGTM from @quis so I'm merging....

PR #2374, opened 2018-10-11
Show if a letter has been cancelled

quis:
At the moment we are manually cancelling letters for people when they ask us to. Once’s we’ve done this there is no indication that it’s happened except for the date going red on the list of letters.

This commit adds some error messaging and styling to show when a letter is cancelled.

Letting people cancel their own letters will be a future enhancement.

***

image

***

screen shot 2018-10-11 at 10 02 00

***

screen shot 2018-10-11 at 10 22 43

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -431,7 +431,9 @@ def format_notification_status_as_field_status(status, notification_type):
     ).get(status, 'error')
 
 
-def format_notification_status_as_url(status):
+def format_notification_status_as_url(status, notification_type):
+    if notification_type == 'letter':
+        return None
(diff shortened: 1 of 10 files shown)


quis:
# Show letters that fail validation as cancelled

At the moment we show precompiled letters that have failed the validation as having been sent. This is confusing.

We should communicate it as having been cancelled (rather than failed), with the implication being that Notify has come along and cancelled the letter before printing it. I think this is conceptually what makes the most sense.

From the user’s point of view any letters that show up as cancelled probably need to be fixed and resent, so it makes sense to group them with the same name.

PR #2379, opened 2018-10-12
Add folder icons

quis:
These are copied from the prototype but with the following changes:
- redrawn to snap exactly to pixels
- slightly thicker border for the blue version, and a thinner border for the black version, so they look better in situ
- run through https://jakearchibald.github.io/svgomg/ for optimal file size

This commit doesn’t do anything with the images yet, it just adds them to the repo.

# Blue variant

## Prototype

image

## Redrawn

image

# Black variant

## Prototype 

image

## Redrawn

image

***

Both versions are designed to be rendered at 20px high. The proportions are such that the folder aligns to the cap height and x-height of New Transport in bold (not exactly because it needed some optical adjustment):

image

image


quis:
# Add thinner version of the blue icon

This is needed for when the icon is displayed at a larger size. The thicker blue icon looks too big if it’s displayed at over 20px high (the use case for this is displaying it at 30px high).

image

PR #2414, opened 2018-10-25
Reduce logging noise in traceback output

quis:
When running tests locally Pytest returns a lot of captured logging info.

This is redundant because Pytest also captures stdout. 

This commit sets the logging to the most minimal level when running tests against the app.

# Before 

image

# After 

image

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 testpaths = tests
+log_level = 999
 env =
     NOTIFY_ENVIRONMENT=test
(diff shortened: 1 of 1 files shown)


quis:
@leohemsted said we should find a way to do this without losing the stack trace, so not going to merge for now.

quis:
@leohemsted I worked out that by suppressing the logging in Pytest’s config (ebea5039) we don’t get any captured logging output (good) but still see the logging captured because it’s still captured by `stdout` (also good).

***

Before (left) and after(right):

image

PR #2421, opened 2018-10-29
Update notifications-utils to 30.5.6

klssmith:
This brings in the change to email validation which removes obscure white space from email addresses.

[Pivotal story](https://www.pivotaltracker.com/story/show/161290942)

### DIFF
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -23,4 +23,8 @@ awscli-cwlogs>=1.4,<1.5
 awscli==1.15.82
 botocore<1.11.0
 
-git+https://github.com/alphagov/notifications-utils.git@30.5.5#egg=notifications-utils==30.5.5
+
+# Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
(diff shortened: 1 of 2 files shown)


requirements-app.txt
```diff
@@ -23,4 +23,8 @@ awscli-cwlogs>=1.4,<1.5
 awscli==1.15.82
 botocore<1.11.0
 
-git+https://github.com/alphagov/notifications-utils.git@30.5.5#egg=notifications-utils==30.5.5
+
+# Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
```
CrystalPea: putting what upgrade on hold?

klssmith: The upgrade to a later version of `itsdangerous`. Version `1.1.0` introduced breaking changes, and even though version `1.1.1` seems to have reverted these it seems best to keep the version change of `itsdangerous` in its own PR.

servingUpAces: The upgrade to itsdangerous from 0.24 to 1.0.0 is a breaking change, we need to do a little work before upgrading

PR #2423, opened 2018-10-29
Move more things from the view layer to the service model 

quis:
We do a lot of logic inside the view methods.

It makes it cleaner to break this logic up into functions. But this would mean passing around variables from one function to another. Putting these methods onto a class (the service model) means that
there’s a place to store this data (rather than having to pass it around a lot).

Making this code more manageable is important so that when we have templates and folders it’s easy to encapsulate the logic around combining the two.

# Before 

img_8604

# After 

img_8605

***

Notes for reviewing:

- There’s a lot of refactoring and moving things around to make this work possible. I‘ve tried to break it into sensible commits, and as such is best reviewed one commit at a time.
- The two view methods I really focused on were `choose_template` and `service_settings`. It would have been too much work to refactor the whole app in one pull request, but hopefully this shows:
    - The advantage and feasibility of architecting things this way
    - The direction to go when building future features or refactoring
- At the moment the user API client calls the `User` model quite a lot, which is the wrong way round. But again, too much work to fix in this one pull request.


app/models/service.py
```diff
@@ -0,0 +1,265 @@
...
+    @property
+    def trial_mode(self):
+        return self._dict['restricted']
+
+    def has_permission(self, permission):
+        return permission in self.permissions
+
+    @cached_property
+    def has_jobs(self):
+        # Can’t import at top-level because app isn’t yet initialised
```
leohemsted: comment no longer applies

app/templates/views/platform-admin/services.html
```diff
@@ -47,7 +47,7 @@
 
         {% call row() %}
           {% call field(border=False) %}
-            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name'] }}</a>
+            <a href="{{ url_for('main.service_dashboard', service_id=service.id) }}" class="browse-list-link">{{ service['name'] }}</a>
```
leohemsted: should this have changed? It's on platform admin, so not handling `current_service`, but instead its own service obj

app/models/service.py
```diff
@@ -0,0 +1,265 @@
...
+    def __bool__(self):
+        return self._dict != {}
+
+    def __getattr__(self, attr):
+        if attr in self.ALLOWED_PROPERTIES:
+            return self._dict[attr]
+        raise AttributeError('`{}` is not a service attribute'.format(attr))
+
+    def __getitem__(self, attr):
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
quis: d3c48339

app/templates/views/platform-admin/services.html
quis: Jinja won’t care, but I’ve squashed it out for the sake of not changing things unnecessarily.

app/main/views/conversation.py
```diff
@@ -44,13 +44,7 @@ def conversation_reply(
...
     notification_id,
 ):
-
-    templates = [
-        template
-        for template in service_api_client.get_service_templates(service_id)['data']
-        if template['template_type'] == 'sms'
-    ]
-
+    templates = current_service.templates_by_type('sms')
```
CrystalPea: neat ☺️

PR #2427, opened 2018-10-30
Add test for the notifications page when there are failed notifications

klssmith:
We already have tests for the `/services/<service_id>/notifications/<notification_type>` page, but these were not testing the page when there were failed notifications, so this adds a test for the page in this situation.

This tests the change made here - https://github.com/alphagov/notifications-admin/pull/2426

### DIFF
--- a/tests/app/main/views/test_activity.py
+++ b/tests/app/main/views/test_activity.py
@@ -485,6 +485,26 @@ def test_html_contains_notification_id(
         assert uuid.UUID(tr.attrs['id'])
 
 
+def test_html_contains_links_for_failed_notifications(
+    client_request,
+    active_user_with_permissions,
+    mock_get_service_statistics,
+    mock_get_service_data_retention_by_notification_type,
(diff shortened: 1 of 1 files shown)


tests/app/main/views/test_activity.py
```diff
@@ -485,6 +485,29 @@ def test_html_contains_notification_id(
...
+def test_html_contains_links_for_failed_notifications(
+    logged_in_client,
+    service_one,
+    active_user_with_permissions,
+    mock_get_service_statistics,
+    mock_get_service_data_retention_by_notification_type,
+    mocker,
+):
+    mock_get_notifications(mocker, active_user_with_permissions, noti_status='technical-failure')
+    response = logged_in_client.get(url_for(
```
quis: This should be using the `client_request` fixture

klssmith: 👍 Updated to use the right fixture

PR #2434, opened 2018-10-31
Let people send one-off letters from the admin app

quis:
When we first made the letters feature we didn’t do this because it would have involved making and processing a single row CSV, and making a job for each letter (that’s how the ‘print a test letter thing worked; it just never let you start a job).

Now we think that there is a user need for sending individual letters. It’s a bit clunky, but in some cases less so than printing, stuffing, franking and posting manually using the office printer.

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2201

***

# Flow for sending a letter 

screen shot 2018-10-31 at 14 10 32

***

screen shot 2018-10-31 at 14 07 41
screen shot 2018-10-31 at 14 04 37

***

screen shot 2018-10-31 at 14 08 02

# If your service is in trial mode

You’ll see this instead of the screen where you can press the ‘send’ button

screen shot 2018-10-31 at 14 12 47
screen shot 2018-10-31 at 14 13 37


app/templates/views/check/column-errors.html
```diff
@@ -109,14 +109,15 @@ <h1 class='banner-title' data-module="track-error" data-error-type="Missing plac
...
-        <p>
-          In <a href="{{ url_for('.using_notify') }}#trial-mode">trial mode</a> you
-          can only preview how your letters will look
-        </p>
+        <div class="bottom-gutter">
+          {% with
+            count_of_recipients=count_of_recipients
+          %}
+            {% call banner_wrapper(type='dangerous') %}
+              {% include "partials/check/trying-to-send-letters-in-trial-mode.html" %}
```
leohemsted: this has two error box outlines

image

PR #2452, opened 2018-11-05
Create template folders

allait:
screen shot 2018-11-06 at 10 48 35

### add new template folder

The add new templates page now has option to add template folders. Tweaked wording of other options and h1 to clarify options since it's not all about templates any more.

Added api client and stuff for it

### show list of template folders on the choose templates page

Not linking anywhere yet. Also not sorted, and not filtered by parent_id.

### Add Service.get_templates method with filters by type and folder

With the addition of template folders we need to filter templates based on a combination of type and parent folder ID.

This replaces the existing `templates_by_type` method with `get_templates`, which supports both type and parent folder filters, avoiding a need to create specific methods for each use case.

We still need the templates property to exist in some way in order to cache it, but it needs to be clear that it's different from `.get_templates`. One option was to make it "private" (i.e. `_templates`), and always use `.get_templates` in the rest of the code, but this requires adding "include all folders" to `.get_templates`, which doesn't have an obvious interface since `parent_folder_id=None` already means "top-level only".

This will probably come up again when we need to look into adding templates from nested folders into the page for live search, but for now renaming `Service.templates` to `.all_templates` makes it clear what the property contains.

### Link template folders to their contents

Clicking on template folder navigates to a page that displays that folder's contents.

This reuses the existing choose template view by adding a filter based on optional `template_folder_id` argument.

Service model methods are rewritten to match `all_templates` and `get_template`. New `get_template_folder_path` method returns a list of folders (from root to the current one) that the selected folder is nested inside.

### Add current template folder path to header and page title

When a folder is selected the full path is displayed in page title and header (for example `Templates / Folder1 / Folder2`). Elements of the path link to the corresponding folder. Current folder is not linked.

### Create new folders inside the current folder

When adding a new folder it is created inside the currently active one. The user is returned to the previously active folder page, which shows the added folder.

This adds a new route to the add template/folder views. Thankfully, `url_for` recognizes when `template_folder_id` is `None` and will use the URL without `/folders/...`, so users without folder permissions should be unaffected by this change.

### Fix template type selector to preserve current folder


app/main/views/templates.py
```diff
@@ -115,11 +119,13 @@ def choose_template(service_id, template_type='all'):
 
     return render_template(
         'views/templates/choose.html',
-        templates=current_service.templates_by_type(template_type),
-        show_search_box=(len(current_service.templates_by_type(template_type)) > 7),
+        template_folder_path=current_service.get_template_folder_path(template_folder_id),
+        template_folders=current_service.get_template_folders(template_folder_id),
+        templates=current_service.get_templates(template_type, template_folder_id),
+        show_search_box=(len(current_service.get_templates(template_type)) > 7),
```
quis: This should depend on the number of visible templates I think, so
```suggestion
        show_search_box=(len(current_service.get_templates(template_type, template_folder_id)) > 7),
```

allait: I haven't touched the search box at the moment, but I think it's more complex than that if we have a lot of templates in nested folders and want the search to show them. Right now they're not added to the page at all, we'll need to come back to this once we can actually put templates inside folders.

tests/app/main/views/test_template_folders.py
```diff
@@ -0,0 +1,166 @@
...
+def test_get_add_template_folder_page(client_request, service_one, parent_folder_id):
+    service_one['permissions'] += ['edit_folders']
+
+    page = client_request.get(
+        'main.add_template_folder',
+        service_id=service_one['id'],
+        template_folder_id=parent_folder_id
+    )
+
+    assert page.find('input', attrs={'name': 'name'}) is not None
```
quis: Prefer using CSS selectors:
```suggestion
    assert page.select_one('input[name=name]') is not None
```

tests/app/main/views/test_template_folders.py
```diff
@@ -0,0 +1,166 @@
...
+    mocker,
+    mock_get_service_templates,
+    mock_get_organisations_and_services_for_user,
+):
+    service_one['permissions'] += ['edit_folders']
+    mocker.patch('app.service_api_client.get_service', return_value={"data": service_one})
+
+    page = client_request.get(
+        'main.add_template_by_type',
+        service_id=service_one['id'],
```
quis: You can use `SERVICE_ONE_ID` here and then you don’t need to use the `service_one` fixture (`client_request` brings it in automatically)

quis: Applies to other uses in this file too.

allait: We use `service_one` to set permissions for editing folders, is there a better way to do this without using the fixture?

quis: Oh yeah, that sucks but I don’t think there’s a better way at the moment.

PR #2453, opened 2018-11-05
Use the model for updating a service and its permissions 

quis:
We have a lot of places in settings where we modify the service by talking to the API client. We can write less code and encapsulate things better by putting these operations on the model.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -108,8 +108,7 @@ def service_name_change_confirm(service_id):
 
     if form.validate_on_submit():
         try:
-            service_api_client.update_service(
-                current_service.id,
+            current_service.update(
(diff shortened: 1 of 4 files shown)


app/main/views/service_settings.py
```diff
@@ -495,7 +457,7 @@ def service_set_inbound_number(service_id):
             is_default=True,
             inbound_number_id=form.inbound_number.data
         )
-        switch_service_permissions(current_service.id, 'inbound_sms')
+        current_service.force_permission('inbound_sms', on=True)
```
quis: This is a functional change, but I think it’s right that if you’re adding an inbound number you should always make sure the service has inbound SMS turned on. Seems like this could cause bugs otherwise.

PR #2456, opened 2018-11-06
Add a URL for method to current_service

quis:
Saves having to type `service_id=current_service.id` 100s of times.

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -108,7 +108,7 @@ def feedback(ticket_type):
         if current_service:
             service_string = 'Service: "{name}"\n{url}\n'.format(
                 name=current_service.name,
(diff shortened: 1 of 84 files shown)


allait:
I'm not sure about this one. I don't feel like the current approach is that repetitive and at the same time it's pretty straightforward.

This makes the `service_id` special, which I'm not convinced is justified just because it is ubiquitous. I also don't think generating URLs is a responsibility of the Service model, so as a new starter for example I wouldn't look for it there (using `url_for` instead) and having noticed it I'd have to figure out what it does (whereas `url_for` I already know).

It'd be a different story if this added safety, but `url_for` already errors if `service_id` is missing, and there's no way for us to enforce the use of `Service.url_for`.

quis:
Yeah, I think you’re right – not a good separation of concerns.

PR #2462, opened 2018-11-07
Sort API keys alphabetically

quis:
Before | After 
---|---
image | image

Also does some refactoring to be able to put this sorting logic in the service model.


leohemsted:
do we want to move revoked keys to the bottom or similar? or do they get hidden/cleaned up after a while?

quis:
They get filtered out after 7 days. Figure it’s better to keep them in the same order until they do.

quis:
(rebased)

PR #2468, opened 2018-11-08
Create a template in a folder.

servingUpAces:
When `Add template` is clicked from a folder page, the template created will be created with inside that folder.

### DIFF
--- a/tests/app/main/views/test_template_folders.py
+++ b/tests/app/main/views/test_template_folders.py
@@ -208,3 +208,44 @@ def test_add_template_by_type_should_redirect_to_view_template_for_letter(
                                    template_id='Untitled',
                                    _external=True),
     )
+    mock_create_service_template.assert_called_once_with('Untitled',
+                                                         'letter',
+                                                         'Body',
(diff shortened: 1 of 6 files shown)


klssmith:
Code looks good and it all worked when I tried it, but could be worth adding some tests that the api client is being called with the right parameters (with parent_id correct). It looks like the tests that check that we are being redirected to the right page would have passed before these changes were added.

tests/app/main/views/test_template_folders.py
```diff
@@ -208,3 +208,44 @@ def test_add_template_by_type_should_redirect_to_view_template_for_letter(
...
+        client_request,
+        service_one,
+        mock_create_service_template
+):
+    data = {
+        'name': "new name",
+        'subject': "Food incoming!",
+        'template_content': "here's a burrito 🌯",
+        'template_type': 'email',
+        'service': service_one['id'],
```
klssmith: Could just use `SERVICE_ONE_ID` here and on line 235, then not include the `service_one` fixture

PR #2469, opened 2018-11-08
Allow our sticky navs to stick to bottom as well as top

tombye:
This changes the existing [sticky JS](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/stick-at-top-when-scrolling.js) so sticky elements can attach the bottom edge of a window instead of the top when scrolling makes them off-screen. It adds functionality to stop an element at a point on the screen so it goes no further. We used to use [a separate script](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/stop-scrolling-at-footer.js) for this.

### How to use

To make an element stick to the bottom, give it a `stick-at-bottom-when-scrolling` class.

To stop it at a certain point, add a `js-header` class to an element above the sticky one. The sticky one should stop when it touches the bottom of that element.

### Explanation of behaviour

The sticky element has vertical boundaries between which will scroll. If scrolling moves it off the screen, and not further than the stopping boundary, it will stick to the window. Depending on how it is set up, it will stick to the top or the bottom edge of the window.

#### Stick to top

stick-to-top

#### Stick to bottom

stick-to-bottom


app/assets/stylesheets/components/stick-at-top-when-scrolling.scss
```diff
@@ -32,17 +29,27 @@
...
+.js-stick-at-top-when-scrolling {
+
+  margin-top: -10px;
+  margin-bottom: 5px;
+  top: 5px;
+  transition: top 0.1s ease-out, box-shadow 1s ease-in-out;
+
+}
+
+.js-stick-at-top-when-scrolling {
```
quis: Should be `js-stick-at-bottom-when-scrolling` I think (otherwise it’s just overriding the above)

quis: Fixed in 55fa85d7

tombye: Good spot!

quis:
Need to remove this fade in on page load, otherwise this is good to go:

fade-load

quis:
Works pretty good…

stick-bottom

app/assets/javascripts/stick-to-window-when-scrolling.js
```diff
@@ -0,0 +1,298 @@
...
+  stickAtBottom.scrolledToOutsideWindow = function (el, windowHeight) {
+    var windowBottom = this.getWindowPositions().scrollTop + this.getWindowDimensions().height;
+
+    return windowBottom < el.scrolledTo;
+  };
+  stickAtBottom.addShimForEl = function ($el, width, height) {
+    $el.after('<div class="shim" style="width: ' + width + 'px height: ' + height + 'px">&nbsp</div>');
+  };
+  stickAtBottom.stop = function (el) {
+    if (!el.stopped()) {
```
quis: This method doesn’t exist.

PR #2470, opened 2018-11-08
Let users move things into folders

quis:
image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -702,7 +702,7 @@ class ServicePostageForm(StripWhitespaceForm):
     )
 
 
-class BrandingStyle(RadioField):
+class RadioFieldWithNoneOption(RadioField):
 
     def post_validate(self, form, validation_stopped):
         if self.data == 'None':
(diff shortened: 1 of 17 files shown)


app/models/service.py
```diff
@@ -36,6 +36,12 @@ class Service():
         'service_callback_api',
     }
 
+    TEMPLATE_TYPES = (
```
allait: Maybe this could be a list of dictionaries?

app/main/forms.py
```diff
@@ -1111,3 +1118,41 @@ class ReturnedLettersForm(StripWhitespaceForm):
 
 class TemplateFolderForm(StripWhitespaceForm):
     name = StringField('Folder name', validators=[DataRequired(message='Can’t be empty')])
+
+
+class TemplateAndFoldersSelectionForm(Form):
+
+    ALL_TEMPLATES_FOLDER = {
+        'name': 'All templates',
+        'id': 'None',
```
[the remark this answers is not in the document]
quis: See c18a38d4f172f0cdf197e70f43ca673a2afaea14

app/models/service.py
quis: See 40f676d3ad86ff5c3b5a7ffc007f5ca46d27860a

PR #2481, opened 2018-11-12
Label and filter folders depending on their contents 

quis:
screen shot 2018-11-12 at 15 06 07<sup>1</sup>

It feels like a solid reckon that knowing what’s in a folder before you click on it will help you navigate around.

However, what do you show in a folder if you’re filtering by template type? We think that:
- if you’re not filtering you should see all folders, even empty ones
- if you’re filtering you should only see folders that will get you to relevant templates

This matches what happens when you filter templates, we don’t ‘grey out’ the non-email templates, we hide them completely.

The logic then extends to how we describe the contents of a folder, ie we won’t count its subfolders if they don’t contain templates of the type we’re looking for. This means that however you see the folder
described (eg ‘3 templates, 1 folder’) will match what you see when you click into it.

1. https://docs.google.com/document/d/1Z8Cch19vHndibx_zMcaMQRga3HutCWrS02q-22KUXEE/edit?ts=5be1c99b#heading=h.txfueg136qtb

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -298,11 +298,30 @@ class Service():
     def all_template_folder_ids(self):
         return {folder['id'] for folder in self.all_template_folders}
 
-    def get_template_folders(self, parent_folder_id=None):
+    def get_template_folders(self, parent_folder_id=None, template_type='all'):
         return [
             folder for folder in self.all_template_folders
-            if folder['parent_id'] == parent_folder_id
+            if (
+                folder['parent_id'] == parent_folder_id and
+                self.show_folder(folder['id'], template_type)
+            )
         ]
+
+    def show_folder(self, template_folder_id, template_type='all'):
+
+        if template_type == 'all':
+            return True
+
+        if self.get_templates(template_type, template_folder_id):
+            return True
+
+        if any(
+            self.show_folder(child_folder['id'], template_type)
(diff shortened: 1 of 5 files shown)


app/models/service.py
```diff
@@ -298,12 +298,31 @@ def all_template_folders(self):
...
         return [
             folder for folder in self.all_template_folders
-            if folder['parent_id'] == parent_folder_id
+            if (
+                folder['parent_id'] == parent_folder_id and
+                self.show_folder(folder['id'], template_type)
+            )
         ]
 
+    def show_folder(self, template_folder_id, template_type='all'):
```
allait: `is_folder_visible`?

app/models/service.py
```diff
@@ -298,12 +298,31 @@ def all_template_folders(self):
     def all_template_folder_ids(self):
         return {folder['id'] for folder in self.all_template_folders}
 
-    def get_template_folders(self, parent_folder_id=None):
+    def get_template_folders(self, parent_folder_id=None, template_type='all'):
```
allait: I think we should reverse the order of the arguments to match `get_templates`

quis: Good idea: 045a4bb

app/models/service.py
quis: Works for me: 63060d9

PR #2482, opened 2018-11-12
Add a new endpoint for managing a folder, use it to rename folder

CrystalPea:
- Add a GET / POST view: manage template folder
- Add a template for that view, where a user can rename their
folder
- Add an API client method for updating a folder

Pivotal ticket: https://www.pivotaltracker.com/story/show/161178427

#### Manage folder page:
screen shot 2018-11-12 at 16 43 21

#### Redirects:
screen shot 2018-11-12 at 16 43 39


tests/app/main/views/test_template_folders.py
```diff
@@ -248,3 +248,58 @@ def test_can_create_email_template_with_parent_folder(
...
+        'main.manage_template_folder',
+        service_id=service_one['id'],
+        template_folder_id=folder_id
+    )
+    assert page.select_one('input[name=name]') is not None
+
+
+def test_get_manage_folder_page_no_permissions(client_request, service_one, mock_get_template_folders):
+    folder_id = str(uuid.uuid4())
+    mock_get_template_folders.return_value = [
```
quis: I don’t think you need to mock this because the page should `403` before it tries to talk to the API.

PR #2487, opened 2018-11-13
Fix templates page in IE10

quis:
# Before 

screen shot 2018-11-13 at 14 02 45

# After 

screen shot 2018-11-13 at 14 02 26

### DIFF
--- a/app/assets/stylesheets/components/pill.scss
+++ b/app/assets/stylesheets/components/pill.scss
@@ -6,6 +6,7 @@
     width: 25%;
     flex-grow: 1;
     text-align: left;
+    float: left;
   }
 
   a,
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/components/pill.scss
```diff
@@ -6,6 +6,7 @@
     width: 25%;
     flex-grow: 1;
     text-align: left;
+    float: left;
```
tombye: I know the reasoning behind this will be preserved in the git history but can we have a CSS comment for it too?

Also: looks like flex-items cannot be floated so this is safe: https://www.w3.org/TR/css-flexbox-1/#flex-containers

quis: Comment added in 807396f

PR #2493, opened 2018-11-14
Add delete folder flow

CrystalPea:
### Work done:

- Add a delete folder link to manage folder page
- Add delete_folder method to template folder api client
- Add delete folder GET/POST endpoint 
- Add Flash messages and error handling to the airport
- Tests

Pivotal ticket: https://www.pivotaltracker.com/story/show/161177455

#### Happy path:

screen shot 2018-11-14 at 17 03 37

screen shot 2018-11-14 at 17 56 26

screen shot 2018-11-14 at 17 04 34

#### Folder not empty:

screen shot 2018-11-14 at 17 00 31


app/main/views/templates.py
```diff
@@ -387,9 +391,45 @@ def manage_template_folder(service_id, template_folder_id):
...
+    if request.method == 'POST':
+        try:
+            template_folder_api_client.delete_template_folder(current_service.id, template_folder_id)
+
+            flash("'{}' folder has been removed".format(template_folder_name), 'info')
+            return redirect(
+                url_for('.choose_template', service_id=service_id)
+            )
+        except HTTPError as e:
+            msg = "Folder is not empty"
```
servingUpAces: We can check that the folder is empty before calling delete.

app/templates/views/templates/manage-template-folder.html
```diff
@@ -25,7 +25,15 @@ <h1 class="heading-large">
...
     {{ textbox(form.name) }}
-    {{ page_footer('Save') }}
+    {{ page_footer(
+      'Save',
+      delete_link=url_for(
+        '.delete_template_folder',
+        service_id=current_service_id,
+        template_folder_id=template_folder_id
+      ),
+      delete_link_text="Delete this folder") }}
```
servingUpAces: Do we want to only show the delete link if the folder is empty?

CrystalPea: We could, and then some kind of message if it's not? Like "If you want to delete this folder, you need to first empty it"...?

app/main/views/templates.py
CrystalPea: You're right 👍

app/main/views/templates.py
```diff
@@ -387,6 +391,51 @@ def manage_template_folder(service_id, template_folder_id):
...
+        flash("'{}' folder is not empty".format(template_folder_name), 'info')
+        return redirect(url_for(
+            '.manage_template_folder', service_id, template_folder_id
+        ))
+
+    if request.method == 'POST':
+        try:
+            template_folder_api_client.delete_template_folder(current_service.id, template_folder_id)
+
+            flash("'{}' folder has been removed".format(template_folder_name), 'info')
```
servingUpAces: We don't need this message

karlchillmaid: `Are you sure you want to delete the [folder name] folder?`
`Confirm`

– as discussed, no need for a confirmation message –

Just finalising the _can't delete_ message

karlchillmaid:
Can't delete message:

`You must empty this folder before you can delete it`
`Move or delete all templates and folders from the [folder name] folder and try again.`

CrystalPea:
So, I changed the content of some messages and got rid of final confirmation message. I updated the screenshots to reflect those changes.

In a separate pull request I will change delete template 'are you sure' message so it is consistent with delete folder message, as per consultation with Karl.

@karlchillmaid @yahoopete @servingUpAces are you happy with how it all looks now?

karlchillmaid:
@CrystalPea Content in the screenshots matches my previous comments and our discussion yesterday.

Good to know the _delete template_ flow will change to match this.

One change – please remove the quote marks around the folder name. We don't use quote marks around template names in the _delete template_ flow.

Not for now – perhaps we can revisit the option to add an extra description to the _can't delete_ message sometime?

app/templates/views/templates/manage-template-folder.html
servingUpAces: We've decided it's better to show the button and let them get the `must be empty` message.

PR #2501, opened 2018-11-15
Make deletion confirmation banner messages consistent across our app

CrystalPea:
Also introduce a way to provide context to a banner / flash message that will be displayed in plain font style.

#### Deleting template:
screen shot 2018-11-16 at 11 03 39

#### Revoking API key:
screen shot 2018-11-16 at 15 12 23

#### Deleting reply-to email:
screen shot 2018-11-16 at 17 07 57

#### Deleting sms sender:
screen shot 2018-11-16 at 17 08 31

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -441,7 +441,7 @@ def delete_template_folder(service_id, template_folder_id):
             else:
                 abort(500, e)
 
-    flash("Are you sure you want to delete the '{}' folder?".format(template_folder_name), 'delete')
+    flash("Are you sure you want to delete the ‘{}’ folder?".format(template_folder_name), 'delete')
     return render_template(
         'views/templates/manage-template-folder.html',
         form=form,
@@ -628,12 +628,9 @@ def delete_service_template(service_id, template_id):
         else:
             raise e
 
+    flash(["Are you sure you want to delete ‘{}’?".format(template['name']), message], 'delete')
     return render_template(
         'views/templates/template.html',
-        template_delete_confirmation_message=(
-            'Are you sure you want to delete {}?'.format(template['name']),
-            message,
-        ),
         template=get_template(
             template,
             current_service,
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -615,7 +615,7 @@ def delete_service_template(service_id, template_id):
         last_used_notification = template_statistics_client.get_template_statistics_for_template(
             service_id, template['id']
         )
-        message = 'It was last used {} ago'.format(
+        message = 'This template was last used {} ago.'.format(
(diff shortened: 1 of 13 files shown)


app/main/views/templates.py
```diff
@@ -628,12 +628,9 @@ def delete_service_template(service_id, template_id):
         else:
             raise e
 
+    flash(["Are you sure you want to delete ’{}’?".format(template['name']), message], 'delete')
```
quis:
```suggestion
    flash(["Are you sure you want to delete ‘{}’?".format(template['name']), message], 'delete')
```
Thanks for trying to use the right one 😉

CrystalPea: Uhh, I will fix it, and meanwhile here's a badge for me :P

CrystalPea: fixed 😊

karlchillmaid:
screen shot 2018-11-15 at 14 53 11

Suggest changing the revoke API content to:

`Are you sure you want to revoke 'API key name'?`
`You will not be able to use this API key to connect to GOV.UK Notify.`
`Yes, revoke this API key`

This is more consistent with the _delete template_ and _delete folder_ flows, where we name the object in the warning message, not in the description.

It's important to do this for the _revoke API key_ flow because you trigger the flow from the page that lists all of the API keys. In the current version, they can't be sure which API key 'this API key' is without reading the description too. Users shouldn't have to do that every time.

karlchillmaid:
Suggest changing the description on the delete template message to say:

`This template was last use more than 7 days ago`

'This template' is more specific than 'it' – which might be helpful now users can delete folders too.

We should also use numerals rather than writing the word 'seven'.

CrystalPea:
@karlchillmaid 
In revoking API key, I changed the content as you wanted except from the button. Button can only takes Yes, [action].

When it comes to template I made updates except from changing the number to numerals, that would require rewriting a whole method for dynamically inserting that content and I feel is out of scope.

I will update the screenshots shortly. Let me know if it's ok.

karlchillmaid:
@CrystalPea 

Fine about the `Yes, revoke` button if that's the restriction we're working within.

`Seven` still needs to change to a numeral (`7`) on the template description.

Thanks

CrystalPea:
Hi @karlchillmaid, I wrote a Pivotal ticket to change how we write our numbers: https://www.pivotaltracker.com/story/show/162013070

I hope it's fine ☺️

CrystalPea:
@servingUpAces sure, at it 👍

CrystalPea:
@servingUpAces /services//service-settings/sms-sender//delete and /services//service-settings/email-reply-to//delete are updated now ☺️

PR #2506, opened 2018-11-16
Fix breadcrumbs on manage_template_folder page

CrystalPea:
Also:
- remove name changing functionality when in folder-deletion mode
- write more tests for delete-folder endpoint

### Screenshot of updated delete-folder page:

screen shot 2018-11-16 at 15 56 17


CrystalPea:
@karlchillmaid if you could maybe help me out: what content should I use for "back" link?

PR #2511, opened 2018-11-19
Use sticky footer on the check one-off page 

quis:
# Before

image

# After 

sticky-send-4r

image

image

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -853,6 +853,7 @@ def _check_notification(service_id, template_id, exception=None):
         db_template,
         current_service,
         show_recipient=True,
+        expand_emails=True,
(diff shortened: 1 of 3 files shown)


app/main/views/send.py
```diff
@@ -853,6 +854,7 @@ def _check_notification(service_id, template_id, exception=None):
         db_template,
         current_service,
         show_recipient=True,
+        expand_emails=True,
```
leohemsted: do we even need the expand_emails flag any more?

quis: For now, yes. It’s still used on the ‘check’ page for sending jobs (which is more complicated to put the sticky on because it has the form for scheduling):

image

PR #2513, opened 2018-11-19
Improve the empty folder state

quis:
# Before 

image

# After 

image

image

### DIFF
--- a/app/assets/stylesheets/components/message.scss
+++ b/app/assets/stylesheets/components/message.scss
@@ -54,6 +54,11 @@
 
   }
 
+  &-empty {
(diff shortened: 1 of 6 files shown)


karlchillmaid:
**Scenario 1**
There are no templates or subfolders in a folder.

Hint: `This folder is empty`

This is consistent with the 'Empty' folder hint on the Templates page.

When a folder is empty _'This folder is empty'_ should display even when a filter is applied.

**Scenario 2**
There are templates and/or subfolders in a folder, but they do not match the applied filter.

Hint: `There are no [type] templates in this folder`

quis:
Updated:

image

image

PR #2531, opened 2018-11-23
Make template search work across folders 

quis:
This makes three main changes to how the template search works:
- it puts items that are in child folders of the current folder on the page, but hidden by default
- it adds the path and folder icon to items that are children of folders at the current level
- it makes the search more specific, so it only looks at each item’s name, not its entire text (which would include the full path and the hint text)

All so that you can still find the thing you’re looking for quickly, even when folders are a thing:

search-in-folder

***

https://www.pivotaltracker.com/story/show/161178531


allait:
How is this supposed to work in combination with moving selected items? Right now, it looks like I can search for a nested template/folder and select the corresponding checkbox, but when I try to move it nothing happens.

quis:
> How is this supposed to work in combination with moving selected items? Right now, it looks like I can search for a nested template/folder and select the corresponding checkbox, but when I try to move it nothing happens.

Good point, I think you were getting tripped up by the form validation. Have fixed in ee991d0.

quis:
Updated selected link state to:
- be full width if an item only contains one link
- to render cleanly in Firefox

# Before

image

# After

image

PR #2541, opened 2018-11-28
Let users add new template from the choose page

quis:
Since we’re letting users add new folders directly from the choose page it makes sense that they should also be able to add templates from there.

This resolves the problem we saw in user research where people found it hard to know where to go to add a new folder when they were all behind one green button.

***

image


quis:
@leohemsted I implemented your suggestion to use a sentinel value for form values that should be interpreted as `None`: 2fcf278

PR #2543, opened 2018-11-28
Remove Whitenoise

quis:
We don’t seem to be using it.


quis:
We are using it. Oops.

PR #2544, opened 2018-11-29
Updated reply-to address content

karlchillmaid:
Updated reply-to address content to better explain why users need to add a reply-to address.

Added a description to the 'Add reply-to address' page.

@quis I commented out the 'can't go live' message because it repeats the new content. We can change it if you think we still need it, or remove it if we don't.

### DIFF
--- a/app/templates/views/service-settings/email-reply-to/add.html
+++ b/app/templates/views/service-settings/email-reply-to/add.html
@@ -14,7 +14,7 @@
     Add email reply-to address
   </h1>
   <p>
-    Let recipients send replies to a shared inbox that's managed by your team.
+    Let recipients send replies to a shared inbox that’s managed by your team.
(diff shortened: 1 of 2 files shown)


quis:
image

***

image

***

image

***

Definite improvement. 

Can you remember what we decided about telling people about ‘autoresponders’?

karlchillmaid:
Setting up an 'autoresponder' for the reply-to address?

I don't think we'd decided what to say about that here. Do you want me to look at that as part of this PR – or merge what we have now and add more later?

quis:
Happy to merge this as is now and refine further. We could even wait until we get a few support tickets off the back of this new content and see what is/isn’t working?

karlchillmaid:
Sounds like the smart thing to do.

PR #2546, opened 2018-11-29
Make Whitenoise serve static assets

quis:
Currently it’s not configured properly, so isn’t having any effect. This change makes it wrap the Flask app, so it intercepts any requests for static content.

Follows the pattern documented in http://whitenoise.evans.io/en/stable/flask.html#enable-whitenoise

### DIFF
--- a/application.py
+++ b/application.py
@@ -12,4 +12,4 @@ STATIC_URL = 'static/'
 app = Flask('app')
 
 create_app(app)
-application = WhiteNoise(app, STATIC_ROOT, STATIC_URL)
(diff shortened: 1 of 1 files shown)


quis:
# Tell browsers not to re-fetch static assets

Since we version our asset filenames there’s no need for a browser to ever fetch the same file twice. It should always cache fetch from its own cache.

The accepted way to effect this behaviour is using the expires header, which is what this argument to `WhiteNoise` does.

quis:
# Before

image

# After

image

PR #2547, opened 2018-11-29
Fix asset path in asset fingeprinter

quis:
The asset fingerprinter was hard coded to always point to `/static`.

It needs to change depending on which environment the app is running in.


quis:
Notice how the last asset path is wrong here:

image

PR #2551, opened 2018-11-30
Back button on one-off message preview takes user to their flow

CrystalPea:
Earlier back button always took user to one-off flow, even when they came
from test flow. This made personalisation disappear when user was coming from test flow.

Also adjust steps calculation for test flow.

Pivotal ticket: https://www.pivotaltracker.com/story/show/161984520

Reservations (please discuss):
- I could not go through letter flow on my local as I couldn’t make antivirus run
- When you go through test flow (use user’s own email address/phone no.), and then go back all the way to choosing recipient, it will be blank. I think it would be better if it was filled with recipient(that is, current_user) data, that this data existed not only as `session['recipient']`, but also as a `placeholders` value. And then also the conditional on `line 393` in `send.py` would not be needed (I think that `current_placeholder` index influences address line numbers in letter template, so might be better to have the same number of placeholders for test flow and one-off flow). The thing is, I couldn’t make it work that way 😳.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -390,7 +390,11 @@ def send_test_step(service_id, template_id, step_index):
     )
 
     try:
-        current_placeholder = placeholders[step_index]
+        if request.endpoint == 'main.send_test_step':
+            current_placeholder = placeholders[step_index - 1]
+        else:
+            current_placeholder = placeholders[step_index]
+
     except IndexError:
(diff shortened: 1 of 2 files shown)


CrystalPea:
why do we even have separate test_flow and one_off_flow: could we just have one flow and rely on comparing recipient data to current user data to check if it’s a test? That would make things easier I think….

PR #2555, opened 2018-12-03
Cache service data retention in Redis

allait:
Adds caching for service data retention. This removes separate API
client methods to retrieve individual data retention records by id
or type in favor of a single method that fetches and caches all
retention settings configured for the service. This makes it much
easier to invalidate cache when settings change.

Lookup by id or type is provided by helper methods in the service
model.

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -3,7 +3,6 @@
 from flask import (
     Response,
     abort,
-    current_app,
     jsonify,
     redirect,
     render_template,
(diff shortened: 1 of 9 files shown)


app/main/views/jobs.py
```diff
@@ -226,8 +226,8 @@ def get_notifications(service_id, message_type, status_override=None):
     service_data_retention_days = None
 
     if message_type is not None:
-        service_data_retention_days = service_api_client.get_service_data_retention_by_notification_type(
-            service_id, message_type
+        service_data_retention_days = current_service.get_data_retention_by_type(
+            message_type
         ).get('days_of_retention', current_app.config['ACTIVITY_STATS_LIMIT_DAYS'])
```
quis: Feels like this line could also be factored into the model if you need to write it each time to get to the actual value in days.

allait: Good call, I've replaced `get_data_retention_by_type` with a method that returns days count directly, since it's the only way it's used at the moment.

PR #2562, opened 2018-12-05
Allow letters to be cancelled

klssmith:
* Added an endpoint to the `NotificationApiClient` for cancelling letters
* Made sure that `cancelled` letters display with the same formatting and in the same way as letters with the status of `permanent failure` (which is the status that we currently use for cancelled letters.)
* Added the link to cancel letters to the letter notification pages if the letter is able to be cancelled.

[Pivotal story](https://www.pivotaltracker.com/story/show/161115614)

### Cancellable letter
screen shot 2018-12-05 at 11 01 32

### Confirmation before cancelling
screen shot 2018-11-21 at 15 36 56

### Notification page after cancelling
screen shot 2018-12-05 at 11 01 55


app/templates/components/table.html
```diff
@@ -162,7 +162,7 @@
           notification.template.template_type
         ) }}
       {% endif %}
-      {% if notification.notification_type == "letter" and notification.status in ['permanent-failure', 'validation-failed'] %}
+      {% if notification.notification_type == "letter" and notification.status in ['permanent-failure', 'validation-failed', 'cancelled'] %}
```
leohemsted: we might want to think about what content we want to show here when validation fails. but that's not relevant for this PR

PR #2572, opened 2018-12-07
Show move errors on the template folder management flow

leohemsted:
depends on https://github.com/alphagov/notifications-api/pull/2253

|||
|-|-|
| cant move to itself | image|
|cannot move subfolders|image|

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -125,7 +125,10 @@ def choose_template(service_id, template_type='all', template_folder_id=None):
     if request.method == 'POST' and templates_and_folders_form.validate_on_submit():
         if not can_manage_folders():
             abort(403)
-        return process_folder_management_form(templates_and_folders_form, template_folder_id)
+        try:
+            return process_folder_management_form(templates_and_folders_form, template_folder_id)
+        except Exception as e:
+            flash(e.message)
 
     return render_template(
         'views/templates/choose.html',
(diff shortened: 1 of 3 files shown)


karlchillmaid:
@leohemsted 

`You cannot move a folder to one of its subfolders`

Does that cover all the possible scenarios?

leohemsted:
@karlchillmaid yeah think that'll do fine 👍

app/main/views/templates.py
```diff
@@ -125,7 +125,13 @@ def choose_template(service_id, template_type='all', template_folder_id=None):
     if request.method == 'POST' and templates_and_folders_form.validate_on_submit():
         if not can_manage_folders():
             abort(403)
-        return process_folder_management_form(templates_and_folders_form, template_folder_id)
+        try:
+            return process_folder_management_form(templates_and_folders_form, template_folder_id)
+        except Exception as e:
+            flash(e.message)
+
+    if 'templates_and_folders' in templates_and_folders_form.errors:
```
quis: Is there a test for this?

leohemsted: nope but there should be 👍

leohemsted: and also for the exception contents from the API showing up as a flash msg

leohemsted: @quis there is now a test for it

PR #2574, opened 2018-12-10
Add counter of selected templates and folders

quis:
selected-count

***

Being able to see how many things you have selected gives you positive feedback that reinforces that what you’ve done has been recognised. It helps you understand the implications of your actions (ie you see ‘3 selected’ before you press the ‘Move’ button). And it gives you an escape hatch the get out of the state you’re in by providing the ‘Clear’ button.

We also found in prototyping that having a ‘Nothing selected’ message helps draws people’s attention to the checkboxes when they first encounter the folders feature.

This PR implements the counter and the cancel button. It tries to follow the existing patterns for this module.

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -74,8 +74,11 @@
       .addClass('js-cancel')
       .attr('tabindex', '0')
       .on('click keydown', event => {
-        event.preventDefault();
-        fn();
+        // space, enter or no keyCode (must be mouse input)
+        if ([13, 32, undefined].indexOf(event.keyCode) > -1) {
+          event.preventDefault();
+          fn();
+        }
       });
 
     this.selectActionButtons = function () {
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -52,7 +52,7 @@
         this.selectActionButtons();
       });
 
-      state.$el.append($cancel);
+      state.$el.find('[type=submit]').after($cancel);
     };
 
     this.addClearButton = function(state) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/templateFolderForm.js
```diff
@@ -41,23 +42,45 @@
...
+
+        // clear existing data
+        state.$el.find('input:radio').prop('checked', false);
+        state.$el.find('input:text').val('');
+
+        // go back to action buttons
+        this.selectActionButtons();
+      });
+
+      state.$el.find('[type=submit]').after($cancel);
```
leohemsted: is `after` the same as `append`?

app/assets/javascripts/templateFolderForm.js
```diff
@@ -102,13 +128,19 @@
...
       </div>
     `;
 
     this.itemsSelectedButtons = `
       <div id="items_selected">
         <button class="button-secondary" value="move-to-existing-folder">Move</button>
         <button class="button-secondary" value="move-to-new-folder">Add to a new folder</button>
+        <div class="template-list-selected-counter">
+          <span class="template-list-selected-counter-count">1</span> selected
+        </div>
```
leohemsted: do we want to show these counts on the form states as well as the grey button states?

app/assets/javascripts/templateFolderForm.js
quis: `after` puts the element outside the target, `append` puts the element inside the target.

app/assets/javascripts/templateFolderForm.js
quis: worth seeing how it feels…

PR #2578, opened 2018-12-12
Note existing live services in go live ticket 

quis:
It’s useful for analysing our growth to know if someone who’s requesting to go live is already a live user of Notify.

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -220,8 +220,8 @@ class UserApiClient(NotifyAdminAPIClient):
         return sorted(all_services, key=lambda service: service['name'])
 
     def user_has_live_services(self, user):
-        return not all(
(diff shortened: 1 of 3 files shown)


app/notify_client/user_api_client.py
```diff
@@ -219,6 +219,11 @@ def get_services_for_user(self, user):
         ), [])
         return sorted(all_services, key=lambda service: service['name'])
 
+    def user_has_live_services(self, user):
+        return not all(
```
allait: It doesn't matter very much, but I think `return any(not service['restricted'] ...` is both potentially clearer and has an opportunity to short-circuit without having to go through the full list.

quis: Good point, refactored in f44ff0b

leohemsted: this'll still short-circuit on the first live service it finds

PR #2580, opened 2018-12-12
Don’t let non-government users request to go live 

quis:
Only users who work for government can accept the terms of use. This will save us from having to email these requesters back telling them they need to find someone else to submit the request.

Some refactoring in separate commits.

Government user | Not a government user 
---|---
image | image


quis:
@karlchillmaid thoughts on the content? Not urgent.

karlchillmaid:
@quis Looks good to me.

I wondered about putting the message higher up the page – but if the user is looking for a big green button to continue at the bottom then it makes sense where it is.

Also, there's nothing stopping a user without a gov.uk email address from completing the required steps.

👍

PR #2589, opened 2018-12-14
Add information about how static assets are served

quis:
This should help developers new to Notify understand how it works.

notify-static


quis:
This will probably be outdated soon…

quis:
Updated with new setup:

notify-static-after

PR #2590, opened 2018-12-14
Fix issues with JS for sticky elements 

tombye:
Fixes the following:

## Changes to sticky element content doesn't update stored references

`templateFolderForm.js` updates the content of a sticky element which changes its dimensions. Sticky JS stores data about the element's dimensions and position to reduce DOM queries. This add a public `recalculate` method to force an update.

## Changes to elements when stuck were not being passed to their shim

When elements are stuck to the window they are positioned absolutely so occupy no space in the document flow. We swap their out for a shim in the document to preserve this space.

Updates to any 'stuck' elements are now passed to the shim.

## Broken code for resize events

The code that runs when the window resizes no longer worked and could be replaced with that introduced for the `recalculate`method.

### DIFF
--- a/app/assets/javascripts/stick-to-window-when-scrolling.js
+++ b/app/assets/javascripts/stick-to-window-when-scrolling.js
@@ -51,13 +51,13 @@
     }
   };
   StickyElement.prototype.stop = function () {
-    this._stopped = true;
+    this._isStopped = true;
   };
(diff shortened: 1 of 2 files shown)


quis:
Also tested in IE11 (our most popular IE) on Browserstack and works just the same.

PR #2601, opened 2018-12-19
Revert "Decouple `upload-static` from cf-deploy"

quis:
Reverts alphagov/notifications-admin#2599

***

Depends on
- [x] https://github.com/alphagov/notifications-aws/pull/476


quis:
Let’s stick with the thing that’s working for now, we can come back to this and tidy up later if we still feel that having this step in the `Makefile` is clearer.

PR #2611, opened 2018-12-21
Set far-future cache headers on S3 assets

quis:
So that browsers will cache them for as long as possible. We invalidate this cache by adding the hash of each file to the query string.

There’s no way of doing this on a whole bucket; it has to be set on each item. Adding this flag does so at the time of uploading the items.


quis:
Value of 10 years in seconds copied from https://github.com/evansd/whitenoise/blob/0ee3bcb1eebdf4cada9024a858fa086d3dac6561/whitenoise/base.py#L19

PR #2617, opened 2018-12-31
Pytest 4: don't call pytest.mark directly

leohemsted:
specifically - don't use `pytest.mark.xfail` directly in parametrize, instead use `pytest.param(*args, marks=pytest.mark.xfail)`. the old way is deprecated in pytest4 - see https://docs.pytest.org/en/latest/deprecations.html#marks-in-pytest-mark-parametrize for rationale

Also, make this an error in pytest.ini so if someone adds a new xfail, it'll crash

### DIFF
--- a/tests/app/main/views/test_email_branding.py
+++ b/tests/app/main/views/test_email_branding.py
@@ -174,10 +174,7 @@ def test_cant_create_new_email_branding_with_unknown_domain(
         'southend.essex.gov.uk',
         'Not an organisation-level domain (use essex.gov.uk if appropriate)',
     ),
-    pytest.mark.xfail(
-        ('voa.gov.uk', ''),
-        raises=AssertionError
-    ),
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_email_branding.py
```diff
@@ -174,10 +174,7 @@ def test_cant_create_new_email_branding_with_unknown_domain(
         'southend.essex.gov.uk',
         'Not an organisation-level domain (use essex.gov.uk if appropriate)',
     ),
-    pytest.mark.xfail(
-        ('voa.gov.uk', ''),
-        raises=AssertionError
-    ),
+    pytest.param('voa.gov.uk', '', marks=pytest.mark.xfail(raises=AssertionError)),
```
quis: Not a dealbreaker, but I’d prefer to keep this on multiple lines so it can be scanned alongside the preceding examples:
```suggestion
    pytest.param(
        'voa.gov.uk',
        '',
        marks=pytest.mark.xfail(raises=AssertionError)
    ),
```

PR #2619, opened 2018-12-31
Stop Redis exception by not passing Redis an empty set to delete

klssmith:
We use `redis_client.delete` to delete multiple keys at once, but this
raises a `redis.exceptions.ResponseError` if it is called with an empty list
or set. We should only call `redis_client.delete` when there is at least
one item to delete.


klssmith:
@leohemsted good point. If we handle it here, it would save calling the Redis client unnecessarily, but maybe it would be better to put the code in the Redis client to make sure it's always handled 🤔 What do you reckon?

leohemsted:
hmm after thinking a bit more I think it's probably better to handle here: this is the difference between calling `foo()` vs `foo(1, 2, 3)`, and `redis.delete()` doesn't mean anything and I think it's fine that it raises an exception (as long as it's obvious that it's because no keys were passed in)

PR #2630, opened 2019-01-04
Nest destination folders

tombye:
Put the folders you can move a template/folder to into a list, nested to show its hierarchy.

Story: https://www.pivotaltracker.com/story/show/162265491

image


quis:
Tested locally and works good. I added a bit of styling:

Before | After 
---|---
image | image

quis:
…and with a bot more horizontal spacing:

image

tests/app/main/views/test_template_folders.py
```diff
@@ -957,7 +1013,7 @@ def test_should_be_able_to_move_a_sub_item(
         'operation': 'add-new-template',
         'templates_and_folders': [],
         'move_to_new_folder_name': '',
-        'move_to': '__NONE__',
+        'move_to': 'ROOT_FOLDER_ID',
```
quis: This should be the constant, not its quoted name:
```suggestion
        'move_to': ROOT_FOLDER_ID,
```

PR #2632, opened 2019-01-04
Do not format letter statuses on admin app level.

CrystalPea:
This work is now done on api level and the data admin app pulls is ready for use.

Here is the API PR: https://github.com/alphagov/notifications-api/pull/2277

Pivotal ticket: https://www.pivotaltracker.com/story/show/162950719


CrystalPea:
@servingUpAces we should, it's good to have single source of truth. Is it spread across both the apps now?

CrystalPea:
We are doing more shallow fix for now.

PR #2644, opened 2019-01-09
Show better error message if precompiled letters can't be opened

klssmith:
Changed how notifications with the status of `validation-failed` are displayed in the table showing all letters - this used to say 'Cancelled' and now says 'Validation failed'

Added a custom page to be shown if a precompiled letter can't be opened, e.g. because the PDF has no EOF marker or because the file is not a PDF. (Previously we would show the `500` status page.)

Changed the error message for precompiled letters which fail validation due to content outside the printable area.

### Letter table after the changes
screen shot 2019-01-09 at 13 14 52

### Validation failed - file can't be opened
screen shot 2019-01-09 at 13 19 07

### Validation failed - content outside the margins
screen shot 2019-01-09 at 13 18 48

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -373,7 +373,8 @@ def format_notification_status(status, template_type):
             'pending-virus-check': 'Pending virus check',
             'virus-scan-failed': 'Virus detected',
             'returned-letter': 'Delivered',
-            'cancelled': 'Cancelled,'
+            'cancelled': 'Cancelled,',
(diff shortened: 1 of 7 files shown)


quis:
Having second thoughts about the error message.

Letters table | Notification page
---|---
Validation failed | Validation failed – this isn’t a PDF file that Notify can read
Validation failed | Validation failed – content is outside the printable area
Validation failed | Validation failed – too many pages

PR #2649, opened 2019-01-10
Add command to fix failing Pyup branches

quis:
Pyup branches often fail because they introduce new requirements but don’t recreate the `requirements.txt` file.

This commit adds a quick Make command to make fixing those branches easy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,18 @@ test-requirements:
 	         echo "Run 'make freeze-requirements' to update."; exit 1; } \
 || { echo "requirements.txt is up to date"; exit 0; }
 
+# Usage:
+# BRANCH=pyup-update-foobar make freeze-requirements-on-branch
+.PHONY: freeze-requirements-on-branch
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -124,6 +124,18 @@ test-requirements:
 	         echo "Run 'make freeze-requirements' to update."; exit 1; } \
 || { echo "requirements.txt is up to date"; exit 0; }
 
+# Usage:
+# BRANCH=pyup-update-foobar make freeze-requirements-on-branch
+.PHONY: freeze-requirements-on-branch
+freeze-requirements-on-branch:
+	git fetch
+	git checkout $(BRANCH)
+	git reset --hard origin/$(BRANCH)
```
allait: Since this will wipe out any working directory changes, I think we should either prompt or stash things beforehand or otherwise make it obvious this target is destructive

quis:
Probably don’t need this now we’re moving to weekly updates.

PR #2651, opened 2019-01-10
Show letters that have gone into ‘technical failure’

quis:
image

image


app/__init__.py
```diff
@@ -362,18 +362,20 @@ def format_notification_status(status, template_type):
...
             'technical-failure': 'Technical failure',
-            'temporary-failure': 'Temporary failure',
-            'permanent-failure': 'Permanent failure',
-            'delivered': 'Delivered',
-            'sending': 'Sending',
-            'created': 'Sending',
-            'sent': 'Delivered',
-            'pending-virus-check': 'Pending virus check',
+            'temporary-failure': '',
+            'permanent-failure': 'Cancelled',
```
klssmith: Since `permanent-failure` for letters is just the status that we used to use for `cancelled`, maybe these two statuses could be formatted the same way (either both display no description, or both display `Cancelled`).

quis:
```suggestion
            'permanent-failure': '',
```

jonathanglassman:
@klssmith does this need to be reflected in the docs?

klssmith:
@jonathanglassman I don't think we need to update anything in the docs - we're not changing or adding any statuses, this is only a small change to what we display in the frontend

jonathanglassman:
@klssmith ok thank you for the quick response

PR #2664, opened 2019-01-15
Stop automatically resending email verification links

klssmith:
This commit stops a new email verification link from being sent to a
user if they click on an email link which has expired or which has
already been used. Instead, they will be see an error message with a
link to the sign in page. This stops the situation where someone could
log in indefinitely (without the needing to enter their password) by
trying to use a used / expired email verification link and receiving a
new, valid link automatically.

#### Page seen when trying to use a link that has expired / already been used
screen shot 2019-01-17 at 11 05 58

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -129,13 +129,7 @@ class UserApiClient(NotifyAdminAPIClient):
             return True, ''
         except HTTPError as e:
             if e.status_code == 400 or e.status_code == 404:
-                if 'Code not found' in e.message:
-                    return False, 'Code not found'
-                elif 'Code has expired' in e.message:
-                    return False, 'Code has expired'
-                else:
-                    # TODO what is the default message?
-                    return False, 'Code not found'
+                return False, e.message
             raise e
 
     def get_users_for_service(self, service_id):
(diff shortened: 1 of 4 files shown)


app/notify_client/user_api_client.py
```diff
@@ -129,13 +129,7 @@ def check_verify_code(self, user_id, code, code_type):
             return True, ''
         except HTTPError as e:
             if e.status_code == 400 or e.status_code == 404:
-                if 'Code not found' in e.message:
```
allait: Is "Code not found" the full message or was there something in the API response we don't want to display?

karlchillmaid:
Suggested content:

>**The link has expired**
>Sign in again to get a new link.

Reasons:
I'm not sure we need 2 variations. 'Expired' works for both a time-based expiry and expiration due to the link already being used.

Saying less also makes the action clearer.

app/notify_client/user_api_client.py
klssmith: That was the full message from API (https://github.com/alphagov/notifications-api/blob/276a9a38283ef1c1e1e3eda14db5b74206fbf232/app/user/rest.py#L141-L150)

klssmith:
I've updated the content - (the screenshot above now shows the new content)

PR #2674, opened 2019-01-18
Ensure non-gov invited users get added to services

klssmith:
We were adding invited users to services in the `main.add_service` view
function as the last step in the process of inviting users. Since this
view function is decorated with `@user_is_gov_user`, invited users with
non-governmental email addresses would never reach this point and would
be able to register an account but would not get linked to a service.

To fix this, we now add the invited user to the service at the point at
which the user gets activated and also ensure that non-gov users don't
get redirected to a page which they don't have permission to view.

[Pivotal story](https://www.pivotaltracker.com/story/show/163230875)

### DIFF
--- a/app/main/views/verify.py
+++ b/app/main/views/verify.py
@@ -16,6 +16,7 @@ from notifications_utils.url_safe_token import check_token
 from app import user_api_client
 from app.main import main
 from app.main.forms import TwoFactorForm
+from app.models.user import InvitedUser
 from app.utils import redirect_to_sign_in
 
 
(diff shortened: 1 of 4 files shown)


app/main/views/verify.py
```diff
@@ -70,10 +71,28 @@ def activate_user(user_id):
...
+    if invited_org_user:
+        user_api_client.add_user_to_organisation(invited_org_user['organisation'], session['user_details']['id'])
+
     if organisation_id:
         return redirect(url_for('main.organisation_dashboard', org_id=organisation_id))
     else:
         return redirect(url_for('main.add_service', first='first'))
+
+
+def _add_invited_user_to_service(invited_user):
```
klssmith: I've moved this function from `app/main/views/add_service.py`, but have left out the `invite_api_client.accept_invite(service_id, invitation.id)` line since I think we were already accepting the invite before this function was called.

PR #2680, opened 2019-01-23
Add an ‘add logo’ button on the letter template

quis:
Some people don’t know they can have their own logos on letters – it’s not very discoverable on the settings page. Some feedback we get:

> The HM Government Logo is at the top of the letter and we can't see a way of putting the [organisation] logo on

> We are intending to use the letter template feature for the first time and wondered whether the branding is configurable or whether the HM Government header is the standard default.

> Can we replace HM Government logo with our own in the letter? IF yes, then how?

> I don't seem to be able to set the branding on the letters to be [organisation]. it's always HM government. Is there something that needs enabling for this account?

No-one actually wants the HM Government logo (no-one is sending real letters using it). So we should leave the space blank and put a button there prompting people to add their own logo.

***

image

***

Depends on:
- [x] replacing the HM Government logo with a blank image.

### DIFF
--- a/app/assets/stylesheets/views/template.scss
+++ b/app/assets/stylesheets/views/template.scss
@@ -45,3 +45,9 @@
   top: 51px; // aligns bottom edge to bottom of postmark
   right: 145px; // Aligns right edge to midpoint of postmark and fold
 }
+
+.edit-template-link-letter-branding {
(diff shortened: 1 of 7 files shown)


leohemsted:
i pressed the wrong button sorry 🙈

PR #2692, opened 2019-01-29
Let users switch channels on and off by themselves

quis:
We let people do this for letters already. We should let them do it for emails and texts too, rather than have to email us.

Before | After 
---|---
image | image
image | image
image | image


tests/app/main/views/test_service_settings.py
```diff
@@ -2542,19 +2542,91 @@ def test_should_set_sms_allowance(
     )
 
 
+def test_old_set_letters_page_redirects(
+    client_request,
+):
+    client_request.get(
+        'main.service_set_letters',
+        service_id=SERVICE_ONE_ID,
+        _expected_status=302,
```
leohemsted: should this return a `301 MOVED PERMANENTLY`? (you can specify it like `redirect(url..., code=301)`)

quis: 2fae31a

PR #2695, opened 2019-01-29
Edit postage per template and not through service settings

CrystalPea:
### Work done:
* move editing postage for a letter template to a new, separate flow
* remove the option to choose postage per service
* get rid of choose_postage switch (all services will be able to choose postage per template)

### TODO:
* redesign postage section (postage information and placement of edit button) in the letter template view
* when everything is ready and tested, probably in a separate PR, remove "choose_postage" feature flag

### FE part of this story:
https://www.pivotaltracker.com/n/projects/1443052/stories/162959391

#### Edit postage screen:
screen shot 2019-01-30 at 15 59 34

#### View letter template screen: (this will probably be redesigned)
screen shot 2019-01-29 at 17 22 22


app/templates/partials/templates/guidance-postage.html
```diff
@@ -0,0 +1,16 @@
...
+<p>
+  Letters sent before 5:30pm are dispatched the next working day (Monday to Friday).
+</p>
+<p>
+  First class letters are delivered one day after they're dispatched. Second class letters are delivered 2 days after they're dispatched.
+</p>
+<p>
+  Royal Mail delivers from Monday to Saturday, excluding bank holidays.
+</p>
+  <a href="{{ url_for('.pricing') + '#letters'}}">See a list of postage prices.</a>
```
karlchillmaid: Full stop should be outside the link text

CrystalPea: I moved it outside... doesn't it look a little silly as an only character outside the link though? 🤔

karlchillmaid: Might be better just to have `postage prices` as the link?

CrystalPea: Here is just `postage prices` as a link, and everything as a link, but without the full stop:

screen shot 2019-01-30 at 10 33 47
   
screen shot 2019-01-30 at 10 34 18

CrystalPea: I think the option you proposed might be the best.

karlchillmaid: @quis When we include 'Back' links, do we do anything to distinguish them from the body content of the page?

CrystalPea: I think we are not very consistent, but we sometimes use grey button:

screen shot 2019-01-30 at 10 44 02

karlchillmaid: @quis Secondary question: would beneath the `Save` button be a better spot for the backlink?

CrystalPea: I guess it would be much more visible this way?

karlchillmaid: I like it.
But I don't know the reasons behind why we don't use the grey button consistently.

CrystalPea: We use it in settings, but not in API integration:
screen shot 2019-01-30 at 10 43 22
screen shot 2019-01-30 at 10 43 13
screen shot 2019-01-30 at 10 43 02

karlchillmaid: My preference would be to use it in this case.

CrystalPea: Do you dig it? :D
screen shot 2019-01-30 at 15 59 34

PR #2701, opened 2019-01-30
Improve display of folder path when deeply nested

quis:
It’s a bit rudimentary to only show the current place in the hierarchy and the parent. You lose a sense of how deep you are.

But we can’t just show the full path, because it can be arbitrarily long. So what this commit does is show the full path, but truncates the display of any items further-up than the current folder or its parent.

This also helps disambiguate between folders and templates, because folders are always shown with the folder icon.

This probably won’t affect many teams, because we don’t anticipate a lot of deep nesting.

# Breadcrumb

## Before 

image

## After 

template nesting

# Search

## Before

image

## After 

image

# Initial sketch

image-1


quis:
# Wrap, don’t truncate display of folder name in H1

This commit applies a carefully selected blend of `white-space`, `display` and `background-position` to ensure that:

- you can always see the full name of the current folder in the heading  (which is useful because people might have multiple folders with the same name, differentiated by some kind of suffix)
image

- the clickable are of a single folder or template in the list is 100% width, for the biggest possible hit area
  image

- the name of a folder or template wraps underneath the folder icon (ie the folder icon is treated like a normal character, not like a bullet point)
  image

PR #2702, opened 2019-01-30
Fix bugs with sticky overflow code

tombye:
The [pull request which adds code for sticky elements whose content overflows](https://github.com/alphagov/notifications-admin/pull/2682) introduced a few bugs:
- filtering the templates/folders with search doesn't update the sticky position
- when the page loads it scrolls the bottom of the window to meet the bottom of the form buttons
- the padding each sticky element has creates too much space between them when stacked

## Filtering the templates/folders with search doesn't update the sticky position

The `recalculate` method was being called by the live-search module but a bug in the `add` method meant the positions and dimensions of all sticky elements we store weren't being updated. This was fixed by https://github.com/alphagov/notifications-admin/commit/177f30248e342c20f4e1848b34d8ee0c510b46b6.

## When the page loads it scrolls the bottom of the window to meet the bottom of the form buttons

Every time the code in `templateFolderForm.js` called `recalculate`, it set the mode to 'dialog'. This kicked in the behaviour where it sorts the stack of sticky elements and scrolls the page if needed.

https://github.com/alphagov/notifications-admin/commit/f4d9c379403d102350cea4e8ad7300b86bd94473 moves setting the mode out of `recalculate`. This meant the `templateFolderForm.js` code could set mode to 'dialog' only when it changed the state of the sticky to one that had multiple items. The state on page load doesn't, so no scroll happens.

## The padding each sticky element has creates too much space between them when stacked

The padding needed to be considered when the CSS for `bottom` is set on each sticky element in the stack. https://github.com/alphagov/notifications-admin/commit/707c426b9aa5b7730ac49dff21763b27507d08bd adds that to the offset calculation.

### DIFF
--- a/app/assets/javascripts/stick-to-window-when-scrolling.js
+++ b/app/assets/javascripts/stick-to-window-when-scrolling.js
@@ -272,6 +272,9 @@
     this.CSS_SELECTOR = selector;
     this.STOP_PADDING = 10;
   };
+  Sticky.prototype.setMode = function (mode) {
+    _mode = mode;
+  };
   Sticky.prototype.getWindowDimensions = function () {
     return {
       height: $(global).height(),
@@ -360,7 +363,7 @@
     }
   };
   // Recalculate stored dimensions for all sticky elements
-  Sticky.prototype.recalculate = function (opts) {
+  Sticky.prototype.recalculate = function () {
     var self = this;
     var onSyncComplete = function () {
       self.setEvents();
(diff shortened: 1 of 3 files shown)


quis:
Not sure if these are new bugs introduced by this PR or further ones I’ve uncovered but…

***

If you select and deselect a template multiple times the page starts to jump:

jump-select

***

Same thing happens when you open/close something in the sticky:

jump-open

tombye:
Looks like those bugs were introduced by this PR.

The first one was because the code in `templateFolderForm.js` changed the mode to 'dialog' but never back.

The second one was because the function that scrolls the page when items pop off the dialog stack was being called even if this didn't happen.

PR #2725, opened 2019-02-04
Warn if a file has been sent already

quis:
We have some teams who have a series of files they have to send each day. It’s easy to get muddled up and accidentally send the same file again, if you lose track of which ones you’ve already sent

This commit blocks you from sending the same combination of template version and filename more than once on the same day[1].

image

This won’t affect teams who re-use the same template to give (for example) updates on an incident for business continuity. These teams edit the template between each send, thereby updating the version
number of the template.

1. This is based on how the `limit_days` argument to the API works - you can dig into the code here: https://github.com/alphagov/notifications-api/blob/2bd4f74ad0e24f7066e1a6bd4d0e515b4b925e10/app/dao/jobs_dao.py#L50

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -609,6 +609,9 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
         )),
         required_recipient_columns=OrderedSet(recipients.recipient_column_headers) - optional_address_columns,
         preview_row=preview_row,
+        sent_previously=job_api_client.has_sent_previously(
+            service_id, template.id, db_template['version'], request.args.get('original_file_name', '')
+        )
     )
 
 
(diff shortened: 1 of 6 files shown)


app/main/views/send.py
```diff
@@ -609,6 +609,9 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
         )),
         required_recipient_columns=OrderedSet(recipients.recipient_column_headers) - optional_address_columns,
         preview_row=preview_row,
+        sent_previously=job_api_client.has_sent_previously(
+            service_id, template.id, db_template['version'], request.args.get('original_file_name', '')
```
allait: When is `original_file_name` not present?

quis: If the user/browser has modified the URL.

PR #2726, opened 2019-02-04
$.inArray returns the same value as Array.indexOf

tombye:
Should have checked the docs. Didn't and so assumed it returned a boolean value by the name.

https://api.jquery.com/jQuery.inArray/

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -147,7 +147,7 @@
       );
 
       // use dialog mode for states which contain more than one form control
-      if ($.inArray(this.currentState, ['move-to-existing-folder', 'add-new-template']) !== -1) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/templateFolderForm.js
```diff
@@ -147,7 +147,7 @@
       );
 
       // use dialog mode for states which contain more than one form control
-      if ($.inArray(this.currentState, ['move-to-existing-folder', 'add-new-template'])) {
+      if ($.inArray(this.currentState, ['move-to-existing-folder', 'add-new-template']) !== -1) {
```
quis: Is there any benefit to the jQuery in that case?

```suggestion
      if (['move-to-existing-folder', 'add-new-template'].indexOf(this.currentState) !== -1) {
```

tombye: Not if our version of babel polyfills `Array.indexOf`. I'll check...

PR #2728, opened 2019-02-04
Fix bug that shows edit postage button on template to everyone

CrystalPea:


quis:
Merge now, add tests later 😅

PR #2730, opened 2019-02-05
Fix hover folder icon in IE10/11

quis:
Duplicates e0ecc95ac634a6a300d919af59d5364b464ce2fd

Copies the code from the normal folder icon, and manually tweaks the colour, to also get the benefits of minification.

***

IE 10 supports using SVG[1] but has some buggy behaviour when they’re used as background images.

Without an explicit width/height it stretches the viewBox of the SVG to fill the containing element. This causes the content of the file to display centered within the viewBox.

Explicitly setting the width and height seems to be the thing that fixes this. Out of the suggested fixes on Stackoverflow[2] this one seems to be the most straightforward.

1. https://caniuse.com/#feat=svg
2. https://stackoverflow.com/questions/17944354/svg-background-image-position-is-always-centered-in-internet-explorer-despite-b


quis:
image

PR #2734, opened 2019-02-05
Set letter branding for a service

leohemsted:
Now we read from the new table, not from DVLA table.

We update the services' letter_branding field. It correctly removes the row from the mapping table if you set to `None`.

Note: Previously, HM Government was the "default" option. now you can have no branding, so we've had to change some fields that expect it to be set.

For now, when you actually generate a letter with no letter branding set it'll fall back to the dvla_organisation_id to get the logo, we'll change it to properly have a blank logo in a future story.


app/notify_client/letter_branding_client.py
```diff
@@ -1,10 +1,18 @@
...
 
-    def get_letter_branding(self):
-        return self.get(url='/dvla_organisations')
+    @cache.set('letter_branding-{branding_id}')
+    def get_letter_branding(self, branding_id):
+        return self.get(url='/letter-branding/{}'.format(branding_id))
+
+    @cache.set('letter_branding')
+    def get_all_letter_branding(self, sort_key=None):
+        brandings = self.get(url='/letter-branding')
```
klssmith: Not sure we need the optional `sort_key` here - the brandings are already sorted by name in api, and the domain is optional so won't always be available to sort on

app/templates/views/service-settings/set-letter-branding.html
```diff
@@ -13,12 +13,14 @@
   <h1 class="heading-large">Set letter branding</h1>
   {% call form_wrapper() %}
     {{ live_search(target_selector='.multiple-choice', show=True, form=search_form, label='Search by name') }}
```
klssmith: `show` should be set to `show_search_box` so it only shows with more than 6 brands

leohemsted: i'm not really too worried about that since we've got like 41 rows in dvla_organisation so that's never going to be a concern

PR #2741, opened 2019-02-06
Template form focus

tombye:
Improvements to how focus moves around when we're changing the HTML on the templates page.

https://www.pivotaltracker.com/story/show/163524995

## Context

Browsers keep a sequence of elements that can be focused. When you tab through a page you're moving through that sequence.

The JS on the templates page changes the HTML which sometimes removes elements from the sequence, effecting tabbing.

For example, if a user clicks 'new folder', the current focus (the 'new folder' button) has been removed from the DOM but we don't move focus to another element.

These changes ensure focus is moved in the way we intend, that helps users complete the task they're on.

## Changes

- when 'new template' is clicked, we move focus to the radios for selecting the type of template
- when templates/folders are selected and 'move' is clicked, we move focus to the radios for selecting the folder to move them to
- when 'new folder' or 'add to new folder' is clicked, we move focus to the textbox for the folder name

## How to test this

Try all the tasks that are effected by the change in HTML, using the keyboard only:
- creating a new template
- creating a new folder
- moving a template/folder to an existing folder
- moving a template/folder to a new folder

...and for the above, try clicking the 'cancel' link when you see it.

You should be able to do all of them without making any extra tabbing movements to correct not being in the right place.

### DIFF
--- a/app/assets/javascripts/stick-to-window-when-scrolling.js
+++ b/app/assets/javascripts/stick-to-window-when-scrolling.js
@@ -576,9 +576,8 @@
 
       el.removeStickyClasses(this);
       $el.css('width', '');
-      if (_mode === 'dialog') {
-        dialog.releaseEl(el, this);
-      }
+      // clear styles from any elements stuck while in a dialog mode
+      dialog.releaseEl(el, this);
       el.removeShim();
       el.release(this);
     }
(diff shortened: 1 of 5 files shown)


tombye:
I pushed 2 new commits since @quis approved. They fix bugs found in testing I ended up doing after this pull request was opened.

They're non-destructive and just fix things that were causing edge-case problems.

This has now been tested with all the tasks in the description across our browser matrix.

PR #2750, opened 2019-02-11
Embiggen all the things

quis:
Bigger hit areas and font sizes make things easier to click and read.

# Team members

image

image

# Settings

image

image

# Your profile 

image

image

### DIFF
--- a/app/assets/stylesheets/components/navigation.scss
+++ b/app/assets/stylesheets/components/navigation.scss
@@ -23,6 +23,10 @@
       right: 0;
       padding: 13px 0 9px 15px;
 
+      &:hover {
+        color: $link-hover-colour;
+      }
+
       &:focus {
         outline: none;
         border-bottom: 4px solid $yellow;
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -219,12 +219,38 @@
   a {
 
     display: block;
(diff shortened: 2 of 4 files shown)


quis:
@tombye what does the code for the pseudo element technique look like?

tombye:
GOVUK Frontend uses one for their buttons. Sorry, should have posted this with my previous comment:

https://github.com/alphagov/govuk-frontend/blob/47407edcb5694afb36b6aa6b2a6050a5e92f9ef4/src/components/button/_button.scss#L89

quis:
The pseudo-element technique expands the click area without expanding the area of the `:active` state:

image

I prefer expanding the active area too because it helps you understand how big the clickable area is for next time.

This is consistent with how the template navigation works, ie we do this:
image

…not this:
image

tombye:
Maybe I'm missing something but can't that be achieved by including a selector with the `:active` state in it?

I mean like GOVUK Frontend does with the pseudo-element when their button is active:

https://github.com/alphagov/govuk-frontend/blob/47407edcb5694afb36b6aa6b2a6050a5e92f9ef4/src/components/button/_button.scss#L115

quis:
@tombye yeah, that works. Added in 1e82b6f8. It’s more code but if you think that technique is better…

<sup>Also I ended up having to use borders for the active state anyway because a pseudo element can’t underlap it’s parent.</sup>

PR #2766, opened 2019-02-14
Proposal: how the data passes between admin and API for folder permissions

quis:
Outlines:
- the new methods we’d need on the service model
- the (one) new API endpoint we’d need to make this possible


quis:
Presuming y’all have got this now.

PR #2775, opened 2019-02-15
Update message status descriptions

karlchillmaid:
Update message status descriptions so they're consistent with the descriptions in the technical documentation. Add letter statuses.

### DIFF
--- a/app/templates/views/using-notify.html
+++ b/app/templates/views/using-notify.html
@@ -43,27 +43,41 @@
              temporary failures"
         style="width: 100%;"
     >
+    <h2 class="heading-medium">Message states – emails and text messages</h2>
+    <h3 class="heading-small">Created (API only)</h3>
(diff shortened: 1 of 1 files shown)


app/templates/views/using-notify.html
```diff
@@ -43,27 +43,40 @@ <h2 id="messagedeliveryandfailure" class="heading-medium">Sending messages</h2>
              temporary failures"
         style="width: 100%;"
     >
+    <h2 class="heading-medium">Message statuses – emails and text messages</h2>
     <h3 class="heading-small">Sending</h3>
-    <p>This means that we’ve sent the message to our providers and are waiting for delivery information.</p>
+    <p>GOV.UK Notify has sent the message to the provider. The provider will try to deliver the message to the recipient. GOV.UK Notify is waiting for delivery information.</p>
+    <h3 class="heading-small">Pending (text messages only)</h3>
```
klssmith: We display the status of 'Pending' text messages as 'Sending', so we can leave Pending out of the list of statuses.

PR #2780, opened 2019-02-18
Replace platform admin service setting buttons with forms

allait:
### Before
screen shot 2019-02-18 at 16 48 48

### After
screen shot 2019-02-18 at 16 49 02

### Add a new boolean radios fields and change forms to use it

This adds a new OnOffField class that implements a boolean field that is rendered as two On / Off radio buttons. This allows us to avoid comparing 'on' and 'off' string values in the views since the field takes care of transforming form data into python booleans.

This also adds a form class that can be used for any single On / Off switch forms (e.g. service permissions).

### Replace platform admin service setting buttons with forms

Most of the existing platform admin buttons on the service settings page used to issue GET requests to switch service settings. This means they weren't protected by CSRF. On top of that as our number of service permissions increases over time a lot of buttons on the page made it hard to work with.

To fix these issues we replace most of the buttons with rows in the platform admin settings table. Each setting has a 'Change' link that leads to a page with an On/Off switch form.

This removes "research mode" switch completely since we're planning to deprecate it in the future and we don't expect to switch any new services into research mode at the moment.

Most service permissions are now handled by a shared endpoint that is parameterized with the permission name. Some permissions that require some additional logic before they can be toggled (like document upload, which requires setting a contact address) have separate initial endpoints that redirect to `set_service_permission`.

"Archive", "Suspend" and "Resume" actions are kept as buttons since they display a confirmation banner (which is a CSRF-protected form) and they're not easily represented as an On/Off switch.


app/main/views/service_settings.py
```diff
@@ -211,63 +224,73 @@ def submit_request_to_go_live(service_id):
...
+        name="Make service live",
+        enabled=not current_service.trial_mode
     )
-    return redirect(url_for('.service_settings', service_id=service_id))
 
+    if form.validate_on_submit():
+        current_service.update(
+            # TODO This limit should be set depending on the agreement signed by
+            # with Notify.
+            message_limit=250000 if form.enabled.data else 50,
```
allait: @quis is this correct? I think the previous logic was the opposite of this, but I'm also not suer if this setting does anything

app/main/forms.py
```diff
@@ -696,23 +696,37 @@ def validate_letter_contact_block(self, field):
...
+        super().__init__(label, choices=[
+            (True, 'On'),
+            (False, 'Off'),
+        ], *args, **kwargs)
 
-    def __init__(self, channel, *args, **kwargs):
+    def process_formdata(self, valuelist):
+        if valuelist:
+            value = valuelist[0]
+            self.data = (value == 'True') if value in ['True', 'False'] else value
```
leohemsted: should this raise an exception if it's not True/False, rather than the field being one of `[True, False, 'dumb_typo']`? (i'm not too sure how these field functions work though)

allait: The values still get compared to the available choices in the form after this function is called, so it will return a validation error.

PR #2793, opened 2019-02-20
Remove redundant API call

quis:
The current user already has a list of service IDs. The current user
- is an API call we have to make anyway to render this page
- is usually cached in Redis


quis:
Need to check if this still works if any of your services belong to an organisation.

PR #2796, opened 2019-02-21
Update new template titles

karlchillmaid:
To keep the H1 titles of new templates consistent with the sticky menu options:

* change `Add email template` to `New email template`
* change `Add text message template` to `New text message template`
* change default letter template name `Untitled` to `New letter template`


karlchillmaid:
@quis Do these changes mean I need to update any tests?

PR #2819, opened 2019-02-28
Edit user folder permissions

allait:
screen shot 2019-02-28 at 14 03 34

### Make radio form components reusable for nested checkboxes

For the template folders permission editing we need a nested checkboxes form that is similar to "move folder" input, except it's using checkboxes instead of radio buttons.

This moves most of the macros into a shared "select-input" components file, which are wrapped by the existing radios.html by setting the required input type.

### Display nested folders permissions form on user permissions page

We're reusing the logic for the `move_to` nested radios field for the user folder permissions nested checkboxes.

The main difference between the two forms (aside from the different input type) is that "Move" form contains the root "Templates" as an option, whereas the folder permissions doesn't.

It turns out that, because of the way NestedFieldMixin.children and select_nested macro are implemented the easiest way to get the desired folder permissions behaviour is to add the root folder as a choice with a `None` value and `NONE_OPTION_VALUE = None` set on the field, which allows the `child_map` to be constructed but doesn't display the root folder checkbox itself since it gets overwritten in the final `child_map`.

### Send updated user folder permissions to the API

Integrates the folder permissions form with the updated API endpoint to store changes in the user folders.

Since user folder permissions are returned in the full list of template folders for the service we need to invalidate the cache key for it each time we update user permissions.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -418,6 +418,15 @@ PermissionsAbstract = type("PermissionsAbstract", (StripWhitespaceForm,), {
 
 
 class PermissionsForm(PermissionsAbstract):
+    def __init__(self, all_template_folders=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if all_template_folders:
+            self.folder_permissions.all_template_folders = all_template_folders
+            self.folder_permissions.choices = [
+                (item['id'], item['name']) for item in ([{'name': 'Templates', 'id': None}] + all_template_folders)
+            ]
+
+    folder_permissions = NestedCheckboxesField('Folder permissions')
 
     login_authentication = RadioField(
         'Sign in using',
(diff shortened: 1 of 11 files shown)


app/main/forms.py
```diff
@@ -276,6 +348,15 @@ def __init__(self, invited_org_user):
...
 class PermissionsForm(PermissionsAbstract):
+    def __init__(self, all_template_folders=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if all_template_folders is not None:
+            self.folder_permissions.all_template_folders = all_template_folders
+            self.folder_permissions.choices = [
+                (item['id'], item['name']) for item in ([{'name': 'Templates', 'id': None}] + all_template_folders)
+            ]
+
+    folder_permissions = NestedCheckboxesField('Folder permissions')
```
karlchillmaid: @quis Did we discussing making the label more descriptive? `Folders this team member can see` or something similar.

PR #2823, opened 2019-03-01
Only show folders that user has permission to see

CrystalPea:
When a user accesses their templates, only show the folders (and templates within those folders) where the user has permission to see those folders.

If the child folder is visible, but the parent is not, move child folder up in the tree and add parent name to the child name, like this:
screen shot 2019-03-07 at 16 45 33

Pivotal ticket: https://www.pivotaltracker.com/story/show/163755337

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -385,6 +385,19 @@ class Service():
         return {folder['id'] for folder in self.all_template_folders}
 
     def get_user_template_folders(self, user_id):
+        """Returns a modified list of folders a user has permission to view
+
+        For each folder, we do the following:
+        - if user has no permission to view the folder, skip it
+        - if folder is visible and its parent is visible, we add it to the list of folders
+        we later return without modifying anything
+        - if folder is visible, but the parent is not, we iterate through the parent until we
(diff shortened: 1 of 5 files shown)


app/models/service.py
```diff
@@ -379,16 +384,61 @@ def all_template_folders(self):
...
+                user_folders.append(folder)
+            else:
+                folder_attrs = {
+                    "id": folder["id"], "name": folder["name"], "parent_id": folder["parent_id"],
+                    "users_with_permission": folder["users_with_permission"]
+                }
+                while folder_attrs["parent_id"] is not None:
+                    folder_attrs["name"] = parent["name"] + " / " + folder_attrs["name"]
+                    if parent["parent_id"] is None:
+                        parent = None
```
leohemsted: don't think you need this line? next time we loop it'll break out because folder_attrs['parent_id'] will be None, so we'll never touch parent again (and if we did touch it it'd crash right)

```suggestion
```

CrystalPea: You are right, this statement is not needed, I will remove it 👍 . And yeah, writing some of these tests was a wild ride, even with Alexey 😆

PR #2827, opened 2019-03-04
When going live, make reply-to address and text message sender dependent on estimated sending volumes

quis:
We have a number of go live requests where people have said they’re sending text messages, but haven’t changed the text message sender from the default of `GOVUK` (we ask teams who aren’t central government to do this). Or they say they’re sending emails but have no email templates, or any email reply-to addresses.

At the moment we don’t prompt them to, because we look at what kinds of templates they have to indicate what kinds of messages they’re going to send.

Now that we explicitly ask for their estimated sending volumes we should use this to determine what we prompt them about because it’s a stronger signal of intent than what templates they’ve set up.

For templates we’re going to ask that they at least have one kind of template of the sort they intend to send. We could be even stricter and say they have to have at least one of each type they
expect to send, but we think that we can get a pretty good idea from just looking at one template whether they’re doing something sensible.

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -269,7 +269,7 @@ class Service():
     @property
     def needs_to_change_sms_sender(self):
         return all((
-            self.has_sms_templates,
+            self.volume_sms,
             self.shouldnt_use_govuk_as_sms_sender,
             self.sms_sender_is_govuk,
         ))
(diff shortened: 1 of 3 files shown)


quis:
Updates:
- Roll back eff86d2ac (Make required templates dependent on volumes). We’ve decided that it’s confusing that you can’t make the templates checklist item complete until you’ve told us the volumes, and that it would also be confusing to hide it until you’ve completed the volumes.
- Show the email reply to address and text message sender tasks if you have relevant templates but haven’t yet told us your volumes (on the assumption that you probably intend to send emails if you have some email templates). See 87c62e6 for more detail.

PR #2828, opened 2019-03-04
Hide ‘go live’ pages from already live services

quis:
They shouldn’t be able to revise their estimates or make another request once they’re live.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -147,6 +147,13 @@ def service_name_change_confirm(service_id):
 @user_has_permissions('manage_service')
 def estimate_usage(service_id):
 
+    if not current_service.trial_mode:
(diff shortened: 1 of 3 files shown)


quis:
We shouldn’t do this while we still might need people to retroactively fill in their estimated volumes.

app/main/views/service_settings.py
```diff
@@ -180,6 +187,13 @@ def estimate_usage(service_id):
 @user_has_permissions('manage_service')
 def request_to_go_live(service_id):
```
servingUpAces: It's a bit weird that if you visit this page when the service is live you just end up back on the service-settings page. Do you not want to be able to visit the page at all once the service is live?

quis:
Don’t think we actually care that much about this.

PR #2850, opened 2019-03-15
Set folder permissions when creating and accepting invites

klssmith:
#### Set folder permissions when inviting users
Added a folder permissions form to the page to invite users to services.
This only shows if the service has 'edit_folder_permissions' enabled,
and all folder checkboxes are checked by default. This change means that
InviteApiClient.create_invite now sends folder_permissions through to
notifications_api (so invites get created with folder permissions).

#### Set folder permissions when accepting invites
Started passing the folder_permissions through to notifications-api when
accepting an invite. This changes UserApiClient.add_user_to_service to
send folder_permissions to notifications_api so that new users get folder
permissions when they are added to the service.

[Pivotal story](https://www.pivotaltracker.com/story/show/163756170)

### DIFF
--- a/app/main/views/invites.py
+++ b/app/main/views/invites.py
@@ -70,7 +70,8 @@ def accept_invite(token):
                 user_api_client.update_user_attribute(existing_user.id, auth_type=invited_user.auth_type)
             user_api_client.add_user_to_service(invited_user.service,
                                                 existing_user.id,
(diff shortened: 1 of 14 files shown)


klssmith:
Needs to be merged after https://github.com/alphagov/notifications-api/pull/2398

PR #2865, opened 2019-03-20
Check folder permissions

allait:
### Add a user method to check folder permission

User model is the most natural place for a permission check method, however this means that we need to pass the full user object to service model methods and TemplateList instead of user_id.

### Add a helper Service method to get a template given user has permission

Checks if the user has access to the template's parent folder and either returns the template or a 403 response.

This method should be used instead of calling service_api_client from the views.

### Replace sevice api client get template calls with Service methods

Instead of using the API client directly views are now calling one of two Service model methods:

`get_template` is used for view actions, where the user should see the template page even if they don't have access to the template folder (since all templates are still inked from the dashboard or the sent notifications pages).

`get_template_with_user_permission_or_403` will check if the user has access to the template's folder first and return 403 otherwise. This method is used for any endpoints that result in an action: editing template attributes, deleting templates or sending messages.

### Add folder permission check to copy template endpoint

Copying a template from another service is one place where we can't use the `current_service` method since the source template can belong to a different service the user has access to, so we're using an API client method.

### Hide template and folder action links if user doesn't have folder access

Hides action links ('Send', 'Edit', 'Delete' and 'Redact' fro templates and 'Manage' for template folders) and buttons ('New template', 'New folder') if the user doesn't have permission to view current folder or template's parent folder.

Screen Shot 2019-03-20 at 15 42 09

Screen Shot 2019-03-20 at 16 06 45

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -203,7 +203,7 @@ def init_app(application):
 
     @application.context_processor
     def _attach_current_user():
-        return{'current_user': current_user}
+        return {'current_user': current_user}
 
     @application.context_processor
     def _nav_selected():
--- a/app/main/views/conversation.py
+++ b/app/main/views/conversation.py
@@ -52,7 +52,7 @@ def conversation_reply(
         templates_and_folders=TemplateList(
             current_service,
             template_folder_id=from_folder,
-            user_id=current_user.id,
+            user=current_user,
             template_type='sms'
         ),
         template_folder_path=current_service.get_template_folder_path(from_folder),
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -100,7 +100,7 @@ def edit_user_permissions(service_id, user_id):
         service_id,
         folder_permissions=[
             f['id'] for f in current_service.all_template_folders
-            if user_id in f.get('users_with_permission', [])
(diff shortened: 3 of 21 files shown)


quis:
Code looks good. Could probably do with some tests to make sure that stuff which is supposed to be hidden is hidden.

And maybe a different message for the folder you don’t have permission to see, so instead of:
> This folder is empty

Something like:
> You don’t have permission to see the templates in this folder

cc/ @karlchillmaid

karlchillmaid:
> Code looks good. Could probably do with some tests to make sure that stuff which is supposed to be hidden is hidden.
> 
> And maybe a different message for the folder you don’t have permission to see, so instead of:
> 
> > This folder is empty
> 
> Something like:
> 
> > You don’t have permission to see the templates in this folder
> 
> cc/ @karlchillmaid

We should use `do not` instead of `don't`.

`You do not have permission to see the templates in this folder`

allait:
As discussed with @quis, I've added a change to unlink folders the user doesn't have a permission for from the template / folder path. So the page with "This folder is empty" is no longer linked to.

Screen Shot 2019-03-29 at 16 38 08

quis:
Functionally this is good to go now.

PR #2866, opened 2019-03-20
fix user permissions save button sometimes deleting

leohemsted:
when you hit the delete button, it flashes the delete button and takes you to the `/service/../user/../delete` url. If you then click the save button, it would make a POST to the delete URL... and delete the user.

now the page stays on the edit url, but adds a `?delete=yes` query string. The dangerous flash banner now has an action field which defines where the browser will make the POST to (which remains at /delete).


app/templates/components/banner.html
```diff
@@ -16,14 +16,14 @@ <h1 class="banner-title">{{ subhead }}</h1>
     </p>
     {% endif %}
     {% if delete_button %}
-      <form method='post'>
+      <form method='post' {% if action %}action="{{action}}"{% endif %}>
```
quis: Would be better to use:
```suggestion
      {% call form_wrapper(action=action) %}
```

app/templates/views/edit-user-permissions.html
```diff
@@ -2,12 +2,21 @@
...
   {{ user.name or user.email_localpart }}
 {% endblock %}
 
 {% block maincolumn_content %}
+  {% if delete %}
+    {{ banner(
+      'Are you sure you want to remove {}?'.format(user.name),
+      type='dangerous',
+      delete_button='Yes, remove',
+      action=url_for('.remove_user_from_service', service_id=current_service.id, user_id=user.id)
```
allait: Can we use `.confirm_remove_user_from_service` here and then unless I'm missing something I don't think we need the `remove_user_from_service` endpoint at all

PR #2869, opened 2019-03-21
Update pyexcel, pyexcel-io and werkzeug

klssmith:
Upgraded pyexcel-io and pyexcel, which don't introduce any significant changes.

However, these cause the Werkzeug version to increase from 0.14.1 to 0.15.0 which requires some changes:

* ProxyFix now needs to be imported from a different location
* The status code of RequestRedirect has changed from 301 to 308. Since
status code 308 is not currently supported on Internet Explorer with
Windows 7 and 8.1, this subclasses RequestRedirect to keep the status
code of 301.

[Werkzeug changelog](https://werkzeug.palletsprojects.com/en/0.15.x/changes/#version-0-15-0)

Closes #2847 

Closes #2857

### DIFF
--- a/app/proxy_fix.py
+++ b/app/proxy_fix.py
@@ -3,7 +3,7 @@ from werkzeug.middleware.proxy_fix import ProxyFix
 
 class CustomProxyFix(object):
     def __init__(self, app, forwarded_proto):
-        self.app = ProxyFix(app)
+        self.app = ProxyFix(app, x_for=1, x_proto=1, x_host=1, x_port=0, x_prefix=0)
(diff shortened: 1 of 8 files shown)


app/proxy_fix.py
```diff
@@ -1,4 +1,4 @@
-from werkzeug.contrib.fixers import ProxyFix
+from werkzeug.middleware.proxy_fix import ProxyFix
```
allait: It sounds like ProxyFix defaults might've changed, might be worth checking if we're affected https://github.com/pallets/werkzeug/issues/1484

PR #2897, opened 2019-04-04
Precompiled show preview with overlay when validation fails

CrystalPea:
Has to go after this template-preview PR: alphagov/notifications-template-preview#315
And this API PR: https://github.com/alphagov/notifications-api/pull/2437

Part of this story: https://www.pivotaltracker.com/story/show/164798738

#### Screenshots:
cursor on the postage sign:
Screen Shot 2019-04-04 at 11 57 14

cursor not on the postage sign:
Screen Shot 2019-04-04 at 11 57 05

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -60,7 +60,9 @@ def view_notification(service_id, notification_id):
 
     if notification['template']['is_precompiled_letter']:
         try:
-            file_contents = view_letter_notification_as_preview(service_id, notification_id, "pdf")
+            file_contents = view_letter_notification_as_preview(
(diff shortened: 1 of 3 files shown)


quis:
Nice work 😎 

Does it show up on the PDF download as well?

CrystalPea:
@quis thanks 😊. It doesn't show on the download for now, should I add that?

quis:
@CrystalPea Yeah, I think that will help people find the problem if their text is a tiny bit into the red.

CrystalPea:
@quis this is done now ☺️ 

 I still need to write some unit tests and hopefully do some refactoring with Leo when I'm back though :)

PR #2907, opened 2019-04-11
Block incomplete requests to go live

quis:
Dealing with users who request to go live but haven’t completed all the steps still represents a significant support overhead for our team. We’ve made some improvements to the percentage of incomplete requests with a better page design, but ultimately because it still shows the button people think it’s OK to press the button while some of the items on the page still say [Not completed].

We can do this now because organisations are in the database, which means we can mark the agreement signed as soon as we get it back, without having to deploy code.

***

image

### DIFF
--- a/app/templates/views/service-settings/request-to-go-live.html
+++ b/app/templates/views/service-settings/request-to-go-live.html
@@ -58,7 +58,7 @@
         </p>
       {% elif (not current_service.go_live_checklist_completed) or (show_agreement and not agreement_signed) %}
         <p>
-          No go
+          You must complete these steps before you can request to go live.
         </p>
       {% else %}
         <p>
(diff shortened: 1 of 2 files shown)


app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -52,7 +52,15 @@ <h1 class="heading-large">Before you request to go live</h1>
...
         {% endif %}
       {% endcall %}
-      {% if current_user.is_gov_user %}
+      {% if not current_user.is_gov_user %}
+        <p>
+          Only team members with a government email address can request to go live.
+        </p>
+      {% elif (not current_service.go_live_checklist_completed) or (show_agreement and not agreement_signed) %}
+        <p>
+          No go
```
karlchillmaid:
```suggestion
          You must complete these steps before you can request to go live.
```

PR #2909, opened 2019-04-12
move manifest to single jinja template

leohemsted:
uses the same pattern as other apps. Infers route based on app name, you can declare more in the dict at the top if you need to.


klssmith:
Does this change mean that we can delete the `scripts/generate_manifest.py` file now?

leohemsted:
@klssmith removed and squashed into 9547fed

PR #2912, opened 2019-04-12
Add Jest framework for testing JavaScript

tombye:
This adds the [Jest](https://jestjs.io/en/) testing framework to let us write tests for our JavaScript.

Part of this story: https://www.pivotaltracker.com/story/show/164777921

Jest was chosen because it:
1. is well supported and [documented](https://jestjs.io/docs/en/getting-started)
2. is easy to pick up and get started with
3. has good [support for testing JavaScript that uses jQuery](https://jestjs.io/docs/en/tutorial-jquery)
4. allows you to swap out the default Node runner for others, like [puppeteer](https://jestjs.io/docs/en/puppeteer)
5. does other useful things like code coverage, support for promise-based testing
6. is used by [GOVUK Frontend](https://github.com/alphagov/govuk-frontend)

This pull request includes a sample test of one our simplest JavaScript modules (which went to prove 2.).

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,11 @@
     "gulp-jshint": "2.1.0",
     "gulp-prettyerror": "1.2.1",
     "gulp-sass-lint": "1.4.0",
+    "jest": "24.7.1",
     "jshint": "2.10.2",
     "jshint-stylish": "2.2.1"
+  },
+  "peerDependencies": {
+    "request": "2.88.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "node": "10.15.3"
   },
   "scripts": {
-    "test": "gulp lint",
+    "test": "gulp lint && jest tests/javascripts",
     "build": "gulp",
     "watch": "gulp watch"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,11 @@
     "gulp-jshint": "2.1.0",
     "gulp-prettyerror": "1.2.1",
     "gulp-sass-lint": "1.4.0",
+    "jest": "24.7.1",
     "jshint": "2.10.2",
     "jshint-stylish": "2.2.1"
+  },
+  "peerDependencies": {
+    "request": "2.88.0"
```
leohemsted: this should be a devDependency not a peerDependency. or it won't install requests when you run `npm install`

tombye: I don't think it needs to be a devDependency as the admin app doesn't depend on it directly, it just has it in its dependency tree. To that end, I've removed the peerDependencies completely in https://github.com/alphagov/notifications-admin/pull/2912/commits/b11ab2c384abdb169b9940aa55d77fcbb723720f

PR #2917, opened 2019-04-16
Update service with the user that requested to go live.

servingUpAces:
When a service is marked as live update the service with the go live datetime.

- [ ] https://github.com/alphagov/notifications-api/pull/2465

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 
 from app.notify_client import NotifyAdminAPIClient, _attach_current_user, cache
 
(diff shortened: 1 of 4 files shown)


app/notify_client/service_api_client.py
```diff
@@ -102,6 +105,7 @@ def update_status(self, service_id, live):
             service_id,
             message_limit=250000 if live else 50,
             restricted=(not live),
+            go_live_at=str(datetime.utcnow()) if live else None
```
quis: Confirmed that toggling this back and forth is the right thing to do, because it’s more likely that we’d put a service live by accident than revert one that really is live. In other words we want to store the date the service went live for real, not the first time it went live.

PR #2931, opened 2019-04-24
Allow more space for values in settings tables

quis:
Before | After
---|---
b4 | localhost_6012_services_42a9d4f2-1444-4e22-9133-52d9e406213f_service-settings(iPad) (1)

Most of the content of our ‘settings’ tables is in the value, not the key. The value is in the middle column. So we should allocate the most space to the value.

The previous layout was based on the premise that most pages divided the grid like this:
```
 _______ _______ _______ _______ _______ _______ _______ _______
|  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |
|               |               |               |               |
|      2/8      |      2/8      |      2/8      |      2/8      |
|               |               |               |               |
|–Navigation––––|–Main column–––––––––––––––––––––––––––––––––––|
|               |                       |                       |
|               |          3/8          |          3/8          |
|               |                       |                       |
|               | Label                 | Value            Link |
|               |                       |                       |
|_______________|_______________________|_______________________|
```

This was because a lot of pages had a left column for emails, and a right column for text messages, so it felt consistent for tables to always default to 50% of the width of the main column.

This consistency has faded with time, especially as we added letters.

So this commit changes these tables to allocate more space to the central column, but still sticking to the grid like this:

```
 _______ _______ _______ _______ _______ _______ _______ _______
|  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |  1/8  |
|       |       |       |       |       |       |       |       |
|      2/8      |      2/8      |              4/8              |
|               |               |                               |
|–Navigation––––|–Main column–––––––––––––––––––––––––––––––––––|
|               |               |                       |       |
|               |      2/8      |          3/8          |  1/8  |
|               |               |                       |       |
|               | Label         | Value                 |   Link|
|               |               |                       |       |
|_______________|_______________|_______________________|_______|
```

Since there’s more space to display the value of a setting this commit also truncates settings that are too long to fit in the width of the column (for example a long email address) rather than the previous
behaviour of truncating them. This all just makes things look a bit cleaner.

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -59,7 +59,7 @@
 
   th {
 
-    &.table-field-heading-first {
+    &:first-child {
       width: 35%;  // 33.33% + fudge
     }
 
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -85,6 +85,10 @@
       text-overflow: ellipsis;
     }
 
+    ul li {
+      margin-bottom: 5px;
+    }
+
   }
 
 }
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -69,7 +69,7 @@
 
   }
 
-  td {
+  td.table-field-center-aligned {
 
     &:first-child {
 
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -69,7 +69,7 @@
 
   }
 
-  td.table-field-center-aligned {
+  td.table-field-left-aligned {
 
     &:first-child {
 
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -51,6 +51,44 @@
 
 }
 
+.settings-table {
+
+  table {
+    table-layout: fixed;
+  }
+
+  th {
+
+    &.table-field-heading-first {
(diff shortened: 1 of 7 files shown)


quis:
Need to figure out why this has made the clickable area of the ‘Edit’ links smaller…

app/assets/stylesheets/components/table.scss
```diff
@@ -51,6 +51,48 @@
...
 
+.settings-table {
+
+  table {
+    table-layout: fixed;
+  }
+
+  th {
+
+    &.table-field-heading-first {
```
tombye: Could this use `:first-child` instead? It would fit nicer with the block below and with the CSS further down that applies certain styles for the first cell using `:first-child` too.

app/assets/stylesheets/components/table.scss
```diff
@@ -51,6 +51,48 @@
...
+      width: 35%;  // 33.33% + fudge
+    }
+
+    &:last-child {
+      width: 17.5%  // 16.67% + fudge
+    }
+
+  }
+
+  td.table-field-center-aligned {
```
tombye: It's not to do with these changes but, reviewing this in the browser, it doesn't seem to be centrally aligned. Is this a problem we could fix now? Not sure if it's in-scope for this or not.

app/assets/stylesheets/components/table.scss
quis: Addressed in 5340c4c2e

PR #2932, opened 2019-04-24
Update documentation.html

jonathanglassman:
There have been multiple support tickets asking how to use the API directly instead of the API clients. This PR adds a statement recommending that users use the API clients instead of the API directly due to lack of documentation.


quis:
Superseded by https://github.com/alphagov/notifications-admin/pull/3096

PR #2934, opened 2019-04-25
Add Live services and Performance platform csv reports

CrystalPea:
Lets platform admins download 2 csv reports:

- Live Services report with following details about live services:

"Service ID", "Organisation", "Service name", "Consent to research", "Main contact",
"Contact email", "Contact mobile", "Live date", "SMS volume intent", "Email volume intent",
"Letter volume intent", "SMS sent this year", "Emails sent this year", "Letters sent this year"

- Performance Platform report;

This PR needs to be deployed first: https://github.com/alphagov/notifications-api/pull/2483

pivotal ticket: https://www.pivotaltracker.com/story/show/164829877

#### Page:
Screen Shot 2019-04-30 at 17 48 42

#### Live Services report:
Screen Shot 2019-04-25 at 11 48 02

#### Performance Platform report:
Screen Shot 2019-04-30 at 17 47 30


app/main/views/platform_admin.py
```diff
@@ -187,6 +200,68 @@ def platform_admin_services():
...
+        "Contact email", "Contact mobile", "Live date", "SMS volume intent", "Email volume intent",
+        "Letter volume intent", "SMS sent this year", "Emails sent this year", "Letters sent this year"
+    ]
+    live_services_data = []
+    live_services_data.append(live_services_columns)
+    for service in services:
+        if service["count_as_live"]:
+            main_contact = None
+            if service["go_live_user"]:
+                main_contact = user_api_client.get_user(service["go_live_user"])
```
quis: This is going to get quite slow for a large number of services. As a rough estimate you can say that a call to the API will take 100ms. So for 800 services that’s 80 seconds. I think our requests are set to time out after 30 seconds, in which case the user will get a `502` error.

CrystalPea: Oh no, you are right :(. Should we make a query in the api that gets all the info we need, then?

CrystalPea: On it 👍

quis: Yeah, @servingUpAces said you should make new query in the API that also returns the main contact user and yearly usage info 👍

app/main/views/platform_admin.py
```diff
@@ -187,6 +197,83 @@ def platform_admin_services():
...
+            row["sms_totals"],
+            row["email_totals"],
+            row["letter_totals"],
+        ])
+    return Response(
+        Spreadsheet.from_rows(live_services_data).as_csv_data,
+        mimetype='text/csv',
+        headers={
+            'Content-Disposition': 'inline; filename="{} live services report.csv"'.format(
+                format_date_numeric(datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")),
```
quis: You should avoid putting colons in file names; some operating systems don’t like them, for example:
> Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
>
> - : (colon)
>
> …

https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#naming-conventions

app/main/views/platform_admin.py
```diff
@@ -187,6 +197,83 @@ def platform_admin_services():
...
+            row["contact_mobile"],
+            datetime.strptime(row["live_date"], '%a, %d %b %Y %X %Z').strftime("%d-%m-%Y"),
+            row["sms_volume_intent"],
+            row["email_volume_intent"],
+            row["letter_volume_intent"],
+            row["sms_totals"],
+            row["email_totals"],
+            row["letter_totals"],
+        ])
+    return Response(
```
quis: You don’t need to wrap this in a `Response`, Flask will accept a tuple of `data, status, headers`, like we do here:
https://github.com/alphagov/notifications-admin/blob/8ab859138bc7eb769b1f3cb178278debb9637301/app/main/views/send.py#L194-L200

You could even add a method (eg `.as_download(filename='foo.csv)`) to the `Spreadsheet` class to avoid having to duplicate this code.

app/main/views/platform_admin.py
CrystalPea: The colon does not show up in the filename, check this out:

Screen Shot 2019-05-01 at 11 41 16

:)

app/main/views/platform_admin.py
CrystalPea: Good suggestion, I dropped the wrappers :)

app/main/views/platform_admin.py
quis: Ah, I missed the call to `format_date_numeric`, my bad.

PR #2936, opened 2019-04-26
Show monthly usage by SMS provider on Providers page

klssmith:
This adds a `% monthly usage` column to the SMS table on the Providers page to let us see how much traffic is being sent to each SMS provider.

Screen Shot 2019-04-26 at 13 30 51

Needs to be merged after:
- [x] https://github.com/alphagov/notifications-api/pull/2484

[Pivotal story](https://www.pivotaltracker.com/story/show/165382068)

### DIFF
--- a/app/templates/views/providers/providers.html
+++ b/app/templates/views/providers/providers.html
@@ -17,13 +17,15 @@ Providers
       caption="Domestic SMS providers",
       caption_visible=False,
       empty_message='No domestic sms providers',
-      field_headings=['Provider', 'Priority', 'Active', 'Last Updated', 'Updated By'],
+      field_headings=['Provider', 'Priority', '% monthly traffic', 'Active', 'Last Updated', 'Updated By'],
       field_headings_visible=True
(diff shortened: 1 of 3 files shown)


app/templates/views/providers/providers.html
```diff
@@ -33,8 +35,8 @@ <h2 class="heading-medium">SMS</h2>
           {{ text_field('None') }}
       {% endif %}
 
-      {% if item.created_by %}
-          {{ text_field(item.created_by.name) }}
+      {% if item.created_by_name %}
```
quis: You don’t need an `if` here. As long as the attribute always exists then Jinja will handle printing a value of `None` as None.

Even better would be to use the optional text field macro which gives some styling, eg
`optional_text_field(item.created_by_name, default='None')` will print ‘None’ in greyed-out text.

klssmith: 👍 Good spot - I've updated it to use `optional_text_field`

PR #2937, opened 2019-04-29
Rationalise heading sizes and ‘Back’ links

quis:
This pull request makes two main changes to rationalise our page layouts:
- consistency in heading sizes
- use of GOV.UK Design System back links

The reason for making these changes now is to de-clutter the area around the green buttons. The folder permissions work presents a design challenge where multiple levels of interaction are happening in the same form. Removing the dark grey ‘Back’ buttons should mean that, in these complicated forms, there’s one fewer thing to compete for the user’s attention.

# Consistency in heading sizes 

We were using `heading-large` (`36px`) everywhere, except on templates pages. On template pages we started using `24px` sized headings, to make space for displaying long folder names. This meant that when surfing the navigation, headings 

# Use of GOV.UK Design System back links

The Design System has standardised on back links being at the top of the page, decorated with a small text-coloured arrow.

I think this makes more sense than having them as buttons at the bottom of the page, because it suggests, in some way, being able to go back before committing to any of the forms on the page. The things at the bottom of the page feel like they should be performing actions on what’s in the page.

I’ve componentised the back links into a `page_header` macro so that the change is easier to roll out and maintain.

# Result

This consolidates our page templates into one of two fundamental types:

## ‘Hub‘ pages

These have `24px` headings, and no back link. They are:
- linked to from the left-hand navigation (or are filtered views of pages linked to from the left hand navigation)
- pages you move through to get to another page
- pages where where you don’t modify 

## ‘Spoke’ pages 

These have `36px` headings, and a back link. The back link either goes to the associated ‘hub’ page, or the previous page in the journey (if there are multiple steps). They are:
- not linked to directly from the left-hand navigation
- pages which modify state, for example by changing a setting or sending a message

# Screenshots

## Settings pages 

Before | After
---|---
image | image
image | image

## Team pages

Before | After
---|---
image | image
image | image

## Templates and sending pages

Before | After
---|---
image | image
image | image
image | image
image | image

### DIFF
--- a/app/templates/views/check/ok.html
+++ b/app/templates/views/check/ok.html
@@ -3,6 +3,7 @@
 {% from "components/radios.html" import radio_select %}
 {% from "components/table.html" import list_table, field, text_field, index_field, hidden_field_heading %}
 {% from "components/file-upload.html" import file_upload %}
+{% from "components/page-header.html" import page_header %}
 {% from "components/page-footer.html" import page_footer %}
 {% from "components/message-count-label.html" import message_count_label %}
 
@@ -19,9 +20,11 @@
 
 {% block maincolumn_content %}
 
-  <h1 class="heading-large">
-    Preview of {{ template.name }}
-  </h1>
+  {{ page_header(
+    'Preview of {}'.format(template.name),
+    back_link=back_link
+  ) }}
+
   {{ skip_to_file_contents() }}
 
   {{ template|string }}
@@ -41,7 +44,6 @@
       {% else %}
         <a href="{{ url_for('main.check_messages_preview', service_id=current_service.id, template_id=template.id, upload_id=upload_id, filetype='pdf') }}" download class="button">Download as a PDF</a>
       {% endif %}
-      <a href="{{ back_link }}" class="page-footer-back-link">Back</a>
     </form>
   </div>
 
--- a/app/templates/views/edit-user-permissions.html
+++ b/app/templates/views/edit-user-permissions.html
@@ -1,5 +1,6 @@
 {% extends "withnav_template.html" %}
 {% from "components/textbox.html" import textbox %}
+{% from "components/page-header.html" import page_header %}
 {% from "components/page-footer.html" import page_footer %}
 {% from "components/form.html" import form_wrapper %}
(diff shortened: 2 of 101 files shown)


app/templates/views/check/ok.html
```diff
@@ -19,6 +19,8 @@
 
 {% block maincolumn_content %}
 
+  <a class="govuk-back-link" href="{{ back_link }}">Back</a>
```
tombye: Could this, and the `<h1>` below be swapped out for the `page-header` macro?

app/templates/views/edit-user-permissions.html
```diff
@@ -18,6 +18,9 @@
     ) }}
   {% endif %}
 
+
```
tombye: Could this, and the `<h1>` below be swapped out for the `page-header` macro?

app/templates/views/manage-users/edit-user-email.html
```diff
@@ -9,19 +9,13 @@
 
 {% block maincolumn_content %}
 
+  <a class="govuk-back-link" href="{{ url_for('main.edit_user_permissions', service_id=current_service.id, user_id=user.id) }}">Back</a>
```
tombye: Could this, and the `<h1>` below be swapped out for the `page-header` macro?

app/templates/views/manage-users/edit-user-mobile.html
```diff
@@ -9,17 +9,15 @@
 
 {% block maincolumn_content %}
 
+  <a class="govuk-back-link" href="{{ url_for('main.edit_user_permissions', service_id=current_service.id, user_id=user.id) }}">Back</a>
```
tombye: Could this, and the `<h1>` below be swapped out for the `page-header` macro?

app/templates/views/service-settings/set-inbound-number.html
```diff
@@ -10,26 +11,18 @@
 
 {% block maincolumn_content %}
     <h1 class="heading-large">Set Inbound Number</h1>
```
tombye: Should this still be here?

PR #2942, opened 2019-04-30
Fix heading size on empty templates page

quis:
This is a special case which I missed when doing the rationalisation in https://github.com/alphagov/notifications-admin/pull/2937

***

image


quis:
Also add the template folder navigation even for teams who don’t have folders, because these pages don’t have a `◀ Back` link

image

PR #2943, opened 2019-05-01
Add service-specific versions of agreement page 

quis:
*This pull request is mostly moving things around – substantive changes are in 35bf42b*

***

Our usability testing found that jumping out of the service when going to download the agreement made it difficult for people to find their way back to the ‘Request to go live’ page.

This commit adds a duplicate, service-specific versions of these pages which have the same content but:
- keep the service navigation
- have a link back to the ‘Request to go live’ page

This gives us a place to – in the future – let people accept the agreement online.

***

Before | After 
---|---
image | image


app/main/views/agreement.py
```diff
@@ -10,12 +10,21 @@
...
 def agreement():
     return render_template(
-        'views/{}.html'.format(current_user.default_organisation.as_jinja_template),
+        'views/agreement/{}.html'.format(current_user.default_organisation.as_jinja_template),
         owner=current_user.default_organisation.name,
         navigation_links=features_nav(),
     )
 
 
+@main.route('/service/<uuid:service_id>/agreement')
```
servingUpAces:
```suggestion
@main.route('/services/<uuid:service_id>/agreement')
```

PR #2944, opened 2019-05-01
Allow replacing a revoked key

quis:
We had someone wondering why they couldn’t do this. Seems legit to allow it, especially since revoked API keys never get completely deleted.


quis:
@leohemsted I knew there was a reason this wasn’t so simple… Can `expiry_date` be made part of the constraint? That should mean you can only have one non-revoked key with a given name at any one time.

PR #2951, opened 2019-05-03
Provide Performance Platform report as Excel file

quis:
This is the format we need to upload it to Performance Platform in, so this means not having to manually re-save it as Excel.


CrystalPea:
I think tests need updating, otherwise it should be ready to go :)

PR #2953, opened 2019-05-07
Don’t convert Excel reports to CSV before output

quis:
Converting Python data to CSV makes every field a string. This means that in the report we return to the user every field will be a string, even if it’s come from an `int` type in Python. This is because the CSV ‘standard’ doesn’t support any kind of typing.

Excel does support types for fields, so we can make our reports more useful by preserving these types. This is particularly relevant in the report we generate for performance platform, which needs the `count` column to be a number type.

This commit adds extra code paths to the `Spreadsheet` class which mean that it can be instantiated from either CSV data or a list of Python data. Previously we were converting the Python data to CSV as an intermediate step, before instantiating the class.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -226,7 +226,10 @@ class Spreadsheet():
     def __init__(self, csv_data, filename=''):
         self.filename = filename
         self.as_csv_data = csv_data
-        self.as_dict = {
+
+    @property
+    def as_dict(self):
+        return {
             'file_name': self.filename,
             'data': self.as_csv_data
         }
--- a/app/utils.py
+++ b/app/utils.py
@@ -280,7 +280,7 @@ class Spreadsheet():
         extension = cls.get_extension(filename)
 
         if extension == 'csv':
-            return cls(Spreadsheet.normalise_newlines(file_content), filename=filename)
+            return cls(csv_data=Spreadsheet.normalise_newlines(file_content), filename=filename)
 
         if extension == 'tsv':
             file_content = StringIO(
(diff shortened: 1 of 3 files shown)


app/utils.py
```diff
@@ -245,27 +264,23 @@ def normalise_newlines(file_content):
...
+            filename=filename,
         )
 
     @classmethod
     def from_file(cls, file_content, filename=''):
         extension = cls.get_extension(filename)
 
         if extension == 'csv':
-            return cls(Spreadsheet.normalise_newlines(file_content), filename)
+            return cls(Spreadsheet.normalise_newlines(file_content), filename=filename)
```
leohemsted: can you make this `csv_data=Spreadsheet....` to make it clear that this is different to the rows=rows call above?

app/utils.py
```diff
@@ -223,14 +223,33 @@ class Spreadsheet():
...
     allowed_file_extensions = ['csv', 'xlsx', 'xls', 'ods', 'xlsm', 'tsv']
 
-    def __init__(self, csv_data, filename=''):
+    def __init__(self, csv_data=None, rows=None, filename=''):
+
         self.filename = filename
-        self.as_csv_data = csv_data
-        self.as_dict = {
+
+        if csv_data and rows:
```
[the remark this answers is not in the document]
quis: Cool, but I don’t want it to raise an exception when called with an empty file (because this code is used in the ‘upload recipients’ journey and it’s totally possible that someone could upload an empty file).

app/utils.py
quis: Good idea, changed in 20e774ca3

PR #2954, opened 2019-05-07
Allow Welsh characters in SMS templates

klssmith:
- Brings in the latest version of notifications-utils which allows Welsh characters in SMS templates.
- Renamed some code to make it clear that Welsh characters are allowed in addition to GSM characters.
- Updated the pricing page to show the new prices for SMS with certain Welsh characters:

Screen Shot 2019-05-15 at 14 20 22

[Pivotal story](https://www.pivotaltracker.com/story/show/165672985)

Needs to be merged after:
- [x] https://github.com/alphagov/notifications-utils/pull/606

### DIFF
--- a/app/templates/views/pricing.html
+++ b/app/templates/views/pricing.html
@@ -40,10 +40,10 @@
     <p>See <a href="#paying">how to pay</a>.
 
     <h3 class="heading-small">Long text messages</h3>
-    <p>If a text message is beyond a certain length, it’ll be charged as more than one message:</p>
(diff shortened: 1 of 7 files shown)


app/templates/views/pricing.html
```diff
@@ -61,6 +61,29 @@ <h3 class="heading-small">Long text messages</h3>
         {% endfor %}
       {% endcall %}
     </div>
+    <h3 class="heading-small">Text messages containing Welsh characters</h3>
```
quis: Don’t need the heading here:
```suggestion
```

PR #2955, opened 2019-05-08
Set org type to NHS if user has NHS email address

quis:
We get people signing up for Notify who work for the NHS, but whose organisation we don’t know about. For example `name@gloshospitals.nhs.uk` will be someone working for Gloucestershire
Hospitals NHS Foundation Trust, which is not an organisation we have in the database.

Currently we rely on knowing the specific organisation (NHS as a whole isn’t an organisation) in order to set the organisation type for any services they create. This commit adds a special case for anyone with an NHS email address to set the organisation type to be NHS, even when we don’t know which specific part of the NHS they work for.

This is the same thing we do on the API side for NHS email and letter branding:
https://github.com/alphagov/notifications-api/blob/a4ae5a0a90494ab338561010ed279bf198b15a72/app/dao/services_dao.py#L310-L313

### DIFF
--- a/tests/app/main/views/test_add_service.py
+++ b/tests/app/main/views/test_add_service.py
@@ -62,6 +62,11 @@ def test_get_should_not_render_radios_if_org_type_known(
     assert not page.select('.multiple-choice')
 
 
+@pytest.mark.parametrize('email_address', (
+    # User’s email address doesn’t matter when the organisation is known
+    'test@example.gov.uk',
+    'test@example.nhs.uk',
+))
 @pytest.mark.parametrize('inherited, posted, persisted, sms_limit', (
     (None, 'central', 'central', 250000),
     ('central', None, 'central', 250000),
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_add_service.py
```diff
@@ -62,6 +62,11 @@ def test_get_should_not_render_radios_if_org_type_known(
     assert not page.select('.multiple-choice')
 
 
+@pytest.mark.parametrize('email_address', (
+    # User’s email address doesn’t matter when the organisation is known
+    'test@example.gov.uk',
+    'test@example.nhs.uk',
+))
 @pytest.mark.parametrize('inherited, posted, persisted, sms_limit', (
```
CrystalPea: Maybe we could add an example where persisted type is None, too?

quis: We don’t let users persist a value of `None`. But I’ve added an extra test to confirm that they’re prevent from doing so in e6d03ec.

CrystalPea: Awesome, thanks 👍

PR #2959, opened 2019-05-09
Use standard back link on new API key page

quis:
Missed this one before because it was using a link, not a dark grey button.


quis:
Actually I think this should stay as a normal link… you’re not really going ‘back’ here, you’re going ‘forward’ to the the next thing… 🤔

PR #2962, opened 2019-05-09
Fix broken links

klssmith:
There were still a few links pointing to the old `/features/using-notify` page.

### DIFF
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -31,7 +31,7 @@
     <p>These are the types of message status you’ll see when you’re signed in to Notify.</p>
     <p>If you’re using our API, some of the statuses you’ll see will be different. Read our <a href="{{ url_for('.documentation') }}">documentation</a> for a detailed list of API message statuses.</p>
 
-    <h3 class="heading-small">Emails</h3>
+    <h3 id="email-statuses" class="heading-small">Emails</h3>
     <div class="bottom-gutter-3-2">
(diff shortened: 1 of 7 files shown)


app/templates/views/message-status.html
```diff
@@ -31,7 +31,7 @@ <h2 id="messagestatus" class="heading-medium">Types of message status</h2>
     <p>These are the types of message status you’ll see when you’re signed in to Notify.</p>
     <p>If you’re using our API, some of the statuses you’ll see will be different. Read our <a href="{{ url_for('.documentation') }}">documentation</a> for a detailed list of API message statuses.</p>
 
-    <h3 class="heading-small">Emails</h3>
+    <h3 id="email-statuses" class="heading-small">Emails</h3>
```
quis: Can you add a test that loads this page and checks for this `id` to ensure that no-one accidentally removes it?

app/templates/views/message-status.html
```diff
@@ -54,7 +54,7 @@ <h3 class="heading-small">Emails</h3>
       {% endcall %}
     </div>
 
-    <h3 class="heading-small">Text messages</h3>
+    <h3 id="sms-statuses" class="heading-small">Text messages</h3>
```
quis: Same here.

app/templates/views/message-status.html
klssmith: 👍I've added a test for the two ids

PR #2974, opened 2019-05-16
Verify email reply-to address

CrystalPea:
### Work done:
- when adding new email reply-to address, send request to api to send verification notification to check if it's a working address
- new endpoint that handles a waiting page, auto-refreshes using ajax, adds the new email address when notification is delivered, announces failure if delivery fails, and also announces failure if notification not delivered within 5 minutes
- edit email reply-to address endpoint also uses verification feature now

### API PR:
https://github.com/alphagov/notifications-api/pull/2511 (needs to be deployed first)

### Story ticket:
 https://www.pivotaltracker.com/story/show/165791343

#### This page didn't change:
Screen Shot 2019-05-21 at 16 10 34

#### This is a new waiting page while we await delivery receipt:
Screen Shot 2019-05-21 at 16 10 47

#### A failure:
Screen Shot 2019-05-21 at 16 11 47

#### A success:
Screen Shot 2019-05-21 at 16 11 19

#### A different header when users are changing an existing email reply-to address:
Screen Shot 2019-05-20 at 14 32 50


app/main/views/service_settings.py
```diff
@@ -402,19 +405,100 @@ def service_email_reply_to(service_id):
...
+                service_api_client.add_reply_to_email_address(
+                    current_service.id,
+                    email_address=notification["to"],
+                    is_default=is_default
+                )
+    created_at_no_tz = notification["created_at"][:-6]
+    seconds_since_sending = (datetime.utcnow() - datetime.strptime(created_at_no_tz, '%Y-%m-%dT%H:%M:%S.%f')).seconds
+    if notification["status"] in [
+        "failed", "permanent-failure", "technical-failure", "temporary-failure"
+    ] or (notification["status"] == "sending" and seconds_since_sending > 300):
```
quis: I think 90 seconds is better. It’s close to ‘a minute’, which is what we say on the page, but with a bit of breathing room.
```suggestion
    ] or (notification["status"] == "sending" and seconds_since_sending > 90):
```

app/main/views/service_settings.py
```diff
@@ -434,13 +518,26 @@ def service_edit_email_reply_to(service_id, reply_to_email_id):
...
-            email_address=form.email_address.data,
-            is_default=True if reply_to_email_address['is_default'] else form.is_default.data
-        )
-        return redirect(url_for('.service_email_reply_to', service_id=service_id))
+        try:
+            notification_id = service_api_client.verify_reply_to_email_address(
+                service_id, form.email_address.data
+            )["data"]["id"]
+        except HTTPError as e:
+            error_msg = "Your service already uses '{}' as an email reply-to address.".format(form.email_address.data)
```
quis:
```suggestion
            error_msg = "Your service already uses ‘{}’ as an email reply-to address.".format(form.email_address.data)
```

app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -0,0 +1,64 @@
...
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/form.html" import form_wrapper %}
+
+<div class="ajax-block-container">
+  {% if verification_status == "pending" %}
+    <p>
+      We’re checking that ‘{{ reply_to_email_address }}’ is a real email address.
+    </p>
+    <p>
+  <span class='loading-indicator'>This can take a minute </span>
```
quis:
```suggestion
  <span class='loading-indicator'>This can take a minute</span>
```

quis: (the trailing whitespace here puts too much space between the text and the animated dots)

app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -0,0 +1,64 @@
...
+    </p>
+
+    <p>
+      <a href="{{ url_for('main.service_verify_reply_to_address', service_id=service_id, notification_id=notification_id) + "?is_default={}".format(is_default) }}"">Refresh</a>
```
quis: `url_for` will generate the query string for you:
```suggestion
      <a href="{{ url_for('main.service_verify_reply_to_address', service_id=service_id, notification_id=notification_id, is_default=is_default) }}">Refresh</a>
```

app/templates/views/service-settings/email-reply-to/_verify-updates.html
CrystalPea: Ahh, ok, I added it specifically because I thought there is not enough space without it 😅. But I will get rid of it :)

app/main/views/service_settings.py
CrystalPea: 👍

CrystalPea: how long does it usually take for us to get a delivery receipt?

quis: I did some analysis of a random sample of email notifications in the `notifications` table.

– | Time to get ‘delivered‘ status (seconds) | Time to get ‘failure‘ status (seconds)
---|---|---
Mean | 11.6 | 22,142.1
Median | 2.9 | 8.7

In percentile terms it looks like this:

image

You can see that failure is slower than success. So the longer a notification takes to get a status, the more likely it is for that status to be a failure anyway. This increases dramatically after 45 seconds. The percentage of emails that go to `delivered` in less than 90 seconds is 98.92%. To get to 99% we’d need to increase the timeout to 178 seconds (3 minutes). We could still get 98.7% of notifications by dropping the timeout to 45 seconds, and improve the experience for notifications that are likely to fail by returning an error more quickly.

Source: https://docs.google.com/spreadsheets/d/1oWttUHbT1eWzzvljeFu0w7VbekD7dlRbjNyYiMj68QU/edit#gid=1850531376

CrystalPea: nice one with the analysis, I updated it now and included part of your above comment in the commit message for future generations :P

app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -0,0 +1,64 @@
...
+  {% if verification_status == "pending" %}
+    <p>
+      We’re checking that ‘{{ reply_to_email_address }}’ is a real email address.
+    </p>
+    <p>
+  <span class='loading-indicator'>This can take a minute</span>
+    </p>
+
+    <p>
+      <a href="{{ url_for('main.service_verify_reply_to_address', service_id=service_id, notification_id=notification_id, is_default=is_default, replace=replace) }}"">Refresh</a>
```
quis: Duplicate closing quote here is making the HTML invalid and confusing the Javascript:
```suggestion
      <a href="{{ url_for('main.service_verify_reply_to_address', service_id=service_id, notification_id=notification_id, is_default=is_default, replace=replace) }}">Refresh</a>
```

app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -0,0 +1,64 @@
...
+<div class="ajax-block-container">
+  {% if verification_status == "pending" %}
+    <p>
+      We’re checking that ‘{{ reply_to_email_address }}’ is a real email address.
+    </p>
+    <p>
+  <span class='loading-indicator'>This can take a minute</span>
+    </p>
+
+    <p>
```
quis: Adding this class means that the refresh link will only show to users who don’t have Javascript:
```suggestion
    <p class="js-hidden">
```

PR #2980, opened 2019-05-20
Try to render AJAX updates one final time if error

quis:
When we return a HTTP error we still want to display what the error is. So we should assume that when returning some JSON with an error status that we should still try to render the updates to the page before switching off the polling (which is what setting it to an empty function does).

For a `500` error the JSON will be `undefined`, so the Javascript will not try to update the page.


quis:
We’re going to try an approach where we add a key to the JSON rather than HTTP status codes.

PR #2985, opened 2019-05-22
Allow a user to be archived

klssmith:
Users can only be archived by Platform Admin from the user page `/users/<user_id>`. This removes them from all services and orgs and updates their details.

We also add a new type of event, `archive_user`, which stores the id of the archived user and the id of the user who is doing the archiving.

[Pivotal story](https://www.pivotaltracker.com/story/show/166027696)

Screen Shot 2019-05-22 at 14 11 58


app/main/views/find_users.py
```diff
@@ -36,3 +38,21 @@ def user_information(user_id):
...
+@login_required
+@user_is_platform_admin
+def archive_user(user_id):
+    if request.method == 'POST':
+        try:
+            user_api_client.archive_user(user_id)
+        except HTTPError as e:
+            raise e
+        else:
+            create_archive_user_event(str(user_id), current_user.id)
```
leohemsted: unless i've missed something, I'm not sure we need this try else here, since all exceptions are just being raised anyway.

I think just

```python
user_api_client.archive_user(user_id)
create_archive_user_event(str(user_id), current_user.id)
```

should be fine

klssmith: Yep, you're right - I didn't need to catch the error here.

PR #2987, opened 2019-05-22
move trial mode services from org dashboard to separate page

leohemsted:
the new page is platform admin only, and also has search built in also, split test_organisation_invites into two files. The old page doesn't have search because we're overhauling that page imminently.

# Live services page 

image

# Trial mode services page

image


app/main/views/organisations.py
```diff
@@ -70,7 +70,10 @@ def add_organisation():
 @login_required
 @user_has_permissions()
 def organisation_dashboard(org_id):
-    organisation_services = organisations_client.get_organisation_services(org_id)
+    organisation_services = [
+        service for service in organisations_client.get_organisation_services(org_id)
+        if service['active'] and not service['restricted']
+    ]
```
quis: Can we move this kind of logic to the [`Organisation` model](https://github.com/alphagov/notifications-admin/blob/master/app/models/organisation.py)? For example adding a property like `organisation.live_services`.

app/main/views/organisations.py
```diff
@@ -70,14 +70,24 @@ def add_organisation():
...
+
+
+@main.route("/organisations/<org_id>/trial-services", methods=['GET'])
+@login_required
+@user_is_platform_admin
+def organisation_trial_mode_services(org_id):
+    return render_template(
+        'views/organisations/organisation/trial-mode-services.html',
+        search_form=SearchByNameForm(),
+        services=[x for x in current_organisation.services if not x['active'] or x['restricted']]
```
quis: This list comprehension could move to the model as well.

app/main/views/organisations.py
```diff
@@ -86,8 +96,8 @@ def organisation_dashboard(org_id):
 @user_has_permissions()
 def manage_org_users(org_id):
     users = sorted(
-        user_api_client.get_users_for_organisation(org_id=org_id) + [
-            invite for invite in org_invite_api_client.get_invites_for_organisation(org_id=org_id)
+        current_organisation.users + [
```
quis: This could be one attribute on the model as well; see how `.team_members` works for a service: https://github.com/alphagov/notifications-admin/blob/238c56faec8b3abdbc0a6df5c719c56dba971e2d/app/models/service.py#L108-L113

PR #2988, opened 2019-05-23
Remove spacing tweak on loading indicator

quis:
It’s not needed when there’s no whitespace around the text.


tombye:
Was whitespace removed from the text? The changes make sense in isolation but I'm not sure about approving without understanding that bit.

quis:
@tombye the whitespace was removed here: https://github.com/alphagov/notifications-admin/pull/2974/files#diff-d95533b53fbba399c5e998e920e480b2R14

I think it might make sense to make a `loading_indicator` macro to make sure we always have control of the whitespace.

quis:
Going to merge this for now to fix the weird looking thing, can revisit later…

PR #2991, opened 2019-05-28
Make user API client return JSON, not a model

quis:
# What kind of pull request is this?

Refactoring only – it shouldn’t change how anything works.

# Why now?

The aim of this refactor is to let us, in the future, write view-layer code like:
```jinja
{% for service in current_user.trial_mode_services %}
    …
```
This kind of code is very difficult to write now because of the way the user model is architected.

# Tell me more

The data flow of other bits of our application looks like this:
```
                         API (returns JSON)
                                  ⬇
          API client (returns a built in type, usually `dict`)
                                  ⬇
          Model (returns an instance, eg of type `Service`)
                                  ⬇
                         View (returns HTML)
```
The user API client was architected weirdly, in that it returned a model directly, like this:

```
                         API (returns JSON)
                                  ⬇
    API client (returns a model, of type `User`, `InvitedUser`, etc)
                                  ⬇
                         View (returns HTML)
```

This mixing of different layers of the application is bad because it makes it hard to write model code that doesn’t have circular dependencies. As our application gets more complicated we will be relying more on models to manage this complexity, so we should make it easy, not hard to write them.

It also means that most of our mocking was of the User model, not just the underlying JSON. So it would have been easy to introduce subtle bugs to the user model, because it wasn’t being comprehensively tested. A lot of the changed lines of code in this commit mean changing the tests to mock only the JSON, which means that the model layer gets implicitly tested.

For those reasons this pull request changes the user API client to return JSON, not an instance of `User` or other models.

# Misc

This pull request also does some small refactoring of the `User` model itself to make initialising it work the same way as our `Service` and `Organisation` models.


app/models/user.py
```diff
@@ -24,23 +25,36 @@ def _get_org_id_from_view_args():
...
+        'failed_login_count',
+        'state',
+        'logged_in_at',
+        'platform_admin',
+        'current_session_id',
+        'services',
+        'organisations',
+    }
+
+    DEFAULTS = {
```
leohemsted: the api _always_ returns exactly:
```
id
name
email_address
auth_type
current_session_id
failed_login_count
logged_in_at
mobile_number
organisations
password_changed_at
permissions
platform_admin
services
state
```

leohemsted: it does this through `models.py::User.serialize` - there is an old marshmallow user_schema in schemas.py but this isn't used for dumping return data, only parsing the json in the create user rest endpoint.

quis: 👍 – removed the defaults in f34a252e7

app/models/user.py
```diff
@@ -307,38 +437,105 @@ class InvitedOrgUser(object):
...
+
     def logged_in_elsewhere(self):
         return False
 
     @property
     def default_organisation(self):
         return Organisation(None)
+
+
+class Users():
```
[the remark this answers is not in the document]
quis: I’d still prefer they didn’t return a built-in type because then it’s not possible to add extra methods (for example `Users.with_permission('manage_service')` might be something we want in the future).

But I’ve made them behave more like lists by inheriting from `collections.abc.Sequence` – see f8fb2d3c8

PR #2993, opened 2019-05-28
Fix error where sending test message flow skipped first placeholder

CrystalPea:
first way round and then collected placeholders again. Now the flow
collects all placeholders in one round.

Also move operators around following flake8's advice :)

Part of this story:
https://www.pivotaltracker.com/story/show/166030958

TODO:
- discuss writing a functional test that will test this flow and ensure that it works well for different types of templates and with different number of placeholders

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -386,7 +386,7 @@ def send_test_step(service_id, template_id, step_index):
     )
 
     try:
-        if "test" in request.endpoint:
+        if request.endpoint == 'main.send_test_step':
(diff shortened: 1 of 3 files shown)


app/main/views/send.py
```diff
@@ -386,7 +386,10 @@ def send_test_step(service_id, template_id, step_index):
     )
 
     try:
-        current_placeholder = placeholders[step_index]
+        if "test" in request.endpoint:
```
servingUpAces: For consistency it would be good to use `request.endpoint == 'main.send_test_step'`

PR #2996, opened 2019-05-31
Show free sms fragments in live services billing report

leohemsted:
- [x] https://github.com/alphagov/notifications-api/pull/2524


leohemsted:
@servingUpAces works for me - it opens inline in google sheets. what happens if you right click and save link as?

servingUpAces:
It was not working on an incognito window - but works otherwise.

PR #3017, opened 2019-06-17
Fix rendering of scrollable tables

quis:
The scrollable tables code styles some of the cells in the target table by looking for the `table-field-center-aligned` class.

This class was renamed in 0512f40ad3d7fdb4d285a4f7e44659cca6f3e09f

This commit updates the scrollable tables code to refer to the new classname, which means that things should line up properly when drawing the table.

***

image

### DIFF
--- a/app/assets/stylesheets/components/fullscreen-table.scss
+++ b/app/assets/stylesheets/components/fullscreen-table.scss
@@ -25,7 +25,7 @@
 
     th,
     .table-field-error-label,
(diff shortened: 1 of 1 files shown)


quis:
Tests are already using the new class name (although they wouldn’t have caught the problem): https://github.com/alphagov/notifications-admin/blob/da29d722f9b474a8ed5ada4a990000fadb9118c9/tests/javascripts/fullscreenTable.test.js#L53

PR #3019, opened 2019-06-19
Allow online acceptance of the contract/memorandum of understanding

quis:
Story: https://www.pivotaltracker.com/story/show/166497058

Content: https://docs.google.com/document/d/1ELqNt15CUW9Laihhl72RcPrfphFlc4xUNhAPuH9FeYA/edit#heading=h.nv75d9tmh567

_This shouldn’t be deployed until the latest versions of the agreement have been uploaded into S3._

# Before 

image

image

# After 

image

image

image

image

image


app/main/forms.py
```diff
@@ -1462,3 +1462,68 @@ class GoLiveNotesForm(StripWhitespaceForm):
...
+            field.data = ''
+
+    validate_on_behalf_of_name = __validate_if_nominating
+    validate_on_behalf_of_email = __validate_if_nominating
+
+    def validate_version(self, field):
+        try:
+            float(field.data)
+        except (TypeError, ValueError):
+            raise ValidationError("Must be a number")
```
leohemsted: i spent a while thinking about this and if we want to bind it more tightly to the pdf version numbers but i can't think of a nice way to do that. i think accepting `-inf` and `NaN` and `1e9` is an acceptable quirk.

app/main/views/agreement.py
```diff
@@ -26,6 +29,50 @@ def service_agreement(service_id):
...
+@login_required
+def service_accept_agreement(service_id):
+
+    if not current_service.organisation:
+        abort(404)
+
+    form = AcceptAgreementForm.from_organisation(current_service.organisation)
+
+    if form.validate_on_submit():
+        current_service.organisation.update(
```
leohemsted: do we have a process in place to validate/audit the data? especially in terms of the version number i'm thinking.

do we have a plan to notify ourselves when someone signs an agreement for an organisation?

app/main/forms.py
quis: I guess far future would be having a platform admin page for uploading new contracts, which would give us a UI for adding version numbers to the database.

app/main/views/agreement.py
quis: Discussed with @yahoopete. We’re happy to keep an eye on the data for bad version numbers or email addresses for the moment.

quis:
# Direct users to contact us if crown status unknown

If we’re not sure whether a user belongs to a crown organisation or not we want to fix that first, before we let them sign the agreement.

This is a last-ditch fallback because we shouldn’t be adding new organisations without also setting their crown status. Looks like this:

image

PR #3021, opened 2019-06-20
Show organisation breadcrumbs

klssmith:
Added a breadcrumb link to a service's organisation to the `withnav_template.html`. This will only show if a service has an organisation and the current user is also a member of that org, or the current user is a platform admin user.

Also removed a couple of unused fixtures from the `client_request` fixture.

[Pivotal story](https://www.pivotaltracker.com/story/show/166056773)

### DIFF
--- a/tests/app/main/views/test_dashboard.py
+++ b/tests/app/main/views/test_dashboard.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from unittest.mock import call
 
 import pytest
+from bs4 import BeautifulSoup
 from flask import url_for
 from freezegun import freeze_time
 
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_dashboard.py
```diff
@@ -1333,3 +1336,81 @@ def test_should_show_all_jobs_with_valid_statuses(
...
+    active_caseworking_user,
+):
+    # active_caseworking_user is not an org member
+
+    service_one_json = service_json(SERVICE_ONE_ID,
+                                    users=[active_caseworking_user['id']],
+                                    organisation_id=ORGANISATION_ID)
+    mocker.patch('app.service_api_client.get_service', return_value={'data': service_one_json})
+
+    client.login(active_caseworking_user, mocker, service_one_json)
```
quis: You don’t have to stop using `client_request` in order to log in as a different user:
```suggestion
    client_request.login(active_caseworking_user, service=service_one_json)
```

klssmith: Didn't know about that - this test and the one below now use `client_request`

PR #3023, opened 2019-06-21
Link to dashboard if user has access to service

quis:
It feels weird to deep link into the usage page. Normally navigating to a service takes you to the dashboard. It should only work differently if you don’t have access to the dashboard.


quis:
I’m not sure this change is right actually. Going to let the existing journey sit and marinate a while first.

PR #3024, opened 2019-06-21
Adjust type and spacing to accommodate organisation breadcrumb

quis:
Before | After 
---|---
image | image
image | image

### DIFF
--- a/app/assets/stylesheets/components/navigation.scss
+++ b/app/assets/stylesheets/components/navigation.scss
@@ -2,16 +2,21 @@
 
   padding: 0 $gutter 0 0;
 
+  $padding-top: 14px;
+  $padding-bottom: 11px;
+
   &-service {
 
     border-bottom: 1px solid $border-colour;
     margin: 0 0 10px;
     position: relative;
+    font-size: 0;
 
     &-name {
       @include bold-19;
-      margin: 0 0 0 0;
+      padding: $padding-top 0 $padding-bottom 0;
       display: inline-block;
+      vertical-align: top;
     }
 
     &-switch {
@@ -21,7 +26,7 @@
       position: absolute;
       top: 0;
       right: 0;
-      padding: 13px 0 9px 15px;
+      padding: $padding-top 0 $padding-bottom $gutter-half;
 
       &:hover {
         color: $link-hover-colour;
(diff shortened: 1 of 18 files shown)


quis:
# Using breadcrumb icon from the GOV.UK Design System

Before | After 
---|---
image | image

# Making the folder path separators also use arrows for consistency
Before | After
---|---
image | image
image | image

app/assets/stylesheets/components/navigation.scss
```diff
@@ -2,16 +2,21 @@
 
   padding: 0 $gutter 0 0;
 
+  $padding-top: 14px;
```
tombye: A suggestion rather than an issue: how about setting the `padding-top` and `padding-bottom` in a single block with all the different selectors rather than in variables and spreading them across the file?

I mean something like:

```
&-service-name,
&-service-switch,
&-back-to,
&-organisation-link {
  padding-top: 14px;
  padding-bottom: 11px;
}
```

Intention would be to be able to see all the navigation modules that need to line up in one place.

app/assets/stylesheets/components/navigation.scss
```diff
@@ -50,6 +51,17 @@
     background-repeat: no-repeat;
   }
 
+  &-organisation-link {
+    @include core-19;
+    display: inline-block;
+    max-width: 25%;
```
tombye: We might want to have a variant `max-width` for mobile views. On Desktop the smallest width this yields is 240px but mobile, assuming a virtual pixel width of 320, is around 80px.

Seeing as there's only ever two items in the nav, we could maybe stack them vertically...?

quis: Added some mobile-specific CSS in d364ce5f9.

Looks like this:

Service only | Service and organisation
---|---
localhost_6012_services_42a9d4f2-1444-4e22-9133-52d9e406213f_usage(iPhone 6_7_8) (1) | localhost_6012_services_42a9d4f2-1444-4e22-9133-52d9e406213f_usage(iPhone 6_7_8)

app/assets/stylesheets/components/navigation.scss
quis: I think the addition of the mobile-specific CSS (see https://github.com/alphagov/notifications-admin/pull/3024#discussion_r300360737) has made it a bit too complicated to factor this out into a specific declaration.

app/assets/stylesheets/components/navigation.scss
tombye: Nice!

PR #3037, opened 2019-06-27
Fix a bug with inviting existing users to an organisation.

servingUpAces:
The method to add the user to the organisation was missing the user id. This PR fixes that.


app/main/views/invites.py
```diff
@@ -105,7 +105,7 @@ def accept_org_invite(token):
     if existing_user:
         invited_org_user.accept_invite()
         if existing_user not in organisation_users:
-            invited_org_user.add_to_organisation()
+            invited_org_user.add_to_organisation(existing_user.id)
```
quis: You can change this without having to change the model
```suggestion
            existing_user.add_to_organisation()
```

PR #3038, opened 2019-07-01
Require more information when creating organisations

quis:
Currently we set not-very-useful defaults for organisation type and crown status when creating an organisation. This commit adds two field to the form (in addition to the existing name field) to explicitly ask for:
- organisation type
- crown status

We need these for all organisations before we can make any of their services live.

This commit also records any new organisation as not having accepted the data sharing and financial agreement, because if we don’t know about the organisation already then they definitely won’t have signed it.

# Before 

image

# After 

image


quis:
Updated to not use platform admin layout:

image

PR #3039, opened 2019-07-02
Fix hard coded organisation name

quis:
😳


quis:
This also fixes a `404` when users were trying to download the agreement.

PR #3041, opened 2019-07-04
Let users delete letter contact blocks

quis:
Because they can delete email reply to addresses and text message senders.

***

image

***

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2554


quis:
When a contact block is being used by a template, we need to decide whether to:

1. remove it from the template when deleting the contact block
2. stop the contact block from being deleted until it’s de-assigned

Favouring 1. currently, which will require API changes.

servingUpAces:
Food for thought: this PR and the api PR doesn't allow a service to delete the default contact block. We probably should allow them to delete the default, because having no address (or blank) is valid. If they can delete the default and they have more than one address what is the default? Blank or the second address? If we don't set the default to the second address it looks like they don't have any addresses from the `service-settings` page. 
We may want to deploy this PR as is or resolve this question first.

quis:
Added some edge cases for UI around ‘blank’ and deleting defaults:

image

***

image

PR #3045, opened 2019-07-09
Stop allowing service org type to be changed

klssmith:
The service  organisation type will either be the same as the org type of the service's organisation or will be set by a user when creating a new service. This removes the ability to change it from the platform admin settings table.

[Pivotal story](https://www.pivotaltracker.com/story/show/166293397)

Deploy after
- [x] https://github.com/alphagov/notifications-api/pull/2553
- [x] Data is cleaned up so that there are no services which have a different org_type to their organisation


tests/app/models/test_service.py
```diff
@@ -173,3 +174,18 @@ def test_get_template_folders_shows_all_folders_when_user_id_not_passed_in(
             'users_with_permission': [active_user_with_permissions['id']],
         }
     ]
+
+
+def test_organisation_type_when_service_has_no_org(mocker, service_one, organisation_one):
```
CrystalPea: I don't understand this test: in the title it says that service has no org, but when mocking `get_service_organisation` we return `organisation_one`. Shouldn't it return None?

tests/app/main/views/test_service_settings.py
```diff
@@ -163,9 +164,34 @@ def test_no_go_live_link_for_service_without_organisation(
...
+                               permissions=['sms', 'email'],
+                               organisation_id=ORGANISATION_ID)
+    mocker.patch('app.service_api_client.get_service', return_value={'data': service_one})
+
+    client_request.login(platform_admin_user, service_one)
+    response = client_request.get(
+        'main.service_settings', service_id=SERVICE_ONE_ID
     )
 
+    org_row = response.select('tr')[18]
```
CrystalPea: selecting row by number might be pretty brittle if we add another row or delete a row.

But then I'm not sure how to do it better without searching by xpath which, I think, we can't to? 🤔

tests/app/models/test_service.py
klssmith: This test title wasn't quite right - I've renamed this and the one below to hopefully explain what they're trying to do a bit better

klssmith:
@CrystalPea Yes, it means that we use the `organisation_type` of the service's org (if there is one) and the `organisation_type` of the service if there isn't.

PR #3049, opened 2019-07-10
Update support thanks page

karlchillmaid:
The thank you page content does not reflect our latest commitment to user support.

This PR updates the:

- thank you page content so it's consistent with the support page
- _if-then_ rules so that the correct response time is shown in and out of office hours


quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/3051

PR #3053, opened 2019-07-12
Add js tests for template folder form

tombye:
Adds tests for the JS template folder form module which controls all actions you can perform on the templates/folders hierarchy.

image

This work forms part of this story:

https://www.pivotaltracker.com/story/show/165409973

## How's it meant to work?

The initial PR for the module explains it pretty well: https://github.com/alphagov/notifications-admin/pull/2559

## Where can I find it?

This module only exists on the templates page: /services/<service_id>/templates

## How to run these tests

From the root of the repository, run:

`npx jest --config tests/javascripts/jest.config.js tests/javascripts/templateFolderForm.test.js`

### Other work included

[A bug was found when using the JAWS screenreader](https://www.pivotaltracker.com/story/show/165565088) when the /templates page was tested by the accessibility team. When focus is shifted to the legend of a fieldset, its contents should be announced but this doesn't happen.

This includes [some work to move focus to the parent fieldset](https://github.com/alphagov/notifications-admin/commit/9eed032f9f0ed8986dae58e39cd329041b318dd4) instead. This will need more testing but, even if it doesn't work, won't effect the current experience negatively.

### DIFF
--- a/tests/javascripts/templateFolderForm.test.js
+++ b/tests/javascripts/templateFolderForm.test.js
@@ -231,7 +231,7 @@ describe('TemplateFolderForm', () => {
   
     // We need parts of the module to be made sticky, but by the module code,
     // not the sticky JS code that operates on the HTML at page load.
-    // Because of this, they wll need to be marked with classes
+    // Because of this, they will need to be marked with classes
     test("the HTML for the module should contain placeholder classes on each part that needs to be sticky", () => {
 
       expect(templateFolderForm.querySelectorAll('#move_to_folder_radios > .js-will-stick-at-bottom-when-scrolling').length).toEqual(2);
--- a/tests/javascripts/templateFolderForm.test.js
+++ b/tests/javascripts/templateFolderForm.test.js
@@ -1,0 +1,645 @@
+const helpers = require('./support/helpers');
+
+function setFixtures (hierarchy) {
+
+  function templatesAndFoldersCheckboxesHTML () {
+    let result = '';
+
+    hierarchy.forEach((node, idx) => {
+
+      result += `
+        <div class="template-list-item template-list-item-with-checkbox  template-list-item-without-ancestors">
+          <div class="multiple-choice">
+            <input id="templates-or-folder-${idx}" name="templates_and_folders" type="checkbox" value="templates-or-folder-${idx}">
+            <label></label>
+          </div>
+          <h2 class="message-name">
+            <a href="/services/6658542f-0cad-491f-bec8-ab8457700ead/templates/all/folders/3d057d9a-51fc-45ea-8b63-0003206350a6" class="template-list-${node.type === 'folder' ? 'folder' : 'template'}">
+              <span class="live-search-relevant">${node.label}</span>
+            </a>
+          </h2>
+          ${node.meta}
+        </div>`;
+
+    });
+
+    return result;
+
+  };
+
+  const foldersCheckboxesHTML = function (filter) {
+    let count = 0;
+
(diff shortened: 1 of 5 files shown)


tests/javascripts/templateFolderForm.test.js
```diff
@@ -0,0 +1,753 @@
...
+
+  function getHiddenCounter () {
+    return formControls.querySelector('[role=status]');
+  };
+
+  describe("Before the page loads", () => {
+  
+    // We need parts of the module to be made sticky, but by the module code,
+    // not the sticky JS code that operates on the HTML at page load.
+    // Because of this, they wll need to be marked with classes
```
CrystalPea: a little typo here "wll" -> "will" :)

tombye: Fixed and now rebased. Thanks for the check :)

tests/javascripts/templateFolderForm.test.js
```diff
@@ -0,0 +1,753 @@
...
+      // .recalculate should have been called so the sticky JS picks up the controls
+      expect(GOVUK.stickAtBottomWhenScrolling.recalculate.mock.calls.length).toEqual(1);
+
+      // mode should have been set to 'default' as the controls only have one part
+      expect(GOVUK.stickAtBottomWhenScrolling.setMode.mock.calls.length).toEqual(1);
+      expect(GOVUK.stickAtBottomWhenScrolling.setMode.mock.calls[0][0]).toEqual('default');
+
+    });
+
+    describe("Clear link", () => {
```
CrystalPea: I think if "Clear link" was named "'Clear selection' link" it would be hmm... clearer 🥁

tests/javascripts/templateFolderForm.test.js
```diff
@@ -0,0 +1,753 @@
...
+        // .recalculate should have been called so the sticky JS picks up the controls
+        expect(GOVUK.stickAtBottomWhenScrolling.recalculate.mock.calls.length).toEqual(1);
+
+        // the mode should be set to 'dialog' so both parts can be sticky
+        expect(GOVUK.stickAtBottomWhenScrolling.setMode.mock.calls.length).toEqual(1);
+        expect(GOVUK.stickAtBottomWhenScrolling.setMode.mock.calls[0][0]).toEqual('dialog');
+
+      });
+
+      describe("When the 'Cancel' link is clicked", () => {
```
CrystalPea: I think you have two describes "When the 'Cancel' link is clicked" in this file, are they the same? I assume not, then how are they different? Should we highlight this difference in the describe maybe?

tombye: They aren't and that's not clear. I think I was assuming their nesting explained the interaction journey their test was in. It kind of does in the command-line when you run the tests, through indentation, but not when you're reading the code.

I've changed the descriptions to make this more obvious in https://github.com/alphagov/notifications-admin/pull/3053/commits/0543dca23d314d38b329047d77b9efe8ff6ef293

Really good spot and exactly the kind of thing I was hoping a code review would turn up 👍

tests/javascripts/templateFolderForm.test.js
tombye: Agreed, and fixed in https://github.com/alphagov/notifications-admin/pull/3053/commits/b6b1e7682bd6e88cefcc22905e43ce97a76a76f2

tombye:
I was looking for the clarity of understanding more than the quality of code so this is really useful

PR #3057, opened 2019-07-16
Introduce new org types

CrystalPea:
#### In this PR:
- change org types to choose from to new org types (with nhs_local value coming under two labels to increase usability following Chris' advice)
- change displayed org types to new org types but leave generic `'nhs'` for now until we migrate it out of existence
- for new users with nhs email, we cannot assign org type automatically anymore, because they can either be `'nhs_local'` or `'nhs_central'` so we ask them which NHS-y org type are they

#### TODO  after we migrate all orgs and services to new org types:
- remove all references to generic `'nhs'` org type


app/models/service.py
```diff
@@ -268,7 +268,9 @@ def needs_to_add_email_reply_to_address(self):
 
     @property
     def shouldnt_use_govuk_as_sms_sender(self):
-        return self.organisation_type in {'local', 'nhs'}
+        return self.organisation_type in {
+            'local', 'nhs', 'nhs_local', 'nhs_central', 'emergency_service', 'school_or_college', 'other'
```
servingUpAces: Is `central` missing? Is there a global list we can use for the list of org types?

quis: `central` can use `GOVUK` as a text message sender – that’s why it’s not in the list.

It might be clearer to say 
```python
return self.organisation_type != 'central'
```

CrystalPea: Ahh, good suggestion @quis 👍

CrystalPea: @servingUpAces there is no global list in admin. I thought about doing one, but then in the forms I need to repeat `'nhs_local'` value twice, with different labels, so I couldn't use that global list to iterate over that, so I wasn't sure. If you think that global list would be useful I could still add it.

PR #3065, opened 2019-07-22
Remove ft from method and url, it doesn't add any meaning.

servingUpAces:


app/notify_client/billing_api_client.py
```diff
@@ -3,15 +3,15 @@
 
 class BillingAPIClient(NotifyAdminAPIClient):
 
-    def get_billable_units_ft(self, service_id, year):
+    def get_billable_units(self, service_id, year):
         return self.get(
-            '/service/{0}/billing/ft-monthly-usage'.format(service_id),
```
CrystalPea: have we changed this and below address in API?

CrystalPea: ahh ok I see both addresses work in API, nice 👍

servingUpAces: Both exist - another PR in API will remove the one that's no longer used.

PR #3068, opened 2019-07-24
Delete service cache when changing an organisation's sector

klssmith:
When we change an organisation's sector we now also change the sector of all its services, so we need to delete those services from Redis.

This also changes the order of the code to ensure that we only add / delete things from Redis if an api request was successful.

### DIFF
--- a/app/notify_client/organisations_api_client.py
+++ b/app/notify_client/organisations_api_client.py
@@ -2,6 +2,7 @@ from itertools import chain
 
 from notifications_python_client.errors import HTTPError
 
+from app.extensions import redis_client
 from app.notify_client import NotifyAdminAPIClient, _attach_current_user, cache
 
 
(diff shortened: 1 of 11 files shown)


app/notify_client/organisations_api_client.py
```diff
@@ -45,7 +46,10 @@ def create_organisation(self, name, crown, organisation_type, agreement_signed):
 
     @cache.delete('domains')
     @cache.delete('organisations')
-    def update_organisation(self, org_id, **kwargs):
+    def update_organisation(self, org_id, cached_service_ids=None, **kwargs):
+        if kwargs.get('organisation_type') and cached_service_ids:
```
servingUpAces: I think this needs to happen after the response from api

klssmith: 👍 I've changed the order

PR #3070, opened 2019-07-25
Say ‘team member’ not ‘user’

quis:
This is currently inconsistent with the language we use everywhere else.

### DIFF
--- a/app/templates/views/edit-user-permissions.html
+++ b/app/templates/views/edit-user-permissions.html
@@ -40,7 +40,7 @@
         {{ page_footer(
           'Save',
           delete_link=url_for('.edit_user_permissions', service_id=current_service.id, user_id=user.id, delete='yes'),
-          delete_link_text='Remove user from service'
+          delete_link_text='Remove team member from service'
         ) }}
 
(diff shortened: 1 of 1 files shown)


app/templates/views/edit-user-permissions.html
```diff
@@ -40,7 +40,7 @@
         {{ page_footer(
           'Save',
           delete_link=url_for('.edit_user_permissions', service_id=current_service.id, user_id=user.id, delete='yes'),
-          delete_link_text='Remove user from service'
+          delete_link_text='Remove team member from service'
```
karlchillmaid: Should we say 'this service'?

quis: Options:
***
1. 
  image
2.
  image
3.
  image
4.
  image
5.
  image

I think 2. is my preference…

karlchillmaid: Agreed.

quis:
```suggestion
          delete_link_text='Remove this team member'
```

PR #3082, opened 2019-08-09
Add 'Uploads' hub and navigation

klssmith:
The uploads hub is just a page with text for now - there are no actions available on the page. It is linked to from a new 'Uploads' menu item on the left of the page which is only visible if your service has the `letter` and `upload_letters` permissions and if the current user has permissions to send messages.

[Pivotal story](https://www.pivotaltracker.com/story/show/167533429)

Screen Shot 2019-08-09 at 11 23 37

### DIFF
--- a/tests/app/main/views/test_uploads.py
+++ b/tests/app/main/views/test_uploads.py
@@ -1,0 +1,5 @@
+from tests.conftest import SERVICE_ONE_ID
+
+
+def test_get_upload_hub_page(client_request):
+    client_request.get('main.uploads', service_id=SERVICE_ONE_ID)
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -33,6 +33,7 @@ from app.main.views import (  # noqa isort:skip
     styleguide,
     templates,
     two_factor,
+    uploads,
     user_profile,
     verify,
 )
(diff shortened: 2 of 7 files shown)


tests/app/main/views/test_uploads.py
```diff
@@ -0,0 +1,5 @@
+from tests.conftest import SERVICE_ONE_ID
+
+
+def test_get_upload_hub_page(client_request):
+    client_request.get('main.uploads', service_id=SERVICE_ONE_ID)
```
CrystalPea: We could also assert that header is correct in here, what do you think?

klssmith: Do you mean the heading on this page or the request headers?

CrystalPea: @klssmith just heading on this page, sorry for ambiguity.

klssmith: Hmm, could do, but I'm not sure that testing the exact words of the heading would add much - there's no logic around the heading text and this test will fail if the heading is not there

CrystalPea: I think the purpose of that assertion would be to check that you are indeed on the right page.

PR #3085, opened 2019-08-12
Add js tests for api key

tombye:
Adds tests for the API key module which allows you to copy your key to the clipboard.

api_key

This work forms part of this story:

https://www.pivotaltracker.com/story/show/165409926

## How's it meant to work?

There's a good explanation in the initial PR: https://github.com/alphagov/notifications-admin/pull/80.

## How to run these tests

From the root of this repository, run:

`npx jest --config tests/javascripts/jest.config.js tests/javascripts/apiKey.test.js`

### DIFF
--- a/tests/javascripts/apiKey.test.js
+++ b/tests/javascripts/apiKey.test.js
@@ -1,0 +1,199 @@
+const helpers = require('./support/helpers');
+
+afterAll(() => {
+
+  require('./support/teardown.js');
+
+  // clear up methods in the global space
+  document.queryCommandSupported = undefined;
+
+});
+
+describe('API key', () => {
+
+  let apiKey;
+  let thing;
+  let component;
+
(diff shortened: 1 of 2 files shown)


tombye:
> Very thorough. Only thing I can see that isn’t covered is the line which sets the `min-height`:
> https://github.com/alphagov/notifications-admin/blob/1e248d92051b90b2425e165176804fd0aeb70850/app/assets/javascripts/apiKey.js#L38
> 
> There’s an explanation for why we set this here: #2428
> 
> Would be good to add a test to prevent regression.

Thanks for that! I've added a test for it in [this commit](https://github.com/alphagov/notifications-admin/pull/3085/commits/ddd8da016302ce86a4d1fbb877b4294c2331a978). Let me know if it needs any more work.

PR #3089, opened 2019-08-13
Find services by name or partial name

CrystalPea:
A new platform admin view that allows internal user to find services by name or partial name

Needs to be deployed after: https://github.com/alphagov/notifications-api/pull/2587

Part of this story: https://www.pivotaltracker.com/story/show/167806249


CrystalPea:
@servingUpAces ahh, yes, thanks, I committed the missing files now :)

app/main/views/find_services.py
```diff
@@ -0,0 +1,23 @@
...
+
+@main.route("/find-services-by-name", methods=['GET', 'POST'])
+@user_is_platform_admin
+def find_services_by_name():
+    form = SearchByNameForm()
+    services_found = None
+    status = 200
+    if form.validate_on_submit():
+        services_found = service_api_client.find_services_by_name(service_name=form.search.data)['data']
+    elif request.method == 'POST':
```
servingUpAces: If you add a `DataRequired` validator to the SearchByNameForm, then the view will render the validation error on the page, rather than from the server. 
That way you don't need this else statement, nor do you need the `status` variable.

tests/app/main/views/test_find_services.py
```diff
@@ -0,0 +1,73 @@
...
+    client_request.login(platform_admin_user)
+    mocker.patch('app.service_api_client.find_services_by_name', return_value={"data": []})
+    document = client_request.post(
+        'main.find_services_by_name', _data={"search": "Nabuchodonosorian Empire"}, _expected_status=200
+    )
+
+    assert document.find('p', {'class': 'browse-list-hint'}).text.strip() == 'No services found.'
+
+
+# def test_find_users_by_email_validates_against_empty_search_submission(
```
servingUpAces: Commented out test?

CrystalPea: Ahh, you're right, thanks, it's uncommented and working now ☺️

app/main/views/find_services.py
CrystalPea: Ahh, that's smart, I've done it now, thanks :)

PR #3096, opened 2019-08-20
Update Documentation content

karlchillmaid:
This PR updates the Documentation page to:
- make the target user and purpose of the page clearer
- tell users how long it might take to integrate with the API
- tell users they can integrate directly with the API
- provide important information about integrating directly with the API

### DIFF
--- a/app/templates/views/documentation.html
+++ b/app/templates/views/documentation.html
@@ -7,9 +7,9 @@
 {% block content_column_content %}
 
   <h1 class="heading-large">Documentation</h1>
-    <p>Use the GOV.UK Notify API to send messages automatically.</p>
+    <p>This documentation is for developers who want to integrate the GOV.UK Notify API with a web application or back office system.</p>
   <h2 class="heading-medium">Clients</h2>
-    <p>Choose a client to integrate our API with your web application or back-office system. Links to documentation open in a new tab.</p>
+    <p>Links to documentation open in a new tab.</p>
   <ul class="list list-bullet">
     {% for key, label in [
       ('java', 'Java'),
@@ -22,5 +22,5 @@
       <li><a href="https://docs.notifications.service.gov.uk/{{ key }}.html" target="_blank" rel="noopener">{{ label }}</a></li>
     {% endfor %}
   </ul>
-
+    <p>You can work directly with the API, but this may be complicated to set up. <a href="{{ url_for('.feedback', ticket_type='ask-question-give-feedback') }}">Contact us</a> for more information.</p>
 {% endblock %}
--- a/app/templates/views/documentation.html
+++ b/app/templates/views/documentation.html
@@ -30,5 +30,5 @@
     <li>what the API client does</li>
     <li>what development work you'll need to do</li>
   </ul>
-  <p>To send a message you'll need to create an HTTPS request and add an authorisation header with your API key. The API key is encoded with JSON Web Tokens (JWT).</p>
+  <p>To send a message you’ll need to create an HTTPS request and add an authorisation header with your API key. The API key is encoded with JSON Web Tokens (JWT).</p>
 {% endblock %}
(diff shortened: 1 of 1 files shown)


karlchillmaid:
Updates made after discussing with @servingUpAces and @jonathanglassman.

yahoopete:
I think we need to say that we recommend using one of our clients here. This will invite requests for the API spec I think...  Also think saying a full integration takes a couple of days is a bit misleading for teams who aren't just changing supplier, because most of the work is customising their systems as a result of the new functionality. Whether that's worth stating or not is another story.....

karlchillmaid:
@yahoopete 
Content updated based on your comments.

quis:
In each of the client documentation we say:

> We recommend that you use this client library rather than use the [GOV.UK Notify API](https://github.com/alphagov/notifications-api) directly, as there is no documentation for using the API in this way.

I’d rather we just said it here, rather than there, now that we’re explaining it better here.

app/templates/views/documentation.html
```diff
@@ -22,5 +22,13 @@ <h2 class="heading-medium">Clients</h2>
...
     {% endfor %}
   </ul>
-
+    <p>A developer should be able to set up the API client and start sending test messages in around 30 minutes. A full integration can take a few days, depending on the other systems you're using.</p>
+  <h2>Integrating directly with the API</h2>
+    <p>We recommend using our client documentation rather than integrating directly with the API.</p>
+    <p>There's no documentation for using the API in this way. You'll still need to read the client documentation to understand:</p>
+  <ul class="list list-bullet">
+    <li>what the API client does</li>
+    <li>what development work you'll need to do</li>
```
quis:
```suggestion
    <li>what development work you’ll need to do</li>
```

app/templates/views/documentation.html
```diff
@@ -22,5 +22,13 @@ <h2 class="heading-medium">Clients</h2>
...
+  </ul>
+  <p>To send a message you'll need to create an HTTPS request and add an authorisation header with your API key. The API key is encoded with JSON Web Tokens (JWT).</p>
```
quis:
```suggestion
  <p>To send a message you’ll need to create an HTTPS request and add an authorisation header with your API key. The API key is encoded with JSON Web Tokens (JWT).</p>
```

app/templates/views/documentation.html
```diff
@@ -22,5 +22,13 @@ <h2 class="heading-medium">Clients</h2>
       <li><a href="https://docs.notifications.service.gov.uk/{{ key }}.html" target="_blank" rel="noopener">{{ label }}</a></li>
     {% endfor %}
   </ul>
-
+    <p>A developer should be able to set up the API client and start sending test messages in around 30 minutes. A full integration can take a few days, depending on the other systems you're using.</p>
+  <h2>Integrating directly with the API</h2>
+    <p>We recommend using our client documentation rather than integrating directly with the API.</p>
+    <p>There's no documentation for using the API in this way. You'll still need to read the client documentation to understand:</p>
```
quis:
```suggestion
    <p>There’s no documentation for using the API in this way. You'll still need to read the client documentation to understand:</p>
```

app/templates/views/documentation.html
```diff
@@ -22,5 +22,13 @@ <h2 class="heading-medium">Clients</h2>
       <li><a href="https://docs.notifications.service.gov.uk/{{ key }}.html" target="_blank" rel="noopener">{{ label }}</a></li>
     {% endfor %}
   </ul>
-
+    <p>A developer should be able to set up the API client and start sending test messages in around 30 minutes. A full integration can take a few days, depending on the other systems you're using.</p>
```
quis:
```suggestion
    <p>A developer should be able to set up the API client and start sending test messages in around 30 minutes. A full integration can take a few days, depending on the other systems you’re using.</p>
```

karlchillmaid:
@quis 
> I’d rather we just said it here, rather than there, now that we’re explaining it better here.

I'll get PRs raised for each of the clients

PR #3101, opened 2019-08-23
Usage for all services

servingUpAces:
New report on platform-admin/reports page to get the usage for all services.


tests/app/main/views/test_platform_admin.py
```diff
@@ -1079,6 +1079,54 @@ def test_get_notifications_sent_by_service_validates_form(mocker, client_request
...
+        'attachment; filename="Usage for all services from {} to {}.csv"'.format('2019-01-01', '2019-03-31')
+    )
+    # I don't know how to get this assert to work
+#     assert response.get_data(as_text=True) == (
+#         """organisation_id,organisation_name,service_id,service_name,sms_cost,sms_fragments,
+# letter_cost,letter_breakdown
+# 7832a1be-a1f0-4f2a-982f-05adfd3d6354,Org for a - with sms and letter,48e82ac0-c8c4-4e46-8712-c83c35a94006,
+# a - with sms and letter,0,0,3.4,"6 second class letters at 45p\n
+# 2 first class letters at 35p"
+# """
```
leohemsted: * new lines inside a docstring need to reflect the new lines in the actual data - so you can't split the string eg in the headers between sms_fragments and letter_cost. If you want to split the test data over lines to make it more readable you'll need to use string concatentation `'blah' + 'blah'` and then declare the new lines individually
* there's also a new line at the end of the "2 first class letters", but inside the data, so inside the double quotation mark
* the lines are actually separated by `\r\n` (thanks to https://docs.python.org/3/library/csv.html#csv.Dialect.lineterminator) - so we need to directly assert that those characters are here

```python
    assert response.get_data(as_text=True) == (
        'organisation_id,organisation_name,service_id,service_name,' +
        'sms_cost,sms_fragments,letter_cost,letter_breakdown' +

        '\r\n' +
        
        '7832a1be-a1f0-4f2a-982f-05adfd3d6354,' +
        'Org for a - with sms and letter,' +
        '48e82ac0-c8c4-4e46-8712-c83c35a94006,' +
        'a - with sms and letter,' +
        '0,' +
        '0,' +
        '3.4,' +
        '"6 second class letters at 45p' +
        '\n' +
        '2 first class letters at 35p"' +
        
        '\r\n'
    )
```

app/templates/views/platform-admin/usage_for_all_services.html
```diff
@@ -0,0 +1,20 @@
+{% extends "views/platform-admin/_base_template.html" %}
```
leohemsted: i dont think it's in scope to change here but i hate the mix of `-` and `_` all over the admin template filenames

tests/app/main/views/test_platform_admin.py
```diff
@@ -1079,6 +1079,54 @@ def test_get_notifications_sent_by_service_validates_form(mocker, client_request
...
+    client_request.login(platform_admin_user)
+
+    mocker.patch("app.main.views.platform_admin.billing_api_client.get_usage_for_all_services",
+                 return_value=[])
+
+    page = client_request.post('main.usage_for_all_services',
+                               _expected_status=200,
+                               _data={'start_date': '2019-01-01', 'end_date': '2019-03-31'})
+
+    errors = page.select('.error-message')
```
leohemsted: you can use `error = page.select_one('.error-message')` to avoid having to loop through a one-long array

leohemsted: also this was hiding an error - errors was empty, so the loop below wasn't being entered and no asserts were given

PR #3108, opened 2019-08-29
Remove expandCollapse JS

tombye:
It's no longer used so worth removing. Includes removal of associated Sass.


quis:
Think you can delete these lines too: https://github.com/alphagov/notifications-admin/blob/c9c3f90fd8158bb6177266b503ea1de944401d6e/app/assets/stylesheets/components/email-message.scss#L73-L81

tombye:
> Think you can delete these lines too:
> 
> https://github.com/alphagov/notifications-admin/blob/c9c3f90fd8158bb6177266b503ea1de944401d6e/app/assets/stylesheets/components/email-message.scss#L73-L81

☝️ should be gone with 26e1e3a

PR #3114, opened 2019-09-03
Remove code specific to IE8 and below

quis:
Our usage for these browsers in the last month is down to 0.2% of all users, or 14 individual users, according to Google Analytics.

These users also visit about half the number of pages per sessions, suggesting that they’re not signed in.


app/templates/admin_template.html
```diff
@@ -2,21 +2,10 @@
 {% from "components/banner.html" import banner %}
 
 {% block head %}
-  <!--[if gt IE 8]><!-->
```
tombye: Not specifically about this line but is it worth wrapping the `<script>` tag for `all.js` in a conditional to not deliver it to IE<9 as part of this?

The JS errors on IE8 and so users see the HTML + CSS version anyway but that would mean they don't download the JS.

quis: Added in df2742f

PR #3116, opened 2019-09-04
Add tests for prevent duplicate form submissions JS

tombye:
Adds tests for the script that prevents forms being submitted more than once by blocking subsequent clicks on the submit button.

## How to run the tests

`npx jest --config tests/javascripts/jest.config.js tests/javascripts/preventDuplicateFormSubmissions.test.js`

### DIFF
--- a/tests/javascripts/preventDuplicateFormSubmissions.test.js
+++ b/tests/javascripts/preventDuplicateFormSubmissions.test.js
@@ -1,0 +1,68 @@
+const helpers = require('./support/helpers.js');
+
+beforeAll(() => {
+  jest.useFakeTimers();
+});
+
+afterAll(() => {
+  require('./support/teardown.js');
+});
+
+describe('Prevent duplicate form submissions', () => {
+
+  let form;
+  let button;
+  let formSubmitSpy;
+
+  beforeEach(() => {
+
+    // set up DOM
+    document.body.innerHTML = `
+      <form action="/" method="post">
+        <button class="button" type="submit">Continue</button>
+      </form>`;
+
+    form = document.querySelector('form');
+    button = document.querySelector('button');
+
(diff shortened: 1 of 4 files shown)


tests/javascripts/preventDuplicateFormSubmissions.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+    expect(formSubmitSpy.mock.calls.length).toEqual(1);
+
+  });
+
+  test("It should allow clicks again after 1.5 minutes", () => {
+
+    helpers.triggerEvent(button, 'click');
+
+    jest.advanceTimersByTime(1500);
```
CrystalPea: if 1.5 minute is 1500 units, what are those units? :o

tombye: Good spot! I've added a comment explaining they're milliseconds. Will update all other tests with this in too.

CrystalPea: Thanks so much, today I've learnt ☺️

CrystalPea: But isn't 1 second == 1000 milliseconds? And then 1500 milliseconds would be 1.5 second not 1.5 minute... right?

tombye: Sorry, missed that completely. I've rebased this to just add a commit fixing the test description as I don't think it needs any more explanation if it's talking about the right units.

CrystalPea: Ahh, ok, no worries and thanks, I think that was what bamboozled me at first :D

CrystalPea: doggo

PR #3117, opened 2019-09-04
Let GP surgeries create their own organisations before going live

quis:
image

***

image

***

image


karlchillmaid:
Suggested content:

>**Accept our data sharing and financial agreement**
>
>Is your practice called ((service_name))?
>- Yes
>- No
>  What’s your practice called?
>
>Continue

quis:
image

***

image

***

image

PR #3118, opened 2019-09-04
Handle new and old response from template preview's validate endpoint

CrystalPea:
We are changing response format in template preview ( here: alphagov/notifications-template-preview#367 ). We want admin app to handle both new and old response type for a smooth deploy of those changes.

Story ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/168057359

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -425,7 +425,14 @@ def letter_validation_preview(from_platform_admin):
             response = validate_letter(pdf_file)
             response.raise_for_status()
             if response.status_code == 200:
-                pages, message, result = response.json()["pages"], response.json()["message"], response.json()["result"]
+                pages, result = response.json()["pages"], response.json()["result"]
+                if "message" in response.json():
+                    message = response.json()["message"]
+                else:
+                    message = "Validation failed, "
+                    for k, v in response.json()["errors"].items():
+                        if v:
+                            message += k.replace("_", " ") + " on pages: {}".format(v)
         except RequestException as error:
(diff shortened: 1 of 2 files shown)


app/main/views/platform_admin.py
```diff
@@ -425,7 +425,14 @@ def letter_validation_preview(from_platform_admin):
             response = validate_letter(pdf_file)
             response.raise_for_status()
             if response.status_code == 200:
-                pages, message, result = response.json()["pages"], response.json()["message"], response.json()["result"]
+                pages, result = response.json()["pages"], response.json()["result"]
```
leohemsted: minor, (and appreciate old code was like this) but it'd be nice to store `resp = response.json()` or something to stop having to call response.json every time. i'm sure flask caches the dict, but still it's a good habit to be in

app/main/views/platform_admin.py
```diff
@@ -425,7 +425,14 @@ def letter_validation_preview(from_platform_admin):
...
+                if "message" in response.json():
+                    message = response.json()["message"]
+                else:
+                    message = "Validation failed, "
+                    for k, v in response.json()["errors"].items():
+                        if v:
+                            message += k.replace("_", " ") + " on pages: {}".format(v)
```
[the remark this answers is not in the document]
CrystalPea: That's a good idea 👍. I will implement it ☺️

app/main/views/platform_admin.py
CrystalPea: Good catch :)

CrystalPea:
Not needed anymore

PR #3119, opened 2019-09-05
Retire old static content pages

quis:


app/templates/views/api/callbacks/delivery-status-callback.html
```diff
@@ -19,7 +19,7 @@
 
       <p>
         When you send an email or text message, we can tell you if Notify was able to deliver it.
-        Check the <a href="{{ url_for('.callbacks') }}"> callback documentation </a> for more information.
+        Read the Callbacks section of our <a href="{{ url_for('main.documentation') }}">API documentation</a> for more information.
```
tombye: Should this reference the api clients directly, like "Read the Callbacks section of our API documentation for the client library you're using."?

quis: Could do, but it’s going to make the sentence a right mouthful, ie _…for the client library you’re using for more information._

PR #3122, opened 2019-09-12
Replace negative contractions

karlchillmaid:
Some users find negative contractions harder to understand. We use them a lot on Notify.

The latest [guidance from GOV.UK](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#contractions) says:

>Avoid negative contractions like can’t and don’t. Many users find them harder to read, or misread them as the opposite of what they say. Use cannot, instead of can’t.
>Avoid should’ve, could’ve, would’ve, they’ve too. These can also be hard to read.

This PR replaces:
- can't with cannot
- don't with do not
- won't with will not
- isn't with is not
- didn't with did not
- wasn't with was not
- couldn't with could not
- doesn't with does not
- aren't with are not
- hasn't with has not
- haven't with have not

It also replaces one instance of they've with they have.

For more information, read: [Using contractions could be making your writing inaccessible](https://medium.com/@joanne.schofield/using-contractions-could-be-making-your-writing-inaccessible-4034bb3cd76a).


app/__init__.py
```diff
@@ -376,7 +376,7 @@ def format_notification_status(status, template_type):
             'failed': 'Failed',
             'technical-failure': 'Technical failure',
             'temporary-failure': 'Inbox not accepting messages right now',
-            'permanent-failure': 'Email address doesn’t exist',
+            'permanent-failure': 'Email address does not exist',
```
quis: Check wrapping…

app/__init__.py
```diff
@@ -386,7 +386,7 @@ def format_notification_status(status, template_type):
             'failed': 'Failed',
             'technical-failure': 'Technical failure',
             'temporary-failure': 'Phone not accepting messages right now',
-            'permanent-failure': 'Phone number doesn’t exist',
+            'permanent-failure': 'Phone number does not exist',
```
quis: Check wrapping…

app/templates/views/manage-users/permissions.html
```diff
@@ -26,7 +26,7 @@
     {{ radios(
       form.login_authentication,
       disable=['sms_auth'],
-      option_hints={'sms_auth': 'Not available because this team member hasn’t added a phone&nbsp;number to their profile'|safe}
+      option_hints={'sms_auth': 'Not available because this team member has not added a phone&nbsp;number to their profile'|safe}
```
quis: Check wrapping…

app/templates/views/notifications/invalid_precompiled_letter.html
```diff
@@ -12,6 +12,6 @@ <h1 class="heading-large">Letter</h1>
       Provided as PDF on {{ created_at|format_datetime_short }}
     </p>
     <p class="notification-status-cancelled">
-      Validation failed – this isn’t a PDF file that Notify can read
+      Validation failed – this is not a PDF file that Notify can read
```
quis: Would this be better?
```suggestion
      Validation failed – Notify cannot read this PDF file
```

app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -29,7 +29,7 @@ <h2 class='banner-title' data-module="track-error" data-error-type="reply-to-ema
           There’s a problem with your reply-to address
         </h2>
         <p>
-          We sent an email to ‘{{ reply_to_email_address }}’ but it wasn’t delivered.
+          We sent an email to ‘{{ reply_to_email_address }}’ but it cannot be delivered.
```
quis: Feel like this should be past tense, for example

```suggestion
          We sent an email to ‘{{ reply_to_email_address }}’ but it was not delivered.
```
or
```suggestion
          We sent an email to ‘{{ reply_to_email_address }}’ but it could not be delivered.
```

app/templates/views/service-settings/sms-sender/edit.html
```diff
@@ -20,7 +20,7 @@
     {% if inbound_number %}
       <p>
         <span class="bottom-gutter-1-3"> {{ sms_sender.sms_sender }} </span>
-        <span class="hint"> This phone number receives replies and can’t be changed </span>
+        <span class="hint"> This phone number receives replies and cannot be changed </span>
```
quis: Check wrapping…

app/templates/views/signedout.html
```diff
@@ -49,7 +49,7 @@ <h2>
           Control your content
         </h2>
         <p>
-          You don’t need any technical knowledge to create email,
+          You do not need any technical knowledge to create email,
```
quis: Check wrapping…

app/templates/views/signedout.html
```diff
@@ -66,7 +66,7 @@ <h2>
         </h2>
         <p>
           Track how many messages you’ve sent and
-          find out which ones aren’t getting delivered.
+          find out which ones are not being delivered.
```
quis: Check wrapping…

app/templates/views/notifications/invalid_precompiled_letter.html
karlchillmaid: Yes, definitely. For this PR I stuck to just replacing negative contractions. I wanted us to pick up the rewording of some of our validation errors as a separate thing, but we can make this change now.

PR #3123, opened 2019-09-13
Let users give us better requests for email branding

quis:
Users who work in local government can’t have GOV.UK branding on their emails. And only those working for Companies House (for example) can request the Companies House branding.

This commit adds:
- new choices of email branding, which offer the name of the branding, rather than the style
- logic to filter this list to only the applicable options, based on what we know about the user, service and organisation

This is a change from the previous approach which gave users the choice of branding style they wanted, when we might already know that a lot of the options weren’t available to them, or would be inconsistent with the branding of other services in their organisation.

# Before

image

# After 

image

image

image

image


app/main/forms.py
```diff
@@ -1357,24 +1357,75 @@ def __init__(self, *args, **kwargs):
...
+    @property
+    def something_else_is_only_option(self):
+        return self.options.choices == (self.FALLBACK_OPTION,)
+
+    def validate_something_else(self, field):
+        if (
+            self.something_else_is_only_option or
+            self.options.data == self.FALLBACK_OPTION_VALUE
+        ) and not field.data:
+            raise ValidationError('Can’t be empty')
```
karlchillmaid: Replace negative contraction with `Cannot`

quis:
```suggestion
            raise ValidationError('Cannot be empty')
```

PR #3125, opened 2019-09-17
Create the app once per run, not once per test

quis:
Turns out our tests spent a lot of time recreating the app for each test case, which is quite intense.

This commit moves the app-creating code outside the fixture, so the fixture re-uses the same app for every test.

This cuts down the time taken to run the test suite to about 50 seconds.

It also makes the tests more parallelizable. Before this change going from 4 to 8 processes made the tests slower. Now it cuts them down from about 50 seconds to about 35 seconds<sup>1</sup>. So this commit also lets Pytest choose the best number of processes to run, since on my machine it chooses 8, which is the fastest.

Overall this means the tests take about 15% of the time they used to. Assuming each person on the teams runs the tests on average once a day, this will save 15 hours per person, per year.

# Before 

image

image

# After 

image

image

***

1. Based on a 2.2GHz quad-core Intel Core i7 processor on a 2015 MacBook Pro

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,7 +36,7 @@ class ElementNotFound(Exception):
     pass
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
 def app_(request):
     app = Flask('app')
     create_app(app)
@@ -46,8 +46,6 @@ def app_(request):
 
     app.test_client_class = TestClient
     yield app
-
-    ctx.pop()
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -31,22 +31,20 @@
...
+create_app(app)
+app.test_client_class = TestClient
+ctx = app.app_context()
+
 
 class ElementNotFound(Exception):
     pass
 
 
 @pytest.fixture
```
leohemsted: you can just change this to `@pytest.fixture('session')` to get it to execute once, and save on module level code

servingUpAces:
One small change for Admin, one giant savings for devs. 
Yahoo!

tests/conftest.py
quis: Re-wrote in 1a353d90ebf32285f2f7e53807bfe095fcfe8099

PR #3126, opened 2019-09-18
Add 'Printed' status for letters and update postage information

karlchillmaid:
This PR:

- updates the postage information on the Features page
- adds a description of the 'Printed' status for letters on the Message statuses page

These small changes should make it easier for users to understand what happens when they send a letter.

### DIFF
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -81,7 +81,7 @@
     ) %}
       {% for message_length, charge in [
         ('Sent', 'Notify has sent the letter to the provider to be printed.'),
-        ('Printed', 'The provider has printed the letter. Letters are printed at 5:30pm and dispatched the next working day.'), 
+        ('Printed', 'The provider has printed the letter. Printing starts at 5:30pm and letters are dispatched the next working day.'), 
         ('Cancelled', 'Sending cancelled. Your letter will not be printed or dispatched.'),
         ('Technical failure', 'Notify had an unexpected error while sending the letter to our printing provider.'),
       ] %}
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -81,6 +81,7 @@
     ) %}
       {% for message_length, charge in [
         ('Sent', 'Notify has sent the letter to the provider to be printed.'),
+        ('Printed', 'The provider has printed the letter. Letters are printed at 5:30pm and dispatched the next working day.'), 
         ('Cancelled', 'Sending cancelled. Your letter will not be printed or dispatched.'),
         ('Technical failure', 'Notify had an unexpected error while sending the letter to our printing provider.'),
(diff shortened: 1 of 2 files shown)


app/templates/views/message-status.html
```diff
@@ -81,6 +81,7 @@ <h3 class="heading-small">Letters</h3>
     ) %}
       {% for message_length, charge in [
         ('Sent', 'Notify has sent the letter to the provider to be printed.'),
+        ('Printed', 'The provider has printed the letter. Letters are printed at 5.30pm and dispatched the next working day.'),
```
CrystalPea: Do we have a notification status called `Printed` in the database? 🤔

karlchillmaid: Sent = Accepted (GOV.UK Notify has sent the letter to the provider to be printed.)
Printed = Received (The provider has printed and dispatched the letter.)

CrystalPea: Ahh, ok, thanks for the explanation, makes sense 👍

CrystalPea: Still it feels wrong to have sent before printed. A letter is usually first printed and then sent.

CrystalPea: > I guess my question is, what triggers the change from the 'Printing starts...' message/status to 'Printed at...'?

this is triggered at 17:30 after we got the letter from the user. So I think `printed` and `sent to DVLA` is the same status.

karlchillmaid: The definition of 'Sent' is `Notify has sent the letter to the provider to be printed.` The timestamp we show for Sent displays the time the user submitted the letter. Letters are always printed _after_ this, at 5.30pm, and dispatched the following day.

On its own, I think the word 'Sent' is confusing, because it can mean different things. Which is why we need to make some of these labels and descriptions clearer.

Chris and I are talking about whether we need to add a `dispatched` status, to make it clearer to users that their letters have actually been posted.

But for now, explaining the 'printed' thing is a start.

CrystalPea: >The definition of 'Sent' is Notify has sent the letter to the provider to be printed.
We send all letters to provider at 5:30pm

> The timestamp we show for Sent displays the time the user submitted the letter.
That's at odds with the above definition, no?

karlchillmaid: Technically, yes.
If we wanted to be more accurate we'd call the status 'Submitted' (or something else), and use this page to explain that the letter has been submitted to Notify, but that we wait until 5.30pm to pass it to the provider – along with all the other letters submitted in the last 24 hours – for printing.
However, there does come a point where exposing all the detail about the process gets confusing.
I'm not against revising the existing app statuses for letters – I just think we need to approach it with caution. Let me know if you think it would be helpful to chat some more in person.

PR #3127, opened 2019-09-18
Update spreadsheet validation errors

karlchillmaid:
This PR updates:
- some validation errors to make them clearer
- the button text

---

Screen Shot 2019-09-27 at 11 41 39

---

Screen Shot 2019-09-27 at 11 43 32

---

Screen Shot 2019-09-27 at 11 45 53

### DIFF
--- a/app/templates/views/check/column-errors.html
+++ b/app/templates/views/check/column-errors.html
@@ -146,7 +146,7 @@
         {{ file_upload(
           form.file,
           action=url_for('.send_messages', service_id=current_service.id, template_id=template.id),
-          button_text='Upload a new file'
+          button_text='Upload your file again'
         ) }}
       {% endif %}
     </div>
--- a/app/templates/views/check/column-errors.html
+++ b/app/templates/views/check/column-errors.html
@@ -38,7 +38,7 @@
       {% elif not count_of_recipients %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="No rows" data-error-label="{{ upload_id }}">
-          Your file is missing rows
+          Your file is missing some rows
         </h1>
         {% if recipients.missing_column_headers %}
           <p>
@@ -56,7 +56,7 @@
       {% elif not recipients.has_recipient_columns %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="Missing recipient columns" data-error-label="{{ upload_id }}">
-          Your file is missing columns
+          Your file is missing some columns
         </h1>
           <p>
             It needs {{ (
@@ -95,7 +95,7 @@
       {% elif recipients.missing_column_headers %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="Missing placeholder columns" data-error-label="{{ upload_id }}">
-          Your file is missing columns
+          Your file is missing some columns
         </h1>
           <p>
             The column names need to match the double brackets in your template.
--- a/app/templates/views/check/column-errors.html
+++ b/app/templates/views/check/column-errors.html
@@ -38,7 +38,7 @@
       {% elif not count_of_recipients %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="No rows" data-error-label="{{ upload_id }}">
-          Your file is missing some rows
+          Your file is missing rows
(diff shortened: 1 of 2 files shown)


app/templates/views/check/column-errors.html
```diff
@@ -56,49 +56,57 @@ <h1 class='banner-title' data-module="track-error" data-error-type="No rows" dat
...
         <p>
           Delete or rename one of these columns and try again.
         </p>
 
       {% elif recipients.missing_column_headers %}
 
         <h1 class='banner-title' data-module="track-error" data-error-type="Missing placeholder columns" data-error-label="{{ upload_id }}">
-          The columns in your file need to match the double brackets in
-          your template
+          Your file is missing some columns
```
quis: This is technically accurate, but less clear I think. Often the problem is that the columns are there, but mislabelled. So saying the columns are ‘missing’ might confuse people.

app/templates/views/check/column-errors.html
```diff
@@ -141,7 +149,7 @@ <h1 class='banner-title' data-module="track-error" data-error-type="Missing plac
         {{ file_upload(
           form.file,
           action=url_for('.send_messages', service_id=current_service.id, template_id=template.id),
-          button_text='Re-upload your file'
+          button_text='Upload a new file'
```
quis: The behaviour here is often to fix the problem with the file (eg rename a column) and then upload it again. I don’t think ‘new file’ describes this well.

karlchillmaid: Re-upload isn't a word, so we shouldn't use it.
My previous draft was 'Upload your file again'.
It shifted to this version after discussions with other content designers. There was a concern that encouraging users to upload again might not make it clear enough that they need to fix the file and save it again first.
Hopefully the wording of the error summary mitigates that risk, though.
Happy to go with 'Upload your file again'.

app/templates/views/check/column-errors.html
```diff
@@ -56,49 +56,57 @@ <h1 class='banner-title' data-module="track-error" data-error-type="No rows" dat
...
 
         <h1 class='banner-title' data-module="track-error" data-error-type="Missing recipient columns" data-error-label="{{ upload_id }}">
-          Your file needs {{ (
-            recipients.missing_column_headers
-            if template.template_type == 'letter' else required_recipient_columns
-          ) | formatted_list(
-            prefix='a column called',
-            prefix_plural='columns called'
-          ) }}
+          Your file is missing some columns
```
quis: I think this has the same problem as the ‘double brackets‘ error message.

If the columns in someone’s file are:
```
email | name | date
```
and you say ‘Your file is missing some columns‘ their reaction will often be ‘no it’s not, all three columns are there‘. And this causes them to lose faith in the system’s ability to help them fix the problem.

An error message that says
> Your file needs a column called ‘email address’

Helps people understand that it’s about the naming of the column, not just that there is a column with some email addresses in it.

karlchillmaid: What if they're missing multiple columns? How long can that first line of the error get, realistically?

quis: Longest example would be:
> Your file needs columns called ‘Address line 1’, ‘Address line 2’ and ’Postcode’

karlchillmaid: Screen Shot 2019-09-26 at 14 36 55

*vs*

Screen Shot 2019-09-26 at 14 38 25

app/templates/views/check/column-errors.html
karlchillmaid: Screen Shot 2019-09-26 at 14 53 27

app/templates/views/check/column-errors.html
karlchillmaid: Screen Shot 2019-09-26 at 14 54 34

app/templates/views/check/column-errors.html
karlchillmaid: Screen Shot 2019-09-26 at 15 17 49

Makes it a bit easier to compare when the column names are only _slightly_ wrong?

PR #3130, opened 2019-09-24
Get notification count from api instead of getting list of notifications from api

CrystalPea:
Get notification count from api instead of getting list of notifications from api when checking if job can be cancelled.

This makes query lighter, less likely to fall over if job is large and removes pagination bug (where only first page of notifications is returned form API and when job is large, this makes it impossible to cancel as the count from job.notification_count does not match).

API part: https://github.com/alphagov/notifications-api/pull/2617
Bug: https://www.pivotaltracker.com/story/show/168734905

### DIFF
--- a/app/notify_client/notification_api_client.py
+++ b/app/notify_client/notification_api_client.py
@@ -127,5 +127,8 @@ class NotificationApiClient(NotifyAdminAPIClient):
             }
         )
 
+    def get_notification_count_for_job_id(self, *, service_id, job_id):
+        return self.get(url='/service/{}/job/{}/notification_count'.format(service_id, job_id))["count"]
+
 
 notification_api_client = NotificationApiClient()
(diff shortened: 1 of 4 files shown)


app/notify_client/notification_api_client.py
```diff
@@ -127,5 +127,8 @@ def get_notification_status_by_service(self, start_date, end_date):
             }
         )
 
+    def get_notification_count_for_job_id(self, *, service_id, job_id):
```
quis: Don’t you need the `*` here? I can’t see where you’re using it.
```suggestion
    def get_notification_count_for_job_id(self, service_id, job_id):
```

CrystalPea: The star makes it necessary to name the arguments following it when calling function. It's a style choice, I like named arguments, they are readable and leave less place for mistakes.

quis: Cool, I learned something! Makes a lot sense when you’re passing in two UUIDs.

CrystalPea: Yay, I learnt on Pycon UK about the star :)

PR #3137, opened 2019-10-04
Update the errors seen when uploading a letter

klssmith:
The error content has been updated and we now use the pattern of showing a box at the top of the page with the error which has a heading and can have additional details.

Also updated the version of the letter specification everywhere from 2.3 to 2.4.

[Pivotal story](https://www.pivotaltracker.com/story/show/168869338)

## Errors before
Screen Shot 2019-10-04 at 09 37 59

Screen Shot 2019-10-04 at 09 37 31

## Errors after
Screen Shot 2019-10-04 at 09 35 12

Screen Shot 2019-10-04 at 09 35 27

Screen Shot 2019-10-04 at 09 35 41

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1111,8 +1111,8 @@ class PDFUploadForm(StripWhitespaceForm):
     file = FileField_wtf(
         'Upload a letter in PDF format',
         validators=[
-            FileAllowed(['pdf'], 'Letters must be saved as a PDF'),
-            DataRequired(message="You need to upload a file to submit")
+            FileAllowed(['pdf'], 'Save your letter as a PDF and try again.'),
+            DataRequired(message="You need to choose a file to upload")
         ]
     )
 
--- a/app/templates/views/uploads/preview.html
+++ b/app/templates/views/uploads/preview.html
@@ -1,4 +1,5 @@
 {% extends "withnav_template.html" %}
+{% from "components/banner.html" import banner_wrapper %}
 {% from "components/page-header.html" import page_header %}
 
 {% block service_page_title %}
@@ -6,16 +7,19 @@
 {% endblock %}
 
 {% block maincolumn_content %}
-    {{ page_header(
+  {% if status == 'invalid' and error %}
+    {% call banner_wrapper(type='dangerous') %}
+      <h1 class="banner-title">{{ error.title }}</h1>
+      {% if error.detail %}
+        <p>{{ error.detail | safe }}</p>
+      {% endif %}
+    {% endcall %}
+  {% else %}
+      {{ page_header(
         original_filename,
         back_link=url_for('main.upload_letter', service_id=current_service.id)
-    ) }}
-
-    {% if status == 'invalid' %}
-    <p class="notification-status-cancelled" id="validation-error-message">
-      Validation failed
-    </p>
-    {% endif %}
+      ) }}
+  {% endif %}
 
     <div class="letter-sent">
       {{ template|string }}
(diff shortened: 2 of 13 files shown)


app/utils.py
```diff
@@ -543,6 +546,54 @@ def get_letter_printing_statement(status, created_at):
...
+        'title': 'We cannot print your letter',
+        'detail': 'The content appears outside the printable area on {invalid_pages} <br>'
+                  'Files must meet our <a href="https://docs.notifications.service.gov.uk/documentation/images/'
+                  'notify-pdf-letter-spec-v2.4.pdf" target="_blank">letter specification</a>.'
+    },
+    'letter-too-long': {
+        'title': 'Your letter is too long',
+        'detail': 'Letters must be 10 pages or less. <br>Your letter is {page_count} pages long.'
+    },
+    'no-encoded-string': {
```
klssmith: The `no-encoded-string` and `unable-to-read-the-file` errors aren't used for uploaded letters (because it's not possible to upload an empty file and we don't use the template preview sanitise endpoint to check if the file can be read). Have included them here so that the dictionary can be used with any template preview error later.

app/templates/views/uploads/preview.html
```diff
@@ -1,21 +1,25 @@
...
 {% from "components/page-header.html" import page_header %}
 
 {% block service_page_title %}
   {{ original_filename }}
 {% endblock %}
 
 {% block maincolumn_content %}
-    {{ page_header(
+  {% if status == 'invalid' and error %}
+    {% call banner_wrapper(type='dangerous') %}
```
klssmith: It may be worth making partials for each error type once we show more detailed errors for precompiled letters sent from the API (but have left that for now since we're only using the errors for uploaded letters)

app/main/forms.py
```diff
@@ -1111,8 +1111,8 @@ class PDFUploadForm(StripWhitespaceForm):
     file = FileField_wtf(
         'Upload a letter in PDF format',
         validators=[
-            FileAllowed(['pdf'], 'Letters must be saved as a PDF'),
-            DataRequired(message="You need to upload a file to submit")
+            FileAllowed(['pdf'], 'Save your letter as a PDF and try again.'),
+            DataRequired(message="You need to choose a file to upload")
```
karlchillmaid: @klssmith is there a reason 1114 has a full stop and 1115 does not?

klssmith: @karlchillmaid I assumed that `You need to choose a file to upload` deliberately didn't have a full stop in the Google doc since it's an error message title (not additional detail), but I can add one if that's not the case.

karlchillmaid: No, that sounds right. Just checking because I hadn't realised 1114 was additional detail.

tests/app/test_utils.py
```diff
@@ -404,3 +405,25 @@ def test_get_letter_printing_statement_for_letter_that_has_been_sent(created_at,
...
+    assert get_letter_validation_error('Unknown error') == {
+        'title': 'Validation failed'
+    }
+
+
+@pytest.mark.parametrize('error_message, expected_title, expected_content', [
+    ('letter-not-a4-portrait-oriented', 'We cannot print your letter', 'A4 portrait size on page 2'),
+    ('content-outside-printable-area', 'We cannot print your letter', 'outside the printable area on page 2'),
+    ('letter-too-long', 'Your letter is too long', 'letter is 13 pages long.')
+])
```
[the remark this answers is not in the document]
klssmith: Might add them in - I didn't include them in this test originally because I was really trying to test the error messages that have string formatting in (to check that the message was being constructed properly).

klssmith:
@CrystalPea I've rebased and pushed a new commit to only show the overlay when content is over the margins

PR #3138, opened 2019-10-04
Rough little script to get Zendesk tickets for support analysis.

servingUpAces:
If this is needed a lot over time we might want to add something to the ZendeskClient.


get_zendesk_tickets.py
```diff
@@ -0,0 +1,140 @@
...
+                                     'Subject line': row['subject'],
+                                     'Date ticket created': row["created_at"],
+                                     'Tags': row.get('tags', '')
+                                     })
+            next_page = data["next_page"]
+
+
+def get_tickets_with_description():
+    ZENDESK_TICKET_URL = 'https://govuk.zendesk.com/api/v2/search.json?query={}'
+    query_params = 'type:ticket group:{}, created>2019-07-01'.format(NOTIFY_GROUP_ID)
```
idavidmcdonald: Does the created at flag here indicate some sortof changeover time when we now have descriptions?

Not sure, if it's related or just something left in?

PR #3140, opened 2019-10-07
Store all data about uploaded letters in the S3 metadata

klssmith:
We had been storing whether or not a file was valid in the S3 metadata, but using the query string of the URL to store the original filename and the page count. This meant that if you tried to view the preview letter page without the query string you would see a `500`. It was possible for this to happen if you were signed out of Notify while on the preview page - you would be redirected back to the preview page but without the query string, causing an error.

### DIFF
--- a/app/s3_client/s3_letter_upload_client.py
+++ b/app/s3_client/s3_letter_upload_client.py
@@ -7,17 +7,22 @@ def get_transient_letter_file_location(service_id, upload_id):
     return 'service-{}/{}.pdf'.format(service_id, upload_id)
 
 
-def upload_letter_to_s3(data, file_location, status):
+def upload_letter_to_s3(data, *, file_location, status, page_count, filename):
     utils_s3upload(
         filedata=data,
         region=current_app.config['AWS_REGION'],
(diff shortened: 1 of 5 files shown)


app/s3_client/s3_letter_upload_client.py
```diff
@@ -7,21 +7,34 @@ def get_transient_letter_file_location(service_id, upload_id):
     return 'service-{}/{}.pdf'.format(service_id, upload_id)
 
 
-def upload_letter_to_s3(data, file_location, status):
+def upload_letter_to_s3(data, file_location, status, page_count, filename):
```
quis: This method might now benefit from [requiring named arguments](https://stackoverflow.com/questions/2965271/forced-naming-of-parameters-in-python/14298976#14298976) to prevent someone calling it with the arguments in the wrong order (cool little trick that @CrystalPea taught me).
```suggestion
def upload_letter_to_s3(data, *, file_location, status, page_count, filename):
```

klssmith: 👍 Good idea, updated to require named args now

PR #3143, opened 2019-10-09
Merge ‘uploaded files’ and ‘uploads’ pages

quis:
‘Uploaded files‘ is the `…/jobs` page.

‘Uploads’ is the new page where you can upload a letter.

Eventually this page will show:
- jobs as in spreadsheets
- contact lists
- uploaded letters (individual and batches)

It will be the route for users to upload:
- contact lists
- uploaded letters

This pull request makes the two pages one, and adds the button to give users (whose services have permission) to enter the ‘upload letters’ journey.


app/templates/main_nav.html
```diff
@@ -8,15 +8,16 @@
   {% endif %}
   {% if current_user.has_permissions() %}
     <li><a href="{{ url_for('.choose_template', service_id=current_service.id) }}" {{ main_navigation.is_selected('templates') }}>Templates</a></li>
-    {% if not current_user.has_permissions('view_activity') %}
+    {% if current_user.has_permissions('view_activity') %}
+      {% if current_service.can_upload_letters %}
```
CrystalPea: So now only users with "View activity" will be able to upload letters. Do we know if there are users without that permission, but with sending permission?

app/templates/views/dashboard/_jobs.html
```diff
@@ -6,14 +6,16 @@
     jobs,
     caption="Recent files uploaded",
     caption_visible=False,
-    empty_message='You have not uploaded any files recently',
+    empty_message=(
+      'Upload a letter and Notify will print, pack and post it for you.' if current_service.can_upload_letters else 'You have not uploaded any files recently'
```
CrystalPea: So if the service has uploaded csv jobs, but has not uploaded letters, they will not see:
```
'Upload a letter and Notify will print, pack and post it for you.'
```
Is it not a bit confusing?

app/templates/main_nav.html
quis: Users who don’t have the `view_activity` permission will still be able to see the uploads page because we link to it again here: https://github.com/alphagov/notifications-admin/pull/3143/files#diff-c2f9cd0234e25159b7da926354229d2dR17

CrystalPea: ok, thanks 👍

quis:
@CrystalPea Yes, this is an in-between state. It’s the first step towards having one page that shows all the things you’ve uploaded. Uploading CSVs will always happen from the templates page, because you have to start with a template.

PR #3144, opened 2019-10-09
Choose postage

idavidmcdonald:
https://www.pivotaltracker.com/story/show/167534014

See commit messages for details.

I haven't written any python in a long time so please do have a good review, in particular I might have missed something with tests, validation, error handling or WTForms.

Screenshot 2019-10-09 at 17 33 34

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -765,6 +765,19 @@ class LetterTemplatePostageForm(StripWhitespaceForm):
     )
 
 
+class LetterUploadPostageForm(StripWhitespaceForm):
+    postage = RadioField(
+        'Choose the postage for this letter',
+        choices=[
+            ('first', 'First class post'),
+            ('second', 'Second class post'),
+        ],
+        default='second',
+        validators=[DataRequired()]
+    )
+    file_id = HiddenField()
+
+
 class ForgotPasswordForm(StripWhitespaceForm):
     email_address = email_address(gov_user=False)
 
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -775,7 +775,9 @@ class LetterUploadPostageForm(StripWhitespaceForm):
         default='second',
         validators=[DataRequired()]
     )
-    file_id = HiddenField()
+    file_id = HiddenField(
+        validators=[DataRequired()]
+    )
 
 
 class ForgotPasswordForm(StripWhitespaceForm):
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -153,6 +153,10 @@ def uploaded_letter_preview(service_id, file_id):
 
     error = get_letter_validation_error(error_message, invalid_pages, page_count)
     template_dict = service_api_client.get_precompiled_template(service_id)
+    # Override pre compiled letter template postage to none as it has not yet been picked even though
+    # the pre compiled letter template has its postage set as second class as the DB currently requires
+    # a non null value of postage for letter templates
+    template_dict['postage'] = None
 
     form = LetterUploadPostageForm()
 
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -18,7 +18,7 @@ from requests import RequestException
 from app import current_service, notification_api_client, service_api_client
 from app.extensions import antivirus_client
 from app.main import main
-from app.main.forms import PDFUploadForm
+from app.main.forms import LetterUploadPostageForm, PDFUploadForm
 from app.main.views.jobs import view_jobs
 from app.s3_client.s3_letter_upload_client import (
     get_letter_metadata,
@@ -154,6 +154,8 @@ def uploaded_letter_preview(service_id, file_id):
     error = get_letter_validation_error(error_message, invalid_pages, page_count)
     template_dict = service_api_client.get_precompiled_template(service_id)
 
+    form = LetterUploadPostageForm()
+
     template = get_template(
         template_dict,
         service_id,
@@ -172,6 +174,7 @@ def uploaded_letter_preview(service_id, file_id):
         status=status,
         file_id=file_id,
         error=error,
+        form=form,
     )
 
 
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -202,13 +202,14 @@ def send_uploaded_letter(service_id):
         abort(403)
 
     file_id = request.form['file_id']
+    postage = request.form['postage']
     metadata = get_letter_metadata(service_id, file_id)
     filename = metadata.get('filename')
 
     if metadata.get('status') != 'valid':
         abort(403)
 
-    notification_api_client.send_precompiled_letter(service_id, filename, file_id)
+    notification_api_client.send_precompiled_letter(service_id, filename, file_id, postage)
 
     return redirect(url_for(
         '.view_notification',
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -202,8 +202,11 @@ def send_uploaded_letter(service_id):
         abort(403)
 
     form = LetterUploadPostageForm()
-
     file_id = form.file_id.data
+
+    if not form.validate_on_submit():
+        return uploaded_letter_preview(service_id, file_id)
+
     postage = form.postage.data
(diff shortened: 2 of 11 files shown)


app/main/views/uploads.py
```diff
@@ -153,13 +160,14 @@ def send_uploaded_letter(service_id):
         abort(403)
 
     file_id = request.form['file_id']
+    postage = request.form['postage']
```
quis: This should probably use a WTForms `form` instance, not a Flask `request`. Not sure why the previous code is using `request`, this isn’t how we normally do form handling.

app/templates/views/uploads/preview.html
```diff
@@ -27,8 +28,9 @@
           'main.send_uploaded_letter',
           service_id=current_service.id,
         )}}" class='page-footer'>
-        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <input type="hidden" name="file_id" value="{{ file_id }}" />
+          {{ form.csrf_token }}
+          {{ radios(form.postage, hide_legend=true, inline=True) }}
+          {{ form.file_id(value=file_id) }}
         <button type="submit" class="button">Send 1 letter</button>
```
quis: If you use the `page_footer` macro here you won’t need to manually print the CSRF token above.
```suggestion
            {{ page_footer('Send 1 letter') }}
```

leohemsted:
There is technically a service using this page (though they haven't looked at the page in a week) - I'd say it's better flow to do three PRs:

1. admin: Always send through `'postage': 'second'`
2. api: Use the postage supplied by the admin
3. admin: this PR.

tests/app/main/views/test_uploads.py
```diff
@@ -88,9 +88,7 @@ def test_post_upload_letter_shows_letter_preview_for_valid_file(mocker, client_r
             _follow_redirects=True,
         )
 
-    assert len(page.select('.letter-postage')) == 1
-    assert normalize_spaces(page.select_one('.letter-postage').text) == ('Postage: second class')
-    assert page.select_one('.letter-postage')['class'] == ['letter-postage', 'letter-postage-second']
+    assert len(page.select('.letter-postage')) == 0
```
CrystalPea: Also here I would test that the new radio elements are present.

app/main/forms.py
```diff
@@ -765,6 +765,21 @@ class LetterTemplatePostageForm(StripWhitespaceForm):
     )
 
 
+class LetterUploadPostageForm(StripWhitespaceForm):
```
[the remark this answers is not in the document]
idavidmcdonald: Good question. I've taken a look. Whilst they are similar I think they have too many differences e.g. choice text, that mean it's not worth trying to share code

app/templates/views/uploads/preview.html
idavidmcdonald: Oh nice. Have changed.

app/main/views/uploads.py
idavidmcdonald: So I've updated it to use the `form` instance which all works fine.

This however now has me thinking that we should be calling `form.validate_on_submit`, like we generally do for forms. It's very unlikely there will be a user error (potentially only if they edited the HTML/request themselves but still) as they are only shown one radio button and it is preselected. If an error was to be raised then we would want to pass it back to the original page they came from i.e. the previous GET request on `/services/<service_id>/preview-letter/<file_id>`. But this route is on a different method. It leaves me with a few options

1). Don't call `validate_on_submit` and assume everything will be fine
2) As part of this method, if `validate_on_submit` fails, render the original preview upload template by duplicating a bunch of code from `uploaded_letter_preview`.
3) Move this POST request to be handled by the `uploaded_letter_preview` method too. 

Naturally 3 feels the most 'correct' way of doing things but changes our URL structure and how the current methods are laid out. Just looking for a second opinion that this is a good decision and worth my time?

quis: You can achieve 2) without any code duplication by doing:
```python
if not form.validate_on_submit():
    return uploaded_letter_preview(service_id, file_id)
```

This should even show the right error message if, for example, you manually uncheck the pre-checked radio.

idavidmcdonald: Ah good point. I've looked into this and found the following:

- `file_id` is not available in the URL for the POST request and is taken from a hidden field in the form so if you remove the file_id from the HTML, things still break as you don't have a valid file_id to pass to the `uploaded_letter_preview` function
- I tested unchecking the radio button and also deleting the radio button element. In both cases, the form seems to validate successfully. I think this is because there is a default value for the radio buttons in the WTForm and it just takes that so the letter gets sent second class in this case.

These two bits of behaviour are exactly the same as what would happen if we did not call `validate_on_submit`. If you screw with the radios, it chooses second class. If you screw with the file_id, it breaks.

Regardless, I feel that your suggestion is better than not calling it `validate_on_submit` and also removes the code repetition. I'm still questioning whether moving this POST request in to the original GET method is still the thing we should be aiming for though. 🤔

app/notify_client/notification_api_client.py
```diff
@@ -59,10 +59,11 @@ def send_notification(self, service_id, *, template_id, recipient, personalisati
         data = _attach_current_user(data)
         return self.post(url='/service/{}/send-notification'.format(service_id), data=data)
 
-    def send_precompiled_letter(self, service_id, filename, file_id):
+    def send_precompiled_letter(self, service_id, filename, file_id, postage):
```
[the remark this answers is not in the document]
idavidmcdonald: Good question. I went and did some reading to remind myself all about key word arguments.

https://treyhunner.com/2018/04/keyword-arguments-in-python/

They had one example in particular that helped me make an opinion (although not strongly held).

image

Here, they keyword arguments are particulary useful as when calling the function, you may put arguments in with little context, such as numbers.

In the usage (so far) of `send_precompiled_letter`, all the arguments passed in are passed in as variables with nice names so if we were to use key word arguments it would be called like

```notification_api_client.send_precompiled_letter(service_id, filename=filename, file_id=file_id, postage=postage)```

which I don't think is giving us much more in terms of readability. The only advantage would be remembering the order, which I think is a fair point but given we would probably want to look up the method anyway we would find out the order that way too.

So yeah, I think I'm OK with leaving them positional unless you feel strongly?

tests/app/main/views/test_uploads.py
idavidmcdonald: Good idea. I've added this.

app/notify_client/notification_api_client.py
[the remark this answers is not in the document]
CrystalPea: I don't feel strongly, I usually make the parameters of a function mandatory keyword when there are more than three for order purposes, it makes me feel safer, but yeah, I don't feel strongly about it :)

app/main/views/uploads.py
idavidmcdonald: I checked Leo's opinion on this too and have decided to follow your suggestion for option 2) by the way

karlchillmaid:
> Screen Shot 2019-10-14 at 16 58 46

`Provided as *a* PDF` would read better.

CrystalPea:
> Also, as Leo said, I would do changes to api before merging the changes you made here, so once this is deployed, everything works nicely. I think right now 1 live service has access to the uploads, but they are not using that access. Still, probably good not to have confusing content (right now all uploaded letters are sent with 2nd class).

Is the API part merged?

karlchillmaid:
The content discussed with @quis a while back was:

- `Uploaded` for PDF letters uploaded through the app
- `Provided` for PDF letters sent via API
- `Sent` for templated letters

idavidmcdonald:
Chatting with Karl, content should be as follows given that the letter has been uploaded through the app:

`Uploaded by Pea...`

idavidmcdonald:
> > Also, as Leo said, I would do changes to api before merging the changes you made here, so once this is deployed, everything works nicely. I think right now 1 live service has access to the uploads, but they are not using that access. Still, probably good not to have confusing content (right now all uploaded letters are sent with 2nd class).
> 
> Is the API part merged?

Yep @CrystalPea

PR #3148, opened 2019-10-15
Add a 'Get started' page

karlchillmaid:
This PR adds a _Get started_ page, similar to those on Pay and Verify. The page is an opportunity to:
- show users how easy it is to start using Notify
- reduce the number of support requests we get

The content should:
- outline the process users will go through
- answer common questions that can't be answered elsewhere on Notify
- signpost existing content, including the Service Manual content guidance

**Why are we adding another page?**

We built Notify so that users can just create an account and start playing, but the reality is they often want to understand the process up front as part of their research.

We get quite a few support requests about how to get started using Notify.

Common questions include:
- how long it takes before you can send messages for real
- what needs to be done before you can send messages for real
- whether their public sector organisation can use Notify
- whether a third party organisation can use Notify

This information doesn't fit neatly into the Features pages, and it's too detailed for the main product page.

This new page is **not** FAQ. It needs a clear purpose, it needs to be brief, and it should form part of users' onboarding journey. It should not become a dumping ground for information.

***

Todo:
- [x] get answer on correct semantic HTML for this list
- [x] write a test that makes sure the page loads when the user is signed in, signed out, and signed in but without a current service
- [x] style the list to look like https://www.payments.service.gov.uk/getstarted/


app/templates/views/get-started
```diff
@@ -0,0 +1,80 @@
...
+    {% if not current_user.is_authenticated %}
+      <p><a href="{{ url_for('.register') }}">Create an account</a> for free and set up your first Notify service. When you set up a new service, it will start in <a href="{{ url_for('main.trial-mode') }}">trial mode</a>.</p>
+    {% else %}
+      <p>Create an account for free and set up your first Notify service. When you set up a new service, it will start in <a href="{{ url_for('main.trial_mode_new') }}">trial mode</a>.</p>
+    {% endif %}
+  </li>
+  
+  <li>
+    <h2 class="heading-small">Write some messages</h2>
+    <p>Add <a href="{{ url_for('main.features', _anchor='templates') }}">reusable message templates</a> with examples of the content you plan to send.</p>
```
karlchillmaid: When signed in, the 'templates' link could take users to their Templates page instead of features. This would be consistent with what we do for Settings, Team members etc further down the page.
BUT, it feels confusing to have the same link text behave differently depending on whether you're signed in/out?

karlchillmaid:
Still need to add a link to this page from the footer and decide how it fits into the left-hand navigation.

quis:
localhost_6012_features_get-started(iPad Pro)

karlchillmaid:
Will make some content changes based on that screenshot.

quis:
localhost_6012_features_get-started(iPad Pro) (2)

app/templates/views/get-started
```diff
@@ -0,0 +1,83 @@
...
+    <h2 class="heading-small">API integration (optional)</h2>
+    <p>Share our <a href="{{ url_for('main.documentation') }}">documentation</a> with your developers to help them set up an API integration.</p>
+
+    <p>You only need to do this if you’re using the API to send messages automatically, rather than signing in to Notify.</p>
+  </li>
+  
+  <li>
+    <h2 class="heading-small">Start sending messages</h2>
+    {% if not current_user.is_authenticated %}
+    <p>You should request to go live when you’re ready to send messages to people outside your team. We’ll process your request within one working day.</p>
```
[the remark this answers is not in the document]
karlchillmaid: Change it to `approve`

app/templates/views/get-started.html
```diff
@@ -0,0 +1,85 @@
...
+    {% if not current_user.is_authenticated %}
+      <p><a href="{{ url_for('.register') }}">Create an account</a> for free and add your first Notify service. When you add a new service, it will start in <a href="{{ url_for('main.trial_mode_new') }}">trial mode</a>.</p>
+    {% else %}
+      <p>Create an account for free and add your first Notify service. When you add a new service, it will start in <a href="{{ url_for('main.trial_mode_new') }}">trial mode</a>.</p>
+    {% endif %}
+  </li>
+
+  <li>
+    <h2 class="heading-small">Write some messages</h2>
+    <p>Add <a href="{{ url_for('main.features', _anchor='templates') }}">message templates</a> with examples of the content you plan to send. You do not need to add letter templates if you’re planning to <a href="{{ url_for('main.features_letters', _anchor='upload-letters') }}">upload your own letters</a>.</p>
```
karlchillmaid: We could lose `You do not need to add letter templates if you’re planning to...`

app/templates/views/get-started.html
```diff
@@ -0,0 +1,85 @@
...
+  </li>
+
+  <li>
+    <h2 class="heading-small">Start sending messages</h2>
+    {% if not current_user.is_authenticated %}
+    <p>You should request to go live when you’re ready to send messages to people outside your team. We’ll approve your request within one working day.</p>
+    {% else %}
+    <p>You should <a href="{{ url_for('.request_to_go_live', service_id=current_service.id) }}">request to go live</a> when you’re ready to send messages to people outside your team. We’ll approve your request within one working day.</p>
+    {% endif %}
+    <p>Once your service is live, we’ll send you an invoice if you:</p>
```
karlchillmaid: We could lose lines 78-82 and just keep the `See how to pay` line.

app/templates/views/get-started.html
```diff
@@ -0,0 +1,85 @@
...
+    <p>Add team members and check their <a href="{{ url_for('main.features', _anchor='permissions') }}">permissions</a>.</p>
+    {% else %}
+    <p>Review your <a href="{{ url_for('.service_settings', service_id=current_service.id) }}">service settings</a> to make sure you have the right branding and sender information.</p>
+    <p>Add <a href="{{ url_for('.manage_users', service_id=current_service.id) }}">team members</a> and check their <a href="{{ url_for('main.features', _anchor='permissions') }}">permissions</a>.</p>
+    {% endif %}
+  </li>
+
+  <li>
+    <h2 class="heading-small">API integration (optional)</h2>
+    <p>Share our <a href="{{ url_for('main.documentation') }}">documentation</a> with your developers to help them set up an API integration.</p>
```
karlchillmaid: Could replace lines 66-68 with `Your developers can use our documentation to help them integrate with the Notify API.`

app/templates/views/get-started.html
```diff
@@ -0,0 +1,80 @@
...
+    <ul class="list list-bullet">
+      <li><a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#how-to-write-emails-and-text-messages">how to write emails and text messages</a></li>
+      <li><a href="https://www.gov.uk/service-manual/design/writing-effective-letters">writing effective letters</a></li>
+    </ul>
+  </li>
+
+  <li>
+    <h2 class="heading-small">Set up your service</h2>
+    {% if not current_user.is_authenticated %}
+    <p>Review your settings to make sure your messages have the right branding.</p>
```
karlchillmaid: Review your settings to add message branding, reply-to addresses and sender names.

app/templates/views/get-started.html
```diff
@@ -0,0 +1,80 @@
...
+    <p>The GOV.UK Service Manual has advice on:</p>
+    <ul class="list list-bullet">
+      <li><a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#how-to-write-emails-and-text-messages">how to write emails and text messages</a></li>
+      <li><a href="https://www.gov.uk/service-manual/design/writing-effective-letters">writing effective letters</a></li>
+    </ul>
+  </li>
+
+  <li>
+    <h2 class="heading-small">Set up your service</h2>
+    {% if not current_user.is_authenticated %}
```
quis: If they go straight to this page after logging in they might not have a most-recently-view service, so you need to guard against that as well.
```suggestion
    {% if not current_user.is_authenticated or not current_service %}
```

karlchillmaid: 👍

quis:
localhost_6012_features_get-started(iPad Pro) (3)

PR #3149, opened 2019-10-17
Let textareas resize automatically without having to highlight placeholders

quis:
Scrolling within `<textarea>`s on the page is a bit grim. Which is why we don’t do it for the textboxes that people use to edit templates.

This pull request extends the auto-resizing of `<textarea>`s to those which don’t need the highlighting of placeholders.

***

awanp

***

image

### DIFF
--- a/app/assets/javascripts/enhancedTextbox.js
+++ b/app/assets/javascripts/enhancedTextbox.js
@@ -7,7 +7,7 @@
 
   const tagPattern = /\(\(([^\)\((\?)]+)(\?\?)?([^\)\(]*)\)\)/g;
 
-  Modules.HighlightTags = function() {
+  Modules.EnhancedTextbox = function() {
 
     this.start = function(textarea) {
 
--- a/app/assets/javascripts/highlightTags.js
+++ b/app/assets/javascripts/highlightTags.js
@@ -11,6 +11,13 @@
 
     this.start = function(textarea) {
 
+      let visibleTextbox;
+
+      this.highlightPlaceholders = (
+        typeof textarea.data('highlightPlaceholders') === 'undefined' ||
+        !!textarea.data('highlightPlaceholders')
+      );
+
       this.$textbox = $(textarea)
         .wrap(`
           <div class='textbox-highlight-wrapper' />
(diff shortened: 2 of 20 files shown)


app/assets/javascripts/enhancedTextbox.js
```diff
@@ -0,0 +1,85 @@
...
+        )
+      );
+
+      if ('stickAtBottomWhenScrolling' in GOVUK) {
+        GOVUK.stickAtBottomWhenScrolling.recalculate();
+      }
+
+    };
+
+    this.contentEscaped = () => $('<div/>').text(this.$textbox.val()).html();
```
tombye: Is the content escaped here? I can't find reference to it in the docs for `.val()`.

quis: There some detail in this commit: https://github.com/alphagov/notifications-admin/commit/836fe77c372eaa10bf6273b0efef54f6bcd709bb

TL;DR is that it’s the `$('<div/>').text(…)` part that does the escaping.

app/assets/javascripts/enhancedTextbox.js
```diff
@@ -0,0 +1,85 @@
...
+  const tagPattern = /\(\(([^\)\((\?)]+)(\?\?)?([^\)\(]*)\)\)/g;
+
+  Modules.EnhancedTextbox = function() {
+
+    this.start = function(textarea) {
+
+      let visibleTextbox;
+
+      this.highlightPlaceholders = (
+        typeof textarea.data('highlightPlaceholders') === 'undefined' ||
```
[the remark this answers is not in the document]
quis: `this.highlightPlaceholders` should be `true` if:
- `data-highlightPlaceholders` is explicitly set to `"true"`
- `data-highlightPlaceholders` is not set (i.e. it’s true by default)

app/assets/javascripts/enhancedTextbox.js
tombye: Gotcha, sorry, should have checked those docs.

app/assets/javascripts/enhancedTextbox.js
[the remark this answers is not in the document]
tombye: Right, I didn't know the second point but see there's a test for it so 👍

quis:
@tombye I fixed the calls to the textbox macro, and added a test to make sure the right data attributes are present on the edit templates page: https://github.com/alphagov/notifications-admin/compare/resize-all-textbox-dem#diff-f2eb304b93cc383727c0ab7fc8fbd464R349-R350

PR #3158, opened 2019-10-23
Replace govuk template with govuk frontend components

tombye:
Makes all pages use [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), the frontend library for the [GOV.UK Design System](https://design-system.service.gov.uk) as their base layout. This replaces the [GOV.UK Template](https://github.com/alphagov/govuk_template).

https://www.pivotaltracker.com/story/show/161594893

#### Current

image

#### New

image

#### Current

image

#### New

image

## Code changes

This makes [template.njk](https://github.com/alphagov/govuk-frontend/blob/v2.13.0/package/template.njk) from GOV.UK Frontend the base layout for all pages.

### Changes specific to this app

The header and footer have been changed to match the Notify design (see https://github.com/alphagov/notifications-admin/commit/9da1937ee6b00e097b62038840836e31f45e5ad8).

### Use of GOV.UK template code

The following files have been moved across from the GOV.UK template repository:
- the JS to set/read cookies
- the print CSS, adapted to the GOV.UK Frontend HTML, as GOV.UK Frontend does not provide any

#### Sass

This only includes the Sass for the GOV.UK Frontend components used.

It also includes [some global CSS](https://github.com/alphagov/notifications-admin/pull/3158/commits/bbf87e169736daa2b816a0a41aeb2175eadad917) (rightly banned under GOV.UK Frontend) to style our links and form controls. This will be removed when we add the `govuk-link` class to all our links and when each form control is replaced with a GOV.UK Frontend component.

#### JS

It includes the JS for all GOV.UK Frontend components. Including only those used is possible but requires changing the frontend build so it can process JS modules, which is enough work to need its own pull request.

## Performance

This increases the page weight from 504KB to 562KB over all requests made (/sign-in page tested).

The increase should only be temporary because:
- the JS will be reduced when a module-bundler is plugged into the frontend build so we can only include the code we need
- [GOV.UK Elements](https://govuk-elements.herokuapp.com/) code will be removed with each component we replace from there with one from GOV.UK Frontend

## Testing

I did a rough audit of the pages in this app and found around 200 individual pages and 24 types of page:

https://docs.google.com/spreadsheets/d/1Vdgb1M3t4988DpswhsCZJi75TCG3iUZ42EUrtFtgTos/edit#gid=0

Based on this, I tested one of each page type with the following browser/OS combinations:

### Windows 7
- Chrome 77
- Firefox 69
- IE9
- IE8

### Windows 10
- Edge 17
- Chrome 77
- Firefox 60
- IE11

### IOS 12.4.1
- Safari

### Android 9 (pie)
- Samsung Internet

Note: I tested against the current version of each page, not for any issues with the current version.


app/navigation.py
```diff
@@ -327,6 +327,12 @@ class HeaderNavigation(Navigation):
         'whitelist',
     }
 
+    # header HTML now comes from GOVUK Frontend so requires a boolean, not an attribute
+    def is_selected(self, navigation_item):
+        if request.endpoint in self.mapping[navigation_item]:
```
quis: 🐍 
```suggestion
        return request.endpoint in self.mapping[navigation_item]
```

app/templates/admin_template.html
```diff
@@ -1,114 +1,225 @@
...
-  <nav class="footer-nav">
-    Built by the <a href="https://www.gov.uk/government/organisations/government-digital-service">Government Digital Service</a>
-    <a href="{{ url_for("main.privacy") }}">Privacy</a>
-    <a href="{{ url_for("main.cookies") }}">Cookies</a>
-    {% if current_service and current_service.research_mode %}
-    <span id="research-mode" class="research-mode">research mode</span>
-    {% endif %}
-  </nav>
+  {% if current_service and current_service.research_mode %}
+    {% set meta_suffix = 'Built by the <a href="#" class="govuk-footer__link">Government Digital Service</a><span id="research-mode" class="research-mode">research mode</span>' %}
```
quis:
```suggestion
    {% set meta_suffix = 'Built by the <a href="https://www.gov.uk/government/organisations/government-digital-service" class="govuk-footer__link">Government Digital Service</a><span id="research-mode" class="research-mode">research mode</span>' %}
```

app/templates/admin_template.html
```diff
@@ -1,114 +1,225 @@
...
+  {% else %}
+    {% set meta_suffix = 'Built by the <a href="#" class="govuk-footer__link">Government Digital Service</a>' %}
```
quis:
```suggestion
    {% set meta_suffix = 'Built by the <a href="https://www.gov.uk/government/organisations/government-digital-service" class="govuk-footer__link">Government Digital Service</a>' %}
```

gulpfile.js
```diff
@@ -70,14 +71,32 @@ const copy = {
...
+  govuk_frontend: {
+    fonts: () => {
+      return src(paths.govuk_frontend + 'assets/fonts/**/*')
+        .pipe(dest(paths.dist + 'fonts/'));
+    }
   }
 };
 
 
+const html5shivs = () => {
```
quis: Do we still need the HTML 5 shiv now we’ve dropped IE8 support?

quis:
Do we no longer have control over which items appear in which columns in the footer? The previous grouping was deliberate.

app/navigation.py
tombye: Put these fixes together with the others in https://github.com/alphagov/notifications-admin/pull/3158/commits/15a4cf5c632dc0845b2661f0e9a70afb61355b78

app/templates/admin_template.html
tombye: This was fixed in https://github.com/alphagov/notifications-admin/pull/3158/commits/15a4cf5c632dc0845b2661f0e9a70afb61355b78

app/templates/admin_template.html
tombye: This was fixed in https://github.com/alphagov/notifications-admin/pull/3158/commits/15a4cf5c632dc0845b2661f0e9a70afb61355b78

gulpfile.js
tombye: I put it in to support the GOV.UK Frontend IE8 code but actually we don't need to pull that in at all. Deleted [the commit that adds IE8 code](https://github.com/alphagov/notifications-admin/pull/3158/commits/58965ed3445c6d37c9a4c0695b1aab150985effa).

tombye:
> Do we no longer have control over which items appear in which columns in the footer? The previous grouping was deliberate.

We do and the GOV.UK Frontend footer expects it, I just did it as one column. Now changed in https://github.com/alphagov/notifications-admin/pull/3158/commits/56a869a3920e6edee480b101f6766d0e48a7aca1

tests/app/main/views/test_service_settings.py
```diff
@@ -1172,8 +1172,8 @@ def test_non_gov_user_is_told_they_cant_go_live(
     assert normalize_spaces(page.select_one('main p').text) == (
         'Only team members with a government email address can request to go live.'
     )
-    assert page.select('form') == []
-    assert page.select('button') == []
+    assert len(page.select('form')) == 0
+    assert len(page.select('button')) == 1
```
idavidmcdonald: Could you remind me what causes the increase in button elements again?

tests/app/main/views/test_notifications.py
```diff
@@ -227,10 +227,9 @@ def test_notification_page_shows_page_for_letter_notification(
     letter_images = page.select('main img')
 
     assert len(letter_images) == count_of_pages
-
-    for index in range(1, count_of_pages + 1):
+    for index in range(count_of_pages):
         assert page.select('img')[index]['src'].endswith(
-            '.png?page={}'.format(index)
+            '.png?page={}'.format(index + 1)
```
idavidmcdonald: Could you explain what is going on here please

idavidmcdonald:
One potential issue. Your change to return our footer to 4 columns with varying numbers of items in them in d4fc8846cdc48ac1b49e0b93125c7b960d351eee introduces the following effect on smaller window sizes:

image


Not sure if this is OK, it's kind of trying to put them into 4 distinct columns still rather than what happens at the moment which is they all go into 1 column on smaller window sizes.

idavidmcdonald:
On the platform-admin page, if you open and close the 'apply filters' link twice it should return to

image


However, on your branch, the arrow does not return to horizontal:

image

idavidmcdonald:
Also if I open your branch in a private window I do not get the cookie banner. If I open the production site in a private window I do get the cookie banner.

I had a quick glance to see if we were disabling this for development but didn't spot anything at a very lazy glance.

idavidmcdonald:
Found a small bug in production (not related to your changes). for (roughly) tablet size viewports, not all the numbers for number of notifications sent fit in the box:

image

tests/app/main/views/test_notifications.py
tombye: I think so:

`page.select('img')` returned all the images in the page. This used to result in two:
1. the GOV.UK logo image
2. the image of the PDF letter

The GOV.UK logo is now an `<svg>` so only one image is returned.

The `for index in range(1, count_of_pages + 1):` loop used to start at 1 to ignore the logo image. It can now start at 0, but that means the number sent to `'.png?page={}'.format()` needs to incremented as the range of pages start at 1.

tests/app/main/views/test_service_settings.py
tombye: Sorry, it's not clear.

The header has a 'Menu' button which appears on smaller screens:
image

This used to be an anchor element but is now, correctly, a `<button>`.

app/templates/components/notify-footer.html
```diff
@@ -0,0 +1,90 @@
+{% macro notify_footer(params) %}
```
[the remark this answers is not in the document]
tombye: You are exactly correct and I totally agree. We have a WIP pull request up at the moment:

https://github.com/alphagov/notifications-admin/pull/3181

They've agreed on new headings, I just wanted this pull request to start by replicating what we have at present.

app/assets/stylesheets/main.scss
```diff
@@ -35,6 +35,7 @@ $path: '/static/images/';
 @import './govuk-frontend/all';
 
 // Specific to this application
+@import 'globals';
```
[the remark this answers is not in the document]
tombye: To add some explanation, some of the imports from GOVUK Elements, unlike those from GOVUK Frontend Toolkit, add actual CSS to this file, not just Sass helpers.

`@import globals` contains CSS that used to exist in the GOVUK Template CSS, which was loaded above the CSS file this ends up as. In other words, it needs to appear above all actual CSS in this file to replicate the cascade that existed when GOVUK Template was part of this app.

tombye:
> One potential issue. Your change to return our footer to 4 columns with varying numbers of items in them in [d4fc884](https://github.com/alphagov/notifications-admin/commit/d4fc8846cdc48ac1b49e0b93125c7b960d351eee) introduces the following effect on smaller window sizes:
> 
> image
> 
> Not sure if this is OK, it's kind of trying to put them into 4 distinct columns still rather than what happens at the moment which is they all go into 1 column on smaller window sizes.

Looks like this is because we don't have headings for our columns yet. We do have a [PR](https://github.com/alphagov/notifications-admin/pull/3181) open to add them but I need confirmation from those involved in the work before merging.

I've added a commit to keep the columns in the same row until they stack vertically on smaller screens:

https://github.com/alphagov/notifications-admin/pull/3158/commits/3e7b9f68a917d6d604be2560bc052ca0d50203e3

This should fix the issue until we can update the content.

tombye:
> On the platform-admin page, if you open and close the 'apply filters' link twice it should return to
> 
> image
> 
> However, on your branch, the arrow does not return to horizontal:
> 
> image

The 'Apply filters' link is actually a [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element. It's a native element but we use some JS to make it more accessible.
The bug you spotted was caused by the JS we currently use for this being broken by similar JS brought in by GOV.UK Frontend. This PR pulls in all the GOV.UK Frontend JS, to be run as is configured with no changes.

I raised a PR to change how we bring in the GOV.UK Frontend JS so only that needed for the [header component](https://design-system.service.gov.uk/components/header/) is included:

https://github.com/alphagov/notifications-admin/pull/3168

tombye:
> Also if I open your branch in a private window I do not get the cookie banner. If I open the production site in a private window I do get the cookie banner.
> 
> I had a quick glance to see if we were disabling this for development but didn't spot anything at a very lazy glance.

This was my mistake, forgetting to update a DOM API call in the cookie JS. Fixed in https://github.com/alphagov/notifications-admin/pull/3158/commits/74488fd2e9cb49a0346257f46d18559182e1340b.

tombye:
@idavidmcdonald I've added some fixes to the issues you raised (with explanations above).

I'll re-test the effected pages before merging but be useful to know what you think.

app/assets/javascripts/modules/all.js
```diff
@@ -0,0 +1,745 @@
+var GOVUKFrontend = (function (exports) {
```
idavidmcdonald: Where has all this code for this file come from? I'm assuming you didn't write it yourself and it is taken from the GOV.UK frontend? However I can't find it in there at a glance

I'm also not clear on what exactly the purpose of this file is. What is this code doing for us? Thanks for any enlightenment.

idavidmcdonald:
Thanks for addressing my previous comments Tom. I think if you retest these new changes, in particular the javascript side of things, and you are happy, then so am I.

I'm a little unclear on some of the implementation of https://github.com/alphagov/notifications-admin/commit/838cb862e874a87b1e1364136d86c276477b2058 (I left a comment with a few questions regarding this) but overall I understand the problem it is trying to solve.

app/assets/javascripts/modules/all.js
tombye: This file is an artefact produced by part of the frontend build that turns JS Modules into script that can run in all browsers. Your comment made me realise that process isn't clearly explained in the existing code.

I've refactored it to (hopefully) be clearer, added loads of comments to the bit in `gulpfile.js` and removed `modules/all.js` from the repo'.

I'm really keen for this to be understandable so please say if any of it isn't clear.

app/assets/stylesheets/main.scss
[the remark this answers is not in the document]
tombye: Now moved in https://github.com/alphagov/notifications-admin/pull/3158/commits/e7f6d530947d6e14c4e71adeaa30c5ffaec0ea88.

tombye:
I

> SHIP IT! 🚢 :shipit:

Since this, I changed how the part of the frontend build that handles JavaScript works. This was to get around the first changes creating a temporary file (see https://github.com/alphagov/notifications-admin/pull/3158#discussion_r347986578 for details). I went over this change, and the other fixes requested with @idavidmcdonald in person, who OK'ed them for merge, subject to further cross-browser testing (which I did afterwards).

PR #3160, opened 2019-10-25
Remove min-height from sticky submit buttons

quis:
It makes them get out of alignment with other things in sticky footers.

# Before 

nojump-jump

# After 

nojump


quis:
This messes up the sticky footer with no button:
image

quis:
Added some fixes so that all the variants look aligned, and the latter two still don’t jump:

image

image

image

image

PR #3165, opened 2019-10-29
Always show letters on the dashboard and usage page 

quis:
We hid letters originally because it wasn’t a mature feature. We rolled it out by letting teams choose to use it (#1803) and then automatically giving it to new teams (https://github.com/alphagov/notifications-api/pull/1600).

This commit doesn’t change who has access to letters, but it does make it more discoverable by revealing it in the UI. This is the same thing we do for emails/texts, where even if you switch them off they still show up on the dashboard and usage page.

# Before 

image

With letters switched on | With letters switched off
---|---
image | image

# After 

image

image


app/templates/views/usage.html
```diff
@@ -87,10 +104,16 @@ <h2 class='heading-small'>Text messages</h2>
                 <li class="tabular-numbers">{{ "{:,}".format(month.free) }} free {{ message_count_label(month.free, 'sms', '') }}</li>
               {% endif %}
               {% if month.paid %}
-                <li class="tabular-numbers">{{ "{:,}".format(month.paid) }} {{ message_count_label(month.free, 'sms', '') }}at
+                <li class="tabular-numbers">{{ "{:,}".format(month.paid) }} {{ message_count_label(month.paid, 'sms', '') }}at
```
idavidmcdonald: Is this change inadvertently fixing a bug? Struggling to see another reason why we are moving from `month.free` to `month.paid` here?

app/assets/stylesheets/components/table.scss
```diff
@@ -98,6 +98,20 @@
 
 }
 
+.usage-table {
+
+  .table-field-heading-first {
+    width: 35%;  // 33.33% + fudge
```
idavidmcdonald: Could you explain for me why we add the extra amount of fudge on these please?

app/templates/views/usage.html
quis: Yeah, this is a bug. It means if you had used 1 paid text message and 250,000 free text messages you would see, the wrong pluralisation:
> 1 text messages at 1.58p

app/assets/stylesheets/components/table.scss
quis: I don’t actually understand why it needs the fudge to visually appear as 33%. Something to do with how the width of tables is calculated differently to normal elements. Copied from here so maybe we can look seeing if there’s a better way of defining these measures?

https://github.com/alphagov/notifications-admin/blob/6c5255f30b700a199c9e1633863ed565016a66f0/app/assets/stylesheets/components/table.scss#L61-L71

PR #3167, opened 2019-10-30
Don’t show template statistics on dashboard if only one template has been used

quis:
If you’ve only used one template then this section of the page isn’t doing its job, which is to show a comparison of the different kinds of message you’re showing.

I think our initial assumption was that everyone would be using multiple templates, so it was good to show this part of the page during the onboarding, to show users where the information was going to appear.

But we have lots of services who only send one template now, typically where they’re populating the contents of the template themselves. In which case this part of the page doesn’t offer them any value.

***

Stops this from happening:

image

***

This is part of a wider piece of work to refresh the dashboard page now that jobs are moving to their own page.

***

I recommend reviewing these changes [with `w=1`](https://github.com/alphagov/notifications-admin/pull/3167/files?w=1) for a cleaner diff.

### DIFF
--- a/tests/app/main/views/test_dashboard.py
+++ b/tests/app/main/views/test_dashboard.py
@@ -496,6 +496,45 @@ def test_should_show_recent_templates_on_dashboard(
     assert '100' in table_rows[3].find_all('td')[0].text
 
 
+@pytest.mark.parametrize('stats', (
+    pytest.param(
+        [stub_template_stats[0]],
+    ),
+    pytest.param(
+        [stub_template_stats[0], stub_template_stats[1]],
+        marks=pytest.mark.xfail(raises=AssertionError),
+    )
+))
+def test_should_not_show_recent_templates_on_dashboard_if_only_one_template_used(
+    client_request,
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_dashboard.py
```diff
@@ -496,6 +496,45 @@ def test_should_show_recent_templates_on_dashboard(
...
+
+    page = client_request.get('main.service_dashboard', service_id=SERVICE_ONE_ID)
+    main = page.select_one('main').text
+
+    mock_template_stats.assert_called_once_with(SERVICE_ONE_ID, limit_days=7)
+
+    assert stats[0]['template_name'] == 'one'
+    assert stats[0]['template_name'] not in main
+
+    # count appears as total, but not per template
```
idavidmcdonald: I might argue that if these following assertions are useful then they should be covered in a different test that tests that the value for total number of text messages/letters/emails sent is shown and correct (with a few cases ie one template stat and multiple stats). I don't think we have a test that does this however?

PR #3168, opened 2019-10-30
[WIP] Add JS modules support & use for GOVUKFrontend

tombye:
# Do not merge yet - depends on https://github.com/alphagov/notifications-admin/pull/3158

Add support for JS modules to the frontend build and rewrite GOV.UK Frontend JS we include to only pull in the components used.

https://www.pivotaltracker.com/story/show/169344114

## Changes to JS folder structure

This introduces the `app/assets/javascripts/modules` folder to hold module code and the following conventions for its contents:
- any JS with the `.mjs` extension will be linted
- any JS with a `.js` extension will contain bundler boilerplate code so will not be linted

## Choice of bundler

I looked at Webpack and Rollup for bundling up JS modules into JS code for browsers.

Both introduced boilerplate code but Webpack had more and it was harder to read. The size of `app/assets/javascripts/modules/all.js` with each was:
- Webpack: 33KB
- Rollup: 26KB

We might use Webpack in future for splitting up our static assets into more optimised bundles. For now Rollup lets us work with JS modules without introducing any bundler-specific code outside of the frontend build so should be easier to swap out if needed than the other way around.


tombye:
Moving this commit into https://github.com/alphagov/notifications-admin/pull/3158 so closing this PR.

PR #3170, opened 2019-10-31
Show users why their precompiled letter failed validation

CrystalPea:
### This introduces dynamic validation failed messages when:
- user goes to notification page for a precompiled letter sent through API that failed valdiation
- admin / user visits an endpoint to check if pdf meets our validation conditions (I asked Pete and he said we want to keep this endpoint so I updated it in line with other endpoints)

### Other work:
- Refactor upload preview to use shared components
- Checking if overlay is needed for notification preview now happens in API, based on metadata, so I removed redundant code from admin

#### Notification page:
Screenshot 2019-10-31 at 12 42 45
Screenshot 2019-10-31 at 12 39 33
Screenshot 2019-10-31 at 12 39 18

##### for services:
Screenshot 2019-10-31 at 12 51 59
Screenshot 2019-10-31 at 12 51 25

##### for platform admins:
Screenshot 2019-10-31 at 12 52 23

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import base64
 import io
+import json
 import os
 from datetime import datetime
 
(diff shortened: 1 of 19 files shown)


CrystalPea:
@servingUpAces yes, it's by design. I first made them all in the box, but then consulted with @quis, and he explained to me how notification page is different because the letter already was sent, so it's different place in the journey and should be treated in line with other errors we show after the notification has been created.

PR #3176, opened 2019-11-04
Be stricter about the format of URL parameters 

quis:
Before, these would need view-layer code or the result of an API call in order to respond with a 404:
```
/services/foo/templates
```
```
/services/1234/templates/fax
```

Now Flask will 404 them straight away because it knows that `service_id` should always be a UUID (not `foo` or `1234`) and `template_type` should always be `email`, `sms`, or `letter` (not `fax`).

***

This is what the tests look like if you miss marking an ID as being a UUID:

image

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -213,7 +213,7 @@ def view_job_updates(service_id, job_id):
 
 
 @main.route('/services/<uuid:service_id>/notifications', methods=['GET', 'POST'])
-@main.route('/services/<uuid:service_id>/notifications/<message_type>', methods=['GET', 'POST'])
+@main.route('/services/<uuid:service_id>/notifications/<template_type:message_type>', methods=['GET', 'POST'])
 @user_has_permissions()
 def view_notifications(service_id, message_type=None):
     return render_template(
@@ -237,7 +237,7 @@ def view_notifications(service_id, message_type=None):
 
 
 @main.route('/services/<uuid:service_id>/notifications.json', methods=['GET', 'POST'])
-@main.route('/services/<uuid:service_id>/notifications/<message_type>.json', methods=['GET', 'POST'])
+@main.route('/services/<uuid:service_id>/notifications/<template_type:message_type>.json', methods=['GET', 'POST'])
 @user_has_permissions()
 def get_notifications_as_json(service_id, message_type=None):
     return jsonify(get_notifications(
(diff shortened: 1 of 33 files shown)


app/main/views/jobs.py
```diff
@@ -237,23 +237,22 @@ def view_notifications(service_id, message_type=None):
...
+@main.route('/services/<uuid:service_id>/notifications/<template_type:message_type>.json', methods=['GET', 'POST'])
 @user_has_permissions()
 def get_notifications_as_json(service_id, message_type=None):
     return jsonify(get_notifications(
         service_id, message_type, status_override=request.args.get('status')
     ))
 
 
-@main.route('/services/<uuid:service_id>/notifications/<message_type>.csv', endpoint="view_notifications_csv")
+@main.route('/services/<uuid:service_id>/notifications.csv', endpoint="view_notifications_csv")
```
idavidmcdonald: Could you explain why this change (the addition of a new route) is being made here please. I'm not sure of the context.

app/main/views/templates.py
```diff
@@ -405,9 +404,16 @@ def _get_template_copy_name(template, existing_templates):
...
+    '/services/<uuid:service_id>/templates/action-blocked/'
+    '<template_type:notification_type>'
+))
+@main.route((
+    '/services/<uuid:service_id>/templates/action-blocked/'
+    '<template_type:notification_type>/<return_to>/<uuid:template_id>'
+))
 @user_has_permissions('manage_templates')
-def action_blocked(service_id, notification_type, return_to, template_id):
+def action_blocked(service_id, notification_type, return_to='add_new_template', template_id=None):
```
idavidmcdonald: Could you explain the context behind the change of `return_to='add_new_template'` please

app/main/views/jobs.py
quis: This URL was sometimes being generated with `message_type=None` (to get all types of messages):
```
/services/128b91b6-2996-4107-bb65-51b7c24a728d/notifications/sms.csv
/services/128b91b6-2996-4107-bb65-51b7c24a728d/notifications/None.csv
```

This changes it to not have `None` in the URL, which matches what we do for the HTML-responding equivalent:
https://github.com/alphagov/notifications-admin/blob/265931d21746918c4ddfc19c4ad3f8cb5683c1bf/app/main/views/jobs.py#L215-L216

***

I’ll add this detail to the commit message.

app/main/views/templates.py
quis: Because we’ve removed the `return_to` parameter from one of the URLs here it now needs to have a default value. We don’t ever want a default value of `None`, because `None` is not a key in this dictionary:
https://github.com/alphagov/notifications-admin/blob/c8fe1fc715afd04d69aa95beded162ceae8e2ddc/app/templates/views/templates/action_blocked.html#L15-L19

app/main/views/jobs.py
idavidmcdonald: Thanks, well explained

app/main/views/templates.py
quis: The `action_blocked` endpoint needed a variation of the URL without a `template_id` parameter, because `None` is no longer a valid `template_id` (because it’s not a UUID).

This change was made in 265931d21746918c4ddfc19c4ad3f8cb5683c1bf, which also removed the `return_to` parameter, because the back link on the `action blocked` page only ever goes to `add_new_template` if there’s no `template_id` provided.

However this was conflating the two things, so I’ve wound it back a bit so that:
- there’s still a new route, whose URL doesn’t include `template_id` as a parameter
- `return_to` is always required

I’ve also refactored the code a bit to move the looking up of the back link from the Jinja into the view layer, so that the related code is in one place and easier to reason about.

***

These changes are in 8df2a2f2fdff6d0ca946beffdc78980aba2fae38

PR #3177, opened 2019-11-06
Update navigation menus

karlchillmaid:
This PR:

- adds the new `Get started` page to our navigation menus
- reorders the footer links into 4 distinct groups:
  - Support
  - About Notify
  - Pricing and payment
  - Using Notify
- updates the `Features` navigation menu
- adds a new `Using Notify` navigation menu

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -75,6 +75,7 @@
           <ul>
             <li><a href="{{ url_for("main.features") }}">Features</a></li>
             <li><a href="{{ url_for("main.roadmap") }}">Roadmap</a></li>
+            <li><a href="{{ url_for("main.get_started") }}">Get started</a></li>
             <li><a href="{{ url_for("main.trial_mode_new") }}">Trial mode</a></li>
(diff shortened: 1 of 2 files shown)


quis:
Looks like this:

image

app/templates/admin_template.html
```diff
@@ -64,21 +64,19 @@
       <div class="grid-row">
         <div class="column-one-quarter">
           <ul>
-            <li><a href="{{ url_for('main.support') }}">Support</a></li>
+            <li><a href="{{ url_for('main.support') }}">24-hour online support</a></li>
```
karlchillmaid: @quis What do you think about `Contact support`?

It's active, it includes 2 keywords that users are likely to be scanning the page for, and it'll still work when we introduce the the column headings.

karlchillmaid:
@quis @tombye I guess some of the changes in this PR won't apply anymore once the new template/footer stuff is merged?

quis:
Closing now the new footer is live – looks like this:
image

PR #3179, opened 2019-11-08
Validate service and organisation name

CrystalPea:
This is a bug fix, details here: https://www.pivotaltracker.com/story/show/169621019

tl;dr: when changing service name to `.`, a user got an unhandled exception. We want to validate against this instead.

Screenshot 2019-11-20 at 13 32 08

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -40,6 +40,7 @@ from app.main.validators import (
     CsvFileValidator,
     DoesNotStartWithDoubleZero,
     LettersNumbersAndFullStopsOnly,
+    MustContainAlphanumericCharacters,
     NoCommasInPlaceHolders,
     OnlySMSCharacters,
     ValidEmail,
@@ -547,7 +548,8 @@ class RenameServiceForm(StripWhitespaceForm):
     name = StringField(
         u'Service name',
         validators=[
-            DataRequired(message='Cannot be empty')
+            DataRequired(message='Cannot be empty'),
+            MustContainAlphanumericCharacters()
         ])
 
 
@@ -555,7 +557,8 @@ class RenameOrganisationForm(StripWhitespaceForm):
     name = StringField(
         u'Organisation name',
         validators=[
-            DataRequired(message='Cannot be empty')
+            DataRequired(message='Cannot be empty'),
+            MustContainAlphanumericCharacters()
         ])
 
 
@@ -662,7 +665,8 @@ class CreateServiceForm(StripWhitespaceForm):
     name = StringField(
         "What’s your service called?",
         validators=[
-            DataRequired(message='Cannot be empty')
+            DataRequired(message='Cannot be empty'),
+            MustContainAlphanumericCharacters()
         ])
     organisation_type = OrganisationTypeField('Who runs this service?')
 
--- a/tests/app/main/test_validators.py
+++ b/tests/app/main/test_validators.py
@@ -5,6 +5,7 @@ from wtforms import ValidationError
 
 from app.main.forms import RegisterUserForm, ServiceSmsSenderForm
 from app.main.validators import (
+    MustContainAlphanumericCharacters,
     NoCommasInPlaceHolders,
     OnlySMSCharacters,
     ValidGovEmail,
@@ -177,6 +178,19 @@ def test_non_sms_character_validation(data, err_msg, client):
         OnlySMSCharacters()(None, _gen_mock_field(data))
 
     assert str(error.value) == err_msg
+
+
+@pytest.mark.parametrize("string", [".", "A.", ".8...."])
+def test_if_string_does_not_contain_alphanumeric_characters_raises(string):
+    with pytest.raises(ValidationError) as error:
+        MustContainAlphanumericCharacters()(None, _gen_mock_field(string))
+
+    assert str(error.value) == "Must include at least two alphanumeric characters"
+
+
+@pytest.mark.parametrize("string", [".A8", "AB.", ".42...."])
+def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
+    MustContainAlphanumericCharacters()(None, _gen_mock_field(string))
 
 
 def test_sms_sender_form_validation(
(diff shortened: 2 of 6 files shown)


app/main/forms.py
```diff
@@ -547,15 +547,17 @@ class RenameServiceForm(StripWhitespaceForm):
     name = StringField(
         u'Service name',
         validators=[
-            DataRequired(message='Cannot be empty')
+            DataRequired(message='Cannot be empty'),
+            Regexp(r"[a-zA-Z0-9].*[a-zA-Z0-9]", message="Needs to include at least two alphanumeric characters")
```
CrystalPea: @karlchillmaid I wrote this message so you might want to take a look :)

leohemsted: this looks to me like it's testing "needs to start and end with alphanumeric characters, but can have anything in between". Which might be fine, but i think "Needs to include at least two alphanumeric characters" would look the same but with `.*` at the beginning and end of the string.

also is there a specific technical reasoning for this? If underscores are okay then `\w` in regex performs the same as `[a-zA-Z0-9_]` so this could be `r"\w.*\w"` (or `r".*\w.*\w.*`)

CrystalPea: name consisting just of underscores still produces exception, but you are right I need more wildcards :)

klssmith:
You could make a new validator in `validators.py` if you wanted - it would save repeating the regex multiple times.

CrystalPea:
@klssmith That's a good idea, I'll do it 👍

app/main/forms.py
karlchillmaid: `Needs to` is fine – `Must` might be neater.

Just to check I've understood the requirements correctly: the service name can be anything, but it must include a minimum of 2 characters (letters or numbers)?

CrystalPea: > Just to check I've understood the requirements correctly: the service name can be anything, but it must include a minimum of 2 characters (letters or numbers)?

Correct. So for example:
Those names would be allowed: `My service`, `AB`, `A.B`, `_my_service`
But those would be not: `...`, `_A_`

tests/app/main/views/organisations/test_organisation.py
```diff
@@ -160,6 +160,31 @@ def test_create_new_organisation_validates(
     assert mock_create_organisation.called is False
 
 
+@pytest.mark.parametrize("new_organisation_name", [".", "A."])
+def test_create_new_organisation_validates_organisation_name(
```
idavidmcdonald: Optional: Test name could explicitely say organisation name contains alphanumeric characters rather than just 'validates'. Further, the three tests you've added have quite different names, I think like they could be more consistent as they are all testing the same thing but on different forms

tests/app/main/views/organisations/test_organisation.py
```diff
@@ -160,6 +160,31 @@ def test_create_new_organisation_validates(
     assert mock_create_organisation.called is False
 
 
+@pytest.mark.parametrize("new_organisation_name", [".", "A."])
```
idavidmcdonald: Optional: You could add an additional test case for strings longer than 2 characters but not alphanumeric, e.g. "  h " or ".8....."

tests/app/main/views/organisations/test_organisation.py
```diff
@@ -160,6 +160,31 @@ def test_create_new_organisation_validates(
...
+
+    client_request.login(platform_admin_user)
+    page = client_request.post(
+        '.add_organisation',
+        _data={
+            'name': new_organisation_name,
+            'organisation_type': 'local',
+            'crown_status': 'non-crown',
+        },
+        _expected_status=200,
```
idavidmcdonald: I find it a bit weird that we are posting a form, the contents is not valid and there is an error but we get a 200 status code. I would have expected a 4xx. Note, this isn't a code change you've introduced, it just took me a whilst to figure out what the test was testing because of it

tests/app/main/views/organisations/test_organisation.py
```diff
@@ -160,6 +160,31 @@ def test_create_new_organisation_validates(
...
+    )
+    assert mock_create_organisation.called is False
+    page.find("span", {"class": "error-message"})
```
idavidmcdonald: Is it worth asserting the error message text itself?

tests/app/main/views/organisations/test_organisation.py
CrystalPea: Good idea 👍

tests/app/main/views/organisations/test_organisation.py
CrystalPea: 👍

tests/app/main/views/organisations/test_organisation.py
CrystalPea: If form validated correctly, it would be 302. I think 200 status is consistent with other places where form does not validate on submit. If we wanted to change that behaviour, I feel we should discuss it with others and do it as a separate refactor-focused piece of work.

tests/app/main/views/organisations/test_organisation.py
CrystalPea: I mulled over what Rob once said to us and try not to assert on content. But I'm open to discussion on this :)

karlchillmaid: Just to check I've understood correctly: asserting on content means adding the expected content to the test? The knock-on effect being that any content changes need to be reflected in the tests as well?

CrystalPea: Yeah, that's correct 👍

tests/app/main/views/organisations/test_organisation.py
idavidmcdonald: Agreed

tests/app/main/test_validators.py
```diff
@@ -179,6 +180,19 @@ def test_non_sms_character_validation(data, err_msg, client):
     assert str(error.value) == err_msg
 
 
+@pytest.mark.parametrize("string", [".", "A.", ".8...."])
+def test_if_string_contains_alphanumeric_characters_raises(string):
```
idavidmcdonald: This might need to be `does_not_contain`?

tests/app/main/views/organisations/test_organisation.py
idavidmcdonald: I don't feel too strongly on it so fine without 👍

PR #3181, opened 2019-11-11
Update footer links

tombye:
Update the footer links so they are grouped into categories, each with a heading.

### Mobile

footer_links_old_mobile

### Desktop

image

### Mobile
notify-footer-new

### Desktop
notify-footer-new-desktop

These changes include using the [footer component from the GOV.UK Design System](https://design-system.service.gov.uk/components/footer/) rather than our custom version.

Based on work done by @karlchillmaid and @quis.

Note: the 'cookies' and 'privacy' links have a heading which is visually hidden and defaults to 'Support links'. This should probably also be updated.


app/templates/admin_template.html
```diff
@@ -1,114 +1,243 @@
...
+
+  {{ govukFooter({
+    "navigation": [
+      {
+        "title": "Support",
+        "columns": 1,
+        "items": [
+          {
+            "href": url_for('main.support'),
+            "text": "24-hour online support"
```
karlchillmaid: @quis
`24-hour online support` or `Contact support`?

app/assets/stylesheets/govuk-frontend/overrides.scss
```diff
@@ -0,0 +1,30 @@
+// Overrides for certain GOV.UK Frontend styles to make its components match this application's design
+
+// Notify's header navigation aligns to the right
+.govuk-header__navigation {
```
quis: I think there is a class for this, Dave showed me once where it’s documented but I can’t remember.

GovWifi is a good example: https://github.com/alphagov/govwifi-admin/blob/8cb3505b51d5b5452ef4caddd6d9ef6e85ffae81/app/views/layouts/application.html.erb#L85

app/templates/admin_template.html
quis: Out of those two, _Contact support_

tombye: Changed in https://github.com/alphagov/notifications-admin/pull/3181/commits/aa2c095704732b7487836d7389028008513ed406

app/assets/stylesheets/govuk-frontend/overrides.scss
tombye: Good spot, I didn't know that existed. Now swapped over to use it instead of that Sass in https://github.com/alphagov/notifications-admin/pull/3181/commits/62fbc07bff447364f3d0969f9b6c2cb3e44b7fa4

app/templates/views/signedout.html
```diff
@@ -1,5 +1,7 @@
 {% extends "fullwidth_template.html" %}
 
+{% set signed_out = True %}
```
quis: …which would mean you don’t need to do this.
```suggestion
```

app/templates/admin_template.html
```diff
@@ -147,35 +147,73 @@
     {% set meta_suffix = 'Built by the <a href="https://www.gov.uk/government/organisations/government-digital-service" class="govuk-footer__link">Government Digital Service</a>' %}
   {% endif %}
 
-  {{ notify_footer({
+  {% if signed_out %}
```
quis: You can use the `current_user` global so that this logic applies to any page:
```suggestion
  {% if not current_user.is_authenticated %}
```
https://github.com/alphagov/notifications-admin/blob/d73c3a68f9ec3468842b4b18483d82ffa7d211ac/app/templates/admin_template.html#L44

tombye: I made the change (thanks for that, using existing code instead of new == awesome) but with the condition reversed so you don't have to deal with a `not`.

app/templates/views/signedout.html
tombye: Fixed in https://github.com/alphagov/notifications-admin/pull/3181/commits/8a597437463872dbbcedec7b9677c0cceee21d4f so marking this as resolved.

tombye:
@servingUpAces see what you mean. I took the 'get started' link out as it broke [this test](https://github.com/alphagov/notifications-admin/blob/ccd245b287d53ef18f79b82c0037cdfb825b68fb/tests/app/main/views/test_dashboard.py#L140) but I think I assumed part of what it was testing was that once you're logged in you don't need to see the 'get started' page. With hindsight I think that's not true and, anyway, the usecase you mentioned is more important.

I've reverted both commits involved:
- https://github.com/alphagov/notifications-admin/pull/3181/commits/f7d115d0defc2b43bcc7c4c762fe64d09c77cf22
- https://github.com/alphagov/notifications-admin/pull/3181/commits/0375af3441d2150f35b9674f7d66ce99bad06a1c

app/templates/admin_template.html
karlchillmaid: 👍

PR #3182, opened 2019-11-13
Add a specific route for template_type=all

quis:
`all` is not a real template type, so for links to template folders that apply to all template types we have a URL that looks like:
```
/services/<uuid:service_id>/templates/folders/<uuid:template_folder_id>
```

However Flask only generates this url when `url_for` is called with `template_type=None`. If called with `template_type=all` then Flask will generate a URL like
```
/services/<uuid:service_id>/templates/all/folders/<uuid:template_folder_id>
```

However attempting to load this URL will now 404, since `all` is not a template type recognised by the regex introduced in https://github.com/alphagov/notifications-admin/pull/3176

It would be nice to not have URLs with `all` in them at all, but since people might have bookmarked them we need to support them indefinitely.

Also considered but decided against adding `all` to the set of template types because it might cause other problems, for example attempting to create a new template with a type of `all` would never work.


app/main/views/templates.py
```diff
@@ -108,6 +108,7 @@ def start_tour(service_id, template_id):
 @main.route("/services/<uuid:service_id>/templates", methods=['GET', 'POST'])
 @main.route("/services/<uuid:service_id>/templates/folders/<uuid:template_folder_id>", methods=['GET', 'POST'])
 @main.route("/services/<uuid:service_id>/templates/<template_type:template_type>", methods=['GET', 'POST'])
+@main.route("/services/<uuid:service_id>/templates/all/folders/<uuid:template_folder_id>", methods=['GET', 'POST'])
```
leohemsted: i think we still need to add `"/services/<uuid:service_id>/templates/all"`?

if you click a blue button without being a folder you end up there (thanks to https://github.com/alphagov/notifications-admin/blob/master/app/main/views/templates.py#L215)

quis: Added in 0f24fbcda74455b44232de154ea7c84e895f917a

PR #3185, opened 2019-11-14
Only show the breaking change page to teams who use the API

quis:
We introduced the ‘breaking change’ page in #631 partly to help teach people about the relationship between the placeholders in their template and the data they were providing. Data can be provided either by API or by uploading a spreadsheet. The users who we struggled to communicate this relationship to were the ones using the upload a spreadsheet feature.

We made two changes to the context of this feature:

1. Around the same time (in #613) we introduced the interactive tour, which ultimately proved to be the thing that helped people understand the relationship between the data they were providing and the placeholders in the template.

2. We introduced a way for people to send one-off messages without using the API or uploading a spreadsheet in #1293. So for this page to say that you’ll need to update a spreadsheet or change an API call if you change the placeholders in your template is no longer accurate.

Therefore I think it makes sense to only show this page to teams who are using the API to send messages. The best proxy we have for that is to look at whether they’ve created any API keys.

***

Since we’re only showing this page to team who are using the API we don’t have to worry about explaining what’s going on in terms of the spreadsheet any more. So this pull request also removes the example spreadsheet.

This makes the page simpler and more focused.

# Before 

Shown to everyone:

image

# After 

Only shown to teams using the API:

image

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -603,7 +603,12 @@ def edit_service_template(service_id, template_id):
 
         new_template = get_template(new_template_data, current_service)
         template_change = get_template(template, current_service).compare_to(new_template)
-        if template_change.placeholders_added and not request.form.get('confirm'):
+
+        if (
+            template_change.placeholders_added and
+            not request.form.get('confirm') and
+            current_service.api_keys
+        ):
             example_column_headings = (
                 first_column_headings[new_template.template_type] + list(new_template.placeholders)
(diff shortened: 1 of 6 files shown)


quis:
Worked with @karlchillmaid to update the content to make the developer instructions more explicit.

The previous ‘you need to update your API calls’ text wasn’t clear enough once the context of the spreadsheet has been removed.

image

PR #3189, opened 2019-11-19
Update Zendesk tags to reflect new taxonomy

quis:
Requests to go live and email branding requests come through to Zendesk with tags attached automatically.

With the revised taxonomy some of these tags need to be updated, as summarised in this spreadsheet.

In addition, `notify_action` tag has to be added in each of those cases.

Old|New
---|---
`notify_request_to_go_live_complete`|`notify_go_live_complete`
`notify_request_to_go_live_incomplete`|`notify_go_live_incomplete`
`notify_action_add_branding`|`notify_branding`
`notify_request_to_go_live_incomplete_mou`|`notify_go_live_incomplete_mou`
`notify_request_to_go_live`|`notify_go_live`

– https://docs.google.com/spreadsheets/d/1o5ATsFsVK8Qpj7x8QvxX-SfEuBZ75028GEySVcdBFYU/edit#gid=0
– https://www.pivotaltracker.com/story/show/169842970

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1065,7 +1065,7 @@ def branding_request(service_id):
             ticket_type=zendesk_client.TYPE_QUESTION,
             user_email=current_user.email_address,
             user_name=current_user.name,
-            tags=['notify_action_add_branding'],
+            tags=['notify_action', 'notify_branding'],
         )
 
         flash((
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -625,7 +625,6 @@ class Service(JSONModel):
             return
 
         for test, tag in (
-            (True, ''),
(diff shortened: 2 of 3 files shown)


karlchillmaid:
The new taxonomy doesn't have a `notify_go_live_incomplete` tag. We replaced this with `notify_go_live_incomplete_mou` because the only way users can submit an incomplete request is if they do not agree to the MOU. 

These are the incomplete tags:

- `notify_go_live_incomplete_mou`
- `notify_go_live_incomplete_reply_to`
- `notify_go_live_incomplete_shared_email`
- `notify_go_live_incomplete_templates`

Of those, only the first one is applied automatically.

Worth checking the logic of this with JP to check we've captured it correctly in the document.

PR #3192, opened 2019-11-20
Update request to go live content

karlchillmaid:
This PR updates step 6 of the [get started](https://www.notifications.service.gov.uk/features/get-started) page for users who are not signed in to Notify, to make it clearer how to request to go live.

**Before**
Screen Shot 2019-11-20 at 14 24 53

**After**
Screen Shot 2019-11-20 at 14 20 25

### DIFF
--- a/app/templates/views/get-started.html
+++ b/app/templates/views/get-started.html
@@ -71,7 +71,7 @@
   <li class="get-started-list__item">
     <h2 class="get-started-list__heading">Start sending messages</h2>
     {% if not current_user.is_authenticated or not current_service %}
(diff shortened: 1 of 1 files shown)


karlchillmaid:
@idavidmcdonald Because Settings is an account page, I'm not sure we can link to it if the user is not signed in. If they **are** signed in, they'll see different content with a direct link to 'request to go live' instead.

I've suggested this change based on feedback from a user who had read this page while signed out, and wasn't sure **how** to request to go live.

The content isn't as elegant, but it is more explicit. We'll probably see more stuff like this as we add guidance to answer common support questions.

idavidmcdonald:
That makes sense Karl, cheers!

PR #3193, opened 2019-11-21
Validate CSVs against rows with empty messages

CrystalPea:
Screenshot 2019-11-27 at 16 57 47

Screenshot 2019-11-27 at 17 35 49

### DIFF
--- a/tests/app/main/test_errors_for_csv.py
+++ b/tests/app/main/test_errors_for_csv.py
@@ -68,10 +68,24 @@ MockRecipients = namedtuple(
             ]
         ),
         (
+            {}, {}, {3, 12}, [],
+            'sms',
+            [
+                'shorten your messages in 2 rows'
+            ]
+        ),
+        (
             {}, {}, {}, {2},
             'sms',
             [
                 'add content to empty message in 1 row'
+            ]
+        ),
+        (
+            {}, {}, {}, {2, 4, 8},
+            'sms',
+            [
+                'add content to empty messages in 3 rows'
             ]
         ),
     ]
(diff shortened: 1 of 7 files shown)


karlchillmaid:
I'm thinking about writing a more specific error message, controlled by an if statement. This would be more in line with some of our other errors. For example:

>**One of your messages is empty**
>You need to add content to the message in row X.
OR
>**Some of your messages are empty**
>You need to add content to the messages in rows X, Y and Z.

What would the knock-on effect of something like this be for the code and other error messages? Right now it looks as though the error title 'There is a problem with [filename]' is used for multiple errors.

@quis do you have any thoughts about this?

quis:
Yeah, I prefer that title. Not sure the sub-line is totally accurate because the only way this can happen is with the ‘optional‘ placeholders right? So it’s not that you directly need to add content, it’s that you need to make sure some content will show. Bit fiddlier to word that though…

karlchillmaid:
@quis I've had a similar content request from JP. Will need to find a way to explain this. In the meantime, will any of these do?

- `You need content for the message(s) in row(s) ...`
- `You will need content for the message(s) in row(s) ...` (bit too passive?)
- `Check that you have content for the message(s) in row(s) ...`

tests/app/main/test_errors_for_csv.py
```diff
@@ -8,63 +8,80 @@
...
+            {2, 4, 6, 8}, {3, 6, 9, 12}, [], [],
             'sms',
             [
                 'fix 4 phone numbers',
                 'enter missing data in 4 rows'
             ]
-        )
+        ),
+        (
+            {}, {}, {3}, [],
```
idavidmcdonald: Should we test the case for multiple rows which need shortening (in the same way we do for the tests cases above)

karlchillmaid:
When there's only 1 error and the description is a single sentence, it should have a full stop.
For example:

`You need to shorten the message in 1 row.`

PR #3200, opened 2019-11-27
Add re-upload button and back link to the preview page

CrystalPea:
Re-upload button is only shown if file failed validation.

Screenshot 2019-11-27 at 14 43 35


quis:
Looks good ✨ 

Think it needs some extra test assertions to make sure the right form is on the page when the letter is and isn’t valid.

klssmith:
The back link on the preview page appears twice now when the letter is valid:
Screenshot 2019-11-27 at 13 51 24

CrystalPea:
Good find @klssmith

PR #3204, opened 2019-11-27
Replace GOVUK Template anchor focus styles

tombye:
These styles were missed from those moved across when the GOVUK Template CSS was removed.

Taken from https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_accessibility.scss#L52


quis:
Checked out locally and tested:
- header
- left hand navigation
- all organisations page
- templates page
- settings page
- team page
- removing a team member using the red link

👍

PR #3208, opened 2019-11-28
Use GOVUK Frontend component for all back-links

tombye:
This moves all back-links in the app to use the [GOVUK Frontend back-link component](https://design-system.service.gov.uk/components/back-link/).

back-link


app/templates/vendor/govuk-frontend/components/back-link/template.njk
```diff
@@ -0,0 +1,2 @@
+<a href="{%- if params.href %}{{ params.href }}{% else %}#{% endif -%}" class="govuk-back-link{%- if params.classes %} {{ params.classes }}{% endif -%}"
```
idavidmcdonald: This sort of code feels like the sort that I would have assumed lives in the gov uk frontend and does not get commited to our repository?

Is there a reason we can't do something like the example on the design system:

```
{% from "govuk/components/back-link/macro.njk" import govukBackLink %}

{{ govukBackLink({
  text: "Back",
  href: "#"
}) }}
```

leohemsted: yeah the vendor dir should definitely not be committed. this is just a build folder that we copy across from node_modules when we run `npm build`

sorry, should have caught this when we first merged the change to add this folder.

idavidmcdonald: No problem :) 

Tidy it up and I'll re-review when ready

tombye: I'd like to make this change separately and rebase this pull request over the top as it effects more than just this one.

If that's acceptable, here's a pull request for it: https://github.com/alphagov/notifications-admin/pull/3212

tombye: As of https://github.com/alphagov/notifications-admin/pull/3212 being merged, `app/templates/vendor` is no longer in this repository so I'm marking this as resolved.

PR #3211, opened 2019-11-29
Flake8 fix

leohemsted:
flake8 was only running for three actual warnings for the last month due to a misconfiguration, fix that and also fix all the errs we introduced over the last month

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,8 +9,3 @@ multi_line_output=3
 known_third_party=notifications_utils,notifications_python_client
 known_first_party=app,tests
 include_trailing_comma=True
-
-[flake8]
-# W504 line break after binary operator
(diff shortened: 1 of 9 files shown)


setup.cfg
```diff
@@ -9,8 +9,3 @@ multi_line_output=3
 known_third_party=notifications_utils,notifications_python_client
 known_first_party=app,tests
 include_trailing_comma=True
-
-[flake8]
-# W504 line break after binary operator
-extend_ignore=B003, B306, W504
-select=B901, B902, B903
```
quis: These are off by default so they still need enabling explicitly

quis: This is a whitelist. Maintaining it is more effort than it’s worth. 👍to removing this line.

PR #3215, opened 2019-12-03
Update error messages

karlchillmaid:
This PR updates 2 error messages to include contractions. This is consistent with other error message content.

### DIFF
--- a/app/templates/error/401.html
+++ b/app/templates/error/401.html
@@ -1,9 +1,9 @@
 {% extends "withoutnav_template.html" %}
-{% block per_page_title %}You are not authorised to see this page{% endblock %}
+{% block per_page_title %}You’re not authorised to see this page{% endblock %}
 {% block maincolumn_content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-    <h1>You are not authorised to see this page</h1>
+    <h1>You’re not authorised to see this page</h1>
       <p><a href="{{ url_for('.sign_in' )}}">Sign in</a> to GOV.UK Notify and try again.</p>
   </div>
 {% endblock %}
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Oh it looks like some test content also contains these strings so you will need to change any references in the 'test' folder

karlchillmaid:
Thanks @idavidmcdonald 

The [plain English guidance](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#plain-english) on GOV.UK says:

>"use contractions like you’ll (but avoid negative contractions like can’t)"

I think this approach comes from GOV.UK’s aim to write in a way that better reflects how our users speak, and to avoid sounding overly formal. It's not a hard and fast rule, though – context plays a part.

In the second of the 2 examples you found, I think `you are` actually works better. In the first, we could use either, although maybe `you’re` reads better aloud. It's pretty subjective stuff, and not a massive problem if we’re not always consistent.

The main reason for changing the wording in this PR is that there didn’t seem to be a good enough reason to use `there is` in these error messages, when we use `there’s` in all the others.

Interestingly the Guardian style guide (which GOV.UK defers to for anything not covered in the A-Z) says:

>"Do not overuse contractions ... while they might make a piece more colloquial or easier to read, they can be an irritant and a distraction, and make a serious article sound frivolous."

🤷‍♂

karlchillmaid:
> Oh it looks like some test content also contains these strings so you will need to change any references in the 'test' folder

I can't find the relevant test files.

idavidmcdonald:
Check out the two last search results in this:
https://github.com/alphagov/notifications-admin/search?q=%22You+are+not%22&unscoped_q=%22You+are+not%22

idavidmcdonald:
Concourse says you have fixed the tests succesfully :) feel free to merge if you are happy

PR #3217, opened 2019-12-04
Display and set provider priority in a non-independent way

quis:
Some of the code/commits here are maybe a bit scrappy, but it’s only for us and I’ve probably sunk enough time into it already:

# Before 

image

# After 

image


app/assets/stylesheets/components/radios.scss
```diff
@@ -117,3 +117,60 @@
     margin-right: 15px;
   }
 }
+
+.inline-stacked-label {
```
idavidmcdonald: Do we need this class?

app/assets/javascripts/radioSlider.js
```diff
@@ -0,0 +1,26 @@
...
+
+      $(component)
+        .on('click', function() {
+
+          leftRight = $(this).find(':checked').next('label').text().split('/');
+
+          if (leftRight.length === 2) {
+            $(this).find('.radio-slider-left-value').html(leftRight[0]);
+            $(this).find('.radio-slider-right-value').html(leftRight[1]);
+          }
```
idavidmcdonald: I don't find it immediately obvious what these lines are doing. Is there anything you could do to help improve that. Maybe a variable rename or some code comments?

app/main/views/providers.py
```diff
@@ -51,6 +57,65 @@ def edit_provider(provider_id):
...
+
+    providers = sorted([
+        provider
+        for provider in provider_client.get_all_providers()['provider_details']
+        if provider['notification_type'] == 'sms'
+    ], key=itemgetter('identifier'), reverse=True)
+
+    form = ProviderRatioForm(ratio=providers[0]['priority'])
+
+    if len(providers) < 2:
```
idavidmcdonald: Optional: Could be `!=2` just in case it ever found 3

app/main/views/providers.py
```diff
@@ -51,6 +57,65 @@ def edit_provider(provider_id):
...
+        abort(400)
+
+    primary_provider, secondary_provider = providers[0:2]
+
+    if form.validate_on_submit():
+        provider_client.update_provider(primary_provider['id'], form.percentage_left)
+        provider_client.update_provider(secondary_provider['id'], form.percentage_right)
```
idavidmcdonald: So there is a risk here that the first provider update succeeds and then the second one fails (maybe there is a blip in the network) and which point we are left with the priorities not adding up to 100% and stuff will probably be pretty wrong.

It's a pretty small risk but ideally the DB edits would happen as a single transaction (not currently something we can do from the admin app).

I think it's an acceptable risk

app/templates/views/providers/edit-sms-provider-ratio.html
```diff
@@ -0,0 +1,92 @@
...
+                <div class="history-list-time">
+                  {{ version.updated_at|format_time }}
+                </div>
+              {% endif %}
+            </div>
+            <div class="column-two-thirds">
+              <div class="history-list-percentage">
+                <div class="history-list-percentage-left-label">
+                  {{ primary_provider }}<br><br>
+                  {{ version.priority|format_thousands }}%
```
idavidmcdonald: What is `format_thousands` doing for us here?

quis: Renders the number without any decimal points.

app/main/views/providers.py
quis: Python client retries failed network requests, agree the risk is low enough.

app/main/views/providers.py
quis: If it finds 3 it will just take the first two. Think this is OK.

app/assets/javascripts/radioSlider.js
quis: Added some variable names in 292b7df5a

app/assets/stylesheets/components/radios.scss
quis: Removed in b1e9dd3fb

quis:
Removed the link to change international SMS providers in 1bd27b97f

app/templates/views/providers/edit-sms-provider-ratio.html
idavidmcdonald: Is it needed? I've removed these from my local environment and it doesn't seem to produce decimal points

quis: If you ever end up with a non-round percentage like 33.333… then the decimal point would start to show. Feels sensible to future-proof this in case we ever start working on more granular numbers than multiples of 10%.

PR #3219, opened 2019-12-05
Updated the uploads page to include letters. 

servingUpAces:
Now the page shows uploaded letters plus jobs.

New units tests have not been written for this page because it is very like this will be refactor and probably a new template created for the page. Some design needs to go into this page.
But we needed something ready for user research.

This depends on https://github.com/alphagov/notifications-api/pull/2670

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1937,6 +1937,42 @@ def mock_get_jobs(mocker, api_user_active):
 
 
 @pytest.fixture(scope='function')
+def mock_get_uploads(mocker, api_user_active):
+    def _get_jobs(service_id, limit_days=None, statuses=None, page=1):
+        if statuses is None:
+            statuses = ['', 'scheduled', 'pending', 'cancelled', 'finished']
+
+        jobs = [
+            job_json(
+                service_id,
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -1936,6 +1936,42 @@ def _get_jobs(service_id, limit_days=None, statuses=None, page=1):
     return mocker.patch('app.job_api_client.get_jobs', side_effect=_get_jobs)
 
 
+@pytest.fixture(scope='function')
+def mock_get_uploads(mocker, api_user_active):
+    def _get_jobs(service_id, limit_days=None, statuses=None, page=1):
```
idavidmcdonald: Sure, this does fix the tests but is a little confusing as its content is all linked to jobs still and not uploads. Some potential options.

Option 1: Rename to `_get_uploads`, remove `statuses` logic, change pagination links to point to `/uploads`, return more realistic data for uploads rather than just jobs

Option 2: Instead of returning more realistic data, you could also just get the tests passing by returning a mocked response from the API that contains no uploads (assuming the user has done none). Would be a nice small JSON response to hardcode.

PR #3228, opened 2019-12-19
Remove pip-accel

idavidmcdonald:
Not sure how much benefit we are getting from this. Pip-accel has
already not been supported for 3 years now either.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 SHELL := /bin/bash
 DATE = $(shell date +%Y-%m-%dT%H:%M:%S)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -37,7 +36,7 @@ venv: venv/bin/activate ## Create virtualenv if it does not exist
 
 venv/bin/activate:
 	test -d venv || virtualenv venv -p python3
-	. venv/bin/activate && pip install pip-accel
+	. venv/bin/activate
```
leohemsted: can we go one step further and remove this entire makefile command? i don't think anyone uses these (rather, everyone manages their own virtual environments)

PR #3230, opened 2019-12-20
Remove analytics

tombye:


app/assets/error_pages/5xx.html
```diff
@@ -125,21 +125,13 @@ <h1 class="heading-large">
...
-    // strip UUIDs
-    page = (window.location.pathname + window.location.search).replace(
-      /[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}/g, '…'
-    )
-    ga('send', 'pageview');
-  </script>
-
+    <script>
+      if (window.GOVUK.cookie('_ga') !== null) {
+        document.cookie = '_ga' + '=;expires=' + new Date() + ';domain=.' + window.location.hostname.replace(/^www/, '') + ';path=/';
```
quis: This is going to result in `";domain=..notifications.service.gov.uk"` – not sure if that will work.

This will work for www, non-www, and localhost:
```suggestion
        document.cookie = '_ga' + '=;expires=' + new Date() + ';domain=' + window.location.hostname.replace(/^www\./, '.') + ';path=/';
```

tombye: Rats, good point

PR #3233, opened 2019-12-23
Returned letters 

servingUpAces:
# Context
Once a month the postal provider sends a report of all the returned letters. At the moment, the services do not have a way to consume this information.

# What
There are two new endpoints available.

1) `services/<service_id>/returned-letter-summary` 
   Shows a list of report available to the service. A report is list of letters returned on a particular day.
 
2) `services/<service_id>/returned-letters-csv/<reported_at>`
  A CSV is downloaded containing all the letters returned for that service and report date. 

# TODO
There are still unit tests to write for the PR but since https://github.com/alphagov/notifications-api/pull/2688 is available I'm pushing this. But fear not tests are coming.


app/main/views/returned_letters.py
```diff
@@ -0,0 +1,46 @@
+from collections import OrderedDict
+
+from flask import render_template
+
+from app import service_api_client
+from app.main import main
+from app.utils import Spreadsheet, user_has_permissions
+
+
+@main.route("/services/<uuid:service_id>/returned-letter-summary", methods=["GET", "POST"])
```
idavidmcdonald: Do we want POST methods on both of these new routes? Is this usually what we do?

app/templates/views/returned-letter-summary.html
```diff
@@ -0,0 +1,32 @@
...
+        empty_message='If you have returned letter reports they will be listed here',
+        field_headings=[
+          'Report date',
+          'number of letters'
+        ],
+        field_headings_visible=False
+      ) %}
+    {% call field() %}
+        <a href="{{url_for('.returned_letters_report', service_id=current_service.id, reported_at=item.reported_at)}}">
+        Returned letters reported on {{ item.reported_at | format_date}}  ({{ item.returned_letter_count}} letters)
```
idavidmcdonald: Should be dealing with pluralisation on the word `letters` please

app/templates/views/returned-letter-summary.html
```diff
@@ -0,0 +1,32 @@
...
+<h1 class="heading-large">
+    Returned letters
+  </h1>
+<div class="body-copy-table" id='pill-selected-item'>
+    {% call(item, row_number) list_table(
+        data,
+        caption="Report",
+        caption_visible=False,
+        empty_message='If you have returned letter reports they will be listed here',
+        field_headings=[
```
idavidmcdonald: I plan on spinning this up locally (but will need to set up some data in my local DB first) but until then I can quickly ask the question:

Here we have two field headings but I only see one field being called (the long text string)? Is my understanding correct? And if so, should that be the case?

tests/app/main/views/test_returned_letters.py
```diff
@@ -0,0 +1,50 @@
...
+    mock = mocker.patch("app.service_api_client.get_returned_letter_summary",
+                        return_value=summary_data)
+
+    page = client_request.get("main.returned_letter_summary", service_id=SERVICE_ONE_ID)
+
+    mock.assert_called_once_with(SERVICE_ONE_ID)
+
+    expected_text = "Returned letters reported on Tuesday 24 December 2019 (30 letters)"
+    assert page.h1.string.strip() == 'Returned letters'
+    assert normalize_spaces(page.select('.table-field-left-aligned')[0].text) == expected_text
```
idavidmcdonald: Might be good to test the `a` `href` value too to know it's going to the right place. Also wondering if it's worth extending your test to cover multiple returned_letter reports

app/templates/views/returned-letter-summary.html
```diff
@@ -0,0 +1,32 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+<h1 class="heading-large">
+    Returned letters
+  </h1>
+<div class="body-copy-table" id='pill-selected-item'>
+    {% call(item, row_number) list_table(
+        data,
+        caption="Report",
```
idavidmcdonald: Again, I haven't spun this up yet to check, but this will be hidden but I'm assuming used for screenreaders? If so, I'm wondering if 'Report` is enough context for those who rely on a screen reader (given the additional context they got from the h1 heading) or if this should be more verbose? I don't know the answer, Tom might know but he's on holiday and I don't see any precedence in this repo already so don't worry if you can't find out

app/templates/views/returned-letter-summary.html
quis:
```suggestion
        Returned letters reported on {{ item.reported_at | format_date}}  ({{ item.returned_letter_count}} {{ message_count_label(item.returned_letter_count, 'letter') }})
```

app/main/views/returned_letters.py
servingUpAces: No the POST isn't necessary

app/templates/views/returned-letter-summary.html
```diff
@@ -0,0 +1,27 @@
...
+<div class="body-copy-table" id='pill-selected-item'>
+    {% call(item, row_number) list_table(
+        data,
+        caption="Returned letters report",
+        caption_visible=False,
+        empty_message='If you have returned letter reports they will be listed here'
+      ) %}
+    {% call field() %}
+      <a target="_blank"
+         href="{{url_for('.returned_letters_report', service_id=current_service.id, reported_at=item.reported_at)}}">Returned letters reported on {{ item.reported_at | format_date}} - {{ item.returned_letter_count}} {{ message_count_label(item.returned_letter_count, 'letter', suffix='')}}</a>
```
idavidmcdonald: Could you add a test for the pluralisation of the text please

app/templates/views/returned-letter-summary.html
```diff
@@ -0,0 +1,27 @@
...
+  </h1>
+<div class="body-copy-table" id='pill-selected-item'>
+    {% call(item, row_number) list_table(
+        data,
+        caption="Returned letters report",
+        caption_visible=False,
+        empty_message='If you have returned letter reports they will be listed here'
+      ) %}
+    {% call field() %}
+      <a target="_blank"
```
tombye: Would you mind adding `class="govuk-link"` to this link? It won't make a difference visually but will save me some work when I have to do it for the rest of the app.

PR #3236, opened 2019-12-31
Returned letters tweaks

quis:
Detail in the commits, but generally:
- makes the pages more consistent with the rest of Notify
- gives users a better idea of what they’re going to download (with a preview)

# Before 

image

# After 

image

image


app/templates/views/returned-letters.html
```diff
@@ -0,0 +1,57 @@
...
+
+{% block maincolumn_content %}
+
+{{ page_header(
+  'Returned letters for {}'.format(reported_at|format_date_normal),
+  back_link=url_for('main.returned_letter_summary', service_id=current_service.id)
+) }}
+
+<p class="bottom-gutter">
+  <a download href="{{ url_for('.returned_letters_report', service_id=current_service.id, reported_at=reported_at) }}" class="heading-small">Download this report</a>
```
tombye: Can we add a `govuk-link` to this link?

```suggestion
  <a download href="{{ url_for('.returned_letters_report', service_id=current_service.id, reported_at=reported_at) }}" class="govuk-link heading-small">Download this report</a>
```

PR #3240, opened 2020-01-03
Show the first line of the address from the to field.

servingUpAces:
Show the first line of the address from the to field.

An update to the API will be persisting the address to the Notification."to" field, after the notification has been validated.
If the letter is pending validation, then "Checking..." will appear as the identifier for the letter.
If the letter has passed validation, then the first line of the address (now persisted in the "to" field) will be displayed, with the client reference underneath.
If the letter has failed validation the "Provided as PDF" will show be displayed, which is now the initial value of the "to" field.

Screenshot 2020-01-03 at 14 53 38

### DIFF
--- a/app/templates/views/activity/notifications.html
+++ b/app/templates/views/activity/notifications.html
@@ -19,7 +19,7 @@
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
           <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.split('\n')[0] }}</a>
+          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.splitlines()[0] }}</a>
         {% endif %}
         <p class="file-list-hint">
           {{ item.preview_of_content }}
--- a/app/templates/views/activity/notifications.html
+++ b/app/templates/views/activity/notifications.html
@@ -17,9 +17,9 @@
     ) %}
       {% call row_heading() %}
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
-          <span class="file-list-filename">{{ item.to }}</span>
+          <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to }}</a>
+          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.split('\n')[0] }}</a>
         {% endif %}
         <p class="file-list-hint">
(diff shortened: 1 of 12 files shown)


app/templates/views/activity/notifications.html
```diff
@@ -17,9 +17,9 @@
     ) %}
       {% call row_heading() %}
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
-          <span class="file-list-filename">{{ item.to }}</span>
+          <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to }}</a>
+          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.split('\n')[0] }}</a>
```
quis: `str.splitlines`<sup>1</sup> is better because of how it handles different newline characters, and trailing newlines:
```suggestion
          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.splitlines()[0] }}</a>
```

1. https://docs.python.org/3/library/stdtypes.html#str.splitlines

app/templates/views/activity/notifications.html
```diff
@@ -17,9 +17,9 @@
     ) %}
       {% call row_heading() %}
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
-          <span class="file-list-filename">{{ item.to }}</span>
+          <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to }}</a>
+          <a class="file-list-filename" href="{{ url_for('.view_notification', service_id=current_service.id, notification_id=item.id) }}">{{ item.to.splitlines()[0] }}</a>
```
klssmith: If the first line of the address ends with a trailing comma it will be displayed as part of the address - not sure if that's something we want to remove or if it doesn't matter

servingUpAces: I think it's ok - that's how they sent us the data. I don't think it's typical

quis: Didn‘t @klssmith’s original work account for this? Would be a shame to undo it.

servingUpAces: Does that mean you want the full address displayed as `Recipient Name, 1st Line, 2nd Line, postcode`? I didn't use that format on the view activity page because the story just ask for ` Recipient Name. But it easy to show the full address.

tests/conftest.py
```diff
@@ -2879,6 +2879,14 @@ def _create_service_inbound_api(service_id, url, bearer_token, user_id):
...
 
 
+@pytest.fixture(scope='function')
+def mock_delete_service_inbound_api(mocker):
+    return mocker.patch(
+        'app.service_api_client.delete_service_callback_api',
+        side_effect=lambda service_id: None
+    )
+
+
```
quis: Is this code related to this PR?

PR #3241, opened 2020-01-06
Send asset path to Header component

tombye:
The fallback image for the SVG GOV.UK logo has an incorrect path.

We were getting 404's for it from IE11, which seems to load it despite being able to render the SVG.

Sets the `assetsPath` option on the Header component. See [the docs](https://design-system.service.gov.uk/components/header/#https://design-system.service.gov.uk/components/header/) for details.

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -125,7 +125,8 @@
     "homepageUrl": url_for('main.show_accounts_or_dashboard'),
     "productName": "Notify",
     "navigation": navigation,
-    "navigationClasses": "govuk-header__navigation--end"
+    "navigationClasses": "govuk-header__navigation--end",
+    "assetsPath": asset_path + "images"
   }) }}
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


quis:
Where is `asset_path` defined? I can’t see it used anywhere else in this file? Can we add a test to make the sure the complete `href` attribute is what we’d expect?

quis:
Example of a similar test here:
https://github.com/alphagov/notifications-admin/blob/2e108d1877630f12ce44235e00ef33f53b4f1c86/tests/app/main/views/test_index.py#L167-L177

tombye:
> Where is `asset_path` defined? I can’t see it used anywhere else in this file? Can we add a test to make the sure the complete `href` attribute is what we’d expect?

It's defined in `app/__init__.py`as one of a few variables sent to all templates (I think): https://github.com/alphagov/notifications-admin/blob/master/app/__init__.py#L214

PR #3242, opened 2020-01-06
Fix incorrect assertion for letter job CSV link

quis:
The statement `page.select('[download=download]')` was returning an empty list because the link on the page has an empty `download` attribute, rather than it being set to the string `download`.

This pull request:
- updates the assertion to find the element on the page
- makes the test more specific by checking where the link goes to, not  just its existance or lack thereof

### DIFF
--- a/tests/app/main/views/test_jobs.py
+++ b/tests/app/main/views/test_jobs.py
@@ -329,7 +329,11 @@ def test_should_show_letter_job(
     assert normalize_spaces(page.select('.keyline-block')[1].text) == (
         '6 January Estimated delivery date'
     )
-    assert page.select('[download=download]') == []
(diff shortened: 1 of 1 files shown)


quis:
N.B. the link originally had `download` set to the string `download` 

We removed this in https://github.com/alphagov/notifications-admin/commit/a604ed60b675eb69867193f3085cd01aceeaeb4f, which explained:
> We should standardise on `<a download>` rather than `<a download="download">` everywhere. The value of the download attribute tells the browser what filename to use, but is overridden by the
Content-Disposition HTTP header. Since it’s not being used, we should remove it for the sake of disambiguation.

quis:
Note also that we removed the download link in https://github.com/alphagov/notifications-admin/pull/1364, but I think it’s useful to keep it.

PR #3245, opened 2020-01-07
Remove govuk_template compatibility flag

tombye:
Setting the `$govuk-compatibility-govuktemplate` flag (in https://github.com/alphagov/notifications-admin/pull/3243) turned off the New Transport font.

I'll add some reasons for why it did here when I know more.

For now, this unsets the flag and removes a piece of govuk_template CSS turning it on fixed, removing the need for it.


tombye:
> The Arial 😱

Sorry 😞

PR #3247, opened 2020-01-07
Make create account button more buttony

quis:
`role=button` for users of voice control software 

`draggable=false` as per https://github.com/alphagov/govuk-frontend/pull/1020


quis:
@idavidmcdonald Yeah, once the buttons are using the Design System they get this automatically. Because the _Create account_ button isn’t in the Design System we have to fix it separately.

PR #3249, opened 2020-01-08
Update the letter validation messages.

servingUpAces:
# What
This PR updates the way the validation errors are displayed for precompiled letters sent in via the API. They now use "partials/check/letter-validation-failed-banner.html", which is the same banner that is displayed when an uploaded letter failed validation.
Also added a new error message for missing address blocks.
There is an outstanding question about what the page title should be, the standard is to match the H1 on the page with the page title, this PR has changed the H1 from `Letter` to the first line of the error message. We'd like to take some time to answer this question, therefore, leaving the page titles as is for now. 

## Before
Screenshot 2020-01-09 at 10 27 39

# After
### Letter outside content area
Screenshot 2020-01-10 at 10 10 02

### Letter that is missing EOF tag
Screenshot 2020-01-10 at 10 05 22

### Cancelled letters now show message using "govuk-warning-text"
Screenshot 2020-01-10 at 10 04 13
 
### Added a new error message for a letter that is missing an address block.
Screenshot 2020-01-10 at 10 14 56

Pair: becca&karl

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -566,14 +566,14 @@ def get_letter_printing_statement(status, created_at):
 
 LETTER_VALIDATION_MESSAGES = {
     'letter-not-a4-portrait-oriented': {
-        'title': 'We cannot print your letter',
-        'detail': 'Your letter is not A4 portrait size on {invalid_pages} <br>'
+        'title': 'Your letter is not A4 portrait size',
+        'detail': 'You need to change the size or orientation of {invalid_pages} <br>'
                   'Files must meet our <a href="https://docs.notifications.service.gov.uk/documentation/images/'
                   'notify-pdf-letter-spec-v2.4.pdf" target="_blank">letter specification</a>.'
     },
     'content-outside-printable-area': {
-        'title': 'We cannot print your letter',
-        'detail': 'The content appears outside the printable area on {invalid_pages}.<br>'
+        'title': 'Your content is outside the printable area',
+        'detail': 'You need to edit {invalid_pages}.<br>'
                   'Files must meet our <a href="https://docs.notifications.service.gov.uk/documentation/images/'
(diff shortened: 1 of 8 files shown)


app/utils.py
```diff
@@ -566,14 +566,14 @@ def get_letter_printing_statement(status, created_at):
...
+        'title': 'Your letter is not A4 portrait size',
+        'detail': 'You need to change the size or orientation of {invalid_pages} <br>'
                   'Files must meet our <a href="https://docs.notifications.service.gov.uk/documentation/images/'
                   'notify-pdf-letter-spec-v2.4.pdf" target="_blank">letter specification</a>.'
     },
     'content-outside-printable-area': {
-        'title': 'We cannot print your letter',
-        'detail': 'The content appears outside the printable area on {invalid_pages}.<br>'
+        'title': 'Your content is outside the printable area',
+        'detail': 'You need to edit {invalid_pages}.<br>'
```
klssmith: We don't show the title for letters sent through the api (only for uploaded letters), so for letters sent through the api the error message has changed from `Validation failed. The content appears outside the printable area on page 1.` to `Validation failed. You need to edit page 1.` which doesn't explain what the issue is any more.

Not sure if that's an issue or not since we still show the overlay, but it might be less clear what's wrong with the letter.

servingUpAces: It's worth taking sometime to review the overlay page for api uploads.

servingUpAces:
I've created a new PR to add the new content changes, https://github.com/alphagov/notifications-admin/pull/new/new-no-address-message without changing the existing format. It seems a longer conversation is required before implementing a change to the validation errors on precompiled letters uploaded by the API.

quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/3254

PR #3250, opened 2020-01-09
Fix exception in track-error js module that broke our re-upload button

CrystalPea:
Basically since we disabled google analytics, and we run `track-error` module in strict js, it caused an exception. Luckily it's a one-line fix (but it took us a while to find what's actually wrong)

Thanks @tombye for help with this 💖

### DIFF
--- a/app/assets/javascripts/errorTracking.js
+++ b/app/assets/javascripts/errorTracking.js
@@ -5,7 +5,7 @@
 
     this.start = function(component) {
 
-      if (!ga) return;
+      if (!('ga' in window)) return;
 
       ga(
         'send',


quis:
If you remove this it will mean we stop recording what kinds of errors users are seeing in Google Analytics. Which is sometimes useful. Would be good to understand why it’s not working…

CrystalPea:
@quis we fixed it now with @tombye , so we are not getting rid of it ☺️

tombye:
The exception was caused by the combination of the script running in strict mode and the condition testing the `ga` variable.

Strict mode forces an error when global variables are used that weren't previously declared. The analytics code we removed declared the `window.ga` property so `if(!ga)` passed. Without that, the condition was testing an undeclared global variable so broke.

This fixes the condition by avoiding any global variables and, instead just checking for a property on `window`.

PR #3251, opened 2020-01-09
Make models for individual jobs and collections of jobs

quis:
This follows the pattern of what we’ve done with services, users and events.

It gives us a better interface to the data we get back from the API than dealing with the raw JSON directly.

Now is a good time to do this because we’re going to be making a bunch of changes to the jobs pages, and those changes will be easier to code and understand with a sensible model behind them.

***

There are lots of changes here, but very few that change the tests. So assuming the existing tests are good the changes should at least work…

### DIFF
--- a/app/models/job.py
+++ b/app/models/job.py
@@ -1,0 +1,159 @@
+from datetime import datetime
+
+from notifications_utils.letter_timings import (
+    CANCELLABLE_JOB_LETTER_STATUSES,
+    get_letter_timings,
+    letter_can_be_cancelled,
+)
+from werkzeug.utils import cached_property
+
+from app.models import JSONModel
+from app.notify_client.job_api_client import job_api_client
+from app.notify_client.notification_api_client import notification_api_client
+from app.notify_client.service_api_client import service_api_client
+from app.utils import set_status_filters
+
+
+class Job(JSONModel):
+
+    ALLOWED_PROPERTIES = {
+        'id',
+        'service',
+        'template',
+        'template_version',
+        'original_file_name',
+        'created_at',
+        'notification_count',
+        'notifications_sent',
+        'notifications_requested',
+        'job_status',
+        'statistics',
+        'created_by',
+        'scheduled_for',
+    }
+
+    @classmethod
+    def from_id(cls, job_id, service_id):
+        return cls(job_api_client.get_job(service_id, job_id)['data'])
+
+    @property
+    def status(self):
+        return self.job_status
+
+    @property
+    def cancelled(self):
+        return self.status == 'cancelled'
+
+    @property
+    def scheduled(self):
+        return self.status == 'scheduled'
+
+    @property
+    def notification_count(self):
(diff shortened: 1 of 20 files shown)


app/models/job.py
```diff
@@ -0,0 +1,210 @@
...
+            return self.template['postage']
+
+    @property
+    def letter_timings(self):
+        return get_letter_timings(self.created_at, postage=self.postage)
+
+    @property
+    def failure_rate(self):
+        if not self.notifications_delivered:
+            return 100 if self.notifications_failed else 0
```
CrystalPea: this sounds like if the notifications are still in sending it would show 100% failure rate? Is this intended behaviour?

app/models/job.py
```diff
@@ -0,0 +1,210 @@
...
+    model = Job
+
+
+class ScheduledJobs(ImmediateJobs):
+    client = job_api_client.get_scheduled_jobs
+
+
+class PaginatedJobs(ImmediateJobs):
+
+    client = job_api_client.get_page_of_jobs
```
CrystalPea: it feels like `get_page_of_results` might be a better variable name here than `client`, since the variable does not contain the whole client.

app/models/job.py
quis: It means that:

Notifications delivered | Notifications failed | Failure rate
--|--|--
0 | Any | 100%
0 | 0 | 0%
Any | Any | `failed / delivered * (failed + delivered) * 100`%

Without this extra check you’d get divide by zero errors when trying to calculate the failure rate.

quis: (this is the existing behaviour)

app/models/job.py
quis: It has to be `client` to match what’s defined here:

https://github.com/alphagov/notifications-admin/blob/861a29ddd3d50f5830ef114d6662180dc637dc33/app/models/__init__.py#L56-L59

Maybe `client_method` would be a better property name though? What do you think?

CrystalPea: yeah, `client_method` sounds good 👍

app/models/job.py
```diff
@@ -0,0 +1,210 @@
...
+    @property
+    def notifications_requested(self):
+        return self._aggregate_statistics()
+
+    @property
+    def notifications_sent(self):
+        return self.notifications_delivered + self.notifications_failed
+
+    @property
+    def notifications_processed(self):
```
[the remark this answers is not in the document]
quis: Good spot, removed `Job.notifications_processed` in cbf0d90

app/models/job.py
quis: Cool, I’ll do that as a separate PR

app/models/job.py
CrystalPea: Ahh, cool, thanks!

app/models/job.py
CrystalPea: Cheers! :)

PR #3253, opened 2020-01-10
Added a new error message when the letter is missing an address block.

servingUpAces:
Before deploying a change to template-preview to return a validation error for letters that are missing the address block, we need to add the new erorr message to admin.

Some content changes have been made to other messages.
The format of the message has changed.

Here is an example of the new error message. 

Screenshot 2020-01-10 at 14 42 49


quis:
Merging this now to unblock other work. Will revise design shortly…

PR #3254, opened 2020-01-10
Add separate messages for precompiled letters that have failed validation

quis:
We show letter validation errors in two places:

1. **In response to a user uploading a PDF**
   Here we use the error banner pattern because the problem is as a direct consequence of a user’s action, and is blocking them from continuing.

2. **Once a PDF provided through the API has been validated**
   We use a less prominent pattern of red text with no border because the message is reporting on something that’s already happened, and which wasn’t a direct consequence of the user clicking something

Because the context and patterns used are different we need slightly different content in each of these situations. Previously we tried to reuse the same content to make the code cleaner and less repetitive. But ultimately a clear interface trumps clear code.


app/utils.py
```diff
@@ -569,32 +569,61 @@ def get_letter_printing_statement(status, created_at):
...
     },
     'unable-to-read-the-file': {
         'title': 'There’s a problem with your file',
-        'detail': 'Notify cannot read this PDF.<br>Save a new copy of your file and try again.'
+        'detail': (
+            'Notify cannot read this PDF.'
+            '<br>Save a new copy of your file and try again.'
+        ),
+        'summary': (
+            'Letters must be 10 pages or less. <br>'
```
idavidmcdonald: Not sure if this summary matches the detail?

tests/app/main/views/test_index.py
```diff
@@ -254,3 +254,21 @@ def test_letter_template_preview_headers(
...
+    expected_url = (
+        'https://docs.notifications.service.gov.uk'
+        '/documentation/images/notify-pdf-letter-spec-v2.4.pdf'
+    )
+    client_request.get(
+        'main.letter_spec',
+        _expected_status=302,
+        _expected_redirect=expected_url,
+    )
+    client_request.logout()
```
idavidmcdonald: Optional: Split this into a separate test explicitely called `test_letter_spec_redirect_can_be_accessed_with_non_logged_in_user` or similar

app/utils.py
```diff
@@ -603,6 +632,8 @@ def get_letter_validation_error(validation_message, invalid_pages=None, page_cou
     if validation_message not in LETTER_VALIDATION_MESSAGES:
         return {'title': 'Validation failed'}
 
+    invalid_pages_are_or_is = 'is' if len(invalid_pages) == 1 else 'are'
```
idavidmcdonald: I don't think we have a test for this behaviour. Could we add one please

tests/app/main/views/test_index.py
quis: Done in bc7deebcc

app/utils.py
quis: Added in 1fc0f5854

app/utils.py
quis: Fixed in 34f209a08

PR #3258, opened 2020-01-14
Update footer and get started

karlchillmaid:
This PR updates the _Get started_ page and the Notify footer.

**Get started page**
- update list of organisations that can use Notify to remove 'housing associations'
- update the wording of 'companies owned by local or central government...'

**Footer**
- move 'Performance' to the Support column and rename it, 'Performance data'

### DIFF
--- a/app/templates/views/get-started.html
+++ b/app/templates/views/get-started.html
@@ -24,9 +24,8 @@
             <li>central government departments</li>
             <li>local authorities</li>
             <li>state-funded schools</li>
-            <li>housing associations</li>
             <li>the NHS</li>
-            <li>companies owned by local or central government that deliver services on their behalf</li>
+            <li>companies running a service on behalf of a public sector organisation</li>
           </ul>
           <p>Notify is not currently available to charities.</p>
         </div>'''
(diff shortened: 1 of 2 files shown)


karlchillmaid:
Pete and I discussed having 'Performance' in *both* the Support and About columns – using different link text to emphasise different aspects of the service. Suggest we do not do this for now - unless we identify a need.

Support seems like a more natural home for the performance data.

app/templates/views/get-started.html
```diff
@@ -24,9 +24,8 @@ <h2 class="get-started-list__heading">Check if GOV.UK Notify is right for you</h
             <li>central government departments</li>
             <li>local authorities</li>
             <li>state-funded schools</li>
-            <li>housing associations</li>
             <li>the NHS</li>
-            <li>companies owned by local or central government that deliver services on their behalf</li>
+            <li>companies running a service on behalf of a public sector organisation</li>
```
karlchillmaid: In future we might want to include a link to contact us if you want to discuss a specific set of circumstances.

PR #3263, opened 2020-01-16
Use time to determine why notifications don’t exist

quis:
Notifications won’t exist for a job if:
- it’s just started
- it started a long time ago (older than the retention period)

We have a bug where:
1. Job starts processing, puts notifications on queue
2. Job finishes processing, sets status to `finished`
3. First notification gets picked up off the queue and put in the    database

In between 2. and 3. it’s possible for a job to be finished, but also to have no notifications. We’re giving the user a message saying this is because the notifications have been deleted, whereas really it’s because they haven’t been created yet.

This commit fixes that bug by introducing the concept of recency for jobs.

‘Recent’ is defined as 1 day, which is:
- a lot longer than it takes to create any notifications
- a lot shorter than anyone’s retention time

N.B. `processing_started` is defined here:
https://github.com/alphagov/notifications-api/blob/879ba1d5f092dbf6e15bf9adadd53e4657660490/app/models.py#L1194

It can be `None` for scheduled jobs that haven’t started yet.


app/models/job.py
```diff
@@ -48,6 +48,12 @@ def scheduled(self):
     def scheduled_for(self):
         return self._dict.get('scheduled_for')
 
+    @property
+    def processing_started(self):
+        if not self._dict.get('processing_started'):
+            return None
+        return datetime.strptime(self._dict['processing_started'][:-6], '%Y-%m-%dT%H:%M:%S')
```
idavidmcdonald: Optional:

I found us doing `strptime` elsewhere in this codebase (including once in this file itself). Most seem to include the milliseconds instead of not parsing them e.g. `%Y-%m-%dT%H:%M:%S.%f'`. Is there a particular reason we want to get rid of the milliseconds here?

Also if there isn't a particular reason and we are doing this in a few places, I'd be tempted to move `%Y-%m-%dT%H:%M:%S.%f` into a common constant to be used in all those places.

app/models/job.py
```diff
@@ -95,6 +101,17 @@ def still_processing(self):
     def finished_processing(self):
         return self.notification_count == self.notifications_sent
 
+    @property
+    def recently_created(self):
+        if not self.processing_started:
```
idavidmcdonald: Are you aware that in production we have some records (roughly 20 I think) that do not have processing started values but were created a long time ago (2016). They appear to all be testing services, apart from maybe one https://www.notifications.service.gov.uk/services/0acc1677-99c2-4866-a0aa-b5eb6dc62d1a

I'm guessing though it won't cause any issues because people aren't going to be looking at this data for old jobs from 2016?

app/models/job.py
```diff
@@ -95,6 +101,17 @@ def still_processing(self):
     def finished_processing(self):
         return self.notification_count == self.notifications_sent
 
+    @property
+    def recently_created(self):
```
idavidmcdonald: Optional: I'm not super keen on this name. This name to me says it would just be looking at the `created_at` field to see if it's been created recently. Technically this method isn't doing that, whilst its behaviour would be similar, it's actually different and maybe therefore a bit misleading.

Saying that, I can't think of anything that is short and succinct and better at describing this function so if you can't either then I'm fine with keeping it as is.

The best I can come up with is:
`is_awaiting_processing_or_processed_recently` or `not_yet_processed_or_processed_in_last_day` but not sure if I'm keen on either of those.

tests/app/main/views/test_jobs.py
```diff
@@ -329,28 +330,58 @@ def test_should_show_job_without_notifications(
...
 
 
+@freeze_time("2020-01-10 0:0:0")
+@pytest.mark.parametrize('created_at, processing_started, expected_message', (
+    # Recently created, not yet started
+    (datetime(2020, 1, 10, 0, 0, 0), None, (
+        'No messages to show yet…'
+    )),
+    # Just started
+    (datetime(2020, 1, 10, 0, 0, 0), datetime(2020, 1, 10, 0, 0, 1), (
```
idavidmcdonald: This test is a bit weird in the sense that the processing_started is in the future given the time is frozen at 00:00:00. Whilst it's probably not causing any errors it feels like one worth changing just to remove a risk of bugs because of it.

tests/app/main/views/test_jobs.py
```diff
@@ -329,28 +330,58 @@ def test_should_show_job_without_notifications(
...
+        'No messages to show yet…'
+    )),
+    # Created a while ago, just started
+    (datetime(2020, 1, 1, 0, 0, 0), datetime(2020, 1, 10, 0, 0, 1), (
```
idavidmcdonald: Processing at in the future again

tests/app/main/views/test_jobs.py
```diff
@@ -329,28 +330,58 @@ def test_should_show_job_without_notifications(
...
+        'No messages to show yet…'
+    )),
+    # Created a while ago, started a couple of days ago
```
idavidmcdonald: As 1 day is what the code is defining as the swap over condition, I feel like we should test either side of that boundary rather than a few days before it, if that makes sense?

tests/app/main/views/test_jobs.py
```diff
@@ -329,28 +330,58 @@ def test_should_show_job_without_notifications(
...
+    (datetime(2020, 1, 1, 0, 0, 0), datetime(2020, 1, 8, 0, 0, 1), (
+        'These messages have been deleted because they were sent more than 7 days ago'
```
idavidmcdonald: Do we actually delete these messages? As in, technically the messages were only sent 2 days ago, therefore should they not be kept around for another 5 days?

app/models/job.py
quis: Looks like we already have a function in utils to do this conversion. I’ve refactored to use it in d93866bc7

app/models/job.py
quis: Happy with not fixing the data for these old jobs.

tests/app/main/views/test_jobs.py
quis: Technically you wouldn’t see either message in that case, because you’d get some notifications back from the API.

app/models/job.py
quis: Changed to `awaiting_processing_or_recently_processed` in f6a263a7b

tests/app/main/views/test_jobs.py
quis: Changed the frozen time in f5ced76e6

tests/app/main/views/test_jobs.py
quis: Changed the frozen time in f5ced76e6

tests/app/main/views/test_jobs.py
quis: Updated the tests to look at the boundary conditions in d27d400f5

tests/app/main/views/test_jobs.py
quis: Comment to clarify added in a86d2e15a

PR #3266, opened 2020-01-20
Don’t assume a job’s status will be present

quis:
The API response for jobs includes a field called `job_status`. The API response for uploads doesn’t.

The `Job` model handles uploads and jobs, so it needs to account for the possibility of the field not being there.

### DIFF
--- a/app/models/job.py
+++ b/app/models/job.py
@@ -34,7 +34,7 @@ class Job(JSONModel):
 
     @property
     def status(self):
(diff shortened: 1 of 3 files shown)


klssmith:
Is `job_status` still needed in `ALLOWED_PROPERTIES` for the Job model?

quis:
@klssmith good point, removed in 0202f73f9

PR #3268, opened 2020-01-21
Parametrise branding request flow so it serves both email and letter branding

CrystalPea:
This is done so letter branding requests create a specialised Zendesk ticket, not a generic feedback ticket.

Story ticket: https://www.pivotaltracker.com/story/show/170626562

## Before:
Screenshot 2020-01-21 at 12 21 24
Screenshot 2020-01-21 at 12 21 35

### When there are no existing options:
Screenshot 2020-01-21 at 12 20 38

### When there are options to choose from:
Screenshot 2020-01-21 at 12 21 01


app/main/forms.py
```diff
@@ -1392,35 +1392,43 @@ def __init__(self, *args, **kwargs):
...
 
-    options = RadioField('Choose your new email branding')
+    options = RadioField('Choose your new branding')
     something_else = TextAreaField('Describe the branding you want')
 
-    def __init__(self, service, *args, **kwargs):
+    def __init__(self, service, *args, branding_type="email", **kwargs):
         super().__init__(*args, **kwargs)
-        self.options.choices = tuple(self.get_available_choices(service))
+        self.options.choices = tuple(self.get_available_choices(service, branding_type))
```
quis: You could keep the label text for the email choices the same by setting it in the constructor:
```suggestion
        self.options.choices = tuple(self.get_available_choices(service, branding_type))
        self.options.label.text = 'Choose your new {} branding'.format(branding_type)
```

app/main/forms.py
```diff
@@ -1392,35 +1392,44 @@ def __init__(self, *args, **kwargs):
...
+            organisation_branding_id = service.organisation.letter_branding_id if service.organisation else None
+            service_branding_id = service.letter_branding_id
+            service_branding_name = service.letter_branding_name
 
         if (
             service.organisation_type == Organisation.TYPE_CENTRAL and
-            service.organisation.email_branding_id is None and
-            service.email_branding_id is not None
+            organisation_branding_id is None and
+            service_branding_id is not None
```
quis: GOV.UK needs to only be available as email branding:
```suggestion
            service_branding_id is not None and 
            branding_type == "email"
```

app/main/forms.py
```diff
@@ -1392,35 +1392,44 @@ def __init__(self, *args, **kwargs):
...
         ):
             yield ('govuk', 'GOV.UK')
 
         if (
             service.organisation_type == Organisation.TYPE_CENTRAL and
             service.organisation and
-            service.organisation.email_branding_id is None and
-            service.email_branding_name.lower() != 'GOV.UK and {}'.format(service.organisation.name).lower()
+            organisation_branding_id is None and
+            service_branding_name.lower() != 'GOV.UK and {}'.format(service.organisation.name).lower()
```
quis: GOV.UK and organisation should only be available as email branding
```suggestion
            service_branding_name.lower() != 'GOV.UK and {}'.format(service.organisation.name).lower() and 
            branding_type == "email"
```

CrystalPea:
@quis ahh, good catch with GOV.UK branding. Fixed now 👍. As to splitting BrandingOption into two classes, I can't see a neat way to do that. Do you have a vision for that? :)

tests/app/main/views/test_service_settings.py
```diff
@@ -4819,6 +4852,43 @@ def test_submit_letter_branding_request(
...
+            '.branding_request', service_id=SERVICE_ONE_ID, branding_type="letter", from_template=from_template,
+            _data=data,
+        )
+    else:
+        page = client_request.post(
+            '.branding_request', service_id=SERVICE_ONE_ID, branding_type="letter",
+            _data=data,
+        )
+
+    assert page.find('a').attrs['href'] == expected_redirect
```
quis: Not sure this is testing what you think it is. The way we usually do this is by passing  the `_expected_redirect` argument to `client_request.post`. Here’s an example:
https://github.com/alphagov/notifications-admin/blob/ccd245b287d53ef18f79b82c0037cdfb825b68fb/tests/app/main/views/test_user_profile.py#L42-L47

CrystalPea: I'm pretty sure this test tested what I wanted to, but what you proposed is more elegant, so I changed it now :)

PR #3277, opened 2020-01-28
bump utils to fix datetime formatting jinja bug

leohemsted:
see alphagov/notifications-utils#696


tombye:
Are the changes this brings in described in: https://github.com/alphagov/notifications-utils/pull/696 ?

If so, we should add this to the description.

PR #3278, opened 2020-01-28
Refactor how we handle datetimes in model classes

quis:


quis:
Going to close this until we really need it. I don’t have time to unpick the mess I’ve made.

PR #3280, opened 2020-02-03
Swap order of usage and team members in navigation when organisation users are looking at a service

quis:
▩ | Organisation | Service 
---|---|---
Before | image |  image
After | image | image

Organisation users, when looking at the page for their org, see:
> Usage
> Team members

When they click into a service it switches to:
> Team members
> Usage

This is jarring. It should stay consistent. I think it that _Usage_ then _Team members_ is the natural way of ordering the navigation at the organisation level, so let’s follow that through to the service level.

This does mean that if someone is a member of both an organisation and a service that the nav will jump (because it’ll switch to the existing, service-level order of _Team members_ then _Usage_) but it’s going to jump anyway because you get all the extra navigation items when you’re a member of a service.

### DIFF
--- a/app/templates/main_nav.html
+++ b/app/templates/main_nav.html
@@ -3,10 +3,10 @@
 {% else %}
 <nav class="navigation">
   <ul>
-  {% if current_user.has_permissions('view_activity') %}
-    <li><a href="{{ url_for('.service_dashboard', service_id=current_service.id) }}" {{ main_navigation.is_selected('dashboard') }}>Dashboard</a></li>
-  {% endif %}
   {% if current_user.has_permissions() %}
+    {% if current_user.has_permissions('view_activity') %}
+      <li><a href="{{ url_for('.service_dashboard', service_id=current_service.id) }}" {{ main_navigation.is_selected('dashboard') }}>Dashboard</a></li>
+    {% endif %}
     <li><a href="{{ url_for('.choose_template', service_id=current_service.id) }}" {{ main_navigation.is_selected('templates') }}>Templates</a></li>
     {% if current_user.has_permissions('view_activity') %}
       {% if current_service.can_upload_letters %}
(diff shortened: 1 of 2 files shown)


app/templates/main_nav.html
```diff
@@ -18,16 +18,19 @@
         <li><a href="{{ url_for('main.uploads', service_id=current_service.id) }}" {{ casework_navigation.is_selected('uploads') }}>Uploads</a></li>
       {% endif %}
     {% endif %}
-  {% endif %}
     <li><a href="{{ url_for('.manage_users', service_id=current_service.id) }}" {{ main_navigation.is_selected('team-members') }}>Team members</a></li>
-  {% if current_user.has_permissions('manage_service', allow_org_user=True) %}
+    {% if current_user.has_permissions('manage_service', allow_org_user=True) %}
```
idavidmcdonald: Do we need the `allow_org_user=True` on this?

quis: No, it’s effectively unreachable code
```suggestion
    {% if current_user.has_permissions('manage_service') %}
```

idavidmcdonald: 👍

quis: Actually, the above comment is wrong. We do need this check because:
- if you have permission to see the service
- but are also a user of the organisation
- but don’t have the `manage_service` permission
- you should always be able to see the usage page

idavidmcdonald: I'm a little confused. If they don't have the `manage_service` permission though then this whole statement will be false anyway so they won't see the Usage page?

I could see it adding users who have BOTH the manage_service permission and are also org users (if that's possible, I'm very unfamiliar with permissions stuff) but not one who doesn't have the`manage_service` permission?

I'm going to have a look now to see if removing the allow_org_user flag in this case breaks a test

quis: I’ve added an explicit test for this in b7b9e97. The test fails if the above suggestion is applied:

image

idavidmcdonald: Oh cool, you were already thinking along the same lines. I'll take another review

tests/app/main/test_permissions.py
```diff
@@ -373,6 +373,45 @@ def test_service_navigation_for_org_user(
     assert [
         item.text.strip() for item in page.select('nav.navigation a')
     ] == [
+        'Usage',
+        'Team members',
+    ]
+
+
+def test_org_users_can_always_see_usage_page(
```
idavidmcdonald: Could we rename this test to be a more explicit about what it's testing please.

From what I've understood of this so far, I think something like `test_org_user_who_is_also_service_user_sees_correct_navigation`? As the difference between this test and the one above is that they are a member of the service, right?

quis: I think it’s good to call out that this is explicitly about the usage page, since it’s the only special one.
```suggestion
def test_org_user_who_is_also_service_user_can_still_see_usage_page(
```

quis: Addressed in cceb731

PR #3282, opened 2020-02-03
Ensure that the session is logged out server side, not just client side.

servingUpAces:
Anytime a user clicks "sign out" we should be signing them out server side as well. This can be accomplished by setting the Users.current_session_id = null.
I found that the method User.logged_in_elsewhere doesn't need to check if the current_session_id is None. The current_session_ids in the cookie and db (redis or postgres) then the user should be forced to log in again.

### DIFF
--- a/app/main/views/sign_out.py
+++ b/app/main/views/sign_out.py
@@ -1,5 +1,5 @@
-from flask import redirect, session, url_for
-from flask_login import current_user, logout_user
+from flask import redirect, url_for
+from flask_login import current_user
 
 from app.main import main
 
@@ -7,6 +7,4 @@ from app.main import main
 @main.route('/sign-out', methods=(['GET']))
 def sign_out():
     current_user.sign_out()
-    session.clear()
-    logout_user()
     return redirect(url_for('main.index'))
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_sign_in.py
```diff
@@ -60,8 +60,7 @@ def test_doesnt_redirect_to_sign_in_if_no_session_info(
 
 
 @pytest.mark.parametrize('db_sess_id, cookie_sess_id', [
-    pytest.param(None, None, marks=pytest.mark.xfail),  # OK - not used notify since browser signout was implemented
-
+    (None, None),
```
leohemsted: can you add one with `None, uuid.UUID(int=1)` for when the user was logged out but their browser still has a cookie set.

app/main/views/sign_out.py
```diff
@@ -1,11 +1,12 @@
...
+from flask_login import current_user, logout_user
 
 from app.main import main
 
 
 @main.route('/sign-out', methods=(['GET']))
 def sign_out():
+    current_user.sign_out()
     session.clear()
     logout_user()
```
quis: Can we move these two lines inside `User.sign_out` so if it’s ever called from somewhere else we do all three things (cookie, session, and database) required to sign a user out.

PR #3288, opened 2020-02-05
Delete template dialog

servingUpAces:
Change the content for the delete template confirmation dialog. …
ca930e6
As per https://www.pivotaltracker.com/story/show/170796514 we want to make the delete template confirmation dialog box more consistent and clear.
The API has been updated with a new endpoint that only returns the last-used date, this date is more accurate since it goes to the ft_notification_status table, if the notification table is empty.

- [x] depends on https://github.com/alphagov/notifications-api/pull/2708


tests/app/notify_client/test_template_statistics_client.py
```diff
@@ -22,11 +22,11 @@ def test_template_statistics_client_calls_correct_api_endpoint_for_service(mocke
 def test_template_statistics_client_calls_correct_api_endpoint_for_template(mocker, api_user_active):
```
CrystalPea: should the test name maybe change now that method name has changed?

PR #3289, opened 2020-02-06
Change callout to inset text

karlchillmaid:
This PR changes the name of the callout component in the Formatting section of the Edit template page.

**Why 'inset text'?**
The Design System calls this component [inset text](https://design-system.service.gov.uk/components/inset-text/).

'Inset' is a more descriptive name.

**Why not 'callout'?**
Callout seems to refer to other types of formatting. Wikipedia's description of a callout is:

>a short string of text connected by a line, arrow, or similar graphic to a feature of an illustration or technical drawing, and giving information about that feature. The term is also used to describe a short piece of text set in larger type than the rest of the page and intended to attract attention.

Neither of these definitions match this component.

**Things to consider**
Do we have evidence that our users:

* call this component a 'callout' (for any reason other than that's what we call it)?
* would not understand the term 'inset text'?

### DIFF
--- a/app/templates/partials/templates/guidance-formatting.html
+++ b/app/templates/partials/templates/guidance-formatting.html
@@ -18,7 +18,7 @@
   </p>
 </div>
 <p class="bottom-gutter-1-3">
-  To inset text, use a caret:
+  To add inset text, use a caret:
 </p>
 <div class="panel panel-border-wide bottom-gutter">
   <p>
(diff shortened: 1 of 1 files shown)


karlchillmaid:
Thanks @idavidmcdonald.

I'm going to check these changes with Pete and Chris next when they're back before merging.

PR #3290, opened 2020-02-06
Update buttons to use govuk-frontend component

klssmith:
Best reviewed commit by commit, but a few other things to note:

* I've deleted a route and 2 templates we weren't using here 467911f - would be good to check these definitely can be deleted when reviewing(!)
* Lots of the templates import unused components, so I've deleted some of these imports when I was changing a template anyway
* I've deleted a section of code that wasn't ever being reached here https://github.com/alphagov/notifications-admin/commit/05ef9ff4460d429d5c4765c762760778998497e5#diff-1e2264d99011a8198ca49e3137f85b2eL42-L46, but good to double check this is fine to delete

Everything _should_ function in the same way, and look the same 🤞

[Pivotal story](https://www.pivotaltracker.com/story/show/169345904)


app/assets/stylesheets/components/list-entry.scss
```diff
@@ -6,9 +6,9 @@
     position: relative;
   }
 
-  .list-entry-remove {
-    @include core-19;
+  &__button--remove {
     display: block;
+    width: auto;
```
tombye: We could probably move this and the `width: auto` for `&__button--add` into their own block, headed by both selectors, like:

```javascript
&__button--remove,
&__button--add {
  width: auto;
}
```

app/assets/javascripts/fileUpload.js
```diff
@@ -6,13 +6,15 @@
...
     this.showCancelButton = () => $('.file-upload-button', this.$form).replaceWith(`
-      <a href="" class='file-upload-button-cancel'>Cancel upload</a>
+      <a href="" class='govuk-button govuk-button--warning govuk-!-margin-right-1'>Cancel upload</a>
     `);
 
     this.start = function(component) {
 
       this.$form = $(component);
 
+      this.$form.find('label.file-upload-button').addClass('govuk-button govuk-!-margin-right-1');
```
tombye: This should probably have a comment above explaining that the label gets styled like a button to hide the native file upload control and give users a button consistent with those elsewhere.

app/assets/stylesheets/views/product-page.scss
```diff
@@ -1,3 +1,9 @@
+$product-page-blue: #005EA5;
+// Because the shadow (s0) is visually 'part of' the button, we need to reduce
```
tombye: I know we took these comments from the GOVUK Frontend code to help understanding but they don't seem to add anything here and reference content we don't have.

Would be better to link to the original (https://github.com/alphagov/govuk-frontend/blob/v2.13.0/src/components/button/_button.scss#L24) which better explains why this variable exists.

tombye:
I found an issue with the button that sits alongside the searchbox on the `/services/<service_id>/notifications` and other pages:

search_button_when_active

It jumps when the button is clicked because of `top: 2px` being applied by the GOVUK Frontend styles. It's to push the button down visually so it looks like its being depressed. The problem here is these buttons already have `top` set, to push them down level with the searchbox, so this overrides that value and makes them jump.

This searchbox and button component has a few existing problems at various states/screen sizes so its not worth trying to fix all of it here. Instead I'd suggest setting `top` when active to the same position it has by default, like:

```css
.search-form__button {
  @include media(desktop) {
    position: relative;
    top: 32px;
    ...

    &:active {
      top: 32px;
    }
  }
}
```

app/assets/stylesheets/views/product-page.scss
klssmith: Good idea, I've taken out that comment and linked to the original

app/assets/javascripts/fileUpload.js
klssmith: 👍Added a comment

app/assets/stylesheets/components/checkboxes.scss
```diff
@@ -83,12 +75,13 @@
...
+}
 
-  .button-secondary {
-    // revert full-width button for smaller screens
-    display: inline-block;
-    width: auto;
-  }
+.selection-footer__button {
+  // revert full-width button for smaller screens
+  clear: both;
```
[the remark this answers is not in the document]
klssmith: I've removed the `clear: both;` and it all looks and works the same

app/templates/views/choose-account.html
```diff
@@ -39,11 +40,6 @@ <h2>
   {% else %}
     </ul>
   {% endif %}
-  {% if show_add_service_button %}
```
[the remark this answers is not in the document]
klssmith: 👍I'll keep that code removed for now since it was never showing, but will check if it was meant to be there and can add something back if it was.

app/assets/stylesheets/components/list-entry.scss
klssmith: I've made that change

klssmith:
Think I've addressed all the comments now. The Search box button no longer jumps around, but the bottom of the box does look very slightly different, so not sure if we need to fix that

Before:
Screenshot 2020-02-10 at 15 33 28

After:
Screenshot 2020-02-10 at 15 33 33

PR #3295, opened 2020-02-12
Restyle template statistics and received text messages

quis:
We want to add more things to the dashboard, but we can’t reuse the ‘received text messages’ banner because it has too much visual weight.

So we need to add a new style of banner with less visual weight, and restyle the template statistics section of the dashboard so it still feels visually coherent.

Before | After
---|---
image | image

### DIFF
--- a/app/assets/stylesheets/components/banner.scss
+++ b/app/assets/stylesheets/components/banner.scss
@@ -142,3 +142,47 @@
   }
 
 }
+
+.banner-dashboard {
+
+  $baseline-shift: -5px;
+
+  display: block;
+  position: relative;
+  padding: $gutter-two-thirds 0;
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+  margin-bottom: $gutter;
+  text-decoration: none;
+
+  &:focus {
+    border-color: $focus-colour;
+    box-shadow: 0 3px 0 0 $focus-colour, 0 -3px 0 0 $focus-colour;
+    color: $text-colour;
+  }
+
+  &-count {
+    @include bold-36;
+    padding-right: 5px;
+  }
+
+  &-count-label {
+    @include bold-24;
+    text-decoration: underline;
+    position: relative;
+    top: $baseline-shift;
+  }
+
+  &-meta {
+    @include core-19;
+    position: absolute;
+    right: 0;
+    bottom: $gutter-two-thirds - $baseline-shift;
+    text-align: right;
+  }
+
+  & + .banner-dashboard {
+    margin-top: -$gutter;
+    border-top: none;
+  }
+}
--- a/app/assets/stylesheets/components/banner.scss
+++ b/app/assets/stylesheets/components/banner.scss
@@ -151,7 +151,7 @@
   display: flex;
   align-items: baseline;
   flex-wrap: wrap;
-  padding: ($gutter-two-thirds - 1px) 0 ($gutter-two-thirds + 1px) 0;
+  padding: ($gutter-half - 1px) 0 ($gutter-half + 1px) 0;
   border-top: 1px solid $border-colour;
   border-bottom: 1px solid $border-colour;
   margin-bottom: $gutter;
@@ -181,6 +181,7 @@
     @include govuk-font(24, $weight: bold);
     text-decoration: underline;
     padding-right: govuk-spacing(6);
+    margin: 10px 0px 5px; // 10px includes 5px extra to counter the -5px margin-top on the count item
     flex: 2 1 auto;
   }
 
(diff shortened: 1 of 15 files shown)


idavidmcdonald:
image

is it intentional that font size for rows in the `in the next few days` table is smaller than rows in the template table?

Also what's your thinking behind dropping the column heading for `Messages sent`?
And is it OK to have the numbers in the messages sent not be all in the same alignment meaning you can't scan down a list of them as easily?

quis:
> is it intentional that font size for rows in the in the next few days table is smaller than rows in the template table?

The _Next few days_ section is going to be replaced by a banner saying how many files you have queued up so it’s OK if it doesn’t look exactly right for now – it’s an interim state

***

> Also what's your thinking behind dropping the column heading for Messages sent?

Looked a bit neater without it. [Was considering](https://github.com/alphagov/notify-design-archive/issues/15#issuecomment-585212037) losing the column heading for _Templates_ at one point. Will revisit and see if it really needs to go…_

***

> And is it OK to have the numbers in the messages sent not be all in the same alignment meaning you can't scan down a list of them as easily?

Think this is OK because you’re just trying to get a sense of relative volume (which the bar chart gives you) not exact comparison.

quis:
I think it looks better without the _Messages sent_ heading now that the numbers are right-aligned. With it, it looks a bit like the heading of an empty column. Compare:

image

***

image

***

I think it looks better still with no headings and no top border:

image

tombye:
What about giving the *Templates* table a visible caption of 'By template'?

It would let it sit under 'In the last 7 days in the headings hierarchy and would explain the entries.

quis:
@tombye I like it

image

quis:
@tombye updated in 84df70e

app/templates/views/dashboard/_inbox.html
```diff
@@ -2,20 +2,18 @@
...
-        )
-      }}
-      <div class="big-number-meta">
-        {% if inbound_sms_summary.most_recent %}
+    <a id="total-received" class="banner-dashboard" href="{{ url_for('.inbox', service_id=current_service.id) }}">
+      <span class="banner-dashboard-count">
+        {{ inbound_sms_summary.count|format_thousands }}
+      </span>
+      <span class="banner-dashboard-count-label">
+        text messages received
```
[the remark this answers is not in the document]
tombye: Made this change in https://github.com/alphagov/notifications-admin/pull/3295/commits/cd262559ccf0bc2a9a7065bbf4f4108dbd7ac339 so resolving this.

app/assets/stylesheets/components/banner.scss
```diff
@@ -142,3 +142,47 @@
...
+  display: block;
+  position: relative;
+  padding: $gutter-two-thirds 0;
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+  margin-bottom: $gutter;
+  text-decoration: none;
+
+  &:focus {
+    border-color: $focus-colour;
```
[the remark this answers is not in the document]
tombye: I made this change in https://github.com/alphagov/notifications-admin/pull/3295/commits/cd262559ccf0bc2a9a7065bbf4f4108dbd7ac339 so marking this as resolved.

app/assets/stylesheets/components/banner.scss
```diff
@@ -142,3 +142,56 @@
...
+  }
+
+  &-count,
+  &-meta {
+    float: left; // for browsers that don't support flexbox
+  }
+
+  &-count {
+    @include govuk-font(36, $weight: bold);
+    padding-right: 5px;
```
quis: Needs a hair more space
```suggestion
    padding-right: 8px;
```

app/assets/stylesheets/components/banner.scss
```diff
@@ -142,3 +142,56 @@
...
+
+.banner-dashboard {
+
+  $baseline-shift: 5px;
+
+  display: block; // for browsers that don't support flexbox
+  display: flex;
+  align-items: baseline;
+  flex-wrap: wrap;
+  padding: $gutter-two-thirds 0;
```
quis: Needs the spacing tweaking by a fraction to accommodate the baseline shift 
```suggestion
  padding: ($gutter-two-thirds - 1px) 0 ($gutter-two-thirds + 1px) 0;
```

app/assets/stylesheets/components/banner.scss
```diff
@@ -142,3 +142,56 @@
...
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+  margin-bottom: $gutter;
+  text-decoration: none;
+
+  &:focus {
+    border: none;
```
quis: Border size need to be consistent with the non-focused state otherwise it changes height when focused
```suggestion
    border-top: 1px solid transparent;
    border-bottom: 1px solid transparent;    
```

PR #3296, opened 2020-02-12
Add letter specification guidance

karlchillmaid:
This PR adds letter specification information to the _Upload a letter_ guidance page.

We're making these updates so that we can user test the content and downloadable documents. If they're successful we'll keep them.

### DIFF
--- a/app/templates/views/guidance/upload-a-letter.html
+++ b/app/templates/views/guidance/upload-a-letter.html
@@ -37,8 +37,8 @@
   <p>To help you set up your letter, you can download:</p>
 
   <ul class="list list-bullet">
-    <li>our <a href="{{ url_for('.letter_spec') }}">letter specification document</a></li>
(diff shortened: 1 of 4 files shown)


app/templates/views/guidance/upload-a-letter.html
```diff
@@ -21,8 +21,25 @@ <h1 class="heading-large">Upload a letter</h1>
...
+    <li>10 pages or less</li>
+    <li>smaller than 2MB</li>
+  </ul>
+  
+  <p>The content of your letter must appear inside the printable area.</p>
+  
+  <p>To help you set up your letter, you can download:</p>
+
+  <ul class="list list-bullet">
+    <li>our <a href="{{ url_for('.letter_spec') }}">letter specification document</a></li>
```
quis: Do we want to be explicit that this is a PDF file?

karlchillmaid: Yeah, I think we should be including file types in the links – for example, (PDF)

karlchillmaid:
@quis For Friday's usability testing, do we want to update the links in the validation error messages to take users to the guidance?

PR #3297, opened 2020-02-12
Put a count of recently returned letters on the dashboard

quis:
Currently there is no way of getting to the returned letters page. This commit adds a link to it from the dashboard, following the pattern of the new received text messages banner.

***

image


quis:
Content this PR adds to the returned letters summary (words by @karlchillmaid)

image

app/models/service.py
```diff
@@ -480,6 +484,12 @@ def inbound_number(self):
     def has_inbound_number(self):
         return bool(self.inbound_number)
 
+    @cached_property
```
tombye: It's probably my lack of understanding but why is this a `@cached_property`? I'd assume this could update in realtime so curious.

quis: It caches the value for the lifetime of the request. We wrap it around things that get data from external sources (the API or Redis).

It means that if you can call `current_service.inbound_sms_summary` multiple times when building the page you’ll still only make one HTTP request to the API. It’s effectively the same thing as assigning the result to a variable and passing that variable around, but a bit neater.

It’s not very comprehensively documented, but it comes from the Wekzeug project (which is one of the things underlying Flask)
 https://werkzeug.palletsprojects.com/en/1.0.x/utils/#werkzeug.utils.cached_property

Was introduced here: 780b9bb715e586d3154290bf563a78abf74586cf

tombye: Makes sense 👍

PR #3298, opened 2020-02-13
Update Using Notify pages, features and guidance

karlchillmaid:
This PR updates the content on the following Using Notify, guidance and features pages:

- `using-notify/trial-mode` – make the information about trial mode restrictions, request to go live process, and things you need to do beforehand clearer (see screenshots below)
- `using-notify/delivery-status` – make the status descriptions consistent with our API documentation
- `using-notify/get-started` – fix an anchor link
- `/pricing` – add anchor IDs to headings
- `using-notify/get-started`,  `using-notify/delivery-status` and `features/letters` – fix style inconsistencies 

# Trial mode page before
trial-mode-old

# Trial mode page after
trial-mode-updates

### DIFF
--- a/app/templates/views/features/letters.html
+++ b/app/templates/views/features/letters.html
@@ -25,7 +25,7 @@
   <h3 class="heading-small" id="postage">Choose your postage</h3>
   <p>Notify can send letters by first or second class post.</p>
   <p>First class letters are delivered one day after they’re dispatched. Second class letters are delivered 2 days after they’re dispatched.</p>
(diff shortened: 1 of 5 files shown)


karlchillmaid:
#For discussion

On the [Get started](https://www.notifications.service.gov.uk/using-notify/get-started) page, we should remove the links to:

- [the Message templates section of Features](https://www.notifications.service.gov.uk/features#templates) from step 3
- [the Permissions section of Features](https://www.notifications.service.gov.uk/features#permissions) from step 4

When signed in, it feels problematic that these links take the user to features content, rather than guidance or the relevant service page.

When signed out, these links are more useful – but possibly not essential, as step 1 already links to the Features page.

PR #3302, opened 2020-02-14
Guidance simplify numbered lists

karlchillmaid:
This PR simplifies the numbered lists of instructions in our guidance pages by:

- removing the instruction to `Sign in to Notify`
- linking directly to the page users need to visit

### DIFF
--- a/app/templates/views/guidance/branding-and-customisation.html
+++ b/app/templates/views/guidance/branding-and-customisation.html
@@ -1,6 +1,5 @@
 {% extends "content_template.html" %}
-{% from "components/table.html" import mapping_table, row, text_field, edit_field, field with context %}
-{% from "components/sub-navigation.html" import sub_navigation %}
+{% from "components/current-service-link.html" import current_service_link %}
 
 {% block per_page_title %}
   Branding and customisation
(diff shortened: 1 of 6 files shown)


app/templates/views/guidance/branding-and-customisation.html
```diff
@@ -26,9 +26,8 @@ <h2 class="heading-medium" id="email-branding">Change your email branding</h2>
   <p>To change your email branding:</p>
 
   <ol class="list list-number">
-    <li><a href="{{ url_for('.sign_in') }}">Sign in to Notify</a>.</li>
-    <li>Go to the <b class="govuk-!-font-weight-bold">Settings</b> page.</li>
-    <li>In the <b class="govuk-!-font-weight-bold">Email settings</b> section, select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Email branding</b> row.</li>
+    <li>Go to the <b class="govuk-!-font-weight-bold">Email settings</b> section of the <b class="govuk-!-font-weight-bold"><a href="{{ url_for('.settings', service_id=current_service.id) }}">Settings</a></b> page.</li>
```
tombye: All these links:

`<a href="{{ url_for('.settings', service_id=current_service.id) }}">Settings</a>`

...need to be:

`<a href="{{ url_for('main.service_settings', service_id=current_service.id) }}">Settings</a>`

quis:
@karlchillmaid I think the tests were failing because linking to pages within the current service doesn’t work if the user is signed out. I’ve attempted a fix in 46a4e1c61 🤞

karlchillmaid:
@quis do you think linking to these pages even if a user is signed out is enough of an issue for us to provide alternative content/links for signed-out users?

PR #3303, opened 2020-02-14
Convert all links to govuk frontend

tombye:
Does what it says on the can🥫, gives all our links [the styles from GOVUK Frontend](https://design-system.service.gov.uk/styles/typography/#links).

https://www.pivotaltracker.com/story/show/169318662

At its most basic this adds the `govuk-link` and `govuk-link--no-visited-state` classes to all our links. The later one stops them looking different if they've been used before.

## What the `govuk-link` and `govuk-link--no-visited-state` classes do

The new classes give links the New Transport font, some print styles and all the styling for [the various link states](https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Styling_links).

They don't set the font size or weight so you'll see [the font override classes from GOVUK Frontend](https://design-system.service.gov.uk/styles/typography/#font-override-classes) used to add those things where needed.

## Red links

image

We have links coloured red in various places so I made a `govuk-link--destructive` class to make one way of creating them. GOVUK Frontend already has [variants on the default link style](https://github.com/alphagov/govuk-frontend/blob/063cd8e2470b62b824c6e50ca66342ac7a95d2d8/src/govuk/core/_links.scss#L17) so I copied them.

I'm not 100% sure `destructive` is the right word to use (I took it from a link state in the browse-list component) so any comments gratefully received.

## Existing link CSS

I've removed all the CSS used to style links that I could. What's left is a mix of styles which:
- allow links to look OK in certain contexts/components
- make the focus style fill the space they're in, no just surround the link text

## If this seems like a lot to review

Bear in mind that any links with broken styles will revert to the browser defaults. In other words, they'll look out of place but will still look like links 🙂


klssmith:
Tested locally and it mostly looks really good and works well, just found one or two places where it still had the old styles.

**Naming of govuk-link--destructive class**
> I'm not 100% sure destructive is the right word to use (I took it from a link state in the browse-> list component) so any comments gratefully received.

One option would be to go with `warning` - this matches the naming for the destructive red buttons that come from govuk-frontend, but I think what you've got is also fine.

**Remove this team member link**
The 'Remove this team member' link that you see when editing a team member should be red

**Some links on the features pages have the old styles**
There are a number of links on the features page (`/features`) and its 3 subpages (`/features/sms`,  `/features/email`, and `/features/letters`) that still show the old link styles

**All links on the guidance pages have the old styles**
Everything on this page `/using-notify/guidance` and its 3 subpages

**List of live and trial services (only visible to Platform Admin)**
This service names are very small and not bold

**Inbound numbers page (only visible to Platform Admin)**
This service names are smaller and not bold

**JavaScript test set up**
In `liveSearch.test.js` and `updateContent.test.js`, the HTML string in the test set up contains links that could be given the new link class to match the code (not sure if this is important or not, so feel free to ignore)

tombye:
@klssmith I should have addressed all your comments in https://github.com/alphagov/notifications-admin/pull/3303/commits/324434bc446ac9998277b7deb949c5b643c56816

klssmith:
The changes all look good but just spotted a few more places where the links don't have the new class

The link to the letter spec in the error that appears when uploading an invalid file isn't using the new style https://github.com/alphagov/notifications-admin/blob/master/app/utils.py#L569

The links in the `5xx.html` page don't have the new class

There's a form error with a link https://github.com/alphagov/notifications-admin/blob/master/app/main/validators.py#L49

The 'Sign into Notify' link found at `/using-notify/trial-mode` doesn't have the new class

And this test assertion https://github.com/alphagov/notifications-admin/blob/master/tests/app/main/views/test_user_profile.py#L266 will now always pass, so will need to be rewritten

tests/app/main/views/test_user_profile.py
```diff
@@ -263,7 +263,7 @@ def test_non_gov_user_cannot_see_change_email_link(
 ):
     client_request.login(api_nongov_user_active)
     page = client_request.get('main.user_profile')
-    assert '<a href="/user-profile/email">' not in str(page)
+    assert '<a class="govuk-link govuk-link--no-visited-state" href="/user-profile/email">' not in str(page)
```
klssmith: Could write it like this so that the test doesn't need updating every time we change the class, but I'll leave that decision up to you
```suggestion
    assert not page.find('a', {'href': url_for('main.user_profile_email')})
```

PR #3304, opened 2020-02-17
Put day of week on estimated letter delivery page

quis:
This helps people understand how delivery dates fit into the week, because if sending a letter spans the weekend it will take a bit longer.

image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -334,6 +334,10 @@ def format_date_human(date):
     return get_human_day(date)
 
 
+def format_day_of_week(date):
+    return utc_string_to_aware_gmt_datetime(date).strftime('%A')
+
+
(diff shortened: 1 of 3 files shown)


quis:
@karlchillmaid do you think it would be better to also say _Provided/Uploaded today…_ instead of _Provided/Uploaded on 17 February…_?

karlchillmaid:
@quis Yes, that would make it consistent with the `Sent today` content we use for templated messages, wouldn't it?

quis:
@karlchillmaid I’ve updated the PR so the line will now read:

> Uploaded by Test User yesterday at noon

…or for a templated message:

> ‘sample template’ was sent by Test User today at 1:01am

PR #3306, opened 2020-02-17
Show scheduled jobs on page one of the uploads page

quis:
Since we’re going to summarise scheduled jobs on the dashboard instead of listing them they need to be listed here instead (which is where we’ll link to from the dashboard).

This is the same thing we do with the jobs page for caseworking users who don’t have the dashboard. 

Design of this will probably evolve as we work out how to style single letter uploads and letter jobs, but that’s OK for now because no-one has the uploads page at the moment.

***

image

### DIFF
--- a/app/templates/views/jobs/jobs.html
+++ b/app/templates/views/jobs/jobs.html
@@ -9,8 +9,19 @@
 {% block maincolumn_content %}
     <h1 class="heading-large">Uploads</h1>
     <div class="dashboard">
-      {{ scheduled_jobs|safe }}
-      {% if jobs or not current_service.scheduled_jobs %}
+      {% if show_scheduled_jobs %}
+        {% with hide_heading = True %}
+          {% include 'views/dashboard/_upcoming.html' %}
+        {% endwith %}
+      {% endif %}
+      {% if jobs %}
+        {% include 'views/dashboard/_jobs.html' %}
+      {% endif %}
+      {% if not jobs and not show_scheduled_jobs %}
+        {#
+          `_jobs.html` will show the ‘You have no jobs’ message when
+          passed an empty list of jobs
+        #}
         {% include 'views/dashboard/_jobs.html' %}
       {% endif %}
       {{ previous_next_navigation(prev_page, next_page) }}
--- a/app/templates/views/jobs/jobs.html
+++ b/app/templates/views/jobs/jobs.html
@@ -10,7 +10,9 @@
     <h1 class="heading-large">Uploads</h1>
     <div class="dashboard">
       {{ scheduled_jobs|safe }}
-      {% include 'views/dashboard/_jobs.html' %}
+      {% if jobs or not current_service.scheduled_jobs %}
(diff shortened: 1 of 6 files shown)


tests/app/main/views/test_uploads.py
```diff
@@ -589,3 +622,37 @@ def test_send_uploaded_letter_when_metadata_states_pdf_is_invalid(mocker, servic
...
+
+@pytest.mark.parametrize('user', (
+    create_active_caseworking_user(),
+    create_active_user_with_permissions(),
+))
+@freeze_time("2012-12-12 12:12")
+def test_uploads_page_shows_scheduled_jobs(
+    mocker,
+    client_request,
+    mock_get_uploads,
```
idavidmcdonald: Do we still need this mock as I think we are overriding it with `mocker.patch('app.models.job.PaginatedUploads.client_method', return_value={'data': []})`?

app/templates/views/jobs/jobs.html
```diff
@@ -10,7 +10,9 @@
     <h1 class="heading-large">Uploads</h1>
     <div class="dashboard">
       {{ scheduled_jobs|safe }}
-      {% include 'views/dashboard/_jobs.html' %}
+      {% if jobs or not current_service.scheduled_jobs %}
```
idavidmcdonald: Review this logic to see if there is a simpler way to write it

tests/app/main/views/test_uploads.py
idavidmcdonald: or the get_no_uploads

app/templates/views/jobs/jobs.html
quis: Addressed in 77fb459

tests/app/main/views/test_uploads.py
quis: Addressed in 88cbe01

tests/app/main/views/test_uploads.py
```diff
@@ -24,18 +29,46 @@ def test_no_upload_letters_button_without_permission(
...
+    assert not page.select('.file-list-filename')
+
+
 def test_get_upload_hub_page(
+    mocker,
     client_request,
     service_one,
     mock_get_uploads,
 ):
+    mocker.patch('app.job_api_client.get_jobs', return_value={'data': []})
```
idavidmcdonald: Optional: You could introduce a `mock_get_no_jobs` for this for consistency

tests/app/main/views/test_uploads.py
```diff
@@ -24,18 +29,46 @@ def test_no_upload_letters_button_without_permission(
     client_request,
     service_one,
     mock_get_uploads,
+    mock_get_jobs,
```
idavidmcdonald: Optional: do we need this

quis: Yes (depending which permissions are set on the service either `mock_get_uploads` or `mock_get_jobs` will be called)

PR #3308, opened 2020-02-18
Guard against obscure bug in test setup

quis:
If you try to freeze time to before 2011 you get the following obscure exception from somewhere within itsdangerous:

```
timestamp = base64_encode(int_to_bytes(self.get_timestamp()))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

num = -1

def int_to_bytes(num):
>       assert num >= 0
E       AssertionError
```

Something in there is baselining `timestamp` to `0` at the start of 2011. Which means that trying to freeze time to before then results in a negative timestamp.

This is a very non-obvious exception to see, so hopefully raising a specific exception instead will save others some time.

### DIFF
--- a/tests/app/test_utils.py
+++ b/tests/app/test_utils.py
@@ -516,3 +516,16 @@ def test_get_letter_validation_error_for_known_errors(
 @freeze_time('2020-02-14T12:00:00')
 def test_is_less_than_90_days_ago(date_from_db, expected_result):
     assert is_less_than_90_days_ago(date_from_db) == expected_result
+
+
+@pytest.mark.parametrize('method', ('get', 'post'))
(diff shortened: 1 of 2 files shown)


klssmith:
It seems like the timestamp thing was a bug in the version of itsdangerous that we are using that has now been fixed (https://github.com/pallets/itsdangerous/issues/46). So we could just try and bump itsdangerous instead - we shouldn't need to pin it to `0.24` anymore because version `1.1.0` has undone the breaking changes that version `1.0.0` introduced. (Haven't tried unpinning it to see if it all works though...)

PR #3313, opened 2020-02-19
Update upload a letter content

karlchillmaid:
Research shows users:

* are confused about when they would need to upload a letter
* expect an uploaded letter to behave in the same way as a letter template – with placeholders that can be filled
* miss important information about the layout specifications of their letter

This PR updates the _Guidance_ and _Upload a letter_ pages to:

* do a better job of explaining the feature
* provide context, guidance and important information at the right point in the journey

### DIFF
--- a/app/templates/views/dashboard/_jobs.html
+++ b/app/templates/views/dashboard/_jobs.html
@@ -7,7 +7,7 @@
     caption="Recent files uploaded",
     caption_visible=False,
     empty_message=(
-      'Upload a letter and Notify will print, pack and post it for you.' if current_service.can_upload_letters else 'You have not uploaded any files yet'
+      'You have not yet uploaded any files yet.'
     ),
     field_headings=[
       'File',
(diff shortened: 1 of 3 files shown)


quis:
I think the blank state of the uploads page needs a bit more work. Happy to help you code it up (it won’t fit in the existing pattern) but we need to get across that:
- there’s nothing here yet
- what uploading letters is
- how you need to go to templates to upload lists of recipients

Thoughts?

karlchillmaid:
@quis If we're going to use the same (or virtually the same) content on the _Guidance_ and _Upload a letter_ pages, is there a smarter way we can manage that content?

quis:
@karlchillmaid we can make a partial, which is code shared between two pages. But I’m not sure it’s worth the added complexity if we’re only sharing the content between two different places, and there might be slight variations.

quis:
This feels strange (I might have messed it up during the rebase though).

It looks like it’s jumping down the page. Should it link straight to the PDF?

ualjmp

karlchillmaid:
@quis Yes. I forgot to update the link.

PR #3314, opened 2020-02-19
Upgrade itsdangerous

quis:
Version 1.1.0 has reverted the breaking change (moving from sha1 to sha256) that was introduced in version 1.0.0.

Upgrading now so that we can take advantage of this bug fix: https://github.com/pallets/itsdangerous/issues/46

***

Closes #3308


quis:
This breaks the email-revalidation journey (and possibly email sign in).

quis:
Going to test on preview.

PR #3315, opened 2020-02-20
Make analytics send the same title for all pages

tombye:
We don't need the data and this will stop any personal info' being sent.

https://www.pivotaltracker.com/n/projects/1443052/stories/171378392

## How do I test this?

Use Chrome and install the [Google Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna/related) to see the data sent to Google Analytics.

We track pageviews and validation errors (see https://www.notifications.service.gov.uk/cookies) so it's worth checking both.

### DIFF
--- a/tests/javascripts/analytics/init.test.js
+++ b/tests/javascripts/analytics/init.test.js
@@ -111,10 +111,10 @@ describe("Analytics init", () => {
 
     test("A pageview will be registered", () => {
 
-      expect(window.ga.mock.calls.length).toEqual(5);
+      expect(window.ga.mock.calls.length).toEqual(6);
 
-      // The first 4 calls configure the analytics tracker. All subsequent calls send data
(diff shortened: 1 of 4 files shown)


tests/javascripts/analytics/init.test.js
```diff
@@ -111,10 +111,10 @@ describe("Analytics init", () => {
 
     test("A pageview will be registered", () => {
 
-      expect(window.ga.mock.calls.length).toEqual(5);
+      expect(window.ga.mock.calls.length).toEqual(6);
 
       // The first 4 calls configure the analytics tracker. All subsequent calls send data
```
idavidmcdonald: Should this comment now say 'the first 5 calls'

tombye: Yes, it should.

PR #3316, opened 2020-02-20
Summarise scheduled jobs on the dashboard

quis:
Before | After 
---|---
image | image

***

Scheduled jobs push everything else on the dashboard down, which makes them very prominent. This is exacerbated by people scheduling more jobs simultaneously than we expected when we originally designed the feature.

We also want to remove all jobs from the dashboard, in favour of putting them on the uploads page.

So this commit replaces them with one of our new dashboard banners (used for received text messages in returned letters) which summarises:
- how many scheduled jobs you have
- when the first one is going out (i.e. how long you have to stop it, if  you need to)

***

Shouldn’t be merged until:
- [x] #3306

### DIFF
--- a/tests/app/main/views/test_dashboard.py
+++ b/tests/app/main/views/test_dashboard.py
@@ -227,7 +227,7 @@ def test_inbound_messages_shows_count_of_messages_when_there_are_messages(
         'main.service_dashboard',
         service_id=SERVICE_ONE_ID,
     )
-    banner = page.select_one('a.banner-dashboard')
+    banner = page.select('a.banner-dashboard')[1]
     assert normalize_spaces(
         banner.text
     ) == '9,999 text messages received latest message just now'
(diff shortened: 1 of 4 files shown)


quis:
Example of the current dashboard for a service with 3 scheduled jobs:

image

tests/app/main/views/test_dashboard.py
```diff
@@ -825,20 +825,26 @@ def test_should_show_upcoming_jobs_on_dashboard(
...
         service_id=SERVICE_ONE_ID,
     )
-
     second_call = mock_get_jobs.call_args_list[1]
     assert second_call[0] == (SERVICE_ONE_ID,)
     assert second_call[1]['statuses'] == ['scheduled']
 
-    table_rows = page.find_all('tbody')[0].find_all('tr')
-    assert len(table_rows) == 2
+    assert normalize_spaces(
```
idavidmcdonald: Do we also want a test for what should be shown (or not shown) when there are no upcoming jobs? I don't think I spotted one

PR #3319, opened 2020-02-20
Fix links to current service pages from guidance

quis:
This wasn’t working because:
- macros don’t have access to `current_service` from within their scope
- there were no tests for this behaviour

### DIFF
--- a/app/templates/components/service-link.html
+++ b/app/templates/components/service-link.html
@@ -1,6 +1,6 @@
 {% macro service_link(service, endpoint, link_text) %}
   {% if service %}
-    <a class="govuk-link" href="{{ url_for(endpoint, service_id=service.id) }}">{{ link_text }}</a>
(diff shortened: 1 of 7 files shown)


app/templates/components/service-link.html
```diff
@@ -1,6 +1,6 @@
-{% macro current_service_link(endpoint, link_text) %}
-  {% if current_service %}
-    <a class="govuk-link" href="{{ url_for(endpoint, service_id=current_service.id) }}">{{ link_text }}</a>
+{% macro service_link(service, endpoint, link_text) %}
+  {% if service %}
+    <a class="govuk-link" href="{{ url_for(endpoint, service_id=service.id) }}">{{ link_text }}</a>
```
tombye: Can we add the class to remove the visited state from this?
```suggestion
    <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for(endpoint, service_id=service.id) }}">{{ link_text }}</a>
```

quis:
It’s worth having the macro because it does the logic of just showing the text, unlinked, if there isn’t a current service available.

PR #3324, opened 2020-02-24
Change forgot to forgotten

karlchillmaid:
This PR updates the forgot.password links and headings to use the past participle – 'forgotten' – instead.

"Forgotten your password?" is more accurate.

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -50,7 +50,7 @@ def sign_in():
         flash(Markup(
             (
                 "The email address or password you entered is incorrect."
-                " <a href={password_reset}>Forgot your password</a>?"
+                " <a href={password_reset}>Forgotten your password?</a>"
             ).format(password_reset=url_for('.forgot_password'))
         ))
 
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
So this looks good. Note, we currently send users to `/forgot-password` as the url. It would be more work to change this URL and we may need to do it in a two step process but it's not too much work. Is this something you think we should do as well?

Also just ccing @CrystalPea as you are currently working on a functional test that will be affected by this change so will want to be aware

idavidmcdonald:
Also, both GOV.UK Pay and GOV.UK PaaS use `forgot`in their flows so maybe something to be aware of in terms of inconsistency/discussing with them

karlchillmaid:
@idavidmcdonald In an ideal world, yes, the URL would match the H1 and the call-to-action. I guess this comes down to what else we're working on and how soon we want to prioritise making that change.

CrystalPea:
Thanks for tagging me @idavidmcdonald 
This change does influence my tests indeed 😮

PR #3333, opened 2020-02-27
Redesign the uploads page to accommodate different kinds of uploads

quis:
# Before 

Screenshot 2020-02-27 at 14 48 52

# After

image

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2732


app/templates/views/jobs/jobs.html
```diff
@@ -18,11 +18,17 @@ <h1 class="heading-medium">Uploads</h1>
         {% include 'views/dashboard/_jobs.html' %}
       {% endif %}
       {% if not jobs and not show_scheduled_jobs %}
-        {#
-          `_jobs.html` will show the ‘You have no jobs’ message when
-          passed an empty list of jobs
-        #}
-        {% include 'views/dashboard/_jobs.html' %}
+        <p>
```
idavidmcdonald: Think we want `class="govuk-body"` on all the `p` tags please https://design-system.service.gov.uk/styles/typography/#paragraphs

tests/app/main/views/test_uploads.py
```diff
@@ -660,3 +660,21 @@ def test_uploads_page_shows_scheduled_jobs(
...
+):
+    page = client_request.get('main.uploads', service_id=SERVICE_ONE_ID)
+
+    assert normalize_spaces(page.select_one('.next-page').text) == (
+        'Next page '
+        'page 2'
+    )
+    assert normalize_spaces(page.select_one('.previous-page').text) == (
+        'Previous page '
+        'page 0'
```
idavidmcdonald: Are we expecting a page 0? Is it not 1 indexed?

app/models/__init__.py
```diff
@@ -76,5 +76,13 @@ def __add__(self, other):
         return list(self) + list(other)
 
 
+class EmptyModelList(ModelList):
```
idavidmcdonald: What was the problem you were solving/improvement you were making with the introduction of this class?

app/templates/views/dashboard/_jobs.html
```diff
@@ -11,37 +12,79 @@
...
+          smallest=True,
+          label=message_count_label(
+            item.notification_count,
+            item.template_type,
+            suffix=''
+          )
+        ) }}
+      {% elif item.pdf_letter %}
+        <p class="govuk-body govuk-!-margin-bottom-1">
+          {% for line in item.recipient.split(',') %}
```
idavidmcdonald: I don't think we have a text for this behaviour of splitting and outputting the first three lines of the address. Would be nice to add one please

app/templates/views/jobs/jobs.html
quis: Fixed in 57c5c298d

app/models/__init__.py
quis: We’re trying to add `ImmediateJobs` and `ScheduledJobs` together. They are both instances of `ModelList`. But sometimes we get back an empty list (see https://github.com/alphagov/notifications-admin/pull/3333/commits/eed5e0fdd76be211edb41d50869f11d47549c42a#diff-ee386cafea142d002b5bf1fcfc61911dL126). 

Python doesn’t understand how to add a list and a `ModelList` together, so we have to make our own special implementation of an empty list.

This is temporary and can be removed once we’re no longer having to deal with an empty list (see https://github.com/alphagov/notifications-admin/pull/3338/commits/1dfb163ed7d04ee1d7e6cddd67d51c531602c6a3)

quis: I realised there’s a more Pythonic way to deal with this, by implementing `__radd__`: b1a97b0d6

tests/app/main/views/test_uploads.py
quis: It will display the link if the API says there is a page to link to. The admin app doesn’t really deal with the pagination logic.

app/templates/views/dashboard/_jobs.html
quis: Test added in 7a780d115

PR #3339, opened 2020-03-02
don't convert letter logo svgs to pngs

leohemsted:
we have stopped using png logos when rendering png previews of templated letters. we no longer need to convert to pngs when we upload a new letter logo.

- [x] https://github.com/alphagov/notifications-template-preview/pull/371

### DIFF
--- a/app/main/views/letter_branding.py
+++ b/app/main/views/letter_branding.py
@@ -8,7 +8,6 @@ from flask import (
     url_for,
 )
 from notifications_python_client.errors import HTTPError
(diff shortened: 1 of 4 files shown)


quis:
Checked some examples locally and I think the quality is alright. Ones with fine detail look a bit blacker, but certainly a lot better than when we first built this.

# From the original PR when we introduced the PNG intermediate 

https://github.com/alphagov/notifications-template-preview/pull/23

image

image

# From this PR

image

image

image

image

image

image

PR #3340, opened 2020-03-03
State when organisation has no live services

quis:
The page looks broken otherwise.

image

### DIFF
--- a/app/templates/views/organisations/organisation/index.html
+++ b/app/templates/views/organisations/organisation/index.html
@@ -106,5 +106,11 @@
     </div>
   {% endfor %}
   <div class="keyline-block govuk-!-margin-top-2"></div>
+  {% if not services %}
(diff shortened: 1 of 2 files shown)


app/templates/views/organisations/organisation/index.html
```diff
@@ -106,5 +106,11 @@ <h3 class="govuk-!-font-weight-bold govuk-!-font-size-24 govuk-!-margin-bottom-3
     </div>
   {% endfor %}
   <div class="keyline-block govuk-!-margin-top-2"></div>
+  {% if not services %}
+    <p class="govuk-body govuk-hint">
+      {{ current_org.name }} has no live services on GOV.UK Notify
```
idavidmcdonald: Is it necessary to say 'on GOV.UK Notify'?

PR #3341, opened 2020-03-03
Don’t show the postage when validation has failed

quis:
The postage covers up some of the letter, so it can hide the problem. It also implies that the letter has been put in an envelope, which will never happen if it fails validation.

This matches what we do for uploaded letters.


app/utils.py
```diff
@@ -366,6 +366,7 @@ def get_template(
     template,
     service,
     show_recipient=False,
+    show_postage=True,
```
klssmith: Doesn't look like this new argument is being used

quis: Good spot, that was a mistake, I’ve squashed it out.

PR #3345, opened 2020-03-03
Use statistics for returned letters on dashboard

quis:
This should be faster and more accurate than querying all the reports.

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2737


servingUpAces:
One more question... do we need the cache? I realise it only changes once a month, but is the page slow?

quis:
On the cache question, going to Redis is faster than going to the database via the API

quis:
This shouldn’t change the existing behaviour at all (i.e. it will say 0 returned letters in the last 7 days for 3 weeks of the month).

I’m not sure why it’s doing the 14 hours thing rather than saying ‘today’. But guess we don’t have to worry about that problem for another 3 weeks 🤷‍♂

PR #3352, opened 2020-03-05
ensure highlighting div size always matches the relevant textarea

leohemsted:
todo: fix tests

| before | after
| --- | --- |
| image | image |

### DIFF
--- a/tests/javascripts/enhancedTextbox.test.js
+++ b/tests/javascripts/enhancedTextbox.test.js
@@ -107,8 +107,7 @@ describe('Enhanced textbox', () => {
 
         backgroundEl = textarea.nextElementSibling;
 
-        // element has box-sizing: border-box so width includes padding and border
-        expect(backgroundEl.style.width).toEqual('582px');
+        expect(backgroundEl.style.width).toEqual('576px');
         expect(backgroundEl.style.borderWidth).toEqual('1px');
 
       });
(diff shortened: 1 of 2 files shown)


tests/javascripts/enhancedTextbox.test.js
```diff
@@ -289,6 +287,35 @@ describe('Enhanced textbox', () => {
 
     });
 
+    test("If a resize changes the textarea's dimensions, the width of the element below should still match", () => {
```
tombye: To match what you did on line 257.
```suggestion
    test("If a resize changes the textarea's width, the width of the element below should still match", () => {
```

tests/javascripts/enhancedTextbox.test.js
```diff
@@ -289,6 +287,35 @@ describe('Enhanced textbox', () => {
...
+      // start module
+      window.GOVUK.modules.start();
+
+      backgroundEl = textarea.nextElementSibling;
+
+      expect(window.getComputedStyle(textarea).width).toEqual("576px");
+      expect(window.getComputedStyle(backgroundEl).width).toEqual("576px");
+
+      textarea.style.width = "500px"
```
tombye: Could this be moved below line 305 so it reads that you're changing the textarea width after the initial setup?

Reading it, I was a bit confused about whether the resize would change the textarea width too.

Also, it would be good if the new size reduced by the same amount the screen did. I know the test doesn't care but it's one less thing to figure out when reading it.

leohemsted: agreed.

i solved the size reduction issue by changing the resizeTo to just a triggerEvent call and getting rid of the screenMock entirely

PR #3355, opened 2020-03-06
move invite error handler to top level

leohemsted:
ensure we catch org invite errors as well as regular invite errors.

you can test this locally by sending an org invite but only copy+pasting half of the link


CrystalPea:
Ahh, that's why I couldn't replicate it yesterday, it was an org invite, not a service invite :o

PR #3357, opened 2020-03-06
upgrade flask_login to 0.5.0

leohemsted:
flask_login sets a bunch of variables in the session object. We only use one of them, `user_id`. We set that to the user id from the database, and refer to it all over the placye.

However, in flask_login 0.5.0 they prefix this with an underscore to prevent people accidentally overwriting it etc. So when a user logs in we need to make sure that we set user_id manually so we can still use it.

I've manually tested inviting a new user, and the send one off flow to make sure session stuff still behaves reasonably. It seems okay but if anyone can think of anything else to test that'd be appreciated

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -139,6 +139,7 @@ class User(JSONModel, UserMixin):
 
     def login(self):
         login_user(self)
+        session['user_id'] = self.id
 
     def sign_in(self):
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -4,6 +4,7 @@ from unittest.mock import patch
 from urllib.parse import parse_qs, urlparse
 
 import pytest
+from flask import session as flask_session
 from flask import url_for
 from flask.testing import FlaskClient
 from flask_login import login_user
@@ -27,6 +28,9 @@ class TestClient(FlaskClient):
 
         with patch('app.events_api_client.create_event'):
             login_user(model_user)
+        with self.session_transaction() as test_session:
+            for key, value in flask_session.items():
+                test_session[key] = value
 
     def logout(self, user):
         self.get(url_for("main.sign_out"))
--- a/tests/app/main/views/test_api_integration.py
+++ b/tests/app/main/views/test_api_integration.py
@@ -172,13 +172,13 @@ def test_api_documentation_page_should_redirect(
 
 def test_should_show_empty_api_keys_page(
     client,
-    api_user_pending,
+    api_user_active,
     mock_login,
     mock_get_no_api_keys,
     mock_get_service,
     mock_has_permissions,
 ):
-    client.login(api_user_pending)
+    client.login(api_user_active)
     service_id = str(uuid.uuid4())
     response = client.get(url_for('main.api_keys', service_id=service_id))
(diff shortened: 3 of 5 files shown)


app/models/user.py
```diff
@@ -138,6 +138,7 @@ def activate(self):
             return self
 
     def login(self):
+        session['user_id'] = self.id
```
idavidmcdonald: I think this should come after the `login_user` call as if `login_user` either returns False or throws an exception, we probably don't want to be setting the user_id on the session.

https://github.com/maxcountryman/flask-login/blob/ade6bbdc521fa5e4dbed311d3e056f2e64ab4a3c/flask_login/utils.py#L145

There might not be an issue if an exception is not caught as maybe `session['user_id'] = self.id` won't actually "commit" that to the session until the end of the request or something like that.

In general, I might be worrying about something that couldn't happen in practicality but might be good to lean on the side of safety.

Oh and also all your concourse tests are breaking :(

tests/app/main/views/test_api_integration.py
```diff
@@ -172,13 +172,13 @@ def test_api_documentation_page_should_redirect(
 
 def test_should_show_empty_api_keys_page(
     client,
-    api_user_pending,
+    api_user_active,
```
idavidmcdonald: What has prompted this change? Is it an unrelated fix/improvement?

leohemsted: the test broke following this. pending user is the state when you're mid-way through creating your account so makes no sense for a test around the api key page

app/models/user.py
leohemsted: a valid concern, although after looking at the code i don't think this matters in practice for a couple of reasons.

* we only call `User.login` in two places (`log_in_user` and `activate_user`), and both of those times we activate the user immediately before hand, so login will never return false
* it'll never raise since it only raises if you do stuff with the duration parameter
* we don't use user_id for checking login state anywhere - we use the `current_user.is_authenticated` method for that, which calls through to the flask_login base class

still i dont think it hurts to change it, it's a good principle to follow on a sensitive part of the code

tests/__init__.py
```diff
@@ -27,6 +28,9 @@ def login(self, user, mocker=None, service=None):
 
         with patch('app.events_api_client.create_event'):
             login_user(model_user)
+        with self.session_transaction() as test_session:
```
idavidmcdonald: We are using `self.session_transaction` further up.

Is it worth calling them all `test_session` or just `session` for consistency?

Is there a reason this needs to be in a separate with statement or could it go in say the first with statement?

idavidmcdonald: Can we also get rid of line 21 in this method? Is it still needed?

leohemsted: this needs to be below login_user so that it gets the session variables, as i detailed in the commit message.

leohemsted: i've called it test_session to distinguish it from flask session

leohemsted: line 21 needs to stay because `login_user` is the flask_login function, not our own `User.login` function

PR #3358, opened 2020-03-09
Convert checkboxes to GOVUK Frontend

tombye:
Change all checkboxes to use the [GOVUK Frontend checkboxes component](https://design-system.service.gov.uk/components/checkboxes/).

https://www.pivotaltracker.com/story/show/170030071

## Types of checkboxes on the admin app

The admin app has 4 types of checkboxes:
1. single checkboxes, mainly used to mark a setting, like reply-to address, as the default
2. multiple checkboxes, like those for permissions on the Team member page
3. a tree of checkboxes, in a hierarchical list, like those to mark which folders a team member has access to on their page
4. a list of checkboxes where their labels are also links, used on the Templates page

### notifyCheckboxes

The GOVUK Frontend checkboxes can be used for 1. and 2. Types 3. and 4. needed a few things it doesn't yet support.

I added a new `notifyCheckboxes` macro (in https://github.com/alphagov/notifications-admin/pull/3358/commits/30f09b19fc1f0e193ff09f1f556aa101649555dd), based on `govukCheckboxes` but with an extended API to add the missing features.

### notifyCollapsibleCheckboxes

Types 2. and 3. can also be made collapsible with the collapsibleCheckboxes [JS module](https://github.com/alphagov/notifications-manuals/wiki/GOVUK-Modules-pattern) so I added a `notifyCollapsibleCheckboxes` macro (in https://github.com/alphagov/notifications-admin/pull/3358/commits/00ce7e114efe4f617364771999062e669b93afa3) to wrap this functionality.

### Existing checkbox macros

https://github.com/alphagov/notifications-admin/pull/3358/commits/98ec4a5fec3a34e75658710f4522bb4c65f6626b removes the existing macros.

## Global Jinja functions

The existing macros make their checkboxes from instances of [WTForms fields](https://wtforms.readthedocs.io/en/stable/fields.html). The GOVUK Frontend checkboxes component expects Dictionary objects with [these options](https://design-system.service.gov.uk/components/checkboxes/#options-checkboxes-example--items) instead.

I added some [helper functions](https://github.com/alphagov/notifications-admin/pull/3358/commits/b67ca3796d24d04d35c42a58e8bf276353d7f31e) for transforming WTForm fields to the GOVUK Frontend format. The reasoning is in the commit but I'd be interested in hearing some opinions on whether this is a sensible way of approaching the problem.

## Templates page

The templates page has its own data format for each checkbox so [I left the conversion to the GOVUK Frontend format in the Jinja for that page](https://github.com/alphagov/notifications-admin/pull/3358/commits/bb3f6355e990ac32daa59e52edf62d0ef4e0243d).

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -791,6 +791,53 @@ def add_template_filters(application):
         application.add_template_filter(fn)
 
 
+def wtforms_boolean_fields_to_govuk_items(boolean_fields):
+    return [{
+        "name": field.name,
+        "id": field.id,
+        "text": field.label.text,
+        "value": "y",
+        "checked": field.data
+    } for field in boolean_fields]
+
+
+def wtforms_option_field_to_govuk_item(option_field):
+    return {
+        "name": option_field.name,
+        "id": option_field.id,
+        "text": option_field.label.text,
+        "value": option_field.data,
+        "checked": option_field.checked
+    }
+
+
+def wtforms_option_fields_to_govuk_items(option_fields):
+    return [wtforms_option_field_to_govuk_item(field) for field in option_fields]
+
+
+def get_tree_of_govuk_items_from_wtforms_option_fields(option_fields, option_to_children_map):
+    items = []
+
+    for field in option_fields:
+        item = wtforms_option_field_to_govuk_item(field)
+        if field.data in option_to_children_map:
+            item['children'] = get_tree_of_govuk_items_from_wtforms_option_fields(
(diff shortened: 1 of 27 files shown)


tombye:
Closing this because:
1. I want to split out the Jinja helper functions into their own pull request so they can be used in with the GOVUK Frontend radios
2. other things, quite rightly, have focus at the moment and this doesn't have to be done now

PR #3360, opened 2020-03-09
International letters permission

klssmith:
https://www.pivotaltracker.com/story/show/171622699

Add international_letters permission as a permission which can only be turned on / off by Platform Admin.

Also fixes a few tests and updates letter pricing on a few pages.

**Requires this to be merged first**
- [x] https://github.com/alphagov/notifications-api/pull/2743

### DIFF
--- a/app/templates/views/features/letters.html
+++ b/app/templates/views/features/letters.html
@@ -36,7 +36,7 @@
   <p>Read our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.documentation') }}">documentation</a>.</p>
 
   <h2 class="heading heading-medium">Pricing</h2>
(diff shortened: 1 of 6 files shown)


app/templates/views/features/letters.html
```diff
@@ -36,7 +36,7 @@ <h3 class="heading heading-small" id="upload-letters">Upload your own letters</h
   <p>Read our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.documentation') }}">documentation</a>.</p>
 
   <h2 class="heading heading-medium">Pricing</h2>
-  <p>It costs between 30p and 76p (plus VAT) to send a letter. Prices include:</p>
```
idavidmcdonald: Optional: We could move these into constants so we don't run into the same problem of updating them in one place but not the others

PR #3362, opened 2020-03-09
Fix the error message for message-too-long

servingUpAces:
There was a bug that displayed the error message with 2 red error boxes around the error message.
Also need another else to handle a new error message from the API, the old one can be removed after the API is deployed. But this can go first. I tested this branch with API master and the API branch with the change. I tested one off SMS and a CSV upload.

### DIFF
--- a/app/templates/partials/check/message-too-long.html
+++ b/app/templates/partials/check/message-too-long.html
@@ -1,13 +1,7 @@
-{% from "components/banner.html" import banner_wrapper %}
-
-<div class="bottom-gutter">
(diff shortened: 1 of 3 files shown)


app/templates/partials/check/message-too-long.html
```diff
@@ -1,13 +1,9 @@
 {% from "components/banner.html" import banner_wrapper %}
```
CrystalPea: I don't think we need `banner_wrapper` anymore

servingUpAces: good spot - I removed it.

PR #3363, opened 2020-03-10
Add event for remove_user_from_service

idavidmcdonald:
https://www.pivotaltracker.com/story/show/171354636

Follows our existing patterns of events for managing users

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -25,6 +25,15 @@ def create_mobile_number_change_event(user_id, updated_by_id, original_mobile_nu
         new_mobile_number=new_mobile_number)
 
 
+def create_remove_user_from_service_event(user_id, removed_by_id, service_id):
+    _send_event(
+        'remove_user_from_service',
+        user_id=user_id,
+        removed_by_id=removed_by_id,
+        service_id=service_id
+    )
+
+
 def create_archive_user_event(user_id, archived_by_id):
     _send_event(
         'archive_user',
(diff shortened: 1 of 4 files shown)


app/main/views/manage_users.py
```diff
@@ -132,6 +133,8 @@ def remove_user_from_service(service_id, user_id):
                 service_id=service_id))
         else:
             abort(500, e)
+    else:
```
idavidmcdonald: I've copied the `try, else` pattern from other functions in this file. Would be keen to hear if people think it's still the correct thing to do? I guess, what benefit are we getting from having it in an `else` statement rather than in the original `try`?

app/event_handlers.py
```diff
@@ -25,6 +25,14 @@ def create_mobile_number_change_event(user_id, updated_by_id, original_mobile_nu
         new_mobile_number=new_mobile_number)
 
 
+def create_remove_user_from_service_event(user_id, removed_by_id):
```
servingUpAces: Let's include the service id please

app/main/views/manage_users.py
servingUpAces: If the creating the event throws an error that is a different scenario than if removing the user caused the exception. 

I question the use of abort 500 here, is that what we what to happen, what if it's a different 400?

app/event_handlers.py
idavidmcdonald: Added and rebased

PR #3364, opened 2020-03-12
Show uploads link even without letters permission

quis:


app/templates/main_nav.html
```diff
@@ -14,9 +14,7 @@
       {% endif %}
     {% else %}
       <li><a class="govuk-link govuk-link--no-visited-state{{ casework_navigation.is_selected('sent-messages') }}" href="{{ url_for('.view_notifications', service_id=current_service.id, status='sending,delivered,failed') }}">Sent messages</a></li>
-      {% if current_service.has_jobs or current_service.can_upload_letters %}
```
idavidmcdonald: So just to confirm, if a user doesn't have the permission `view_activity` then they can always see the uploads link? Even if their service has no jobs?

servingUpAces: I think we also need to update the `current_service.can_upload_letters` to `return self.has_permission('upload_letters')`.... remove the self.has_permission('letter')

quis: I’ve edited the wrong bit…

quis: I’ve changed it now to remove the check for people who have _Dashboard_ in the navigation. Existing logic works fine for those who don’t.

PR #3366, opened 2020-03-12
Let teams upload an emergency contact list for later use 

quis:
image

***

image

***

image

***

image

***

image

***

image

***

image

***

We increasingly have teams wanting to do business-continuity type messaging. They might be without access to their normal systems, which is where they would otherwise go to get the list of email addresses or phone numbers.

So we want to give them a place in Notify where they can store their spreadsheets and use them at a later date.

For the initial pass we’re going to scope this to only allowing spreadsheets with one column, ie just phone numbers/email addresses. This is because:
- it minimises the amount of personal info we’re storing
- it reduces the chance of getting a placeholder error when you go to send the message, which is probably a high-stress situation where you might not be able to re-generate the file

The code for this is mostly copied from the existing upload CSV journey. It’s quite duplicative, but that’s what I needed to do to get this out quickly. There are opportunities for refactoring later.

Similarly, I would have liked to split this up into smaller PRs, but this is kinda what was needed to ship quick enough.

***

At the moment this is only available for platform admin users, so we can test it out before shipping it.

To do in future pull requests:
- let users delete contact lists
- link jobs to the contact list they were sent from


app/main/views/uploads.py
```diff
@@ -282,3 +298,144 @@ def send_uploaded_letter(service_id):
         service_id=service_id,
         notification_id=file_id,
     ))
+
+
+@main.route("/services/<uuid:service_id>/upload-a-contact-list", methods=['GET', 'POST'])
```
servingUpAces: why `upload-a-contact-list` we don't say `upload-a-letter`

quis: Changed in cad40db05

PR #3367, opened 2020-03-12
Pull in version of utils that validates postcodes for CSVs

idavidmcdonald:
This will mean that a user can not send a CSV for letters with invalid
postcodes and must fix their data.

See https://github.com/alphagov/notifications-utils/pull/704

image


leohemsted:
You'll need to freeze requirements tho

PR #3368, opened 2020-03-13
one off letter address form

leohemsted:
Currently, the starting point of the one off flow is the endpoint `send_one_off` (which sets up the session, etc). It redirects. from there to `send_one_off_step` with a step_index of 0.

As of this PR, it goes to the new page `send_one_off_letter_address`.

this new endpoint has a textarea that it splits into lines (removing blank lines), and checks that there are between 3 and 7. Then, it maps those lines to the old placeholders `address line [1-6]` and `postcode` - the last line always goes into postcode, the rest of the address line placeholders are filled out sequentially.

If a user is half-way through the old flow during the deploy, they won't see any changes. If a user hits back from a later placeholder, even if they filled in the `send_one_off_letter_address` endpoint earlier, they will go back to stepping manually through each field in turn. A future PR will make the first seven step indexes automatically redirect to the address block, but that'll be deployed separately to ensure people don't get the flow disrupted during the deploy. This functionality is already tested by `test_send_test_allows_empty_optional_address_columns`.

If someone has one of the optional later address placeholders in the body of their text, it will not fill that in if they don't supply enough lines from `send_one_off_letter_address`, and will then prompt them to fill in that line. When they fill that line in, it'll update both the value in the main body and also add that line to the address block. I think that's fine because that's such a weird niche case.

What it looks like:

image

Error messages:

| image | image |
| --- | --- |


quis:
Looks tasty. Can you start it at 4 lines high and use the auto-grow Javascript to expand as people type?

app/main/views/send.py
```diff
@@ -328,6 +334,86 @@ def get_notification_check_endpoint(service_id, template):
...
+        for i in range(1, 7):
+            session['placeholders'][f'address line {i}'] = ''
+
+        # unroll the address into lines, and place into the session in the underlying placeholder names
+        # postcode is required so make sure we put the last value in that
+        # TODO: When postcode is no longer a required field, remove this special case and just use `address line #`
+        address_lines, last_address_line = lines[:-1], lines[-1]
+        for i, line in enumerate(address_lines, start=1):
+            session['placeholders'][f'address line {i}'] = line
+        session['placeholders']['postcode'] = last_address_line
```
quis: Would this be nicer encapsulated into the form object? For example:
```python
session.update(form.as_address_lines_1_to_7_with_postcode)
```

tests/app/main/views/test_send.py
```diff
@@ -2066,6 +2063,179 @@ def test_send_test_clears_session(
...
+        'address line 2': 'b',
+        'address line 3': 'c',
+        'address line 4': 'd',
+        'address line 5': 'e',
+        'address line 6': 'f',
+        'postcode': 'g',
+    }),
+    # it ignores empty lines and strips whitespace from each line
+    ('\n  a\ta  \n\n\n      \n\n\n\nb  b   \r\nc', {
+        'address line 1': 'a\ta',
```
quis: Should we be normalising out stuff like tabs, Mongolian vowel separator, etc?

app/templates/views/send-one-off-letter-address.html
```diff
@@ -0,0 +1,45 @@
...
+
+{% block maincolumn_content %}
+
+  {{ page_header(
+    page_title,
+    back_link=back_link
+  ) }}
+
+  {% call form_wrapper(
+    class='js-stick-at-top-when-scrolling send-one-off-form',
```
quis:
```suggestion
    class='send-one-off-form',
```
We probably don’t want the sticky behaviour here because scrolling is likely to obscure the address on the letter, which is the thing you’re trying to match up with the textbox. Still makes sense for later placeholders, which will be further down the page.

tests/app/main/views/test_send.py
leohemsted: lifted the code we use in utils - had to tweak the remove dupe newlines code slightly though

PR #3372, opened 2020-03-16
Increase all ajax calls to 5 seconds

idavidmcdonald:
The default is 2 seconds and this will mean that we are halving traffic
for these ajax calls which can only be good for trying to limit queries
on the database.

I think the user impact on this will likely not be noticable.

Debatable whether we should up them all even further to 10 seconds but
this is definitely a quick although maybe small win.


quis:
You can drop some more load by dropping the `jobs` part of `…/dashboard.json`

I was working to get rid of it as part of cleaning up after the uploads work, but it might help to push through that one particular change for performance reasons.

PR #3373, opened 2020-03-16
Remove jobs from the dashboard

quis:
We’ve done this already for services with the upload letters permission. And all services can upload letters now.

But we’re still returning it in the JSON response we use to AJAX-ify the page.

Since the jobs response can query stats for up to 50 jobs at a time this puts some load on the API/database. Hopefully this might drop that load a bit.


app/templates/main_nav.html
```diff
@@ -8,14 +8,10 @@
       <li><a class="govuk-link govuk-link--no-visited-state{{ main_navigation.is_selected('dashboard') }}" href="{{ url_for('.service_dashboard', service_id=current_service.id) }}">Dashboard</a></li>
     {% endif %}
     <li><a class="govuk-link govuk-link--no-visited-state{{ main_navigation.is_selected('templates') }}" href="{{ url_for('.choose_template', service_id=current_service.id) }}">Templates</a></li>
-    {% if current_user.has_permissions('view_activity') %}
```
servingUpAces: Should this change?

PR #3376, opened 2020-03-17
Use new API endpoint to get single contact list

quis:
When we first built the contact list feature this endpoint didn’t exist. Now it does we can remove the slightly cludgy looping-through-all-the-lists code.

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2750

### DIFF
--- a/app/notify_client/contact_list_api_client.py
+++ b/app/notify_client/contact_list_api_client.py
@@ -25,7 +25,10 @@ class ContactListApiClient(NotifyAdminAPIClient):
         return job
 
     def get_contact_lists(self, service_id):
-        return self.get('/service/{}/contact-list'.format(service_id))
+        return self.get(f'/service/{service_id}/contact-list')
+
+    def get_contact_list(self, *, service_id, contact_list_id):
+        return self.get(f'/service/{service_id}/contact-list/{contact_list_id}')
 
 
 contact_list_api_client = ContactListApiClient()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1870,6 +1870,36 @@ def mock_get_contact_lists(mocker, api_user_active, fake_uuid):
 
 
 @pytest.fixture(scope='function')
+def mock_get_contact_list(mocker, api_user_active, fake_uuid):
+    def _get(*, service_id, contact_list_id):
+        return {
+            'created_at': '2020-03-13 10:59:56',
+            'created_by': 'Test User',
+            'id': fake_uuid,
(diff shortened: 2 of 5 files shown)


app/notify_client/contact_list_api_client.py
```diff
@@ -25,7 +25,10 @@ def create_contact_list(
         return job
 
     def get_contact_lists(self, service_id):
-        return self.get('/service/{}/contact-list'.format(service_id))
+        return self.get(f'/service/{service_id}/contact-list')
+
+    def get_contact_list(self, *, service_id, contact_list_id):
```
idavidmcdonald: What is the `*` for?

quis: It forces `service_id` and `contact_list_id` to be [keyword only arguments](https://www.python.org/dev/peps/pep-3102/). That means the following will raised a `TypeError`:
```python
contact_list_api_client.get_contact_list(contact_list_id, service_id)
```
This is good because the above code looks fine but we’ve inadvertently swapped the two arguments.

Using the single-star syntax means you’re forced to do the following instead, which makes it much more obvious that there’s a problem:
```python
contact_list_api_client.get_contact_list(service_id=contact_list_id, contact_list_id=service_id)
```

idavidmcdonald: Ah yes, this rings some bells now. Thanks for the thorough explanation.

I guess my two cautions with this are:
- I feel like we always pass in service id first for all our calls generally so feel like maybe the risk is lower then in general
- It would be nice to have this done consistently across our methods rather than just in one place here

Neither of those cautions are big enough to stop me approving the PR though

tests/conftest.py
```diff
@@ -1869,6 +1869,36 @@ def _get(service_id, template_type=None):
     )
 
 
+@pytest.fixture(scope='function')
+def mock_get_contact_list(mocker, api_user_active, fake_uuid):
+    def _get(*, service_id, contact_list_id):
```
idavidmcdonald: Do we need to go to the trouble of making a function for `_get` with the forcing of key word arguments? Is that so we are testing the behaviour were we have to be using key word arguments?

Rather than say
```
return mocker.patch(
        'app.models.contact_list.contact_list_api_client.get_contact_list',
        return_value = {'dict_key': 'dict_value_example'}
    )
```

quis: Yeah, I think we do this a lot elsewhere to ensure that, when testing, you’re calling the function with the signature it expects, not just any old rubbish.

PR #3381, opened 2020-03-19
Add status page link to failwhale page

klssmith:


klssmith:
I asked around and couldn't find out if we still need the Terraform failwhale, so haven't deleted it just in case

PR #3382, opened 2020-03-23
Require email address for all support tickets

idavidmcdonald:
We are seeing little benefit of allowing users to not put in their email
address. This will mean that you must provide it for feedback, not just
problems with the site.

There could maybe be some more refactoring of the support templates as
this is now very similar to the report a problem page but this is a
quick fix so haven't gone too in depth.

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -894,7 +894,7 @@ class SupportType(StripWhitespaceForm):
 
 class Feedback(StripWhitespaceForm):
     name = StringField('Name')
-    email_address = email_address(label='Email address', gov_user=False, required=False)
+    email_address = email_address(label='Email address', gov_user=False, required=True)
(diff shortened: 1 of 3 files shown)


app/main/forms.py
```diff
@@ -894,7 +894,7 @@ class SupportType(StripWhitespaceForm):
 
 class Feedback(StripWhitespaceForm):
     name = StringField('Name')
```
quis:
```suggestion
    name = StringField('Name (optional)')
```

> If you ask for optional information, mark the labels of optional fields with ‘(optional)’.
– https://design-system.service.gov.uk/patterns/question-pages/#asking-users-questions

idavidmcdonald: Yep, have added this change. It also does it on both pages already.

idavidmcdonald:
image

PR #3383, opened 2020-03-23
Drop frequency of AJAX updates on notifications page

quis:
This is another page that puts load on the database, so let’s treat it the same as the dashboard.

### DIFF
--- a/app/templates/views/notifications.html
+++ b/app/templates/views/notifications.html
@@ -31,7 +31,8 @@
     {{ ajax_block(
       partials,
       url_for('.get_notifications_as_json', service_id=current_service.id, message_type=message_type, status=status),
-      'counts'
+      'counts',
+      interval=20
     ) }}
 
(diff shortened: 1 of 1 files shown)


app/templates/views/notifications.html
```diff
@@ -30,7 +30,7 @@ <h1 class="heading-large">
 
     {{ ajax_block(
       partials,
-      url_for('.get_notifications_as_json', service_id=current_service.id, message_type=message_type, status=status),
+      url_for('.get_notifications_as_json', service_id=current_service.id, message_type=message_type, status=status, interval=20),
```
idavidmcdonald: Is the interval supposed to be in the url_for? or should it be a separate argument to `ajax_block`?

quis: Good point, I’m moving too fast. I’ve changed and squashed into this PR now.

PR #3386, opened 2020-03-24
Direct members of the public raising support tickets to somewhere more useful

quis:
We can’t give advice to members of the public, but increasingly we’re seeing them try to use our support form to ask.

It would be better for them if we can direct them straight to somewhere more useful, before they have the chance to raise a support ticket. We have a strong hunch that if people get to a page where there’s a box they can type that they’re likely to ignore any other text on the page – we don’t think that signposting is enough here.

This commit replaces the report a problem/ask a question triaging for users who aren’t signed in. It’s not possible for non-signed-in users to raise a priority 1 ticket, so we never need to triage the tickets in this way.

Instead we can triage people based on whether they work in the public sector or not. If they do then we send them on to the feedback form. If not then they go to a new page which contains some useful links. We’ve chosen these links based on [some analysis of the support tickets we’ve received recently](https://docs.google.com/spreadsheets/d/1uBQn-ZnCYfz6ltFaUKZpytgvBF0-MeshCLZ1cD74R0c/edit?usp=sharing).

# Before 

image

***

image

# After 

image

***

image


app/templates/views/support/public.html
```diff
@@ -0,0 +1,48 @@
...
+  The GOV.UK Notify team can’t give advice to members of the public
+{% endblock %}
+
+{% block maincolumn_content %}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    {{ page_header(
+      'The GOV.UK Notify service is for people who work in the government',
```
idavidmcdonald: Optional: On the previous form we use `public sector` and then here we use `government`. Seems slightly inconsistent?

app/main/views/feedback.py
```diff
@@ -1,67 +1,79 @@
...
 from app import convert_to_boolean, current_service, service_api_client
 from app.extensions import zendesk_client
 from app.main import main
-from app.main.forms import Feedback, Problem, SupportType, Triage
-
-QUESTION_TICKET_TYPE = 'ask-question-give-feedback'
-PROBLEM_TICKET_TYPE = "report-problem"
-
-
-def get_prefilled_message():
```
idavidmcdonald: I'm guessing this was no longer being used?

app/main/views/feedback.py
```diff
@@ -1,67 +1,79 @@
...
 
+@main.route('/support/public')
+def support_public():
+    return render_template('views/support/public.html')
+
+
 @main.route('/support/triage', methods=['GET', 'POST'])
-def triage():
+@main.route('/support/triage/<ticket_type:ticket_type>', methods=['GET', 'POST'])
+def triage(ticket_type=PROBLEM_TICKET_TYPE):
```
idavidmcdonald: Do we need the default argument here?

app/main/views/feedback.py
```diff
@@ -121,13 +132,15 @@ def feedback(ticket_type):
...
 
-    if not form.feedback.data:
-        form.feedback.data = get_prefilled_message()
-
     return render_template(
-        'views/support/{}.html'.format(ticket_type),
+        'views/support/form.html',
         form=form,
-        ticket_type=ticket_type,
+        show_status_page_banner=(ticket_type == PROBLEM_TICKET_TYPE),
```
idavidmcdonald: When triaging a general ticket and saying it's an emergency I end up at `support/general?severe=yes`. there is no content about the looking for the status page which feels like should be shown in this case

app/templates/views/support/form.html
```diff
@@ -6,23 +6,25 @@
...
-  Report a problem
+  {{ page_title }}
 {% endblock %}
 
 {% block maincolumn_content %}
 
     {{ page_header(
-      'Report a problem',
+      page_title,
       back_link=url_for('.support')
```
idavidmcdonald: Optional: Should this go back to the triage page if they've come from there (not essential I think if its not a 2 min fix given it's good to get this out earlier than possible)

app/templates/views/support/public.html
quis: This was a deliberate choice.

On the first page we have to say ‘public sector’ because, technically, it’s not just government organisations that can use Notify so we don’t want to put off people from the NHS etc.

But once someone has said they’re trying to get help as a member of the public I feel like it’s more straightforward language to just talk about ‘working for the government‘.

app/main/views/feedback.py
quis: Yeah, I think it became redundant once letter branding requests became self-service (https://github.com/alphagov/notifications-admin/pull/3268)

app/main/views/feedback.py
quis: It preserves the existing behaviour while we’re deploying, for people who end up at `/support/triage`

app/main/views/feedback.py
quis: If you look at that page after 5:30pm (or by modifying `in_business_hours` to always return `False`) then you’ll be redirected to `/support/escalate` which has content about the status page:

image

app/templates/views/support/form.html
quis: Changed in 2ca32019e

PR #3387, opened 2020-03-26
Add ‘Who it’s for’ page

quis:
We have a policy about how suppliers are allowed to use Notify. But we don’t explain it anywhere. Which drives contact to our support form.

This pull request adds a new page that explains the policy.

It then links to the page from the error message that someone gets if they try to sign up with an email domain we don’t recognise.

# Before

image

# After

image

***

image

### DIFF
--- a/app/main/views/sub_navigation_dictionaries.py
+++ b/app/main/views/sub_navigation_dictionaries.py
@@ -53,6 +53,10 @@ def using_notify_nav():
             "link": "main.get_started",
         },
         {
+            "name": "Who it’s for",
+            "link": "main.who_its_for",
+        },
+        {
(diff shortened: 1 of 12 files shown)


app/main/views/sub_navigation_dictionaries.py
```diff
@@ -51,6 +51,13 @@ def using_notify_nav():
         {
             "name": "Get started",
             "link": "main.get_started",
+            "sub_navigation_items": [
```
tombye: Are these needed?

quis: Ooh, good spot. Have squashed out now.

PR #3389, opened 2020-03-26
Fix bug where adding same reply-to-email twice for the same service caused a 5xx

CrystalPea:
This was due to changed quotation marks in error message.

We ended up getting rid of error message check in favour of a more accurate error code that will make sure we are getting the right error.

This PR makes our code accept both 400 and 409 error codes. We can switch to solely accepting 409s after this PR is merged: https://github.com/alphagov/notifications-api/pull/2779

story ticket: https://www.pivotaltracker.com/story/show/171785534

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -391,7 +391,7 @@ def service_add_email_reply_to(service_id):
                 service_id, form.email_address.data
             )["data"]["id"]
         except HTTPError as e:
-            error_msg = "Your service already uses '{}' as an email reply-to address.".format(form.email_address.data)
+            error_msg = "Your service already uses ‘{}’ as an email reply-to address.".format(form.email_address.data)
             if e.status_code == 400 and error_msg == e.message:
                 flash(error_msg, 'error')
                 return redirect(url_for('.service_email_reply_to', service_id=service_id))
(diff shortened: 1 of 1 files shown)


app/main/views/service_settings.py
```diff
@@ -391,7 +391,7 @@ def service_add_email_reply_to(service_id):
                 service_id, form.email_address.data
             )["data"]["id"]
         except HTTPError as e:
-            error_msg = "Your service already uses '{}' as an email reply-to address.".format(form.email_address.data)
+            error_msg = "Your service already uses ‘{}’ as an email reply-to address.".format(form.email_address.data)
```
idavidmcdonald: This is a good enough fix. If we were really wanting to invest time in it, I'd argue that the API should return a HTTP 409 conflict response and the admin would show this error based on the http response code rather than the text of the error message. But I'm happy for what you've done anyway, just thought I'd share my thoughts

CrystalPea: I like this, nice and exact. 
So we would have to:
1. change admin so it recognises both existing and new error from API
2. change API to the new 409 error
3. stop checking for old error and message content in admin

Do you think I missed anything?

idavidmcdonald: You may not need to do step 1. I assume a 409 would also be raised as an HTTPError and assuming the message is kept the same for that error then the admin would still work (you should validate this assumption).

That being the case, you could then just do steps 2 and 3.

Not sure how long this would all take, maybe half a day, so I'm very happy for this PR to go in as is and we could put in the longer term fix at a later date as I don't know if there are more important things to work on

app/main/views/service_settings.py
```diff
@@ -391,9 +391,8 @@ def service_add_email_reply_to(service_id):
                 service_id, form.email_address.data
             )["data"]["id"]
         except HTTPError as e:
-            error_msg = "Your service already uses '{}' as an email reply-to address.".format(form.email_address.data)
-            if e.status_code == 400 and error_msg == e.message:
-                flash(error_msg, 'error')
+            if e.status_code == 409 or e.status_code == 400:
```
idavidmcdonald: Should this be accepting `e.status_code == 409 or (e.status_code == 400 and error_msg == e.message)` to maintain existing behaviour whilst also adding in the 409 behaviour? Otherwise, now it may do different behaviour for all 400s?

Also is there a unit test that we can write/edit for the introduction of the 409 behaviour?

PR #3391, opened 2020-03-27
Cache organisation name in Redis

quis:
A lot of pages in the admin app are now generated entirely from Redis, without touching the API. Talking to Redis is typically 100× faster than talking to the API, so this makes the UI feel nice and snappy.

The one remaining API call that a lot of pages make, when the user is platform admin or a member of an organisation, is to get the name of the current service’s organisation.

This commit adds some code to start caching that as well, which should speed up page load times for when we’re clicking around the admin app (it’s around 100ms to get the organisation, and more than that when the API is under load).

This means changing the service model to get the organisation from the API by ID, not by service ID. Otherwise it would be very hard to clear the cache if the name of the organisation ever changed.

We can’t cache the whole organisation because it has a `count_of_live_services` field which can change at any time, without an update being made.


tests/app/main/views/test_service_settings.py
```diff
@@ -4641,14 +4671,19 @@ def test_show_branding_request_page_when_branding_is_same_as_org(
     branding_type
 ):
     service_one['{}_branding'.format(branding_type)] = sample_uuid()
+    mocker.patch(
+        'app.models.service.Service.organisation_id',
+        new_callable=PropertyMock,
+        return_value=ORGANISATION_ID,
+    )
     if branding_type == 'email':
```
idavidmcdonald, line 4679: Optional: This code is repeated often, could we move it into a fixture?

PR #3397, opened 2020-04-02
Handle postcode validation for precompiled letters

CrystalPea:
Needs to go in before: https://github.com/alphagov/notifications-template-preview/pull/434

I'd love some eyes on the content for this.

Screenshot 2020-04-02 at 11 54 52
Screenshot 2020-04-02 at 11 54 32

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -651,20 +651,12 @@ LETTER_VALIDATION_MESSAGES = {
         ),
     },
     'not-a-real-uk-postcode': {
-        'title': 'Postcode in the address block is not a real UK postcode',
+        'title': 'There’s a problem with the address on this letter',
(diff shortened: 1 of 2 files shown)


quis:
How about this?

> **There’s a problem with the address on this letter**
>
> The last line of the address must be a real UK postcode

***

And for API letters:

> Validation failed because the last line of the address is not a real UK postcode.

***

(no link to the specification in either case – it doesn’t say anything about postcodes)

CrystalPea:
@quis thanks a lot, I made the changes and changed the screenshots, what do you think?

quis:
Looks great, if I do say so myself 😊 

Can you add a test and then this is good to go

PR #3398, opened 2020-04-02
Fix unnecessary call to organisation API endpoint

quis:
We’re caching the organisation name, but still talking to the API to see if the organisation exists.

`Service().organisation_id` only goes to the JSON for the service.

`Service().organisation` makes a separate API call.

We only need the former to know if a service belongs to an organisation.

### DIFF
--- a/tests/app/models/test_service.py
+++ b/tests/app/models/test_service.py
@@ -1,5 +1,6 @@
 import uuid
 
+from app.models.organisation import Organisation
 from app.models.service import Service
 from app.models.user import User
 from tests import organisation_json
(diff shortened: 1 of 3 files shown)


tests/app/models/test_service.py
```diff
@@ -195,3 +196,49 @@ def test_organisation_type_when_service_and_its_org_both_have_an_org_type(mocker
...
+    assert service.organisation_name == 'Test Organisation'
+    mock_redis_set.assert_called_once_with(
+        f'organisation-{ORGANISATION_ID}-name',
+        '"Test Organisation"',
+        ex=604800,
+    )
+
+
+def test_service_without_organisation_doesnt_need_org_api(mocker, service_one):
+    # No mocking of the API or Redis here
```
idavidmcdonald, line 238: Should we be mocking these still and asserting that they are not called? Otherwise, is this test actually testing that it doesn't call out to the API?

quis: Yeah, that’s probably clearer. Fixed and squashed.

PR #3403, opened 2020-04-07
Specify mimetype for PDF letters to be downloaded

klssmith:
We had a report that when clicking on the 'Download this letter' link on
the notification page the file was not being downloaded as a PDF file
but was given a `.htm` file extension instead. We should be able to stop
that happening by using Flask's `send_file` function with the right mimetype.
This change updates the `view_letter_notification_as_preview` to use
`send_file` and splits out code to get the file data into a separate
function.

Mocks in the tests have been updated and some unused mocks removed.

### DIFF
--- a/tests/app/main/views/test_notifications.py
+++ b/tests/app/main/views/test_notifications.py
@@ -254,7 +254,7 @@ def test_notification_page_shows_uploaded_letter(
     fake_uuid,
 ):
     mocker.patch(
-        'app.main.views.notifications.view_letter_notification_as_preview',
+        'app.main.views.notifications.get_letter_file_data',
         return_value=(b'foo', {
             'message': '',
             'invalid_pages': '[]',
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_notifications.py
```diff
@@ -783,14 +770,9 @@ def test_notification_page_has_expected_template_link_for_letter(
...
         mocker.patch(
-            'app.main.views.notifications.view_letter_notification_as_preview',
+            'app.main.views.notifications.get_letter_file_data',
             side_effect=[(b'foo', {"message": "", "invalid_pages": "[]", "page_count": "1"}), b'foo']
         )
-    else:
-        mocker.patch(
-            'app.main.views.notifications.view_letter_notification_as_preview',
-            return_value=b'foo'
-        )
```
idavidmcdonald, line 793: How come we are able to drop all these patches for the non precompiled letter tests? Just struggling to figure it out myself

klssmith: I don't think we needed to mock `view_letter_notification_as_preview` for notifications that weren't precompiled letters. For precompiled letters `view_letter_notification_as_preview` was getting called directly, but for other notifications although we use that endpoint as a link on the page, the function doesn't actually get called.

PR #3407, opened 2020-04-09
Fix one off letter journey if you have a placeholder from the address block

quis:
Currently users are blocked from continuing if they have placeholder like `address line 1` or `postcode` in their template. They get redirected back to filling in the address block when they encounter one of these placeholders. So this commit:
- makes the flow skip over filling in these placeholders individually
- makes the flow skip over filling in these placeholders when clicking back through the sequence

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -3801,6 +3801,58 @@ def test_check_notification_shows_help(
     )
 
 
+def test_check_notification_shows_back_link(
+    mocker,
+    client_request,
+    service_one,
+    fake_uuid,
+    mock_template_preview
+):
+    service_one['restricted'] = False
+    mocker.patch(
+        'app.service_api_client.get_service_template',
+        return_value={'data': template_json(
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_send.py
```diff
@@ -2045,6 +2046,77 @@ def test_send_one_off_back_link_populates_address_textarea(
...
+            'address_line_3': '',
+            'address_line_4': '',
+            'address_line_5': '',
+            'address_line_6': '',
+            'address_line_7': 'SW1A 1AA',
+            'postcode': 'SW1A 1AA',
+        }
+
+    back_link = page.select_one('.govuk-back-link')['href']
+    # assert back_link == 'foo'
```
CrystalPea, line 2112: `# assert back_link == 'foo'`

is this a leftover comment?

quis: Oops yeah, thanks for spotting. I’ve updated to test for the correct URL now and squashed into d31c244363c80e4b3b58f50db292f634d833650e

PR #3411, opened 2020-04-15
Show errors that aren’t specific to a single spreadsheet column across the whole table row

quis:
# Before

image

# After

image

***

The reason for making this improvement now is because, in the future, letter address errors won’t be related to specific cells, but to the combination of all the address columns.

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/719

### DIFF
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -65,12 +65,12 @@
   {% endif %}
 {%- endmacro %}
 
-{% macro field(align='left', status='', border=True) -%}
+{% macro field(align='left', status='', border=True, colspan=None) -%}
 
     {% set field_alignment = 'table-field-right-aligned' if align == 'right' else 'table-field-left-aligned' %}
     {% set border = '' if border else 'table-field-noborder' %}
 
-    <td class="{{ [field_alignment, border]|join(' ') }}">
+    <td class="{{ [field_alignment, border]|join(' ') }}" {% if colspan %}colspan="{{ colspan }}"{% endif %}>
       <div class="{{ 'table-field-status-' + status if status }}">{{ caller() }}</div>
     </td>
 {%- endmacro %}
@@ -81,8 +81,8 @@
   </th>
 {%- endmacro %}
 
-{% macro index_field(text=None) -%}
-  <td class="table-field-index">
+{% macro index_field(text=None, rowspan=None) -%}
+  <td class="table-field-index" {% if rowspan %}rowspan="{{ rowspan }}"{% endif %}>
     {{ text if text != None else caller()  }}
   </td>
 {%- endmacro %}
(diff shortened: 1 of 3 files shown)


tombye:
Is it possible for the error summary to link to the errors in the table? That's the guidance for forms so it would make sense to do it here too.

tombye:
The code looks good here, I've just got one question about the errors.
I was pretty thrown by them referring to errors in individual cells but talking about the 'message', as in 'shorten the message in row 1'. It took me a while to realise it was telling me the contents of a cell was causing the message sent to be too long.
Is it possible to change this to split out the error (the message being too long) from the cell causing it?

quis:
@tombye because there can be multiple errors in the table we can’t link directly to the error (because which one do you choose). But:
- each error summary has a visually hidden link (‘Skip to file contents’) which goes straight to the table
- the table only shows the rows that have errors, so in a way it’s already skipping past the ones that don’t

quis:
> Is it possible to change this to split out the error (the message being too long) from the cell causing it?

The problem is that the error is caused by a combination of all the cells, not any one cell. Can we revisit the design and wording once @karlchillmaid is in and we have the extra address error messages in place?

tombye:
> @tombye because there can be multiple errors in the table we can’t link directly to the error (because which one do you choose). But:
> 
>     * each error summary has a visually hidden link (‘Skip to file contents’) which goes straight to the table
> 
>     * the table only shows the rows that have errors, so in a way it’s already skipping past the ones that don’t

Can we link them to the row then? I mean give each an id and link to its fragment identifier. Might help people fixing CSVs with enough errors that they need to scroll between the summary and row to get all the information. Could also be done with information by naming the row number in the summary error message.

quis:
If there’s more than one error then the message looks like this:
> You need to:
> - shorten the messages in 3 rows
> - fix 3 phone numbers

In which case we can’t link to a specific row because there’s more than one. I suppose we could add a link for the case where there’s only one error, but then it would feel inconsistent because you’d be able to click the first bullet here but not the second:

> You need to 
> - [fix 1 phone number](https://www.example.com)
> - fill in missing data in 3 rows

tombye:
In that case could linking the summary and rows/cells sbe in the summary error content then?

Like 'shorten the content in rows 1, 2 & 3?'

quis:
Hmm, I feel like that’s changing existing behaviour which people rely on, and may end up being less usable.

The average job size this year to date is 254 rows. So if 5% of your rows had problems then the error could look like: 
> shorten the content in rows 1, 54, 55, 56, 57, 124, 127, 180, 181, 252, 253 and 254

Which is repeating, not summarising, the information in the table below. I think it’s more useful to say ‘you need to fix 12 rows’ and then the user can make a judgement about whether that’s realistic to do on an ad-hoc basis or if there’s something more fundamental that needs addressing.

PR #3412, opened 2020-04-15
Bump utils to bring in recent changes to Template classes

quis:
The main things we have to account for are:
- `WithSubjectTemplate` has gone
- every template dictionary must have `template_type` set

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -13,7 +13,11 @@ from flask import (
 )
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
-from notifications_utils.template import Template, WithSubjectTemplate
+from notifications_utils.template import (
+    EmailPreviewTemplate,
+    LetterPreviewTemplate,
+    SMSPreviewTemplate,
+)
 
 from app import (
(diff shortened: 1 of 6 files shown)


app/main/views/jobs.py
```diff
@@ -402,29 +406,37 @@ def get_job_partials(job):
...
+        return EmailPreviewTemplate(
+            notification['template'],
+            notification['personalisation'],
+            redact_missing_personalisation=True,
+        ).subject
+
+    if notification['template']['template_type'] == 'letter':
+        return LetterPreviewTemplate(
+            notification['template'],
+            notification['personalisation'],
```
idavidmcdonald, line 441: Is it OK that we are not redacting missing personalisation like before for letters?

quis: Redacting personalisation is only for email and text message templates:
https://github.com/alphagov/notifications-admin/blob/4a2b216956bf9d26a624a7cd3487f288690d8ae1/app/templates/views/templates/template.html#L74-L82

Setting it to `True` for `LetterPreviewTemplate` has no effect.

PR #3420, opened 2020-04-21
Let users search for letters

quis:
Like we have search by email address or phone number, finding an individual letter is a common task. At the moment users are having to click through pages and pages of letters to find the one they’re looking for.

Users of the API will also be able to search by reference, same as for emails and text messages. But we only show this hint text to users who have some API keys.

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2814

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -169,8 +169,12 @@ def view_notifications(service_id, message_type=None):
         things_you_can_search_by={
             'email': ['email address'],
             'sms': ['phone number'],
-            'letter': [],
-            None: ['email address', 'phone number'],
+            # This should become ‘postal address’ not ‘first line…’ once
+            # we’ve finished populating normalised addresses
+            'letter': ['first line of address', 'file name'],
+            # We say recipient here because combining all 3 types, plus
+            # reference gets too long for the hint text
+            None: ['recipient'],
         }.get(message_type) + {
             True: ['reference'],
             False: [],
--- a/app/templates/views/notifications.html
+++ b/app/templates/views/notifications.html
@@ -34,41 +34,38 @@
       'counts'
     ) }}
 
-    {% call form_wrapper(
-      action=url_for('.view_notifications', service_id=current_service.id, message_type=message_type),
-      class="govuk-grid-row"
-    ) %}
-      <div class="govuk-grid-column-three-quarters {% if message_type == 'sms' %}extra-tracking{% endif %}">
-        {{ textbox(
-          search_form.to,
-          width='1-1',
-          label=things_you_can_search_by|formatted_list(
-            conjunction='or',
-            before_each='',
-            after_each='',
-            prefix='Search by',
-            prefix_plural='Search by'
-          )
-        ) }}
-      </div>
-      <div class="govuk-grid-column-one-quarter">
-        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
-        {{ govukButton({
-          "text": "Search",
-          "classes": "search-form__button"
-        }) }}
-      </div>
-    {% endcall %}
-
-    {% call form_wrapper(id="search-form") %}
-      <input type="hidden" name="to" value="{{ search_form.to.data }}">
-      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
-    {% endcall %}
-  {% else %}
-    {% call form_wrapper(id="search-form") %}
-      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
-    {% endcall %}
   {% endif %}
+
+  {% call form_wrapper(
+    action=url_for('.view_notifications', service_id=current_service.id, message_type=message_type),
+    class="govuk-grid-row"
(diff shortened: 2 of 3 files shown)


tests/app/main/views/test_activity.py
```diff
@@ -615,10 +626,10 @@ def test_redacts_templates_that_should_be_redacted(
     "message_type, tablist_visible, search_bar_visible", [
         ('email', True, True),
         ('sms', True, True),
-        ('letter', False, False)
+        ('letter', False, True)
     ]
 )
-def test_big_numbers_and_search_dont_show_for_letters(
+def test_big_numbers_dont_show_for_letters(
```
idavidmcdonald, line 632: We can drop the `search_bar_visible` pytest paramed arg from this test as it's always `True` and just assert is `True` in the test

tests/app/main/views/test_activity.py
```diff
@@ -437,7 +448,7 @@ def test_api_users_are_told_they_can_search_by_reference_when_service_has_api_ke
 
 
 @pytest.mark.parametrize('message_type, expected_search_box_label', [
-    (None, 'Search by email address or phone number'),
+    (None, 'Search by recipient'),
```
idavidmcdonald, line 451: Do we need to add a test case for letters here?

app/templates/views/notifications.html
```diff
@@ -34,41 +34,38 @@ <h1 class="heading-large">
...
-    {% endcall %}
+  {% endif %}
 
-    {% call form_wrapper(id="search-form") %}
-      <input type="hidden" name="to" value="{{ search_form.to.data }}">
+  {% call form_wrapper(
+    action=url_for('.view_notifications', service_id=current_service.id, message_type=message_type),
+    class="govuk-grid-row"
+  ) %}
+    <div class="govuk-grid-column-three-quarters {% if message_type == 'sms' %}extra-tracking{% endif %}">
```
idavidmcdonald, line 43: I couldn't work out what the `extra-tracking` styling is doing in 2 minutes so thought I'd ask? And just a double check that it's just SMS and we don't wish to apply this to letters too?

app/templates/views/notifications.html
```diff
@@ -34,41 +34,38 @@ <h1 class="heading-large">
...
+          conjunction='or',
+          before_each='',
+          after_each='',
+          prefix='Search by',
+          prefix_plural='Search by'
+        )
+      ) }}
+    </div>
+    <div class="govuk-grid-column-one-quarter">
       <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
```
idavidmcdonald, line 57: I see that we have two hidden inputs for csrf. It looks like we had this before but I'm not sure whether that was taken care of by an `else` statement meaning we only showed one.

I haven't gone to the trouble of spinning this up so haven't looked into it too throughly but worth just double checking that the hidden csrf stuff is still OK?

There is a reasonable chance this observation is just me misunderstanding the jinja and how this form works but thought I'd get it in before standup.

app/main/views/jobs.py
```diff
@@ -169,8 +169,12 @@ def view_notifications(service_id, message_type=None):
         things_you_can_search_by={
             'email': ['email address'],
             'sms': ['phone number'],
-            'letter': [],
-            None: ['email address', 'phone number'],
+            # This should become ‘postal address’ not ‘first line…’ once
+            # we’ve finished populating normalised addresses
+            'letter': ['first line of address', 'file name'],
+            # We say recipient here because combining all 3 types, plus
```
idavidmcdonald, line 175: Good comments

tests/app/main/views/test_activity.py
quis: Dropped in b01bc17db

tests/app/main/views/test_activity.py
quis: Added in c835674df

app/templates/views/notifications.html
quis: ‘Tracking’ is the typographic term for the spacing between every letter. We add extra tracking for fields that expect numeric input, as detailed here: https://github.com/alphagov/notifications-admin/pull/1545

app/templates/views/notifications.html
quis: It’s the same before and after, one hidden CSRF input per form. When you submit a form it only includes the inputs within that form, not all the inputs on the page. So each form needs it’s own hidden input.

tests/app/main/views/test_activity.py
```diff
@@ -639,7 +650,7 @@ def test_big_numbers_and_search_dont_show_for_letters(
     )
 
     assert (len(page.select("[role=tablist]")) > 0) == tablist_visible
-    assert (len(page.select("[type=search]")) > 0) == search_bar_visible
+    assert (len(page.select("[type=search]")) > 0) is True
```
idavidmcdonald, line 653: Optional: I assume this could just be `assert len(page.select("[type=search]"))`

PR #3421, opened 2020-04-21
Show error when user cannot be archived

klssmith:
A user can't be archived if they are the only member of their service with `manage_settings` permission. `notifications-api` returns a `400` and an error message if that is the case, however this PR to remove the `400` error handler #3320 stopped the error message from showing. This meant that instead of seeing a message about why a user couldn't be archived, we would just show a `500` error page instead. This change checks the response from `notifications-api` and shows an error banner with a message if the user can't be archived.

### DIFF
--- a/tests/app/main/views/test_find_users.py
+++ b/tests/app/main/views/test_find_users.py
@@ -4,6 +4,7 @@ import pytest
 from bs4 import BeautifulSoup
 from flask import url_for
 from lxml import html
+from notifications_python_client.errors import HTTPError
 
 from tests import user_json
 from tests.conftest import normalize_spaces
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_find_users.py
```diff
@@ -221,20 +222,48 @@ def test_archive_user_posts_to_user_client(
...
-        response = platform_admin_client.post(
+        platform_admin_client.post(
             url_for('main.archive_user', user_id=api_user_active.id)
         )
 
-        assert response.status_code == 500
-        assert response.location == url_for('main.user_information', user_id=api_user_active['id'], _external=True)
-        mock_user_client.assert_called_once_with('/user/{}/archive'.format(api_user_active['id']), data=None)
-        assert not mock_events.called
+    assert not mock_events.called
```
leohemsted, line 269: good spot on the badly indented insert 👍

klssmith: I am pretty sure I was the original author of that 😄

PR #3422, opened 2020-04-22
Fix preview of text messages on activity page

quis:
Also beefs up the tests a bit so we can be more confident that the right thing is getting displayed.

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/725

### DIFF
--- a/tests/app/main/views/test_activity.py
+++ b/tests/app/main/views/test_activity.py
@@ -86,6 +86,7 @@ from tests.conftest import (
         ('test@example.com', 'test@example.com'),
     ]
 )
+@freeze_time('2020-01-01 01:00')
 def test_can_show_notifications(
     client_request,
     logged_in_client,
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_activity.py
```diff
@@ -129,13 +130,28 @@ def test_can_show_notifications(
...
-    text_of_first_row = page.select('tbody tr')[0].text
-    assert '07123456789' in text_of_first_row
-    assert (
-        'template content' in text_of_first_row or
-        'template subject' in text_of_first_row
+    first_row = page.select_one('tbody tr')
+    assert normalize_spaces(
+        first_row.select_one('a.file-list-filename.govuk-link').text
+    ) == (
+        '07123456789'
```
tombye, line 137: Can this, and the data on lines 144 and 146 be explained in a comment? Everything else here can be inferred from the test config.

quis: @tombye done in 79b32a1

PR #3430, opened 2020-05-01
Tell template preview whether an uploaded letter can be sent internationally

quis:
If a service has permission to send international letters then the admin app should tell template preview, so that template preview knows what rules to apply when it’s validating the address of the letter.

We don’t need to wait for template preview to start looking at this query string argument – it will just ignore it until alphagov/notifications-template-preview#445 is merged and deployed.

### DIFF
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -115,7 +115,12 @@ def upload_letter(service_id):
         file_location = get_transient_letter_file_location(service_id, upload_id)
 
         try:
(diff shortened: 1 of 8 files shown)


klssmith:
This worked well for me locally, just needs the tests to pass

quis:
@klssmith thanks. Should pass now – looks like there was a dependency which was withdrawn so pip couldn’t install what we were specifying in the requirements file.

PR #3436, opened 2020-05-06
Don’t show pagination links when searching

quis:
The search form makes a `post` request, so that phone numbers and email addresses don’t show up in logs or browser history.

At most the API paginates the results, with a page size of 50. We can’t easily give you pagination links to click in the admin app, because links can only perform `get` requests.

Because the value of seeing more than 50 results feels quite low (users will probably make their search more specific before scrolling through all 50) let’s just show a message saying only the first 50 results are displayed.

***

image


app/templates/views/activity/notifications.html
```diff
@@ -33,6 +33,12 @@
     </div>
   {% endif %}
 
-  {{ previous_next_navigation(prev_page, next_page) }}
+  {% if more_than_can_be_shown %}
+    <p class="table-show-more-link">
+      Only showing the first 50 messages
+    </p>
+  {% else %}
+    {{ previous_next_navigation(prev_page, next_page) }}
```
klssmith, line 41: Why is the `else` needed here? If there are less than 50 messages, I thought that `prev_page` and `next_page` would both be `None` so the macro wouldn't display anything.

quis: Yeah, maybe the variable naming/logic wasn’t the clearest. I changed it in 8655e51, any better?

klssmith: Yep, thanks, I understand the logic now

PR #3441, opened 2020-05-13
Change checkboxes to GOVUK frontend

tombye:
Changes all checkboxes to use the [GOVUK Frontend Checkboxes](https://design-system.service.gov.uk/components/checkboxes/) component.

https://www.pivotaltracker.com/story/show/170030071

## Checkboxes changed by this

This changes checkboxes in the following pages:

- `/platform-admin/live-services`
- `/platform-admin/trial-services`
- `/services/<service_id>/service-settings/email-reply-to/add`
- `/services/<service_id>/service-settings/email-reply-to/edit`
- `/services/<service_id>/service-settings/email-reply-to/<notification_id>/verify`
- `/services/<service_id>/service-settings/letter-contact/add`
- `/services/<service_id>/service-settings/letter-contact/<contact_id>/edit`
- `/services/<service_id>/service-settings/sms-sender/add`
- `/services/<service_id>/service-settings/sms-sender/<sender_id>/edit`
- `/services/<service_id>/users/<user_id>`
- `/services/<service_id>/templates/folders/<template_folder_id>/manage`

### Changes to user permissions

image

The checkboxes on the `/services/<service_id>/users/<user_id>` page were a group of BooleanField checkboxes.

This pull request changes them to be a single set of checkboxes, which required some rewriting of the code in `PermissionsForm` (in `app/main/forms.py`) and the controller for the `/services/<service_id>/users/<user_id>` route.

### Changes to functional tests

~~This requires https://github.com/alphagov/notifications-functional-tests/pull/335 to be merged prior to merging for the functional tests to work with the new HTML it introduces.~~ Now merged.


app/assets/stylesheets/components/checkboxes.scss
```diff
@@ -1,3 +1,7 @@
+// Taken from https://github.com/alphagov/govuk-frontend/blob/v2.13.0/src/components/checkboxes/_checkboxes.scss
```
CrystalPea, line 1: this might be a daft question, but if we pull in components from govuk-frontend, why do we have to also add those size dimensions "manualy" to our repository?

tombye: Not daft at all. It's currently necessary because both the `$govuk-touch-target-size` and `$govuk-checkboxes-size` variables are declared [in a local scope](https://github.com/alphagov/govuk-frontend/blob/v2.13.0/package/components/checkboxes/_checkboxes.scss#L12) (inside the [content block](https://sass-lang.com/documentation/at-rules/mixin#content-blocks) sent to the `govuk-exports` mixin so aren't available to our code.

The Design System team do support an [API for their Sass](https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference) but these variables aren't currently part of it.

CrystalPea: Ahh, ok, so it's something we might be able to connect in the future then, depending on the work Design System do?

PR #3446, opened 2020-05-15
Catch errors when user register from invite

klssmith:
API gives an error if it tries to add a user to a service and that user is
already a member of the service. This situation shouldn't occur - admin checks
if an invited user is a member of a service before calling API, but we
have seen this error occurring when there are two requests processing at
the same time.

This change catches the errors from API if a user is already a member of
a service and redirects the user to the service dashboard so that they
don't see an error page.

https://www.pivotaltracker.com/story/show/172622185


klssmith:
I've changed this now to catch api errors in the model and to make sure we're only adding users to a service from the model.

PR #3448, opened 2020-05-18
Log ID when creating Zendesk ticket

quis:
Members of the public are still getting through to our support channel, and we don’t really understand where they’re coming from. We don’t have a way of tying their support ticket to anything in our logs.

By logging the ID of the ticket when we created it we should be able to work back to the request they made to create the ticket.

From there, we should be able to look at what other requests they made, to understand the journey they took to get to our support channel.

***

Depends on:

- [ ] https://github.com/alphagov/notifications-utils/pull/736

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 import pytz
-from flask import redirect, render_template, request, session, url_for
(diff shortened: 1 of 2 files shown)


quis:
Closing in favour of doing the logging within the utils code

PR #3455, opened 2020-05-19
Don’t show postage choice for international letters

quis:
All instances of the app will be posting to the URL with the `file_id` parameter now, so we can remove the old route without the ID now.

This means we can start getting the PDF metadata before building the form. Which means that we hide or show the postage options on the form based on whether or not the address is international.

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/3454

### DIFF
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -344,20 +344,25 @@ def send_uploaded_letter(service_id, file_id=None):
     if not (current_service.has_permission('letter') and current_service.has_permission('upload_letters')):
         abort(403)
 
+    metadata = get_letter_metadata(service_id, file_id)
+
+    if metadata.get('status') != 'valid':
+        abort(403)
+
+    recipient_address = metadata.get('recipient')
+
     form = LetterUploadPostageForm()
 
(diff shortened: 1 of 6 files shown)


app/main/views/uploads.py
```diff
@@ -338,28 +324,33 @@ def view_letter_upload_as_preview(service_id, file_id):
         return TemplatePreview.from_valid_pdf_file(pdf_file, page)
 
 
-@main.route("/services/<uuid:service_id>/upload-letter/send", methods=['POST'])
 @main.route("/services/<uuid:service_id>/upload-letter/send/<uuid:file_id>", methods=['POST'])
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def send_uploaded_letter(service_id, file_id=None):
```
klssmith, line 329: It's not doing any harm, but you could remove the default argument for `file_id` here if you wanted since it's now always provided

quis:
```suggestion
def send_uploaded_letter(service_id, file_id):
```

PR #3457, opened 2020-05-21
Redirect preview to notifications if the notification already exists in the db

leohemsted:
the upload preview page has a file_id - this corresponds to the file in the transient pdf uploads bucket. However, if the user already hit send (and then navigated back) the file's no longer in that bucket, it's been moved to the regular letters-pdf bucket. So the s3 get request fails. To avoid this, simply redirect to the notifications page if the notification already exists. This is better for the user as it'll stop them trying to submit it twice, and will provide more clarity on the status of the notification too


app/main/views/uploads.py
```diff
@@ -273,6 +274,20 @@ def format_recipient(address):
...
+        # if the call succeeds, we've already persisted to the db - the user has hit send, and the letter won't be in
+        # the transient bucket any more. So lets just bounce out to the notification page
+        return redirect(url_for(
+            '.view_notification',
+            service_id=service_id,
+            notification_id=file_id,
+        ))
+    except HTTPError as e:
+        # if it's a 404, that means we haven't saved it yet
+        if not (e.status_code == 404 and e.message == 'No result found'):
```
idavidmcdonald, line 288: Can we drop the check of the message please? I think it's adds no more value than checking the status code and adds more code upkeep if we end up changing error messages in the future

Also I had two other thoughts in this area which I'm 50/50 on so keen to get your thoughts.

- Should we reraise if it's not a 404 or should we just continue?
- Should we check the notification on the start for everytime we use this view or should we instead move to only checking it if the file doesn't exist in s3, this would improve page load times and reduce the number of unnecessary API calls

leohemsted: ### status check

Agreed, happy to drop it, think you're right about that making it a bit more brittle and not adding value.


### reraising on non 404s

The most likely case is a 5xx from the API (probably a 502). Then it'll fail fast and noisy. I think that's preferrable, rather than potentially 500ing immediately when trying to get a letter that might not exist from s3. Also means we'll get logs for those situtations.

### checking s3 vs checking api

I did briefly consider this. My chosen solution was driven by simplicity, especially while working with limited internet access on thurs 😄. Now that I'm thinking about it in the light of day, and now that I've had the chance to actually look at the real error message, I think this is a nicer way of handling things. Updated my PR.

app/main/views/uploads.py
```diff
@@ -257,7 +258,20 @@ def _get_error_from_upload_form(form_errors):
...
     re_upload_form = PDFUploadForm()
 
-    metadata = get_letter_metadata(service_id, file_id)
+    try:
+        metadata = get_letter_metadata(service_id, file_id)
+    except ClientError as e:
+        # if the file's not there, it's probably because we've already created the notification and the letter has been
+        # moved to the normal letters-pdf bucket. So lets just bounce out to the notification page
+        if e.response['Error']['Code'] == 'NoSuchKey':
+            return redirect(url_for(
```
idavidmcdonald, line 267: Optional: I feel (but not super strongly) that we should not blindly redirect, just in case for some reason there is a problem in our system and the file doesn't exist in the bucket but the notification has not progressed. It would be more watertight to then call the API and check the status of the notification. Thoughts?

leohemsted: hmmm, i think at that point i don't really care. if the file isn't where we expect, we would previously have 500'd. If we now redirect to notifications and then that 500s the user sees the same thing, we still get the error in the logs. We create the notification before moving the file in the api, so under normal circumstances this could never happen.

PR #3458, opened 2020-05-22
Bump utils to 39.4.2

quis:
Adds delivery estimates for letters posted to Europe or the rest of the world.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -443,7 +443,8 @@ def notification_json(
     if status is None:
         status = 'delivered'
     links = {}
-    postage = postage or 'second'
+    if template_type == 'letter':
+        postage = postage or 'second'
 
     if with_links:
         links = {
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -443,8 +443,7 @@ def notification_json(
     if status is None:
         status = 'delivered'
     links = {}
-    if template_type == 'letter':
-        postage = postage or 'second'
+    postage = postage or 'second'
 
     if with_links:
         links = {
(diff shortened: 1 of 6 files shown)


tests/app/main/views/test_notifications.py
```diff
@@ -530,13 +530,13 @@ def test_notification_page_does_not_show_cancel_link_for_letter_which_cannot_be_
...
         'letter-postage-international',
-        'Estimated delivery date: Wednesday 6 January',
+        'Estimated delivery date: Friday 8 January',
     ),
     (
         'rest-of-world',
         'Postage: international',
         'letter-postage-international',
-        'Estimated delivery date: Wednesday 6 January',
+        'Estimated delivery date: Monday 11 January',
```
idavidmcdonald, line 539: Very optional: Not to do with your code but you may consider refactoring so that all assertions about postal classes and estimated delivery dates are in a single test. Whether that means moving these three cases into `test_notification_page_shows_page_for_letter_notification` or moving the assertions out of `test_notification_page_shows_page_for_letter_notification` into here I don't mind.

This test structure just seems slightly off to have the different cases belonging in a different test

tests/__init__.py
```diff
@@ -443,8 +443,7 @@ def notification_json(
     if status is None:
         status = 'delivered'
     links = {}
-    if template_type == 'letter':
-        postage = postage or 'second'
+    postage = postage or 'second'
```
idavidmcdonald, line 446: Could you explain why we need to drop the `if` statement please?

quis: Our mock_get_notifications fixture tries to be too clever and work out which fields to return based on what arguments it’s called with. But we don’t specify the `notification_type` when calling `get_notifications` for a job. So I was trying to get around this by returning postage for all notification types. Which is a bit hacky.

In other places we call the mocking code directly with arguments that make it more explicit what the mocked response should be. I’ve changed this PR to do the same for the tests that we’d otherwise have to make changes to the fixture for, because postage can no longer be `None` for letter notifications.

That change is in f40984f

PR #3459, opened 2020-05-22
Delete cached users and templates when archiving a service

idavidmcdonald:
https://www.pivotaltracker.com/story/show/172955889

When the admin app gets user objects from the API, these include a dict
of permissions by service for what the user can do to that services.
Permissions for inactive services are not included in the response as
per:
https://github.com/alphagov/notifications-api/blob/87cb6f25974d394a8fa1b8f9cd7abe8aa28e05a7/app/dao/permissions_dao.py#L66

However, this causes a bug where a service is archived but cached user
data still tells us that the user has permissions to view the service.
This should not be the case and causes errors where users can still see
the archived service page, it's settings, and even request to go live
for it, because they are using old cached data for the user.

We solve this by deleting the users who are part of the service from the
cache.

We also delete the templates for this service from the cache as the
templates are also archived when we ask the API to archive the service
as per:
https://github.com/alphagov/notifications-api/blob/d95c0131e09efb87e535d3b57e49c502824dcabb/app/service/rest.py#L597

Note, one decision I had to make was whether to delete the user cache
for just active team members or also invited users. Assuming an invited
user can't see the service until they've accepted their invite anyway, it
shouldn't make any difference whether we delete their cache or not.

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -130,11 +130,9 @@ class ServiceAPIClient(NotifyAdminAPIClient):
 
     @cache.delete('service-{service_id}')
     @cache.delete('service-{service_id}-templates')
-    def archive_service(self, service_id, service_users):
-        # We need to purge the cache for the services users as otherwise, although they will have had their permissions
-        # removed in the DB, they would still have permissions in the cache to view/edit/manage this service
-        for user in service_users:
-            cache.delete(f'user-{user.id}')
+    def archive_service(self, service_id, cached_service_user_ids):
+        if cached_service_user_ids:
+            cache.delete(*map('user-{}'.format, cached_service_user_ids))
         return self.post('/service/{}/archive'.format(service_id), data=None)
 
     @cache.delete('service-{service_id}')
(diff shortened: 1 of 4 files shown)


app/notify_client/service_api_client.py
```diff
@@ -129,7 +129,12 @@ def update_service_with_properties(self, service_id, properties):
         return self.update_service(service_id, **properties)
 
     @cache.delete('service-{service_id}')
-    def archive_service(self, service_id):
+    @cache.delete('service-{service_id}-templates')
+    def archive_service(self, service_id, service_users):
+        # We need to purge the cache for the services users as otherwise, although they will have had their permissions
+        # removed in the DB, they would still have permissions in the cache to view/edit/manage this service
+        for user in service_users:
+            cache.delete(f'user-{user.id}')
```
quis, line 137: Bit worried about how performant this would be on a service with many users, since this will do a synchronous network call for each user. 

The way we’ve done this elsewhere is by calling the delete method once with multiple keys:
https://github.com/alphagov/notifications-admin/blob/b98f4561fabc00ed1e9abcedfdee29005cdf3e5d/app/notify_client/organisations_api_client.py#L56-L57

idavidmcdonald: Oo, wasn't aware of this. Very nice. I've pushed up a new commit that does this and also changes the layout of my code to match the pattern of passing the ids to cache through as a list

PR #3460, opened 2020-05-22
Handle session expiring during service name change

CrystalPea:
We saw some errors earlier where session expired midway through service name change. This PR handles such edge cases.

Here is the fresh redirect with flash error message:
Screenshot 2020-05-22 at 15 34 21

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -112,6 +112,10 @@ def service_name_change(service_id):
 @main.route("/services/<uuid:service_id>/service-settings/name/confirm", methods=['GET', 'POST'])
 @user_has_permissions('manage_service')
 def service_name_change_confirm(service_id):
+    if 'service_name_change' not in session:
+        flash("Session expired. Try again", 'error')
+        return redirect(url_for('main.service_name_change', service_id=service_id))
+
     # Validate password for form
     def _check_password(pwd):
         return user_api_client.verify_password(current_user.id, pwd)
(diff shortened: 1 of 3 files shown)


CrystalPea:
@servingUpAces not yet, I will ask Karl

app/main/views/service_settings.py
```diff
@@ -112,6 +112,10 @@ def service_name_change(service_id):
 @main.route("/services/<uuid:service_id>/service-settings/name/confirm", methods=['GET', 'POST'])
 @user_has_permissions('manage_service')
 def service_name_change_confirm(service_id):
+    if 'service_name_change' not in session:
+        flash("Session expired. Try again", 'error')
```
CrystalPea, line 116: Hey Karl,

What do you think about this error message? For context it's an error a user would see when they try to change service name, then forget about it during the confirmation step. And by the time they get back to it, the session has expired and we don't know what they wanted to change the service name to anymore.

CrystalPea, line 116: Hey @karlchillmaid 
What do you think about this error message? For context it's an error a user would see when they try to change service name, then forget about it during the confirmation step. And by the time they get back to it, the session has expired and we don't know what they wanted to change the service name to anymore.

karlchillmaid: We've discussed the content and the journey users go through before seeing this error.

'Session expired' or similar makes this sound like a new error. This might confuse the user and make them think their sign in didn't work and that the session has expired again.

Suggested content: `The change you made was not saved. Please try again.`

PR #3462, opened 2020-05-26
Hide pages from search engines using a meta tag instead of robots.txt

quis:
Google’s documentation says:

> robots.txt is not a mechanism for keeping a web page out of Google. To keep a web page out of Google, you should use noindex directives

A noindex directive means adding the following meta tag to pages that shouldn’t be indexed:
```html
<meta name="robots" content="noindex" />
```

It’s also possible to set the directive as a HTTP header, but this seems trickier to achieve on a per-view basis in Flask.

I’ve implemented this as a decorator so it can quickly be added to any other pages that we decide shouldn’t appear in search results.

This PR also adds `/features/email` to the URLs that shouldn’t be indexed, as per alphagov/notifications-aws#796


idavidmcdonald:
Just wondering whether we need to be keeping the robots.txt file for the other search engines that will be scraping us? Or maybe non search engine robots? Did you have any thoughts about this?

quis:
@idavidmcdonald the HTTP header is supported by Bing, Google and Yahoo.

I did manage to find a Czech search engine which [doesn’t support the HTTP header but does support the `<meta>` tag](https://napoveda.seznam.cz/cz/fulltext-hledani-v-internetu/meta-tag-robots/). So I’ve amended this PR to add both. I don’t think it’s worth the maintenance overhead of keeping the `robots.txt` file as well.

quis:
Also worth noting that using `robots.txt` in combination with a meta tag or HTTP header means the page _can_ still show up in search results: 

> For the noindex directive to be effective, the page must not be blocked by a robots.txt file. If the page is blocked by a robots.txt file, the crawler will never see the noindex directive, and the page can still appear in search results, for example, if other pages link to it.

— https://support.stackpath.com/hc/en-us/articles/360001464066-EdgeRules-Adding-an-x-robots-tag-Header

tests/app/main/views/test_index.py
```diff
@@ -62,18 +63,36 @@ def test_logged_in_user_redirects_to_choose_account(
...
+    ('feedback', {'ticket_type': 'general'}),
+    ('feedback', {'ticket_type': 'report-problem'}),
+    ('bat_phone', {}),
+    ('thanks', {}),
+    ('register', {}),
+    ('features_email', {}),
+    pytest.param('index', {}, marks=pytest.mark.xfail(raises=AssertionError)),
+))
+@freeze_time('2012-12-12 12:12')  # So we don’t go out of business hours
+def test_hiding_pages_from_search_engines(
```
idavidmcdonald, line 85: Optional: A test of the inverse, e.g. pick a route that is able to found by search engines and check we aren't sending the header and we don't include the meta tag

PR #3440, opened 2020-05-12
Show jobs grouped within the contact list they were sent from

quis:
image

***

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2841
- [x] https://github.com/alphagov/notifications-api/pull/2855

PR #3464, opened 2020-05-28
Update contact list content

karlchillmaid:
This PR makes the naming of the emergency contact lists (and what you do with them) more consistent.

### DIFF
--- a/app/templates/views/send-contact-list.html
+++ b/app/templates/views/send-contact-list.html
@@ -36,7 +36,7 @@
         caption="Emergency contact lists",
         caption_visible=False,
         empty_message=(
-          'You have not uploaded any emergency contact lists.'
+          'You have not saved any emergency contact lists.'
         ),
         field_headings=[
           'File',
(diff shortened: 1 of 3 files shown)


app/templates/views/send-test.html
```diff
@@ -33,7 +33,7 @@
           {% if link_to_upload %}
             <a class="govuk-link govuk-link--no-visited-state govuk-!-margin-right-3" href="{{ url_for('.send_messages', service_id=current_service.id, template_id=template.id) }}">Upload a list of {{ recipient_count_label(999, template.template_type) }}</a>
             {% if current_service.contact_lists %}
-              <a class="govuk-link govuk-link--no-visited-state govuk-!-margin-right-3" href="{{ url_for('.choose_from_contact_list', service_id=current_service.id, template_id=template.id) }}">Use a saved list</a>
+              <a class="govuk-link govuk-link--no-visited-state govuk-!-margin-right-3" href="{{ url_for('.choose_from_contact_list', service_id=current_service.id, template_id=template.id) }}">Use an emergency list</a>
```
karlchillmaid, line 36: @quis Which do you prefer?

'emergency list' or 'saved list'?

app/templates/views/send-contact-list.html
```diff
@@ -33,10 +33,10 @@
     <div class='dashboard-table ajax-block-container'>
       {% call(item, row_number) list_table(
         contact_lists,
-        caption="Existing contact lists",
+        caption="Emergency contact lists",
         caption_visible=False,
         empty_message=(
-          'You don’t have any contact lists yet'
+          'You have not saved any contact lists yet.'
```
karlchillmaid, line 39: @quis It's probably not necessary to repeat 'emergency' here, right?

quis:
I’m going to wait until after https://github.com/alphagov/notifications-admin/pull/3440 is finished to look at this, so we can review the usage as a whole.

app/templates/views/send-test.html
quis: Emergency for consistency 👍🏻

app/templates/views/send-contact-list.html
quis: This text is never shown to the user. When the uploads page is empty they see this:

image

quis (approved):

PR #3465, opened 2020-05-28
Add scroll to reveal element method

tombye:
Adds a public `scrollToRevealElement` method to:
- `window.GOVUK.stickAtTopWhenScrolling`
- `window.GOVUK.stickAtBottomWhenScrolling`

...to show an element if the sticky nav is covering it.

## Why

At the moment this is for use in the functional tests, to make sure any elements we want to interact with are visible.

It could be useful in future to have this functionality and doesn't change the use of this code to reveal focused elements.

Most of the work for this was done in [the PR to make sure focused elements aren't covered by the sticky nav](https://github.com/alphagov/notifications-admin/pull/2843), it's mainly making that code part of the public API.

### DIFF
--- a/app/assets/javascripts/stick-to-window-when-scrolling.js
+++ b/app/assets/javascripts/stick-to-window-when-scrolling.js
@@ -10,7 +10,10 @@
     var $el = el.$fixedEl;
     var $scrollArea = $el.closest('.sticky-scroll-area');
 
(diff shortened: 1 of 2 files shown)


tombye:
This could be useful but not having it isn't blocking anything so I'm gonna close it for now.

tombye:
Re-opening this PR as it's needed to help make some functional tests pass.

PR #3466, opened 2020-05-29
Let users switch international letters on and off

quis:
We’ve decided that this feature should be self-service, and on by default for all users. Before we can make it on by default we should give users a way to switch it off for themselves. Which is what this PR does.

***

image

***

image

***

Depends on:
- [ ] being able to send international letters for real


app/templates/views/service-settings/set-international-letters.html
```diff
@@ -0,0 +1,33 @@
...
+
+{% block maincolumn_content %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-five-sixths">
+      {{ page_header(
+        'Send international letters',
+        back_link=url_for('main.service_settings', service_id=current_service.id)
+      ) }}
+      <p>
```
idavidmcdonald, line 19: Do we want `class="govuk-body"` on these paragraphs?

quis: Yes. Going to address this more widely in a separate PR.

quis: See https://github.com/alphagov/notifications-admin/pull/3467

PR #3471, opened 2020-06-05
Update permanent failure error message content

karlchillmaid:
This PR updates the error and description content for a permanent failure.

The current error states that a permanent failure happens when a "phone number does not exist". This isn't always the case. Permanent failures can happen when numbers are blacklisted, not accepting text messages, or if the network operator rejects the message.

We've been contacted by users who have received this error when sending to numbers that definitely *do* exist.

Until we can offer more specific reasons for the failure, we're going to update the content to include some of the other possible causes.

### DIFF
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -55,7 +55,7 @@
         ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
         ('Sent internationally', 'The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information.'),
         ('Delivered', 'The message was successfully delivered. Notify will not tell you if a user has opened or read a message.'),
-        ('Phone number does not exist', 'The provider could not deliver the message because the phone number was wrong. You should remove these phone numbers from your database. You’ll still be charged for text messages to numbers that do not exist.'),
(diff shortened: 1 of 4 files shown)


app/templates/views/message-status.html
```diff
@@ -55,7 +55,7 @@ <h2 id="sms-statuses" class="heading-medium">Text messages</h2>
         ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
         ('Sent internationally', 'The message was sent to an international number. The mobile networks in some countries do not provide any more delivery information.'),
         ('Delivered', 'The message was successfully delivered. Notify will not tell you if a user has opened or read a message.'),
-        ('Phone number does not exist', 'The provider could not deliver the message because the phone number was wrong. You should remove these phone numbers from your database. You’ll still be charged for text messages to numbers that do not exist.'),
+        ('Not delivered', 'The provider could not deliver the message. This can happen if the phone number was wrong or if the network operator rejects the message. You should check that these phone numbers are correct. If not, you should remove them from your database. You’ll still be charged for text messages that cannot be delivered.'),
```
idavidmcdonald, line 58: I'm not a word guru so don't take this as an official github approval but this does look really good.

The only thing I'd consider adding would be to encourage them to reach out to our support channel if they believe the number was correct and should have been delivered successfully? Pros: We will know about when our providers may be having issues, can get our providers to fix individual routing issues and can give the user more information about why it failed. Cons: May increase support tickets

karlchillmaid: Thanks for the suggestion @idavidmcdonald. I'll have a chat with JP about this.

PR #3475, opened 2020-06-09
Allow invite form to be pre-populated with an existing user

quis:
At the moment users must be invited to join a service. But this means:
- users must know that a service already exists
- they need to know who to ask for an invite

If the user doesn’t know these thing then sometimes they just go ahead and set up a new service. Which means they have to get all the way to the point of requesting to go live before we tell them that there’s already a service with a similar name or purpose.

So we should let users:
1. discover what other services exist in their organisation
2. apply to join a service
3. automatically notify the service managers of their interest
4. be invited by a service manager
5. accept the invite

This commit implements step 4. We can just link them to the invite form in step 3., but we should make it easy for them to send the invite, without having to copy and paste email addresses.

So this commit let the invite form be pre-populated with an existing user’s email address.

image

### DIFF
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -70,6 +70,10 @@ def invite_user(service_id, user_id=None):
                 'views/user-already-invited.html',
                 user_to_invite=user_to_invite,
             )
+        if not user_to_invite.default_organisation:
+            abort(403)
+        if user_to_invite.default_organisation.id != current_service.organisation_id:
+            abort(403)
         form.email_address.data = user_to_invite.email_address
     else:
         user_to_invite = None
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -65,7 +65,10 @@ def invite_user(service_id, user_id=None):
 
     if user_id:
         user_to_invite = User.from_id(user_id)
-        if user_to_invite.belongs_to_service(current_service.id):
+        if (
+            user_to_invite.belongs_to_service(current_service.id)
+            or current_service.invite_pending_for(user_to_invite.email_address)
+        ):
             return render_template(
                 'views/user-already-invited.html',
                 user_to_invite=user_to_invite,
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -65,10 +65,12 @@ def invite_user(service_id, user_id=None):
 
     if user_id:
         user_to_invite = User.from_id(user_id)
-        if (
-            user_to_invite.belongs_to_service(current_service.id)
-            or current_service.invite_pending_for(user_to_invite.email_address)
-        ):
+        if user_to_invite.belongs_to_service(current_service.id):
+            return render_template(
+                'views/user-already-team-member.html',
+                user_to_invite=user_to_invite,
+            )
+        if current_service.invite_pending_for(user_to_invite.email_address):
             return render_template(
                 'views/user-already-invited.html',
                 user_to_invite=user_to_invite,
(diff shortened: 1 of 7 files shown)


app/templates/views/user-already-invited.html
```diff
@@ -0,0 +1,19 @@
+{% extends "withnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+
+{% block service_page_title %}
+  This person is already a team member
```
CrystalPea, line 5: Has @karlchillmaid looked at this content yet? I understand it will show both when user is already a team member and also when they have been invited but have not accepted?

What will happen if the invite link goes stale? I remember it's only valid some time?

CrystalPea (commented):
Ohh, it's cool we are proceeding with parts of this new flow ☺️ 

Good work, although I am not sure about the organisation checks, seems like they might cause trouble for some edge cases.

app/main/views/manage_users.py
```diff
@@ -62,6 +63,24 @@ def invite_user(service_id):
...
+        user_to_invite = User.from_id(user_id)
+        if (
+            user_to_invite.belongs_to_service(current_service.id)
+            or current_service.invite_pending_for(user_to_invite.email_address)
+        ):
+            return render_template(
+                'views/user-already-invited.html',
+                user_to_invite=user_to_invite,
+            )
+        if not user_to_invite.default_organisation:
```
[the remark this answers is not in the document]
quis: Contractors create their own services or sign up for an account. To start with this feature should be restricted in the same way. I don’t think we want contractors to be able to browse around and see which services they might like to join.

app/templates/views/user-already-invited.html
quis: Content needs looking at. Maybe we need a separate page/message for ‘already invited’ vs ‘already a team member’. Or the ‘already invited‘ message could cover both?

If the link goes stale then `current_service.invite_pending_for(user_to_invite.email_address)` will return `False` so we’re OK there.

app/main/views/manage_users.py
```diff
@@ -62,6 +63,24 @@ def invite_user(service_id):
...
+            abort(403)
+        if user_to_invite.default_organisation.id != current_service.organisation_id:
```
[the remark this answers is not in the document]
quis: They can still do that if they know the person’s email address.

This check stops people modifying the URL to find out the email addresses of a user in another organisation from their `user_id`

app/main/views/manage_users.py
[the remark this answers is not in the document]
CrystalPea: Good point if this will only be used for that flow.

app/main/views/manage_users.py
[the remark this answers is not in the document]
CrystalPea: ahh, ok, that's ok then 👍🏼

app/templates/views/user-already-invited.html
CrystalPea: Sweet, thanks a lot!

> Content needs looking at. Maybe we need a separate page/message for ‘already invited’ vs ‘already a team member’. Or the ‘already invited‘ message could cover both?

Yeah, not sure what would be the best option here 🤔

CrystalPea (approved):
Good to go after content check ☺️

app/templates/views/user-already-invited.html
karlchillmaid: 2 messages seems like a good idea. How about:

>**This person is already a team member**
>
>((user)) is already a member of ‘((service-name))’.

and

>This person has already received an invite
>
>((user)) has not accepted their invitation to ‘((service-name))’ yet.
>You do not need to do anything.

quis: I’ve added the two separate messages in fbd58ef

PR #3477, opened 2020-06-11
cache api keys

leohemsted:
this isn't for the admin app's benefit so much as the api. but by setting it here we ensure it gets revoked when a service is deactivated or a key is revoked


leohemsted:
Closing this as redis doesn't help us for API keys: the json that the api sends to the admin doesn't include api_key.secret (nor should it!) - and that's exactly what we need to cache for the api requests to save time.

PR #3478, opened 2020-06-12
Don’t treat self arg of cache-decorated functions as special

quis:
Our API client caching code assumes that what it’s decorating is a bound method, i.e. it’s first argument is `self`. It then does some trickery to make sure that you can’t use the value of `self` as part of the cache key. While using the value of `self` as part of the cache key would be
useless (because when turned to a string it contains the address of the object in memory, meaning it would not be a long-lived key) it does also stop us from using this decorator on plain functions.

This commit removes the handling of `self` as a special case, and adds some tests proving that it can now be used on plain functions.

This will be useful for when we extract this code for use in the API.

### DIFF
--- a/app/notify_client/cache.py
+++ b/app/notify_client/cache.py
@@ -16,7 +16,7 @@ def _get_argument(argument_name, client_method, args, kwargs):
 
     with suppress(ValueError, IndexError):
         argument_index = list(signature(client_method).parameters).index(argument_name)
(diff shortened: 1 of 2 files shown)


quis:
Probably don’t need to merge this, it just explains why the code going into https://github.com/alphagov/notifications-utils/pull/748 is a bit different

PR #3485, opened 2020-06-16
Remove term ‘blacklist’ from codebase

quis:
‘Commonly used passwords’ is more specific, and avoids the terminology ‘blacklist’ which the National Cyber Security Centre explain to be problematic:

> It's fairly common to say whitelisting and blacklisting to describe desirable and undesirable things in cyber security. For instance, when talking about which applications you will allow or deny on your corporate network; or deciding which bad passwords you want your users not to be able to use.
>
> However, there's an issue with the terminology. It only makes sense if you equate white with 'good, permitted, safe' and black with 'bad, dangerous, forbidden'. There are some obvious problems with this. So in the name of helping to stamp out racism in cyber security, we will avoid this casually pejorative wording on our website in the future. No, it's not the biggest issue in the world - but to borrow a slogan from elsewhere: every little helps.

– https://www.ncsc.gov.uk/blog-post/terminology-its-not-black-and-white

### DIFF
--- a/app/main/_commonly_used_passwords.py
+++ b/app/main/_commonly_used_passwords.py
@@ -1,4 +1,4 @@
-blacklisted_passwords = [
+commonly_used_passwords = [
     'govuknotify',
     'GOVUKnotify',
     'GOV.UK Notify',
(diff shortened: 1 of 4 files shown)


quis:
(well spotted @karlchillmaid)

PR #3486, opened 2020-06-17
Update guest list introduction

karlchillmaid:
This PR adds a description of what the guest list is to the _Guest list_ page.

It also changes the explanation of who is automatically included in the guest list.

# After
Screenshot 2020-06-17 at 13 30 36

### DIFF
--- a/app/templates/views/api/guest-list.html
+++ b/app/templates/views/api/guest-list.html
@@ -40,7 +40,7 @@
   {% endif %}
 
   <p class="govuk-body">
-    You can use the team and guest list key to send real messages to your team and up to 5 other email addresses or phone numbers.
+    You can use a team and guest list key to send real messages to your team and up to 5 other email addresses or phone numbers.
   </p>
 
   <p class="govuk-body">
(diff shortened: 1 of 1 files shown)


app/templates/views/api/guest-list.html
```diff
@@ -40,9 +40,11 @@ <h1 class='banner-title'>
   {% endif %}
 
   <p class="govuk-body">
-    You and members of
-    <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.manage_users', service_id=current_service.id) }}">your team</a>
-    are included in the guest list automatically.
+    You can use the team and guest list key to send real messages to your team and up to 5 other email addresses or phone numbers.
```
karlchillmaid, line 43: @quis 

*the* team and guest list key? Or *a* team and guest list key?

quis: _a_ – you can have more than one

karlchillmaid: I've changed key to 'a' because you can have more than one – but I've kept guest list as 'the', because there's only one list of 5 emails and phone numbers.

PR #3490, opened 2020-06-22
Stop expecting letter contact block in service JSON

quis:
We’re removing it for performance reasons: https://github.com/alphagov/notifications-api/pull/2890

This means removing the old pages that edited the letter contact block when it was stored directly on the service, rather than the current model where a service can have multiple contact blocks.


quis:
@servingUpAces fixed now, thanks

PR #3492, opened 2020-06-24
Switch off statsd

quis:
image


idavidmcdonald:
I've removed statsd completely from the admin app in https://github.com/alphagov/notifications-admin/pull/3495 so this PR is now redundant

PR #3500, opened 2020-07-03
Add flow for choosing an area to broadcast to 

quis:
These are just so we have some pages to click through for now. They don’t use real templates, or any of the broadcast stuff from the database.

But I think it’s useful to get some skeleton pages in first so that we can see the map etc working, then build on that, without having to do it all in one mega PR. It also means we can make progress on the admin side of things and API side independently.

For that reason there are two short term things I’ve done in this commit which should be revisited soon:
- no tests for the endpoints (because they’ll all need re-writing anyway)
- data about which areas are selected is stored in the session (because we don’t have CRUD endpoints in the API for this stuff yet)

No services have the `broadcast` permission yet, so existing users won’t be able to see these pages.

***

image

***

image

***

image

***

image

***

image

***

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/3501
- [x] https://github.com/alphagov/notifications-utils/pull/758

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -612,7 +612,8 @@ def useful_headers_after_request(response):
         "connect-src 'self' *.google-analytics.com;"
         "object-src 'self';"
         "font-src 'self' {asset_domain} data:;"
-        "img-src 'self' {asset_domain} *.google-analytics.com *.notifications.service.gov.uk {logo_domain} data:;"
+        "img-src 'self' {asset_domain} *.tile.openstreetmap.org *.google-analytics.com"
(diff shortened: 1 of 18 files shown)


quis:
@CrystalPea None of these endpoints are talking to the API yet, so there’s nothing to mock. But I liked your idea of  placeholder tests, so I’ve added:
- a test that checks that each endpoint is only accessible to services that have the `broadcast` permission
- a test for each endpoint that checks it loads with a `200` or redirect, when the service does have the correct permission

PR #3501, opened 2020-07-03
Add Leaflet.js for rendering slippy maps

quis:
[Leaflet](https://leafletjs.com/) seems to be the go-to library for rendering maps these days. It will be useful for the broadcast work.

This commit add the Leaflet Javascript and CSS to our asset pipeline. The Javascript is already minified so all we need to do is copy it. The CSS is uncompressed so we put it through the same pipe as our other stylesheets.

I’m keeping these as separate files because they’re quite heavy (or the JS is at least – 38kb minified) so I want them to only be loaded on the pages where they’re used. Most users of Notify will never need to see a map.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -95,6 +95,12 @@ const copy = {
         )
       });
     }
+  },
+  leaflet: {
+    js: () => {
+      return src(paths.npm + 'leaflet/dist/leaflet.js')
+        .pipe(dest(paths.dist + 'javascripts/'))
+    }
   }
 };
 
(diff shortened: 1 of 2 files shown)


quis:
Copying Slack discussion for posterity:
> We could probably do with adding a few things for accessibility at some point. I’d guess:
> - making the focus state of the controls match ours
> - using this plugin to add controls for panning the map: http://kartena.github.io/Leaflet.Pancontrol/
> - giving the map a role of image and a label explaining the contents

PR #3511, opened 2020-07-08
Add some styling to broadcast messages

quis:
This is borrowed from an earlier prototype, and is meant to be temporary – something better than just plain text.

Text in generated content isn’t always announced by screen readers, so we should definitely move away from that once we understand what text will be shown on the phone and where it comes from.

Before | After 
---|---
image | image
image | image


quis:
Here’s what the map page would look like with similar styling to the preview page. Don’t think it’s quite right, but it is closer to the [tag component](https://design-system.service.gov.uk/components/tag/) in the Design System

Screenshot 2020-07-08 at 17 51 41

PR #3512, opened 2020-07-09
Make the flow to create a broadcast talk to the API

quis:
This commit removes the code the puts areas into the session and instead creates and then updates a draft broadcast in the database.

This is so we can avoid session-related bugs, and potentially having a large session when we start adding personalisation etc.

Once a broadcast is ready to go it is set to `broadcasting` straight away with no approval. We’ll revisit this as we learn more about how users might want to manage who can create and approve broadcasts.

The tests are a bit light in terms of checking what’s on the page, but clicking through the pages is probably good enough for now.

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/2917

### DIFF
--- a/app/notify_client/broadcast_message_api_client.py
+++ b/app/notify_client/broadcast_message_api_client.py
@@ -1,0 +1,46 @@
+from app.notify_client import NotifyAdminAPIClient, _attach_current_user
+
+
+class BroadcastMessageAPIClient(NotifyAdminAPIClient):
+
+    def create_broadcast_message(
+        self,
+        *,
+        service_id,
+        template_id,
+    ):
+        data = {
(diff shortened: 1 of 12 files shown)


app/notify_client/broadcast_message_api_client.py
```diff
@@ -0,0 +1,45 @@
...
+            data=data,
+        )
+
+    def update_broadcast_message_status(self, status, *, service_id, broadcast_message_id):
+        self.post(
+            f'/service/{service_id}/broadcast-message/{broadcast_message_id}/status',
+            data={
+                'status': status,
+            },
+        )
```
leohemsted: this needs the `_attach_current_user` treatment

PR #3514, opened 2020-07-09
Broadcast dashboard

quis:
Add broadcasts to the dashboard

Shows current and previous broadcasts with a similar pattern to the uploads page.

Does not add a page for viewing an individual broadcast or cancelling broadcasts.

Broadcasts are split into live and previous. Draft broadcasts are excluded from the dashboard.

***

Screenshot 2020-07-09 at 15 51 10


yahoopete:
For accounts with broadcast permissions they can't send email, sms or letters (to avoid confusion). So there won't be any uploads, costs (so usage not needed), and API integration for broadcast will come later. So this just removes those nav options that can't be used.

PR #3515, opened 2020-07-09
Update usage page to display international letters

klssmith:
The usage page splits letters by postage and rate. The API returns postage as `europe` or `rest-of-world`, so the data for international letters has to be aggregated into a new `international` category before we can display it.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -1,4 +1,5 @@
 import calendar
+from collections import namedtuple
 from datetime import datetime
 from functools import partial
 from itertools import groupby
@@ -416,15 +417,21 @@ def get_free_paid_breakdown_for_billable_units(year, free_sms_fragment_limit, bi
             free_sms_fragment_limit, cumulative, previous_cumulative,
             [billing_month for billing_month in sms_units if billing_month['month'] == month]
         )
-        letter_billing = [(x['billing_units'], x['rate'], (x['billing_units'] * x['rate']), x['postage'])
+
+        LetterDetails = namedtuple('LetterDetails', ['billing_units', 'rate', 'cost', 'postage'])
+
+        letter_billing = [LetterDetails(billing_units=x['billing_units'],
+                                        rate=x['rate'],
+                                        cost=(x['billing_units'] * x['rate']),
+                                        postage=x['postage'])
                           for x in letter_units if x['month'] == month]
 
         if letter_billing:
-            letter_billing.sort(key=lambda x: (x[3], x[1]))
+            letter_billing.sort(key=lambda x: (x.postage, x.rate))
 
         letter_total = 0
         for x in letter_billing:
-            letter_total += x[2]
+            letter_total += x.cost
             letter_cumulative += letter_total
         yield {
             'name': month,
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -456,7 +456,7 @@ def format_letter_details_for_month(letter_units_for_month):
 
         letter_details = LetterDetails(
             billing_units=sum(x['billing_units'] for x in rate_group),
-            rate=format_letter_rate(rate_group[0]['rate']),
+            rate=rate_group[0]['rate'],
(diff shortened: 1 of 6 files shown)


app/main/views/dashboard.py
```diff
@@ -436,6 +436,50 @@ def get_free_paid_breakdown_for_billable_units(year, free_sms_fragment_limit, bi
...
+    LetterDetails = namedtuple('LetterDetails', ['billing_units', 'rate', 'cost', 'postage_description'])
+
+    # Aggregate the rows for international letters which have the same rate
+    result = []
+    for _k, g in groupby(letter_units_for_month, key=itemgetter('postage', 'rate')):
+        g = list(g)
+
+        letter_details = LetterDetails(
+            billing_units=sum(x['billing_units'] for x in g),
+            rate=format_letter_rate(g[0]['rate']),
```
quis, line 459: Optional, but it might be a better separation of concerns to make a template filter for formatting the rate as pounds or pence (then it would be easier to reuse in other places as well).

app/main/views/dashboard.py
```diff
@@ -436,6 +436,50 @@ def get_free_paid_breakdown_for_billable_units(year, free_sms_fragment_limit, bi
...
+    # letter_units_for_month must be sorted before international postage values can be aggregated
+    postage_order = {'first class': 0, 'second class': 1, 'international': 2}
+    letter_units_for_month.sort(key=lambda x: (postage_order[x['postage']], x['rate']))
+
+    LetterDetails = namedtuple('LetterDetails', ['billing_units', 'rate', 'cost', 'postage_description'])
+
+    # Aggregate the rows for international letters which have the same rate
+    result = []
+    for _k, g in groupby(letter_units_for_month, key=itemgetter('postage', 'rate')):
+        g = list(g)
```
quis, line 455: I think a more descriptive variable name like `rate_group` would be better here. I’d only use single letter variables where it’s really obvious, for example `k` for `key`. Also there is `flask.g` which means something completely different.

app/main/views/dashboard.py
```diff
@@ -436,6 +436,50 @@ def get_free_paid_breakdown_for_billable_units(year, free_sms_fragment_limit, bi
...
+        result.append(letter_details)
+
+    return result
+
+
+def get_postage_description(postage):
+    if postage in ('first', 'second'):
+        return f'{postage} class'
+    else:
+        return 'international'
```
quis, line 473: I think leaving out the `else` makes it clearer that this function returns `international` by default:
```suggestion
    if postage in ('first', 'second'):
        return f'{postage} class'
    return 'international'
```

app/templates/views/usage.html
```diff
@@ -108,9 +108,9 @@ <h2 class='heading-small'>Letters</h2>
                 {{- ' {:.2f}p'.format(sms_rate * 100) }}</li>
               {% endif %}
               {% for letter in month.letters%}
-                {% if letter[0] %}
-                  <li class="tabular-numbers">{{ "{:,} {}".format(letter[0], letter[3])}} class {{ message_count_label(letter[0], 'letter', '') }}at
-                  {{ '{:.0f}p'.format(letter[1] * 100) }}</li>
+                {% if letter.billing_units %}
+                  <li class="tabular-numbers">{{ "{:,} {}".format(letter.billing_units, letter.postage_description) }} {{ message_count_label(letter.billing_units, 'letter', '') }}at
+                  {{ letter.rate }}</li>
```
quis, line 113: This is so much nicer 👍

app/main/views/dashboard.py
klssmith: Good point, I've changed this name to `rate_group`

app/main/views/dashboard.py
klssmith: 👍 Updated

app/main/views/dashboard.py
klssmith: I've made a filter for this in 7548ed7

PR #3518, opened 2020-07-10
Add a page to view a single broadcast

quis:
This pull request adds a page to view a single broadcast. This is important for two reasons:
- users need an audit of what happened when, and who else was involved in approving or cancelling a broadcast
- we need a place to put actions (approving, cancelling) on a broadcast so that you can confirm details of the message and the areas before performing the action

This means we can remove the ‘Stop broadcasting’ links from the dashboard.

Later we should add:
- a map preview
- a `get`/confirm/`post` loop for cancelling a broadcast

# Before

Screenshot 2020-07-10 at 15 29 20

# After 

Screenshot 2020-07-10 at 15 28 55

***

image

***

image

***

Screenshot 2020-07-10 at 14 20 50

***

Screenshot 2020-07-10 at 14 21 01

### DIFF
--- a/app/templates/views/broadcast/view-message.html
+++ b/app/templates/views/broadcast/view-message.html
@@ -1,0 +1,54 @@
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import sticky_page_footer %}
+
+{% extends "withnav_template.html" %}
+
+{% block service_page_title %}
+  {{ broadcast_message.template_name }}
+{% endblock %}
+
+{% block maincolumn_content %}
+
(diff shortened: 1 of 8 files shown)


app/templates/views/broadcast/view-message.html
```diff
@@ -0,0 +1,54 @@
...
+      Stopped by {{ broadcast_message.cancelled_by.name }}
+      {{ broadcast_message.cancelled_at|format_datetime_human }}.
+    {% else %}
+      Finished broadcasting {{ broadcast_message.finishes_at|format_datetime_human }}.
+    {% endif %}
+  </p>
+
+  {% for area in broadcast_message.areas %}
+    {% if loop.first %}
+      <ul class="area-list">
```
leohemsted, line 42: ooh interesting didn't know about this magic 🧙

PR #3519, opened 2020-07-13
Remove emails, text messages and letter permissions from broadcast services

quis:
To make the interface as simple as possible we don’t want to mix up sending other types of communication for services that have the broadcast permission.

So this pull request:
- removes the other permissions when a service is given the broadcast permission
- removes references to the other permissions from the settings page 
- blocks the ability to switch the other permissions back on

At the moment we don’t know what the ‘restrictions’ of trial mode will mean for broadcast services, so the page doesn’t completely make sense. But we can address this when we figure out what the restrictions should be.

# Before 

image

# After 

image


tests/app/main/views/test_service_settings.py
```diff
@@ -3679,6 +3712,15 @@ def test_unknown_channel_404s(
...
+        'email',
+        'It’s free to send emails through GOV.UK Notify.',
+        'Send emails',
+        ['broadcast'],
+        'True', 'True',
+        ['email', 'broadcast'],
+        marks=pytest.mark.xfail(raises=AssertionError),
+    ),
 ])
 def test_switch_service_enable_letters(
```
idavidmcdonald, line 3725: This test name no longer makes sense with your added test case (although looks like it hasn't made sense for a while anyway but would be nice to fix please)

idavidmcdonald: I'm super confused by this test (not necessarily by what you've added to it but just in general).

So the arguments appear to be as follows
```
    channel = 'email'
    expected_first_para = 'It's free to send emails through GOV.UK Notify.'
    expected_legend = 'Send emails'
    expected_initial_value = ['broadcast']
    posted_value= 'True'
    initial_permissions = 'True'
    expected_updated_permissions = ['email', 'broadcast']
```

If that is correct, it confuses me that the expected_initial_value is a list rather than a bool and initial_permissions is a bool as opposed to that list?

Also is there anyway we can not have it catch the `AssertionError`? It seems too generic, I've no idea which of the many assertions failed. Instead can we get it catch a 403 exception?

idavidmcdonald: Also this is super minor but it really bothers me that in all the params, they are all on a new line apart from two booleans which are on the same line. It makes it harder to join up the values in the pytest params to the function arguments.

idavidmcdonald: Ah, found the reason for the answer to my second comment. It's because
```
@pytest.mark.parametrize((
    'channel,'
    'expected_first_para,'
    'expected_legend,'
    'initial_permissions,'
    'expected_initial_value,'
    'posted_value,'
    'expected_updated_permissions'
)
```
doesn't match the order of the arguments in the function
```
def test_switch_service_enable_letters(
    client_request,
    service_one,
    mocker,
    mock_get_free_sms_fragment_limit,
    channel,
    expected_first_para,
    expected_legend,
    expected_initial_value,
    posted_value,
    initial_permissions,
    expected_updated_permissions,
):`

I guess technically it's not wrong but I don't think I've ever seen our tests not match that pattern in the past and only succeeded in confusing me for 20 minutes

Comment (idavidmcdonald) [reply tests/app/main/views/test_service_settings.py:3725]: Further to test, I would expect that expected_initial_value should be 'False' rather than 'True' like in your test which I think may be hidden by the generic `AssertionError` catching

Comment (quis) [reply tests/app/main/views/test_service_settings.py]: I have:
- cleaned up the test naming and parameter order in 6d1be9e
- added a custom exception to make sure we’re getting a `403` in 13734af

File: tests/conftest.py [hunk 2]
```diff
@@ -37,6 +37,19 @@ class ElementNotFound(Exception):
     pass
 
 
+class UnexpectedHTTPStatus403(Exception):
+    pass
+
+
+def get_unexpected_status_exception(status_code):
```
idavidmcdonald, line 44: So it took me a few minutes and looking at the python docs, so what I think this does is

If status code is 403 then returns the class `UnexpectedHTTPStatus403`.

If status code is 409 then returns a class generated on the fly, `UnexpectedHTTPStatus403` which inherits from Exception.

In reality, you don't even need to define `UnexpectedHTTPStatus403` on line 40 but you have done as an example of what you could do?

Optional, if the above is correct a doc block comment to clarify that would be great please.

I assume we've had to solve this problem in the API before as well? This is short enough a solution that I'm not too fussed about potential duplication if that is the case but do take a glance at the API if you have time

tests/conftest.py
```diff
@@ -3036,7 +3049,11 @@ def get_url(
             if _expected_redirect and _expected_status == 200:
                 _expected_status = 302
 
-            assert resp.status_code == _expected_status, resp.location
```
idavidmcdonald, line 3039: Optional: We are dropping the use of `resp.location` is that intended? Which supposedly means if the status codes don't match then we raise `AssertionError(resp.location)` which I guess is maybe helpful with debugging but I don't fully understand what it will give us to be honest. Maybe worth changing `f'Expected HTTP status {_expected_status}, reponse was {resp.status_code}'` to
f'Expected HTTP status {_expected_status}, reponse was {resp.status_code} for {resp.location}'

tests/app/main/views/test_service_settings.py
```diff
@@ -3644,53 +3678,68 @@ def test_unknown_channel_404s(
...
+        'True',
+        'True',
         ['email', 'sms', 'letter'],
     ),
+    pytest.param(
+        'email',
+        'It’s free to send emails through GOV.UK Notify.',
+        'Send emails',
+        ['broadcast'],
+        'True',
```
idavidmcdonald, line 3726: Oh how come the expected initial value isn't False as this service doesn't yet have email permissions so it wouldn't be checked?

tests/conftest.py
quis: I realised that this is a bit over-engineered. So I’ve abandoned this approach in favour of writing a separate test that uses the `_expected_status` argument of `client_request`. This is how we typically check for `403`s in other parts of the codebase, so I think it’s better to not invent a new thing just for the sake of keeping the test case in the same `@parametrize`.

quis: New test is here: https://github.com/alphagov/notifications-admin/pull/3519/files#diff-15f7f6c51717d303b04559c8a0d9e81bR3764-R3787

PR #3520, opened 2020-07-13
Make user permissions make sense for services with the broadcast permission

quis:
# Before 

image

# After 

image

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -54,7 +54,11 @@ from app.main.validators import (
 )
 from app.models.feedback import PROBLEM_TICKET_TYPE, QUESTION_TICKET_TYPE
 from app.models.organisation import Organisation
-from app.models.roles_and_permissions import permissions, roles
+from app.models.roles_and_permissions import (
+    broadcast_permissions,
+    permissions,
+    roles,
+)
 from app.utils import guess_name_from_email_address
 
 
@@ -493,7 +497,12 @@ PermissionsAbstract = type("PermissionsAbstract", (StripWhitespaceForm,), {
 })
 
 
-class PermissionsForm(PermissionsAbstract):
+BroadcastPermissionsAbstract = type("BroadcastPermissionsAbstract", (StripWhitespaceForm,), {
+    permission: BooleanField(label) for permission, label in broadcast_permissions
+})
+
+
+class BasePermissionsForm(StripWhitespaceForm):
     def __init__(self, all_template_folders=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.folder_permissions.choices = []
@@ -517,11 +526,14 @@ class PermissionsForm(PermissionsAbstract):
 
     @property
     def permissions(self):
-        return {role for role in roles.keys() if self[role].data is True}
+        return {field.id for field in self.permissions_fields if field.data is True}
 
     @property
     def permissions_fields(self):
-        return (getattr(self, permission) for permission, _ in permissions)
+        return (
+            getattr(self, permission) for permission, field in self.__dict__.items()
+            if isinstance(field, BooleanField)
+        )
 
     @classmethod
     def from_user(cls, user, service_id, **kwargs):
@@ -535,7 +547,18 @@ class PermissionsForm(PermissionsAbstract):
         )
 
 
-class InviteUserForm(PermissionsForm):
+class PermissionsForm(PermissionsAbstract, BasePermissionsForm):
+    pass
+
+
+class BroadcastPermissionsForm(BroadcastPermissionsAbstract, BasePermissionsForm):
+
+    @property
+    def permissions(self):
+        return {'view_activity'} | super().permissions
+
+
+class BaseInviteUserForm():
     email_address = email_address(gov_user=False)
 
     def __init__(self, invalid_email_address, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_manage_users.py
```diff
@@ -266,6 +295,41 @@ def test_service_without_caseworking_doesnt_show_admin_vs_caseworker(
...
+def test_broadcast_service_only_shows_relevant_permissions(
+    client_request,
+    service_one,
+    mock_get_users_by_service,
+    mock_get_template_folders,
+    endpoint,
+    extra_args,
+):
+    service_one['permissions'] = ['broadcast']
+    page = client_request.get(
```
idavidmcdonald, line 317: Can we also write/edit a unit test for the POST cases of these forms please?

app/models/roles_and_permissions.py
```diff
@@ -25,6 +25,12 @@
     ('manage_api_keys', 'Manage API integration'),
 )
 
+broadcast_permissions = (
+    ('send_messages', 'Prepare and approve broadcasts'),
```
idavidmcdonald, line 29: I did briefly consider whether we wanted a new permission for `send_broadcasts` or similar but I think it's fine like this

app/main/forms.py
```diff
@@ -535,6 +539,24 @@ def from_user(cls, user, service_id, **kwargs):
...
 
 
+BroadcastPermissionsAbstract = type("PermissionsAbstract", (StripWhitespaceForm,), {
+    permission: BooleanField(label) for permission, label in broadcast_permissions
+})
+
+
+class BroadcastPermissionsForm(PermissionsForm):
+
+    view_activity = HiddenField('view_activity', default='y')
```
idavidmcdonald, line 549: Wondering if an inline code comment just saying why this is done like this would be useful?

quis:
```suggestion
    # All users of broadcast services should see the dashboard, but it
    # shouldn’t be a choice that the user can select
    view_activity = HiddenField('view_activity', default='y')
```

quis: I’ve changed this to not use a hidden field at all. A hidden field can still be modified with the browser developer tools, it’s better that we hard-code the permission in 🐍

tests/app/main/views/test_manage_users.py
quis: I’ve added tests for the `POST` cases. The test showed that the approach wasn’t working (it was still allowing you to forge a request which gave users the `manage_api_keys` permission for broadcast services). So I’ve re-written things and squashed into a48ea5e

tests/app/main/views/test_manage_users.py
```diff
@@ -787,6 +890,62 @@ def test_invite_user_with_email_auth_service(
...
+        {
+            'view_activity',
+            'send_messages',
+            'manage_templates',
+            'manage_service',
+        },
+    ),
+    (
+        {
+            'view_activity': 'y',
```
idavidmcdonald, line 938: Optional: on the test cases for this test and the test you edited above, it might be nice to have some inline code comments saying what the before and after is as a description of the user behaviour

For example, this test case here seems to be testing that permissions that aren't allowed for broadcast services are ignored.

Would just help when revisiting this functionality in a years time to understand what the intention was behind these test cases.

tests/app/main/views/test_manage_users.py
```diff
@@ -787,6 +890,62 @@ def test_invite_user_with_email_auth_service(
...
+):
+    service_one['permissions'] = ['broadcast']
+    mocker.patch('app.models.user.InvitedUsers.client_method', return_value=[sample_invite])
+    mocker.patch('app.models.user.Users.client_method', return_value=[active_user_with_permissions])
+    mocker.patch('app.invite_api_client.create_invite', return_value=sample_invite)
+    post_data['email_address'] = 'broadcast@example.gov.uk'
+    client_request.post(
+        'main.invite_user',
+        service_id=SERVICE_ONE_ID,
+        _data=post_data,
```
idavidmcdonald, line 966: Optional: Should we assert the status code and redirect url that we expect for this post request? Like in the test you edited above?

app/main/forms.py
```diff
@@ -493,7 +497,12 @@ def __init__(self, invited_org_user):
 })
 
 
-class PermissionsForm(PermissionsAbstract):
+BroadcastPermissionsAbstract = type("BroadcastPermissionsAbstract", (StripWhitespaceForm,), {
```
idavidmcdonald, line 500: Optional: I'm not sure what the benefit is of using the `type` syntax to create our class here? It's a pattern that I don't think we have anywhere else in our codebase and I'd encourage you to go for the more usual approach please as I don't think this will help the ease of understanding our codebase

`class BroadcastPermissionsAbstract(StripWhitespaceForm)...`

app/main/forms.py
```diff
@@ -517,11 +526,14 @@ def __init__(self, all_template_folders=None, *args, **kwargs):
...
     def permissions(self):
-        return {role for role in roles.keys() if self[role].data is True}
+        return {field.id for field in self.permissions_fields if field.data is True}
 
     @property
     def permissions_fields(self):
-        return (getattr(self, permission) for permission, _ in permissions)
+        return (
+            getattr(self, permission) for permission, field in self.__dict__.items()
+            if isinstance(field, BooleanField)
```
idavidmcdonald, line 535: This is making an assumption that all boolean fields are permissions fields which may not be watertight for the future but I think is acceptable

app/main/forms.py
quis: It’s the only way to dynamically create properties on a class before instantiating it. It is used already in this file already. Agree that it’s a bit gnarly but I couldn’t find a way of doing the same thing without hard coding stuff.

idavidmcdonald: Ah yes, you are correct and I hadn't spotted the one above it, apologies :)

PR #3532, opened 2020-07-21
Make live broadcasts feel more prominent on dashboard

quis:
We have a reckon that live broadcasts don’t feel prominent, consequential or active enough on the dashboard.

This commit adds an animated component, similar to an ‘on air’ indicator in a broadcast studio, or a ‘recording’ indicator on a video camera.

This is one option for addressing our reckon. We shouldn’t merge this until we have a better understanding of the problem from another round of user research.

live


yahoopete:
I've seen enough to want to try this out. I don't think the 'approval' or liveness of a broadcast is coming through strongly enough just yet....

PR #3535, opened 2020-07-24
Try introducing govuk checkboxes again

tombye:
NOTE: Depends on https://github.com/alphagov/notifications-admin/pull/3465/commits and https://github.com/alphagov/notifications-functional-tests/pull/342 being merged before it can be merged.

An attempt to apply the changes from https://github.com/alphagov/notifications-admin/pull/3476 again.

My hypothesis is that [this change to the functional test that failed](https://github.com/alphagov/notifications-functional-tests/pull/342) will fix  the problem.

## New code

The difference between this set of changes and those in https://github.com/alphagov/notifications-admin/pull/3476 is:
- the ["Update permissions page" commit](https://github.com/alphagov/notifications-admin/commit/830aeae7b8b83b5263c2e59c2731fd65166bbd9c) has been updated to work with the version of the page users see if they have broadcast permissions
- https://github.com/alphagov/notifications-admin/pull/3535/commits/f3222296140fecde6d752624cba649c86a757125 adds tests for the 'name' attribute of the `permissions_field` checkboxes (suggested in https://github.com/alphagov/notifications-admin/pull/3535#discussion_r460869797)
- https://github.com/alphagov/notifications-admin/pull/3535/commits/e3c434bb8fc3683616ee86eff129ad5973c4c8bd and https://github.com/alphagov/notifications-admin/pull/3535/commits/75bac87a4db0fa6df48b9c62af659b40c0e2e9aa to protect against clients still running pages with the old user permissions checkboxes when the instances roll (pointed out in https://github.com/alphagov/notifications-admin/pull/3535#discussion_r460872903)


tests/app/main/views/test_manage_users.py
```diff
@@ -288,11 +288,11 @@ def test_service_without_caseworking_doesnt_show_admin_vs_caseworker(
...
-    assert page.select('input[type=checkbox]')[0]['name'] == 'view_activity'
-    assert page.select('input[type=checkbox]')[1]['name'] == 'send_messages'
-    assert page.select('input[type=checkbox]')[2]['name'] == 'manage_templates'
-    assert page.select('input[type=checkbox]')[3]['name'] == 'manage_service'
-    assert page.select('input[type=checkbox]')[4]['name'] == 'manage_api_keys'
+    assert page.select('input[type=checkbox]')[0]['value'] == 'view_activity'
+    assert page.select('input[type=checkbox]')[1]['value'] == 'send_messages'
+    assert page.select('input[type=checkbox]')[2]['value'] == 'manage_templates'
+    assert page.select('input[type=checkbox]')[3]['value'] == 'manage_service'
+    assert page.select('input[type=checkbox]')[4]['value'] == 'manage_api_keys'
```
quis, line 295: Can we assert on the name and value here? I’m fine with adding an extra check for the right value of `value`, but not sure why we’re removing the check for the value of `name`?

tests/app/main/views/test_service_settings.py
```diff
@@ -2299,8 +2299,8 @@ def test_incorrect_sms_sender_input(
 
 @pytest.mark.parametrize('reply_to_addresses, data, api_default_args', [
     ([], {}, True),
-    (create_multiple_email_reply_to_addresses(), {}, False),
-    (create_multiple_email_reply_to_addresses(), {"is_default": "y"}, True)
+    (create_multiple_email_reply_to_addresses(), {"is_default": []}, False),
+    (create_multiple_email_reply_to_addresses(), {"is_default": ["y"]}, True)
```
quis, line 2303: How do we feel about how this is going to work during a deploy? You could serve the `GET` request from an instance with the new code, then `POST` to an instance with the old code, which will be expecting `"y"` but will get `["y"]`.

tests/app/main/views/test_manage_users.py
tombye: They all share the same name, in the same way the previous version of this group shared a value. I removed them to match the original point of the assertions, to confirm the attribute that varied was tested. It just switched from being the name to the value. But yeah, I can see the worth in testing the name too as that's part of what makes their POST param valid.

tests/app/main/views/test_service_settings.py
tombye: I confess to not having thought about this, mainly because it's not been something I've had to until now. Will put in some checks to ensure the scenario you described works.

tests/app/main/views/test_manage_users.py
tombye: I added [these tests](https://github.com/alphagov/notifications-admin/pull/3535/commits/f3222296140fecde6d752624cba649c86a757125) to check the `name` as well.

tests/app/main/views/test_service_settings.py
tombye: This test was actually wrong because the `is_default` checkbox now sends the same `y` as before so I've corrected those tests in https://github.com/alphagov/notifications-admin/pull/3535/commits/e266b11ee2b2876d93040130d45b3b5414699660.

The scenario you describe could happen with the user permissions field(s) though as they were converted from multiple `BooleanField`s to one `govukCheckboxesField` so send different params. I've added some tests for that happening in https://github.com/alphagov/notifications-admin/pull/3535/commits/f3222296140fecde6d752624cba649c86a757125 and rewritten the form to handle both types of request in https://github.com/alphagov/notifications-admin/pull/3535/commits/75bac87a4db0fa6df48b9c62af659b40c0e2e9aa.

tombye: Forgot to mention that I've put a draft pull request up to remove the extra code added once deployment is complete: https://github.com/alphagov/notifications-admin/pull/3543

PR #3538, opened 2020-07-27
Make start time explicit when previewing a broadcast

quis:
# Context

We [recently introduced](https://github.com/alphagov/notifications-admin/pull/3527) a form control that lets users choose when a broadcast ends:

image (12)

Based on the most recent research participant, we think:
- there is a specific misunderstanding of what this control does
- there is a general low level of understanding of what a ‘broadcast’ means

People will try to understand what a ‘broadcast’ is by using mental models they have for other kinds of messaging, for example text messages.

Other kinds of messaging are one-to-one, i.e. they go from a sender to a recipient. They are not ongoing in any way.

Emails and texts are sent at a specific time (and for all practicable purposes are received at that same time). So, when we present the user with a form that controls time, they might well assume it controls the time when the message will be sent.

We reinforce this assumption with the labelling of the form control. By front-loading it with the word ‘When’ we are playing to the users confirmation bias. In other words they are interpreting the meaning of the control in a way that confirms their prior beliefs about how messaging works.

Controlling when a message is sent is a feature we already offer to users uploading a spreadsheet of recipients. That’s where we’ve borrowed this pattern and language from:

image

# The change

This pull request does two things:
- re-labels the form to front-load the word ‘End’ not ‘When’
- adds text to the page explaining when the broadcast will start, to interrupt that confirmation bias

image (13)

If we can get users to go through this before sending a broadcast for real, it could help them learn what a broadcast is, and how it differs from sending text messages.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -993,7 +993,7 @@ class ChooseBroadcastDurationForm(StripWhitespaceForm):
         )
 
     finishes_at = RadioField(
-        'When should this broadcast end?',
+        'End time',
     )
 
 
--- a/app/templates/components/radios.html
+++ b/app/templates/components/radios.html
@@ -24,11 +24,12 @@
  field,
  hint=None,
  wrapping_class='form-group',
- show_now_as_default=True
+ show_now_as_default=True,
+ bold_legend=False
 ) %}
  <div class="{{ wrapping_class }} {% if field.errors %} form-group-error{% endif %}">
    <fieldset>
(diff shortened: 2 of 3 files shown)


quis:
# Some earlier ideas 

## Just add some guidance

Probably too easy to miss.

image (10)

## Make start time a single-option radio button

Gives start time and end time equal emphasis, but feels like a bit of a dark/weird pattern.

image (11)

PR #3539, opened 2020-07-28
Check for invalid characters in letter addresses

klssmith:
These characters `@ ( ) = [ ] ” \ / ,` can't be used at the start of address lines
because they cause issues for DVLA. A method to check for these characters
was added here https://github.com/alphagov/notifications-utils/pull/771

This PR brings in the new version of utils and uses the method to add
validation for invalid characters on the `LetterAddressForm` for one-off
letters. It also adds a validation failed message for uploaded letters,
precompiled letters sent through the API, and CSV rows with errors.

[Pivotal story](https://www.pivotaltracker.com/story/show/173892353)

**Sending a one-off letter**
Screenshot 2020-07-28 at 13 56 32

**CSV with an error**
Screenshot 2020-07-28 at 13 56 08

**Uploading a file**
Screenshot 2020-07-28 at 13 57 11

**Viewing a precompiled letter sent through the API**
Screenshot 2020-07-28 at 13 56 44

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -853,6 +853,11 @@ class LetterAddressForm(StripWhitespaceForm):
                 f'Last line of the address must be a real UK postcode'
             )
 
+        if address.has_invalid_characters:
+            raise ValidationError(
+                'Address lines must not start with any of the following characters: @ ( ) = [ ] ” \\ / ,'
+            )
+
 
 class EmailTemplateForm(BaseTemplateForm):
     subject = TextAreaField(
--- a/app/templates/views/check/row-errors.html
+++ b/app/templates/views/check/row-errors.html
@@ -90,6 +90,8 @@
                   {% else %}
                     Last line of the address must be a real UK postcode
                   {% endif %}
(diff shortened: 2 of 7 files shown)


app/main/forms.py
```diff
@@ -853,6 +853,11 @@ def validate_address(self, field):
                 f'Last line of the address must be a real UK postcode'
             )
 
+        if address.has_invalid_characters:
+            raise ValidationError(
+                'Address lines must not start with any of the following characters: @ ( ) = [ ] ” \\ / ,'
```
quis, line 858: It would be nice to copy ‘the too many lines’ error and not hard code the characters:
```suggestion
                f'Address lines must not start with any of the following characters: {" ".join(PostalAddress.INVALID_CHARACTERS)}'
```

…but maybe not worth it since:
- we don’t anticipate these characters changing often
- that would mean releasing utils again because the order of the characters in `PostalAddress.INVALID_CHARACTERS` is different 
- doing `join` inside an f-string is arguably a bit gross

PR #3540, opened 2020-07-30
Tell users that they can search whole postal address

quis:
We’re now [normalising and storing the whole address in the `normalised_to` field](https://github.com/alphagov/notifications-api/pull/2814). Previously we were only storing the first line of the address.

Enough time should now have passed that the field will have been populated for all letters in the database. This means that users can already search by any part of the address, but we’re not telling them that they can.

We can now tell users that it’s not just the first line they can search by.

This also changes the notification page to show the whole address, comma separated, so it’s a bit more obvious why the search is returning the results it is.

### DIFF
--- a/app/templates/views/activity/notifications.html
+++ b/app/templates/views/activity/notifications.html
@@ -19,7 +19,7 @@
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
           <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="govuk-link govuk-link--no-visited-state file-list-filename" href="{{ single_notification_url(notification_id=item.id) }}">{{ item.to.splitlines()[0].lstrip().rstrip(' ,') if item.to else '' }}</a>
+          <a class="govuk-link govuk-link--no-visited-state file-list-filename" href="{{ single_notification_url(notification_id=item.id) }}">{{ item.to.splitlines()|join(', ') if item.to else '' }}</a>
         {% endif %}
         <p class="file-list-hint">
           {{ item.preview_of_content }}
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_activity.py
```diff
@@ -757,8 +757,8 @@ def test_sending_status_hint_displays_correctly_on_notifications_page(
 
 
 @pytest.mark.parametrize("is_precompiled_letter,expected_address,expected_hint", [
-    (True, "Full Name,\nFirst address line\npostcode", "ref"),
-    (False, "Full Name,\nFirst address line\npostcode", "template subject")
```
quis, line 761: We remove the comma in utils now:
https://github.com/alphagov/notifications-utils/blob/74eba53e8358d2e46194811b44997ccb3707d130/notifications_utils/postal_address.py#L41

No address lines should have trailing commas by the time they get to this point

app/templates/views/activity/notifications.html
```diff
@@ -19,7 +19,7 @@
         {% if item.status in ('pending-virus-check', 'virus-scan-failed') %}
           <span class="file-list-filename loading-indicator">Checking</span>
         {% else %}
-          <a class="govuk-link govuk-link--no-visited-state file-list-filename" href="{{ single_notification_url(notification_id=item.id) }}">{{ item.to.splitlines()[0].lstrip().rstrip(' ,') if item.to else '' }}</a>
+          <a class="govuk-link govuk-link--no-visited-state file-list-filename" href="{{ single_notification_url(notification_id=item.id) }}">{{ item.to.splitlines()|join(', ') if item.to else '' }}</a>
```
quis, line 22: We remove the comma in utils now:
https://github.com/alphagov/notifications-utils/blob/74eba53e8358d2e46194811b44997ccb3707d130/notifications_utils/postal_address.py#L41

So don’t need the `rstrip` here any more. No address lines should have trailing commas by the time they get to this point

PR #3547, opened 2020-08-03
Add a tour for users new to broadcast services

quis:
This is an initial, prototype-quality attempt at introducing some kind of tour for users new to broadcasting. A lot of the users we’re speaking to don’t have a good concept of what broadcasting means, which is causing usability problems down the line.

We did a similar thing in the early days of Notify to explain the concept of message templates and personalisation.

image

image

image

image


app/main/views/invites.py
```diff
@@ -40,6 +40,9 @@ def accept_invite(token):
 
     if invited_user.status == 'accepted':
         session.pop('invited_user', None)
+        service = Service.from_id(invited_user.service)
+        if service.has_permission('broadcast'):
+            return redirect(url_for('main.broadcast_tour', service_id=service.id, step_index=1))
```
leohemsted, line 45: i don't think this is enough for "my existing user has been invited to a new broadcast service, and i should be told how broadcasting works"


this code will, i believe, be called if someone clicks on an invite link which has already been accepted. it then bounces to `service_dashboard`, which will either show succesfully if the correct user is signed in, or bounce again to sign-in if the user isn't signed in.

not sure if its useful to have this here, though i don't hate it being here. however, i think we also need to add it ten lines below in the existing_user block:

```
    if existing_user:
        invited_user.accept_invite()
        [....]
        return redirect(url_for('main.service_dashboard', service_id=service.id))
```

if the user doesn't already exist, then they'll get scooped up by verify.py so thats good 👍

PR #3552, opened 2020-08-05
Don’t let users self-approve broadcasts

quis:
At the moment they will get a ‘technical difficulties’ error if they try.

We probably want to do something around letting people self-approve broadcasts in trial mode, but for now just telling them they can’t is a better experience than ‘technical difficulties’ (and will probably be close to what they should see on a live service as well).

# Before 

(this is still what a different user to the one who created the broadcast would see)

image

# After 

image


app/models/broadcast_message.py
```diff
@@ -94,15 +95,15 @@ def status(self):
...
     def created_by(self):
         return User.from_id(self.created_by_id)
 
-    @property
+    @cached_property
     def approved_by(self):
         return User.from_id(self.approved_by_id)
 
-    @property
+    @cached_property
```
leohemsted, line 106: does it make sense to just put a LRU cache around `User.from_id`? We use that in a whole bunch of places

app/templates/views/broadcast/view-message.html
```diff
@@ -17,17 +18,30 @@
...
-        {{ broadcast_message.created_by.name }} wants to broadcast this
-        message until {{ broadcast_message.finishes_at|format_datetime_relative }}.
-      </p>
-      {{ page_footer(
-        "Start broadcasting now",
-        delete_link=url_for('main.reject_broadcast_message', service_id=current_service.id, broadcast_message_id=broadcast_message.id),
-        delete_link_text='Reject this broadcast'
-      ) }}
-    {% endcall %}
+    {% if broadcast_message.created_by == current_user %}
```
servingUpAces, line 21: should we just compare the id of the users?
created_by is a cached property, it's possible the user changed, but the cache has not been invalidated.

app/models/broadcast_message.py
quis: `functools.lru_cache` persists between requests (so could become stale). `cached_property` only lives for the lifetime of the request.

app/templates/views/broadcast/view-message.html
quis: Under the hood it’s comparing the `id`: https://github.com/alphagov/notifications-admin/blob/cd398ab55c17ea3d1d4abe5a7e7a4ab2ad28c2e7/app/models/__init__.py#L25-L26

app/models/broadcast_message.py
leohemsted: ah 👍

PR #3553, opened 2020-08-06
Move broadcast areas into admin

tlwr:
Adds broadcast areas to this app

Also read https://github.com/alphagov/notifications-utils/pull/769

Libraries added:
- Countries
- Counties and unitary authorities
- Regions
- Electoral wards

Scripts added:
- Plotting script to show the different areas in the vendored database
- Recreation script which rebuilds the database from the vendored geojson files

We are using a sqlite db because we do not have to parse the entire geojson file on every request, it can be indexed ahead of time.


tests/app/broadcast_areas/test_broadcast_area.py
```diff
@@ -0,0 +1,191 @@
...
+
+def test_examples():
+    assert re.match(
+        "^([^,]*, ){3}[^,]*",
+        broadcast_area_libraries.get('countries').get_examples(),
+    )
+    assert re.match(
+        "^([^,]*, ){4}and 5 more…$",
+        broadcast_area_libraries.get('regions-of-england').get_examples()
+    )
```
quis, line 59: Can we make these concrete assertions please? There’s too many ways that `[^,]*` could match the wrong thing and give a false positive.

tests/app/main/views/test_broadcast.py
```diff
@@ -226,25 +226,26 @@ def test_choose_broadcast_library_page(
...
-    ]
+    ])
+
+    assert any([
+        all([
+            country in hint.text
+            for country in ('England', 'Northern Ireland', 'Scotland', 'Wales')
+        ])
+        for hint in page.select('.file-list-hint-large')
+    ])
```
quis, line 247: I’d prefer this test to be a little less clever as well:
```suggestion
    assert normalize_spaces(page.select('.file-list-hint-large')[1].text) == (
        'England, Northern Ireland, Scotland, and Wales'
    )
```

PR #3555, opened 2020-08-07
delete template cache when reply_to stuff is updated

leohemsted:
sms sender, email reply to, letter contact blocks.

These are all cached within template, under `template.reply_to` - if the
template doesnt have a specific default, then that field stores the
service default. So when service default changes, we need to clear the
template cache so that it is updated to reflect that.

We already use this pattern for deleting the template cache for a bunch
of templates in `template_folder_api_client.move_to_folder`

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -6,6 +6,14 @@ from app.notify_client import NotifyAdminAPIClient, _attach_current_user, cache
 
 class ServiceAPIClient(NotifyAdminAPIClient):
 
+    def _delete_template_cache_for_service(self, service_id):
+        templates_for_service = self.get_service_templates(service_id)['data']
+        if templates_for_service:
+            redis_client.delete(*[
+                f"template-{x['id']}-version-None"
+                for x in templates_for_service
+            ])
+
     @cache.delete('user-{user_id}')
     def create_service(
         self,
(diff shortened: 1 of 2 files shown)


app/notify_client/service_api_client.py
```diff
@@ -6,6 +6,13 @@
 
 class ServiceAPIClient(NotifyAdminAPIClient):
 
+    def _delete_template_cache_for_service(self, service_id):
+        templates_for_service = self.get_service_templates(service_id)['data']
+        redis_client.delete(*map(
+            'template-{}-version-None'.format,
+            [x['id'] for x in templates_for_service],
+        ))
```
quis, line 14: I’m sure you told me something about Guido not liking `map` at one point
```suggestion
        redis_client.delete(*[
            f'template-{template["id"]}-version-None'
            for template in templates_for_service
        ])
```

app/notify_client/service_api_client.py
```diff
@@ -6,6 +6,13 @@
 
 class ServiceAPIClient(NotifyAdminAPIClient):
 
+    def _delete_template_cache_for_service(self, service_id):
+        templates_for_service = self.get_service_templates(service_id)['data']
+        redis_client.delete(*map(
```
klssmith, line 11: `redis_client.delete` will raise an error if the service has no templates because it can't be called with an empty list.

leohemsted: good spot, fixed and added test

PR #3558, opened 2020-08-10
Services can only create templates of types they have turned on

CrystalPea:
- Services can only create templates of types they have permissions for
- If service only has permission for one type of notification, "New Template" button takes it straight to new template page, without showing drop-down radios menu

### DIFF
--- a/tests/javascripts/templateFolderForm.test.js
+++ b/tests/javascripts/templateFolderForm.test.js
@@ -1,6 +1,6 @@
 const helpers = require('./support/helpers');
 
-function setFixtures (hierarchy) {
+function setFixtures (hierarchy, newTemplateDataModules = "") {
 
   const foldersCheckboxesHTML = function (filter) {
     let count = 0;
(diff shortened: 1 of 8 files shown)


tests/javascripts/templateFolderForm.test.js
```diff
@@ -309,6 +309,36 @@ describe('TemplateFolderForm', () => {
 
   });
 
+  describe("Click 'New template' for single channel service", () => {
+    test("should redirect to new template page", () => {
+      setFixtures(hierarchy, "data-channel='sms' 'data-service='123'")
```
quis, line 314: There’s an extra quotation mark here:
```suggestion
      setFixtures(hierarchy, "data-channel='sms' data-service='123'")
```

tests/javascripts/templateFolderForm.test.js
```diff
@@ -309,6 +309,36 @@ describe('TemplateFolderForm', () => {
 
   });
 
+  describe("Click 'New template' for single channel service", () => {
+    test("should redirect to new template page", () => {
+      setFixtures(hierarchy, "data-channel='sms' data-service='123'")
```
quis, line 314: You need to re-declare the value of `templateFolderForm` after calling `setFixtures`, otherwise it refers to elements which are no longer in the page:

```suggestion
      setFixtures(hierarchy, "data-channel='sms' data-service='123'")
      templateFolderForm = document.querySelector('form[data-module=template-folder-form]');
```

PR #3566, opened 2020-08-12
Reduce risk of commiting files with personal information in

idavidmcdonald:


.gitignore
```diff
@@ -1,3 +1,33 @@
+## We don't allow commiting of certain files types that are at risk they could be user files containg personal information
+## Therefore if you are sure that a file is OK, you can explicitely add it to the list of files to be allowed
```
quis, line 2: Tidied up a bit:
```suggestion
## We don’t allow committing certain file types that risk containing users’ personal information
## If you are sure that a file is OK, you can explicitly add it to the list of allowed files 
```

PR #3572, opened 2020-08-13
Rationalise broadcast libraries into one for national and one for local areas

quis:
This PR does two main things:
- removes the ‘Counties and unitary authorities in England and Wales’ library because it’s redundant now we have the ‘Electoral wards in the United Kingdom’ library
- renames the libraries to local and national because we’ve observed this language during user research, and it’s has less
tongue-twisting ambiguity than county vs country

# Before 

image

# After 

image


quis:
# Update

We’ve now decided to go with ‘Local authorities’ as it’s more accurate, and your ‘local area’ sounds like something different.

So the libraries page looks like this:

image

And the `<h1>`s of the next pages have been updated to match:

image
image

Which leads nicely into the existing local authority areas page:

image

PR #3574, opened 2020-08-14
Add confirmation banner when cancelling user invites

klssmith:
This shows the green banner with a tick when cancelling a user's
invitation to a service or organisation. The accessibility audit noted
that 'When cancelling an invite a new page loads, however, there is no
immediate indication that the invite has been cancelled.'

In order to display the invited user's email address as part of the
flash message, this adds new methods to the api clients for invites to get
a single invite.

[Pivotal story](https://www.pivotaltracker.com/story/show/174294749)

### DIFF
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -283,4 +283,7 @@ def confirm_edit_user_mobile_number(service_id, user_id):
 def cancel_invited_user(service_id, invited_user_id):
     current_service.cancel_invite(invited_user_id)
 
+    invited_user = InvitedUser.by_id_and_service_id(service_id, invited_user_id)
+
+    flash(f'Invitation cancelled for {invited_user.email_address}', 'default_with_tick')
     return redirect(url_for('main.manage_users', service_id=service_id))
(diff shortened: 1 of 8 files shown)


app/main/views/manage_users.py
```diff
@@ -283,4 +283,5 @@ def confirm_edit_user_mobile_number(service_id, user_id):
 def cancel_invited_user(service_id, invited_user_id):
     current_service.cancel_invite(invited_user_id)
 
+    flash('Invitation cancelled', 'default_with_tick')
```
idavidmcdonald, line 286: The flash message to say someone has been invited includes the email for example, `Invite sent to user@gov.uk`. Should we do the same here for this or is `Invitation cancelled` sufficient? The pivotal story doesn't say and I don't have access to the DAC report to see if this has been decided already.

klssmith: I have been checking about the content and we do want the email address to be included in the message, so I'll make that change

klssmith:
I misunderstood the story at first and added a confirmation step before cancelling the invite (https://github.com/alphagov/notifications-admin/pull/3570). We actually only want the confirmation after cancelling - it should confirm the action that has just happened

klssmith:
This has been changed to display the email address in the flash message.

It will now need this new PR to be merged first https://github.com/alphagov/notifications-api/pull/2958

PR #3578, opened 2020-08-17
Update sent status label

karlchillmaid:
We've had some feedback that the status label `sent internationally` can be confusing.

Unless the user checks the list of status descriptions, `sent internationally` doesn't make clear whether:

* the message was sent to an international phone number OR
* the message was sent to a UK number, but the recipient is outside the UK

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -479,7 +479,10 @@ def format_notification_status_as_field_status(status, notification_type):
             'virus-scan-failed': 'error',
             'returned-letter': None,
             'cancelled': 'error',
-        }
+        },
+        'sms': {
+            'sent': 'sent-international'
+        },
     }.get(
         notification_type,
         {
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Holding off merging this at Karls request as some further discussion needed first

quis:
Looks fine here:
image

Doesn’t work so well here:
Screenshot 2020-09-10 at 09 49 04

karlchillmaid:
@quis I thought that might happen. We do have other statuses that wrap when combined with the date and time, though, don't we?

quis:
For other statuses that wrap, we:
- control where the line break occurs
- remove the top margin so the two status lines align with the recipient/preview lines on the left

I’ll dig into this and see if we can do the same here.

quis:
I’ve updated this PR so it displays nicely on two lines now:
image

PR #3580, opened 2020-08-18
Make SVG alert icon show in Internet Explorer 10

quis:
IE 10 supports using SVG<sup>1</sup> but has some buggy behaviour when they’re used as background images.

Without an explicit width/height it stretches the viewBox of the SVG to fill the containing element. This causes the content of the file to display centered within the viewBox.

Explicitly setting the width and height seems to be the thing that fixes this. Out of the suggested fixes on Stackoverflow<sup>2</sup> this one seems to be the most straightforward.

1. https://caniuse.com/#feat=svg
2. https://stackoverflow.com/questions/17944354/svg-background-image-position-is-always-centered-in-internet-explorer-despite-b


tombye (approved):
Just checked on IE11 & 10 and looks good 👍

PR #3582, opened 2020-08-18
Fix or remove skip links on bulk send and contact list upload pages

CrystalPea:
From DAC accessibility audit of Notify:

```
There is a skip link present on the page that is not visible for all users. A keyboard-only user
can tab to this link although the reason for their temporary loss of focus is not clear.
Voice activations users using Dragon NaturallySpeaking can tag the link, but it will also be
unclear what this relates to. Links such as this are highly beneficial for screen readers when
there are large blocks of content present, however, we feel it may not be required in this
instance.
```

We fixed one such link (on the page for bulk-sending notifications by uploading a CSV with personalisation). Now it is visible when highlighted / tabbed to. This will make it usable to keyboard and voice activation users.

We removed the links from other pages, where file contents the link skipped to were directly below the skip link, making the skip link redundant.

### DIFF
--- a/app/templates/views/check/column-errors.html
+++ b/app/templates/views/check/column-errors.html
@@ -6,13 +6,6 @@
 {% from "components/back-link/macro.njk" import govukBackLink %}
 {% from "components/message-count-label.html" import message_count_label %}
 
-{% set file_contents_header_id = 'file-preview' %}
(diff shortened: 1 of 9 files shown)


tombye (approved):
This looks good to me 👍

PR #3585, opened 2020-08-21
Let people approve own broadcasts in training mode

quis:
We want to let users learn the system by trying it out. At the moment they can’t explore it fully by themselves because they’re blocked at the point of approving the broadcast, unless they involve in another member of their team. Having to involve another person is friction that will discourage people from exploring.

So this adds a button that lets people approve their own broadcasts in trial mode, with some rough-and-ready content that explains training mode and how it would be different to trial mode.

# Before

image

# After 

image

image


quis:
@karlchillmaid I’ve accepted all your changes, and squashed them into the original commit.

quis:
# Updated _after_

image
image

servingUpAces (approved):

PR #3586, opened 2020-08-21
Ensure H2 headings follow H1 headings

klssmith:
We had some cases where an H3 heading followed an H1. This was flagged as something to change in the accessibility report.


klssmith:
Thanks, I would definitely have forgotten to check the functional tests if you hadn't mentioned them 👍

PR #3590, opened 2020-08-25
Make simplification of polygons more sophisticated

quis:
Simplifying polygons means reducing the number of points used to render them. This pull request implements simplification such that, for any given input polygons, the combined point count of the simplified polygons is less than 200.

# Rules

When simplifying the polygons we are trying to get the smallest number of points while meeting these two rules: 
1. No part of the area the user has chosen can be cut off
2. The area of the simplified polygon should be as small as possible

# Techniques

This pull request introduces two techniques we weren’t using before:
1. Dilating and eroding the area to fill in concave details of the shape, like inlets and harbours
2. Making the simplification threshold proportionate to the perimeter of all polygons, so bigger and crinklier polygons get more simplification applied

It also shows the estimated bleed as a separate polygon. This lets us make it bigger (so it’s more closer the the approximate bleed) without having to send a bigger area to the CBC and compounding the amount of actual bleed.

# Examples

Before | After 
---|---
image | image 
image | image
image | image
image | image
image | image
image | image
image | image

### DIFF
--- a/app/broadcast_areas/create-broadcast-areas-db.py
+++ b/app/broadcast_areas/create-broadcast-areas-db.py
@@ -1,66 +1,49 @@
 #!/usr/bin/env python
 
-from copy import deepcopy
 from pathlib import Path
 
 import geojson
-import shapely.geometry as sgeom
+from notifications_utils.formatters import formatted_list
 
+from polygons import Polygons
 from repo import BroadcastAreasRepository
 
 package_path = Path(__file__).resolve().parent
-
-
-def convert_shape_to_feature(shape):
-    return {
-        "type": "Feature",
-        "properties": {},
-        "geometry": sgeom.mapping(shape),
-    }
-
-
-def simplify_polygon(series):
-    polygon, *_holes = series  # discard holes
-
-    approx_metres_to_degree = 111320
-    # initially buffer (extend area past perimeter) by ~25m
-    buffer_degrees = 500 / approx_metres_to_degree
-    # initially simplify (snap to closest point) by ~25m
-    simplify_degrees = 50.0 / approx_metres_to_degree
-
-    starting_polygon = sgeom.LineString(polygon).buffer(buffer_degrees)
-    simplified_polygon = None
-    num_polys = len(polygon)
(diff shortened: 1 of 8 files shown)


app/models/broadcast_message.py
```diff
@@ -72,30 +72,24 @@ def initial_area_names(self):
...
-        return broadcast_area_libraries.get_polygons_for_areas_lat_long(
-            *self._dict['areas']
+        return Polygons(
+            list(itertools.chain(*(
+                area.polygons for area in self.areas
+            )))
         )
 
-    @property
+    @cached_property
```
servingUpAces, line 83: Why is this a cached_property?

app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -173,3 +165,16 @@ def simplify_geometry(feature):
...
 
 repo.insert_broadcast_areas(areas_to_add)
+
+
+most_detailed_polygons = formatted_list(
+    sorted(point_counts, reverse=True)[:5],
+    before_each='',
+    after_each='',
+)
+print(  # noqa: T001
```
servingUpAces, line 175: Do we want to keep all the print statements?

app/models/broadcast_message.py
quis: We call it twice:
- once to show the actual broadcast area
- once to show the estimated bleed of the broadcast (we take the polygons for the broadcast area and buffer then by 1.5km)

Making it a cached property means we don’t need to get the JSON from the SQLite database twice, or re-smooth the polygons twice.

`cached_property` stores the result of calling the property against the object, so it’s only cached for the lifetime of that object (i.e. not between requests or anything like that).

app/broadcast_areas/create-broadcast-areas-db.py
quis: This is only run locally, so probably doesn’t hurt? Especially when we’ll adding new areas in https://www.pivotaltracker.com/story/show/174376516

servingUpAces (approved):

PR #3591, opened 2020-08-25
Fix copy to clipboard

tombye:
Changes how the aria live-region is implemented in the 'Copy to clipboard' (AKA `api_key`) component so what is announced is clearer.

https://www.pivotaltracker.com/story/show/174294716

# The problem

The 'copy to clipboard' button failed the Web Content Accessibility Guidelines (WCAG) on [success criteria 4.1.3 Status messages](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html). The tester reported that the content injected into the page is not announced.

The changes _are_ actually announced but the announcement is unclear because the whole component is an [aria live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) so it announces the button and text above it whenever its state changes:

copy_to_clipboard_current_live_region

# Proposed solution

## Before (using Voiceover)

- Focus moves to the button
- `Copy to clipboard, button` announced
- user clicks button
- `Press copy to clipboard, copied to clipboard, show ID` announced
- user clicks button
- `Press show ID, 8b31d0eb-4d15-4a42-9d0a-6aaea0bcdbd1, copy to clipboard` announced

[Audio version of existing flow](https://drive.google.com/file/d/1fnGDgy4OoKJMWSQlebldEyKwuXx-tg2T/view?usp=sharing)

## After (using Voiceover)

- Focus moves to the button
- `Copy to clipboard for tom.byers+notify-reply@digital-cabinet.office.gov.uk, Button` announced
- user clicks button
- `Id copied to clipboard, press button to show in page` announced
- user clicks button
- `Id returned to page, press button to copy to clipboard` announced

[Audio version of new flow](https://drive.google.com/file/d/1HZfXzoxuR_0n2mEEiVCCjrHUfOPdnqzk/view?usp=sharing)

# Changes to the code

There aren't many changes but I think they all need explaining in a bit of depth. They are:
1. move the live region so it only contains text we want announced and add some hidden text to explain how to:
    a. show the ID again when it's copied
    b. copy the ID when it is shown in the page again
2. add hidden text to the ID string to say what it is
3. add some hidden suffix text to the button so it works isolated from its context in the page

## 1. Move the live region so it only contains text we want announced

The live region is now a separate `<span>` which is shown when the ID is copied (as: "Copied to clipboard") and hidden when the ID is returned to the page, to make space for the ID string.

```html
<span class="api-key__notice" aria-live="assertive">
    ID </span>Copied to clipboard<span class="govuk-visually-hidden">, press button to show in page</span> 
</span>
```

The extra, hidden, text was added because the live region is set to `assertive` so it speaks over any other announcements. This means when the state changes it's not clear what the button does now.

The text says "press button to show in page" when copied and "press button to copy to clipboard" when shown again.

## 2. Add hidden text to the ID string to say what it is

If the component doesn't show its heading, we add a hidden prefix to the ID so it is labelled.

```html
<span class="api-key__key">
    <span class="govuk-visually-hidden">ID: </span>8b31d0eb-4d15-4a42-9d0a-6aaea0bcdbd1
</span>
```

## 3. Add extra text to the button so it works isolated from its context in the page

If you have multiple 'Copy to clipboard' components in the page, the buttons all have the same text.

```html
<button class="govuk-button govuk-button--secondary api-key__button--show">
    Show ID<span class="govuk-visually-hidden"> for tom.byers+notify-reply@digital.cabinet-office.gov.uk</span>
</button>
```

Fixes the issue raised in this story: https://www.pivotaltracker.com/story/show/174187762

# Changes to the tests

I rewrote the testsuite for `app/assets/javascripts/apiKey.js` so it tests all the variations of HTML the `api_key` component can give it onload. Seeing their output is the best way to understand the changes I made to the component code.

You can run them on their own by running this in your terminal:
```bash
npx jest --config tests/javascripts/jest.config.js tests/javascripts/apiKey.test.js
```

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -807,6 +807,7 @@ def add_template_filters(application):
         format_notification_status_as_url,
         format_number_in_pounds_as_currency,
         formatters.formatted_list,
+        formatters.normalise_lines,
         nl2br,
         format_phone_number_human_readable,
         format_thousands,
--- a/app/assets/javascripts/apiKey.js
+++ b/app/assets/javascripts/apiKey.js
@@ -6,13 +6,24 @@
   Modules.ApiKey = function() {
 
     const states = {
-      'keyVisible': (key, thing) => `
-        <span class="api-key__key">${key}</span>
-        <input type='button' class='govuk-button govuk-button--secondary api-key__button--copy' value='Copy ${thing} to clipboard' />
+      'keyVisible': (options) => `
+        <span class="api-key__key">
+          ${options.keyLabel ? '<span class="govuk-visually-hidden">' + options.thing + ': </span>' : ''}${options.key}
+        </span>
+        <span class="api-key__notice govuk-visually-hidden" aria-live="assertive">
+          ${options.onload ? '' : options.thing + ' returned to page, press button to copy to clipboard'}
+        </span>
+        <button class="govuk-button govuk-button--secondary api-key__button--copy">
+          Copy ${options.thing} to clipboard${options.name ? '<span class="govuk-visually-hidden"> for ' + options.name + '</span>' : ''}
+        </button>
       `,
-      'keyCopied': thing => `
-        <span class="api-key__key">Copied to clipboard</span>
-        <input type='button' class='govuk-button govuk-button--secondary api-key__button--show' value='Show ${thing}' />
+      'keyCopied': (options) => `
+        <span class="api-key__notice" aria-live="assertive">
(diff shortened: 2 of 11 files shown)


servingUpAces (approved):
I tried this out, it works ok.

servingUpAces:
I did get a little confused by Before and After example. The before is an example of the copy to clipboard for the API key and the after is an example for the reply to address.

PR #3592, opened 2020-08-26
Make ‘Add another area’ button wrap where possible

quis:
Broke this while fixing the spacing between the buttons and the areas.

# Before 

image

# After 

image


tombye (approved):
Looks good to me 👍

PR #3593, opened 2020-08-26
Fix link and button text across pages

tombye:
Some of our buttons and links share the same text so cause confusion to anyone using a screenreader, where you often view them out of the context of the page:

links_with_duplicate_text

This should fix all those with this problem, based on the findings of the report from the Digital Accessibility Centre (DAC).

https://www.pivotaltracker.com/story/show/174187762

All the content here was discussed with @karlchillmaid. We went with the approach of adding some extra words as a hidden suffix to the link/button text. The intention being, when reading left-to-right, you can stop as soon as you get enough of the context to understand.


servingUpAces (approved):
Tried this out - the links are much better for accessibility after this change.

PR #3595, opened 2020-08-26
Add a key to the map

quis:
image

Alternative treatment:

image

### DIFF
--- a/app/templates/views/broadcast/preview-areas.html
+++ b/app/templates/views/broadcast/preview-areas.html
@@ -1,6 +1,7 @@
 {% from "components/button/macro.njk" import govukButton %}
 {% from "components/page-header.html" import page_header %}
 {% from "components/page-footer.html" import page_footer %}
+{% from "views/broadcast/macros/area-map.html" import map %}
 
 {% extends "withnav_template.html" %}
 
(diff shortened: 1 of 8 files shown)


app/templates/views/broadcast/preview-areas.html
```diff
@@ -115,9 +115,19 @@
 
   {% if broadcast_message.areas %}
     <div id="map"></div>
-    <form action="{{ url_for('.preview_broadcast_message', service_id=current_service.id, broadcast_message_id=broadcast_message.id) }}" method="get">
-      {{ sticky_page_footer('Continue to preview') }}
-    </form>
+    <div class="js-stick-at-bottom-when-scrolling">
+      <div class="govuk-!-margin-bottom-3">
+        <div style="width: 1.6em; height: 1.6em; border: 3px solid #3A9DCE; background: #3A9DCE; display: inline-block; box-shadow: inset 0 0 0 2px #0b0b0c;">
```
tombye, line 120: Might be worth putting these in `<p>` tags, or in a list and putting some explanation for the colour ('light blue', 'dark blue'?).

app/templates/views/broadcast/preview-areas.html
```diff
@@ -115,9 +115,19 @@
...
+        </div>
+        <span style="position: relative; top: -0.55em; margin: 0 15px 0 5px">Will get alert</span>
```
tombye, line 122: If you need them to line up vertically, you could also put `display: flex` on the container and use [align-items](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items) to do it.

quis:
Updated design:

image

With hidden text (highlighted in yellow) for non-visual users:

image

idavidmcdonald (approved):
Looks good and works both in and not in a details tag.

Only question I have is if the choice of colours for the two areas is OK accessibility wise. I'm not too familiar but hear things about colour contrast, not just using colour but maybe using a line pattern through it too. Tom may be able to advise if you don't know.

PR #3599, opened 2020-08-27
Fix the autofocus component

quis:
Since moving textboxes to GOV.UK Frontend we’ve started putting the data attribute on the `input` element itself, not a wrapper around it.

This commit updates the Javascript accordingly.


tombye (approved):
Looks good to me 👍

PR #3601, opened 2020-08-27
Broadcast: only approved domain users can be invited

CrystalPea:
Only approved email domain users can be invited to broadcast services. This is done to improve security.

Question: Ok, so if a service is already a broadcast service, they will not be able to invite users from non-govt domains.

What to do in a scenario where a service already has users from random domain, like hotmail, and then they apply to become a broadcast service? Should there be a safeguard that says they have to remove those "unsafe" users first? Or is this scenario too unlikely to be taken into consideration?

Pivotal ticket: https://www.pivotaltracker.com/story/show/174510183

Screenshot 2020-08-27 at 18 07 52

### DIFF
--- a/tests/app/main/views/test_manage_users.py
+++ b/tests/app/main/views/test_manage_users.py
@@ -997,6 +997,37 @@ def test_invite_user_to_broadcast_service(
     )
 
 
+def test_invite_non_govt_user_to_broadcast_service_fails_validation(
+    client_request,
+    service_one,
+    active_user_with_permissions,
+    mocker,
+    sample_invite,
+    mock_get_template_folders,
+    mock_get_organisations,
+):
+    service_one['permissions'] = ['broadcast']
+    mocker.patch('app.models.user.InvitedUsers.client_method', return_value=[sample_invite])
+    mocker.patch('app.models.user.Users.client_method', return_value=[active_user_with_permissions])
+    mocker.patch('app.invite_api_client.create_invite', return_value=sample_invite)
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_manage_users.py
```diff
@@ -997,6 +997,37 @@ def test_invite_user_to_broadcast_service(
...
+        ],
+        'email_address': 'random@example.com'
+    }
+    page = client_request.post(
+        'main.invite_user',
+        service_id=SERVICE_ONE_ID,
+        _data=post_data,
+        _expected_status=200
+    )
+    app.invite_api_client.create_invite.assert_not_called()
```
quis, line 1027: I prefer to avoid `assert_not_called`, because if I make a typo like this, the tests will still pass:
```suggestion
    app.invite_api_client.create_invite.asset_not_called()
```

It’s harder to have a false positive with the statement written this way:
```suggestion
    assert app.invite_api_client.create_invite.called is False
```

It does appear [lots of times](https://github.com/alphagov/notifications-admin/search?q=assert_not_called&unscoped_q=assert_not_called) in this repo, so I’ve made a [separate PR](https://github.com/alphagov/notifications-admin/pull/3604) to remove the other usages.

quis (approved):
This is good to go once the comment is addressed.

I think in the future will we be creating new services for people who want to broadcast, not converting their existing services. So as long as we still restrict creating new services to people with public sector email addresses then it should be fine.

PR #3602, opened 2020-08-28
Fix font-size when Chrome autofills inputs

quis:
For some reason Chrome decides that using its own font declaration is preferable to keeping the input looking as it would without autofill.

This overrides that with our bigger, better font.

Before | After
---|---
image | image

https://stackoverflow.com/questions/57242841/input-text-very-small-when-hovering-on-autofill-suggestion

### DIFF
--- a/app/assets/stylesheets/app.scss
+++ b/app/assets/stylesheets/app.scss
@@ -253,3 +253,7 @@ details .arrow {
   }
 
 }
+
+input:-webkit-autofill::first-line {
+  @include core-19;
+}


servingUpAces (approved):

app/assets/stylesheets/app.scss
```diff
@@ -253,3 +253,7 @@ details .arrow {
   }
 
 }
+
+input:-webkit-autofill::first-line {
```
idavidmcdonald, line 257: How come this is just being added to the first line? rather than for all lines of an input?

Also is this sort of change one that should be in the design system? Or at least a more similar one that overrides any webkit autofill styling for those who don't want to use say core-19 size font?

quis: We have to be as specific as the browser is being, otherwise the browser’s style overrides ours. For some reason the browser is specifiying `::first-line`, which doesn’t even really make sense for an `input` since they can only ever have one line. But it’s what works so 🤷‍♂️

PR #3604, opened 2020-08-28
Remove uses of .assert_not_called

quis:
I prefer to avoid `assert_not_called`, because if I make a typo like this, the test will always pass:
```
    app.invite_api_client.create_invite.asset_not_called()
```

Similarly, if you forget to add the brackets, the test will always pass
```
    app.invite_api_client.create_invite.asset_not_called
```

It’s harder to make a mistake with the statement written like this:
```
    assert app.invite_api_client.create_invite.called is False
```

### DIFF
--- a/tests/app/main/views/test_api_integration.py
+++ b/tests/app/main/views/test_api_integration.py
@@ -301,7 +301,7 @@ def test_cant_create_normal_api_key_in_trial_mode(
         },
         _expected_status=400,
     )
-    mock_post.assert_not_called()
+    assert mock_post.called is False
 
 
 def test_should_show_confirm_revoke_api_key(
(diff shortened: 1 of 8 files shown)


servingUpAces (approved):

idavidmcdonald:
Nice. You may also be interested in this to further remove the risk of typos in assertions.

Article which isn't the most brief but doesn't explain using autospeccing: https://medium.com/python-pandemonium/mocking-has-a-weakness-speccing-removes-it-2d2068a17df8

By adding an autospec to all our mocks, which we might want to do anyway as I think it makes tighter code, would catch the `app.invite_api_client.create_invite.asset_not_called()` case and other cases where we might mispell a method.

And all it requires is a change of
```mock_post = mocker.patch('app.notify_client.api_key_api_client.ApiKeyApiClient.post')```

to

```mock_post = mocker.patch('app.notify_client.api_key_api_client.ApiKeyApiClient.post', autospec=True)```

It's generally good to add it when creating a mock (although I've fallen out of the habit) but there are a few small caveats https://stackoverflow.com/questions/35915703/when-using-unittest-mock-patch-why-is-autospec-not-true-by-default/35921144

PR #3605, opened 2020-08-28
Prevent user name email being cut off

tombye:
https://www.pivotaltracker.com/story/show/174293979

We currently clip the line containing the user name and email address on the /users page. This was picked up in the report done by the Digital Accessibility Centre (DAC) as it caused problems for users who zoomed the page:

image

This pull request (PR) introduces changes to how that text wraps to ensure all the information is visible.

# Changes

This PR:
1. puts the content for `user-list-item`s in a grid for consistency
2. changes the CSS controlling what happens when the text overflows its container

## Changing the CSS for how text behaves when it overflows its container

We changed the CSS so it doesn't clip the text with an ellipsis (...) when it overruns it's container any more. To compensate, we introduced a few rules for that text overflowing:
- we let the line break on whitespace
- if any of the words are larger than the container, we let them break by letter (this one is intended for the email address)

### If all text on the line fits

image

### If any overflow, break on whitespace

image

### If any single word overflows on its own

image

# How to review

This PR is best to review commit-by-commit.

### DIFF
--- a/app/assets/stylesheets/views/users.scss
+++ b/app/assets/stylesheets/views/users.scss
@@ -13,10 +13,8 @@ $item-top-padding: govuk-spacing(3);
     &-heading {
 
       padding-right: govuk-spacing(3);
-      white-space: nowrap;
       overflow: hidden;
-      text-overflow: ellipsis;
-      color: $secondary-text-colour;  // So the ellipsis is grey
+      overflow-wrap: break-word;
 
       .heading-small {
         color: $black;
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/views/users.scss
```diff
@@ -7,17 +7,14 @@ $item-top-padding: govuk-spacing(3);
 
   &-item {
 
-    padding: $item-top-padding 150px govuk-spacing(3) 0;
+    padding: govuk-spacing(3) 0px;
```
CrystalPea, line 10: Is linter happy with px after 0? :D

tombye: There is a warning but actually that rule (all `0` values not having `px`) shouldn't apply here because the `govuk-spacing` helper returns values with `px` so it will end up being consistent with that.

servingUpAces (approved):

PR #3609, opened 2020-09-03
Send simple polygons to API along with areas

CrystalPea:
When creating broadcast message, or updating it. We want to send simple polygons to API so we can later relay them to the broadcast provider.

Ticket: https://www.pivotaltracker.com/story/show/174385015
API PR: https://github.com/alphagov/notifications-api/pull/2965 (they should go together, API first)

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -62,9 +62,7 @@ class BroadcastMessage(JSONModel):
 
     @property
     def areas(self):
-        return broadcast_area_libraries.get_areas(
-            *self._dict['areas']
-        )
+        return self.get_areas(areas=self._dict['areas'])
 
     @property
     def initial_area_names(self):
@@ -82,14 +80,7 @@ class BroadcastMessage(JSONModel):
 
     @cached_property
     def simple_polygons(self):
-        polygons = Polygons(
-            list(itertools.chain(*(
-                area.simple_polygons for area in self.areas
-            )))
-        )
-        # If we’ve added multiple areas then we need to re-simplify the
-        # combined shapes to keep the point count down
-        return polygons.smooth.simplify if len(self.areas) > 1 else polygons
+        return self.get_simple_polygons(areas=self.areas)
 
     @property
     def template(self):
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_broadcast.py
```diff
@@ -510,7 +511,7 @@ def test_add_broadcast_area(
         service_id=SERVICE_ONE_ID,
         broadcast_message_id=fake_uuid,
         data={
-            'areas': ['ctry19-E92000001', 'ctry19-S92000003', 'ctry19-W92000004']
+            'areas': ['ctry19-E92000001', 'ctry19-S92000003', 'ctry19-W92000004'], 'simple_polygons': mocker.ANY
```
idavidmcdonald, line 514: I'm cautious on using mocker.ANY here. It tests that we are sending something to the API but nothing about what type of value we are sending (is it a list of strings or is it just a boolean for example) and is it the correct value?

I'm guessing simple_polygons might be a complicated list of long lat coords that may not be ideal for asserting however. If that is the case, is there something we can do like mocking `get_simple_polygons` and making sure the return value of `get_simple_polygons` is the thing we send to the API? Or a unit test of `get_simple_polygons` or one of the other methods?

Have a think. I'm not ruling out keeping it as is but there is some non trivial code added in this PR that would be nice to add a bit more testing around I think please

app/models/broadcast_message.py
```diff
@@ -122,16 +113,39 @@ def approved_by(self):
     def cancelled_by(self):
         return User.from_id(self.cancelled_by_id)
 
+    def get_areas(self, areas=None):
+        if not areas:
+            areas = self._dict['areas']
+        return broadcast_area_libraries.get_areas(
+            *areas
+        )
```
idavidmcdonald, line 121: Optional: Not sure if this method is trying to do too much. I feel like it may be better as a helper method like

```
def get_areas(self, areas):
        return broadcast_area_libraries.get_areas(
            *areas
        )
```

so it's just acting as an simple abstraction of knowing how to call broadcast_area_libraries properly and the decision for which areas to ask for would be up to the method calling it e.g.

```
def areas(self):
        return self.get_areas(self._dict['areas'])
```

And technically if you implement `get_areas` as suggested above, there is no reason why it actually needs to be tied to an instance of `BroadcastMessage`, or maybe it could be a class method.


Give it a brief bit of thought but this is very optional, I'm not 100% on my suggestion being an improvement or not.

CrystalPea: I like that, makes a lot of sense, thank you!

I also applied it to `get_simple_polygons()` ☺️

tests/app/main/views/test_broadcast.py
CrystalPea: That is a good suggestion, I mocked the `get_simple_polygons()` method now and assert that we are sending the mock coordinates ☺️

tests/app/main/views/test_broadcast.py
```diff
@@ -543,8 +550,14 @@ def test_add_broadcast_sub_area(
     fake_uuid,
     post_data,
     expected_selected,
+    mocker
 ):
     service_one['permissions'] += ['broadcast']
+    polygon_class = namedtuple("polygon_class", ["as_coordinate_pairs_lat_long"])
+    coordinates = [[50.1, 0.1], [50.2, 0.2], [50.3, 0.2]]
+    polygons = polygon_class(as_coordinate_pairs_lat_long=coordinates)
+    mocker.patch('app.models.broadcast_message.BroadcastMessage.get_simple_polygons', return_value=polygons)
```
idavidmcdonald, line 559: Optional: What you've done here works fine. You are using a named tuple as the return value of `get_simple_poloygons` which technically isn't the same type of return value that we'd really get but has the important fact that it has a property called `as_coordinate_pairs_lat_long` which is good enough for this test.

Optional way to do it using mocks and therefore not using the named tuple would be
```
coordinates = [[50.1, 0.1], [50.2, 0.2], [50.3, 0.2]]
    get_simple_polygons_return_value = mocker.MagicMock()
    coordinates_property_mock = mocker.PropertyMock(return_value=coordinates)
    type(get_simple_polygons_return_value).as_coordinate_pairs_lat_long = coordinates_property_mock
    mocker.patch('app.models.broadcast_message.BroadcastMessage.get_simple_polygons', 
    return_value=get_simple_polygons_return_value)
```

It is a bit gnarly though and not sure it gives that much more benefit over what you've done apart from using mocks. I'm not sure which option I suggest but give it a thought and feel free to stick with what you've done.

idavidmcdonald (approved):

tests/app/main/views/test_broadcast.py
CrystalPea: Ahh, cheers, interesting. I think I will leave it as is. Like you say, it's less gnarly and does the job :).

PR #3610, opened 2020-09-04
Fix pill tab problems

tombye:
The 'pill' tabs we have on pages like /templates and /usage are marked up as both links and tabs. This is causing confusion for users of screenreaders.

image

https://www.pivotaltracker.com/story/show/174625880

## Proposed solution

We decided they shouldn't be tabs anyway. Tabs show/hide sections of a page whereas the 'pill's cause navigations so should be grouped in a [navigation element](https://www.w3.org/TR/html52/sections.html#elementdef-nav).

### Changes involved

This pull request:
1. removes all semantics identifying the 'pill's as tabs
2. wraps them in a `nav`
3. labels the `nav` with the h1 (using `aria-labelledby`)
4. makes the selected item a link as well (like the other `nav`s) and uses [aria-current](https://www.w3.org/TR/wai-aria-1.2/#aria-current) to mark it as the one representing the current page

Number 3. is done to help distinguish the `nav` from the other `nav`s in the page (see https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/ for reasons why).


app/templates/components/pill.html
```diff
@@ -9,23 +9,18 @@
...
+            {{ big_number(count, **big_number_args) }}
           {% endif %}
-              <div class="pill-label{% if not show_count %} pill-centered-item{% endif %}">{{ label }}</div>
-        {% if current_value == option %}
-            </div>
-        {% else %}
-            </a>
-        {% endif %}
-          </li>
+            <div class="pill-item__label{% if current_value == option %} pill-item__label--selected{% endif %}{% if not show_count %} pill-item--centered{% endif %}">{{ label }}</div>
```
idavidmcdonald, line 21: is `pill-item__label--selected` a valid class that we have defined styling for?

I can see `pill-item__label` and `pill-item--selected` but I can't see `pill-item__label--selected`?

idavidmcdonald (commented):
From what I can tell it generally looks good. One question to answer and if that is OK then I can give the thumbs up.

I've run locally as well and all appears to be fine, at least from a presentation perspective.

app/templates/components/pill.html
tombye: Good spot. I think what was left over from when we were looking at removing the underline for selected items. It was decided not to so this is now redundant. I just removed it in a new commit.

idavidmcdonald (approved):
Good work

PR #3613, opened 2020-09-07
Show international postage as soon as address is entered

quis:
At the moment we don’t check whether a one-off letter is international until the user’s clicked send. It’s more accurate to show that the letter will be sent internationally as soon as we know the address.

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/786


idavidmcdonald (approved):
Will need a make freeze requirements when utils is merged but looks good

PR #3614, opened 2020-09-08
Select from counties when creating broadcasts

leohemsted:
Took a little longer than expected, with a lot of rebases along the way! Did a fair bit of refactoring along the way, so it's best reviewed commit by commit, and I really do mean that.

You can now create broadcasts with rural counties, as well as metropolitan boroughs (which were already included as they're unitary districts), rural districts (local authorities), and electoral wards.

I implemented this mock-up from chris:

image

| Rural Counties are visible                                            | image |
|-----------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------|
| Local Authorities are visible                                         | image |
| Electoral Wards are NOT visible                                       | image |
|                                                                       |                                                                                                               |
| Ward selection for a local authority is unchanged                     | image |
| NEW! Drill down into districts (local authorities) for a rural county | image |

# TODO list

(I don't think these block review/merging of this, it's spent long enough kicking around that I think getting eyes on the rest of the code is more important)

* [ ] The numbers on the top level library page ("Aberdeen City, Aberdeenshire, Angus, and 213 more…") are wrong
* [ ] Back button doesn't have context of how you got to a page. It always takes you `.../libraries`. As an immediate quick win, it could redirect you to `.../libraries/wd20-lad20-ctyua19`, but ideally it'd have a back link, like `/libraries/wd20-lad20-ctyua19/lad20-E07000008?prev_group_id=ctyua19-E10000003` which it could populate the back link from


app/broadcast_areas/repo.py
```diff
@@ -166,17 +166,25 @@ def get_areas(self, area_ids):
         return areas
 
     def get_all_areas_for_library(self, library_id):
-        # only interested in areas with children - local authorities, counties, unitary authorities. not wards.
-        q = """
-        SELECT id, name
-        FROM broadcast_areas
-        JOIN (
-            SELECT DISTINCT broadcast_area_library_group_id
+        if library_id == 'wd20-lad20-ctyua19':
```
quis, line 170: You could make this generic by looking at the return value of
```
select broadcast_area_library_group_id from broadcast_areas where broadcast_area_library_id = ? limit 1
```

quis (changes requested):
I think this needs tests for `GET` and `POST` on the new counties page, just to verify it’s showing the right list of things. Otherwise good to go, and thanks for splitting into decent commits.

Happy to leave the TODOs for separate PRs.

app/broadcast_areas/repo.py
leohemsted: as in "show me all areas in this library that have a parent"? Yeah that could work. i think this is the correct-est implementation:

```sql
select
exists(
  select *
  from broadcast_areas
  where broadcast_area_library_id = ? and
  broadcast_area_library_group_id is not null
)```

quis: Yeah, that looks tidy. Does mean running two queries, but we only run these queries on app startup anyway so performance isn’t much of a concern.

leohemsted:
@quis tests squashed in to https://github.com/alphagov/notifications-admin/pull/3614/commits/b8fc46c7df9c3ac2186f5ce7e62b4d3f7acc8944

quis (approved):

PR #3617, opened 2020-09-09
Give estimates of the number of phones in a broadcast area

quis:
We need to give people a better feel for the consequences of broadcasting an alert. We’ve seen in research that some users will assume it is subscription based, or opt-in, rather than going to every phone in the area.

I reckon that the most effective way to communicate this is to put some numbers next to the areas, to give people an idea of how many people will get alerted.

We can estimate how many phones are in an area by:
- taking the population<sup>1</sup> of all electoral wards in that area
- multiplying it by the percentage of people<sup>2</sup> who own an internet connected phone<sup>3</sup>

The Office for National Statistics publish both these datasets.

These numbers won’t be exact, but should be enough to give people a feel for the severity of what they’re about to do. We can see if they achieve this aim in user research.

***

1. For the City of London, where no population figures are available, we make some assumptions based on how many people are typically in the City during the day, and how large each ward is
2. The number of people who own an internet connected phone varies a lot by age. Since the population data for each ward is broken down by age we can factor this in. Simplified, the calculation looks like this:
    - take the Abbey ward of Barking and Dagenham
    - in this ward there are 26 people aged 80
    - 40% of people over 65 have an internet-connected phone
    - therefore 10 of these 80-year-olds would be likely to receive a broadcast  
    - (repeat for all other ages)
3. This is a proxy for the number of people who are likely to have a 4G capable phone, because only 4G capable phones will be receiving  broadcasts to begin with 

# Examples

image

***

image

***

image


app/broadcast_areas/__init__.py
```diff
@@ -51,6 +52,16 @@ def sub_areas(self):
...
         ]
 
+    @property
+    def count_of_phones(self):
+        if self.id.endswith(CITY_OF_LONDON.WARDS):
+            return CITY_OF_LONDON.DAYTIME_POPULATION * (
+                self.polygons.estimated_area / CITY_OF_LONDON.AREA_SQUARE_MILES
+            )
+        if self.sub_areas:
+            return sum(area.count_of_phones or 0 for area in self.sub_areas)
```
idavidmcdonald, line 62: So I'm not too familiar with the `sub_areas` concept although I'm guessing its some sort of smaller area. Two questions based on that

Should these two lines be moved up to the top of this method? So that the recursive behaviour also works for city of london wards? Or will it be guaranteed that city of london wards don't have sub areas?

And then how come we need the `or 0`. Is it possible for a area to not have a integer return value for `.count_of_phones`?

app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -54,6 +61,46 @@ def polygons_and_simplified_polygons(feature):
...
+
+    if country_or_ward_code not in area_to_population_mapping:
+        raise ValueError(f'No population data for {country_or_ward_code}')
+
+    population = area_to_population_mapping[country_or_ward_code]
+    population_by_range = {}
+
+    for range, ownership in SMARTPHONE_OWNERSHIP_BY_AGE_RANGE.items():
+        min, max = range
+        population_by_range[range] = sum(
```
idavidmcdonald, line 87: Optional, could call this something more reflective like 'ownership_by_range` or 'phone_ownership_by_age'

app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -187,7 +263,8 @@ def _add_local_authorities(dataset_id):
             dataset_id,
             'ctyua19-' + ctyua_id if ctyua_id else None,
             feature,
-            simple_feature
+            simple_feature,
+            None,
```
idavidmcdonald, line 267: How come we are setting None for local authorities and counties and unitary authorities? Is it that they are larger areas made up of electoral wards so we only store the data for the electoral wards and when we need to calculate the area for a local authority we do so by looking at the sum of it's sub areas?

app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -54,6 +61,46 @@ def polygons_and_simplified_polygons(feature):
     )
 
 
+def estimate_number_of_smartphones_in_area(country_or_ward_code):
```
idavidmcdonald, line 64: This file contains a fair bit of complex code. I would be interested in thinking about what unit testing we can put behind this? We are fairly confident it will run successfully as we can run the script and that will tell us if it passes but cautious about lack of testing for the validity of the data.

app/broadcast_areas/__init__.py
```diff
@@ -51,6 +52,16 @@ def sub_areas(self):
             for row in BroadcastAreasRepository().get_all_areas_for_group(self.id)
         ]
 
+    @property
+    def count_of_phones(self):
```
idavidmcdonald, line 56: Could we get some unit testing of this method please

app/models/broadcast_message.py
```diff
@@ -15,6 +16,13 @@
 from app.notify_client.service_api_client import service_api_client
 
 
+def round_to_significant_figures(x, n):
```
idavidmcdonald, line 19: Optional, could give this more readable variable names like `value` and `num_of_sig_figs`

idavidmcdonald (commented):
Generally looks good but I think could do with some more unit testing to support the code please

app/broadcast_areas/__init__.py
quis: Wards are the lowest-level area, so will never have sub-areas.

The two possible relationships are:
- Upper tier local authority (colloquially a ‘county‘) → Lower tier local authority (colloquially a ‘district‘) → Electoral ward
- Unitary Authority → Ward

`BroadcastArea.sub_areas` always gives you wards.

***

Looks like the `or 0` was a holdover from when the data was incomplete. I’ve checked the data now and we have figures for all wards so I’ve removed it in 2348736c1

app/broadcast_areas/create-broadcast-areas-db.py
quis: Changed to `smartphone_ownership_for_area_by_age_range` in 7511ff8dd

app/broadcast_areas/create-broadcast-areas-db.py
quis: Correct. I think it’s better to keep the data we store normalised.

app/models/broadcast_message.py
quis: Changed in 9db2a35c9

app/broadcast_areas/__init__.py
quis: I’m going to need to back out that change because we do seem to have some lower-tier local authorities that are missing wards (see https://www.pivotaltracker.com/n/projects/1443052)

app/broadcast_areas/create-broadcast-areas-db.py
quis: I refactored things a bit which allowed me to write a unit test in e61092dbb

app/broadcast_areas/__init__.py
quis: Some tests added in b9f75218d1f2ab44e9b111f9c87f914d7642b040

tests/app/broadcast_areas/test_broadcast_area.py
```diff
@@ -147,3 +148,31 @@ def test_repository_has_all_libraries():
...
+        assert area.count_of_phones > 0
+
+
+def test_bryher_has_hard_coded_count():
+    bryher = broadcast_area_libraries.get_areas('wd20-E05011090')[0]
+    assert bryher.name == 'Bryher'
+    assert bryher.count_of_phones == 76.44
+
+
+def test_city_of_london_counts_are_not_derived_from():
```
idavidmcdonald, line 170:
```suggestion
def test_city_of_london_counts_are_not_derived_from_population():
```

app/broadcast_areas/create-broadcast-areas-db.py
idavidmcdonald: Cool. An inline comment may be useful in that case.

app/broadcast_areas/populations.py
```diff
@@ -0,0 +1,62 @@
...
+MEDIAN_AGE_UK = 40
+
+for min, max in SMARTPHONE_OWNERSHIP_BY_AGE_RANGE.keys():
+    if min <= MEDIAN_AGE_UK <= max:
+        MEDIAN_AGE_RANGE_UK = (min, max)
+
+
+class CITY_OF_LONDON:
+    WARDS = (
+        'E05009288', 'E05009289', 'E05009290', 'E05009291', 'E05009292',
```
idavidmcdonald, line 26: I see you added 'E05009288' in one of the new commits. I'm guessing this means you had just forgotten it in the first case?

idavidmcdonald (commented):
A few small comments before I head off for the day but I need to look again in the morning with a fresh head around what wards are and unitary authorities and all that jazz. Generally looks good though and the unit tests are an improvement so thanks!

app/broadcast_areas/populations.py
quis: Yeah, I realised we only had 24 when there should have been 25 (as in the database)

quis:
I’ve added some commits to handle the case where an area has no wards. These areas will return a count of 0 for now:
image

idavidmcdonald (approved):
Thanks for the changes Chris. Looks really good now

PR #3622, opened 2020-09-10
Autofocus off by default on live-search

CrystalPea:
Accessibility audit uncovered issues screen-reader users would have with autofocus on Notify pages. Autofocus can lead to confusion for those types of users and make them miss parts of the page.

Hence, we decided that autofocus will be turned off by default for live-search component.

We decided to keep the autofocus on for live-search text box when:
1. the page is a task page - as opposed to browsing pages, where user wonders freely, task pages have more established flow, so page focusing on textbox can actually be helpful.
2. page does not have actionable elements above the autofocus.

Screenreaders announce the page title when a page loads so it's important this still happens, so users know they're on a new page. The JS that sets focus has been changed so the page title still gets announced when the page loads (see https://github.com/alphagov/notifications-admin/pull/3622/commits/ac96d344a3874070f883b3cddce3db97ddde1f4a for details).

Story ticket: https://www.pivotaltracker.com/story/show/174438066

### The code

Go commit-by-commit. This branch has a lot of history which could confuse otherwise.

### The behaviour

If you're comfortable using Voiceover, try navigating to a page listed here and comparing that with the current behaviour.

If you're not comfortable using Voiceover, look through the JS tests to check if the behaviours added meet the intention of this story.

Do note that we'll need to do extra testing with other screenreaders and users of screenreaders after this is merged so any screenreader issues not caught by the authors should also come up them.

### DIFF
--- a/app/templates/components/live-search.html
+++ b/app/templates/components/live-search.html
@@ -1,21 +1,25 @@
-{% from "components/textbox.html" import textbox %}
-
 {% macro live_search(
     target_selector=None,
     show=False,
     form=None,
-    label=None
+    label=None,
+    autofocus=False
 ) %}
     {%- set search_label = label or form.search.label.text %}
+
+    {%- set param_extensions = {
+      "label": {"text": search_label},
+      "autocomplete": "off",
+    } %}
+
+    {% if autofocus %}
+      {% set x=param_extensions.__setitem__("attributes", {"data-module": "autofocus"}) %}
+    {% endif %}
+
     {% if show %}
-      <div data-module="autofocus">
         <div class="live-search js-header" data-module="live-search" data-targets="{{ target_selector }}">
-          {{ form.search(param_extensions={
-            "label": {"text": search_label},
-            "autocomplete": "off"
-          }) }}
+          {{ form.search(param_extensions=param_extensions) }}
           <div role="region" aria-live="polite" class="live-search__status govuk-visually-hidden"></div>
         </div>
-      </div>
     {% endif %}
 {% endmacro %}
--- a/app/assets/javascripts/autofocus.js
+++ b/app/assets/javascripts/autofocus.js
@@ -3,13 +3,26 @@
 
   Modules.Autofocus = function() {
     this.start = function(component) {
-      var forceFocus = $(component).data('forceFocus');
+      var $component = $(component),
+          forceFocus = $component.data('forceFocus'),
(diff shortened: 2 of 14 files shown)


tombye (commented):
This is a really excellent piece of work and a really clever fix for the pages with autofocus.

I have one suggestion, could we send through the full page title instead of just the page name? I mean "Set email branding – My service  – GOV.UK Notify" instead of "Set email branding".

When you navigate to any other page you get the full title, which ends up signifying a page load. Without it, it's a bit unclear you've moved to a new page.

CrystalPea:
@tombye Thank you 😊 . Your suggestion is good. I will apply it and let you know when I'm finished 👍

tombye:
I've added a few changes so the aria-label is added by JS and the textbox's label is reset when focus shifts. Full explanation is now in the description.

I'm aware I'm pushing into someone else's pull request so please consider these as a proposal for how this should be done and subject to questioning by any reviewers.

app/templates/components/live-search.html
```diff
@@ -1,21 +1,25 @@
...
 ) %}
     {%- set search_label = label or form.search.label.text %}
+
+    {%- set param_extensions = {
+      "label": {"text": search_label},
+      "autocomplete": "off",
+    } %}
+
+    {% if autofocus %}
+      {% set x=param_extensions.__setitem__("attributes", {"data-module": "autofocus"}) %}
```
idavidmcdonald, line 16: This seems a bit of a gnarly way to do it but I'm not sure if there is a nicer way in Jinja

idavidmcdonald (approved):
Nice one. Looks good from what I can tell. Commits make sense etc. Only slight confusion was it appears in the penultimate commit you are adding a test and then in the final commit you are adding the functionality. Makes sense if you are doing test driven development but it was a little confusing when reviewing commit by commit. If you wanted, you could squash those two commits together so they come as a single package but I really don't mind.

🚢

CrystalPea (commented):
@tombye I like the changes you made, much less repetition and it is neat that page title will only be read out once.

tombye:
> Nice one. Looks good from what I can tell. Commits make sense etc. Only slight confusion was it appears in the penultimate commit you are adding a test and then in the final commit you are adding the functionality. Makes sense if you are doing test driven development but it was a little confusing when reviewing commit by commit. If you wanted, you could squash those two commits together so they come as a single package but I really don't mind.
> 
> 🚢

That's a good point. I'd like to leave it here for this bit of work to get this out but I'll make sure to commit like that in future 👍

PR #3624, opened 2020-09-10
Add accessible letter specification

karlchillmaid:
We need to provide an accessible version of the [letter specification PDF](https://docs.notifications.service.gov.uk/documentation/images/notify-pdf-letter-spec-v2.4.pdf).

Because the PDF was created in Illustrator and contains largely visual information, there's probably not much we can do to make the file more accessible. Instead we're going to provide an alternative written description of the key information a user needs to create a letter according to our spec. We can link to this from the API documentation, the guidance and from within the _upload a letter_ journey in Notify.

We've written the information for the logo block, address block and letter contact block as you would in CSS or any Adobe software.


app/templates/views/guidance/letter-specification.html
```diff
@@ -0,0 +1,47 @@
...
+
+  <p class="govuk-body">Page size and layout: A4 portrait (210 × 297 mm)</p>
+  <p class="govuk-body">Maximum file size: 2 MB</p>
+  
+  <p class="govuk-body">Your letter must 10 pages or less (5 double-sided sheets of paper).</p>
+  <p class="govuk-body">Content must be inside the printable area.</p>
+  
+  <h2 class="heading-medium">Page 1 of 10</h2>
+  
+  <p class="govuk-body">Top margin 5.00mm</p>
```
karlchillmaid, line 20: @quis Do we need the .00 on all of these?

app/templates/views/guidance/letter-specification.html
```diff
@@ -0,0 +1,47 @@
...
+  <p class="govuk-body">Left margin 15.00mm</p>
+  <p class="govuk-body">Right margin 15.00mm</p>
+  <p class="govuk-body">Bottom margin 5.00mm</p>
+
+  <h3 class="heading-small">Logo block</h3>
+  
+  <p class="govuk-body">Position: 5mm from top edge, 15mm from left edge</p>
+  <p class="govuk-body">Size: 200mm wide by 25mm high</p>
```
karlchillmaid, line 28: Just checking that vertical position, horizontal position is the right order if the dimensions are given as width by height?

app/templates/views/guidance/letter-specification.html
```diff
@@ -0,0 +1,47 @@
...
+  
+  <h3 class="heading-small">Address block</h3>
+  
+  <p class="govuk-body">Position: 24.6mm from top edge, 39.5mm from left edge</p>
+  <p class="govuk-body">Size: 95.4mm wide by 26.8mm high</p>
+  
+  <h3 class="heading-small">Letter contact block</h3>
```
karlchillmaid, line 35:
```suggestion
  <h3 class="heading-small">Contact block</h3>
```
Doesn't seem necessary to repeat `letter` here

karlchillmaid: `Letter contact block` is what we call it in the UI. I'm sure we talked about changing it at some point, because we saw a few users struggling to understand what it was in usability testing.

karlchillmaid:
We need to decide where to put this page.

Should it be a child of the [upload a letter guidance](https://www.notifications.service.gov.uk/using-notify/guidance/upload-a-letter)? If so, will it appear in the left-hand nav?

app/templates/views/guidance/letter-specification.html
quis: I think it should be left, top, as in x, y.

CSS has broken my brain because it does dimensions in clockwise order, as in top, right, bottom, left (which I always have to remember as TRouBLe).

app/templates/views/guidance/letter-specification.html
quis: The most precise we go is 24.6mm. So options are:
- .0 everywhere, so 5.0mm and 24.6mm (5.0mm looks odd to me though)
- .00 everywhere to match the PDF, so 5.00mm and 24.60mm
- drop the 0, so 5mm and 24.6mm

app/templates/views/guidance/letter-specification.html
quis: I think use _Letter contact block_ for consistency then. If we change it to something better we should do that everywhere (which is out of the scope of these accessibility improvements).

app/templates/views/guidance/letter-specification.html
karlchillmaid: Ok, I'll update them to fit that. Left, top, width, height makes more sense to me.

app/templates/views/guidance/letter-specification.html
karlchillmaid: Thinking about assistive tech reading this out, perhaps option 3 is the clearest/least verbose?

Agree that 5.0 looks odd.

app/templates/views/uploads/choose-file.html
```diff
@@ -48,7 +48,7 @@ <h2 class="heading-medium">Your file must meet our letter specification</h2>
       <li>smaller than 2MB</li>
     </ul>
 
-  <p class="govuk-body">To help you set up your letter, you can download our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.letter_spec') }}">letter specification document (PDF)</a>.</p>
+    <p class="govuk-body">To help you set up your letter, you can <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.letter_spec') }}">check our letter specification</a> or <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.letter_spec') }}">download the letter specification as a PDF</a>.</p>
```
karlchillmaid, line 51: @quis would it be better to say less on this page – remove all the details and the PDF – and simply link out to the letter spec page with all the info **and** the PDF?

quis: Just link to the page I think

app/templates/views/guidance/letter-specification.html
```diff
@@ -0,0 +1,57 @@
...
+  <p class="govuk-body">Maximum file size: 2 MB</p>
+  <p class="govuk-body">Your letter must 10 pages or less (5 double-sided sheets of paper).</p>
+  <p class="govuk-body">The content of your letter must appear inside the printable area.</p>
+
+  <p class="govuk-body">To help you set up your letter, you can download our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.letter_spec') }}">letter specification as a PDF</a>.</p>
+
+  <details class="govuk-details" data-module="govuk-details">
+  <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+      View a text version of the letter specification
```
karlchillmaid, line 22:
```suggestion
      See a text version of the letter specification
```

quis (approved):

PR #3627, opened 2020-09-14
Increase threshold for minimum polygon size

quis:
We filter out very small polygons from the original data to remove glitches. These glitches are caused by trying to subtract the water from a polygon that includes some land and some water, but using two different definitions or resolutions of mean high water line.

If we don’t do this then we end up with a bunch of very small polygons which lie far outside the understood area of a place, causing large overspill.

We need to increase the threshold for this process because we’re still seeing this problem around Bristol and Norwich.

This does mean we lose a few very small polygons in places like Shetland and the Scilly Isles, but not in such a way that we would avoid broadcasting to them (because they’d still be caught by the simplification and overspill).

Before | After
---|---
image | image
image | image
image | image

### DIFF
--- a/app/broadcast_areas/polygons.py
+++ b/app/broadcast_areas/polygons.py
@@ -41,7 +41,7 @@ class Polygons():
     # The threshold for removing very small areas from the map. These
     # areas are likely glitches in  the data where the shoreline hasn’t
     # been subtracted from the land properly
-    minimum_area_size_square_metres = 50 ** 2
+    minimum_area_size_square_metres = 14_000
 
     def __init__(self, polygons):
         if not polygons:


quis:
N.B. the entire Isles of Scilly have one phone mast per provider, so this change won’t have any impact on which masts start broadcasting:

image

Source: [Cellmapper](https://www.cellmapper.net/map?MCC=234&MNC=15&type=LTE&latitude=49.945414305556056&longitude=-6.3601587336737655&zoom=12.052274619394717&showTowers=true&showTowerLabels=true&clusterEnabled=true&tilesEnabled=true&showOrphans=false&showNoFrequencyOnly=false&showFrequencyOnly=false&showBandwidthOnly=false&DateFilterType=None&showHex=false&showVerifiedOnly=false&showUnverifiedOnly=false&showLTECAOnly=false&showENDCOnly=false&showBand=0&showSectorColours=true)

app/broadcast_areas/polygons.py
```diff
@@ -41,7 +41,7 @@ class Polygons():
     # The threshold for removing very small areas from the map. These
     # areas are likely glitches in  the data where the shoreline hasn’t
     # been subtracted from the land properly
-    minimum_area_size_square_metres = 50 ** 2
+    minimum_area_size_square_metres = 14_000
```
idavidmcdonald, line 44: 14000 > 2500 so that looks good.

I'm assuming the underscore in the value is purely presentational? I hadn't seen that done before but looks fine

idavidmcdonald (commented):
We are also deleting a broadcast areas sqlite file in this PR which doesn't appear to be mentioned in the commit/pr message. Is this intentional?

app/broadcast_areas/polygons.py
quis: Yup, see [PEP 515 – Underscores in Numeric Literals](https://www.python.org/dev/peps/pep-0515/)

> The underscores have no semantic meaning, and literals are parsed as if the underscores were absent.

quis:
image

I think what Github is trying to say is:
- 99% of this file has changed 
- the size has reduced by 512kb

I haven’t deleted the file, just regenerated it to update the stored polygons.

idavidmcdonald:
Ah thanks, that makes sense.

How do you regenerate the file? Is there a command to run that is documented anywhere so I know if I need to do it in the future?

idavidmcdonald (approved):

quis:
`./app/broadcast_areas/create-broadcast-areas-db.py` – don’t think it’s documented yet

PR #3628, opened 2020-09-15
Update focus style to GOVUK Frontend 3 version

tombye:
Updates our focus style to GOVUK's new one, to make the contrast compliant.

https://github.com/alphagov/notifications-admin/pull/3303/files

This is a draft proposal, intended to show where this work is so it can have some design work.

This includes:
1. changing the core focus style for links and buttons
2. updating all the variants of links and buttons custom to Notify

Number 2. includes an attempt to update the 'pill' items which probably needs some more work.

## New GOVUK Focus style

The new focus style is explained in [this blog post](https://designnotes.blog.gov.uk/2019/07/29/weve-made-the-gov-uk-design-system-more-accessible/) and [this page on their docs](https://design-system.service.gov.uk/get-started/focus-states/).

The main thing to bear in mind, for any variants we introduce is:
- design system's design actually includes 2 styles (the colour and the underline) so part of it will always contrast with the background colour
- they need to comply with [WCAG 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

This [WCAG techniques article](https://www.w3.org/WAI/WCAG21/Techniques/css/C40) contains some useful tests for compliance of any similar design.

## Focus styles custom to Notify

Here's some screenshots of my attempt at updating the various links/buttons custom to Notify. I based this work on those found in https://github.com/alphagov/notifications-admin/pull/3303 so there's a chance I might have missed some...

### Link in dashboard table

image

### Big numbers

image

### Show more component

image

### Service navigation links

image

### Org' navigation links

image

### Previous/next navigation

image

### Link in table cell

image

### Pill items

image

### Separate pill items

image

### DIFF
--- a/app/assets/stylesheets/components/big-number.scss
+++ b/app/assets/stylesheets/components/big-number.scss
@@ -97,7 +97,10 @@
 
       color: $light-blue-25;
 
-      .big-number {
+      .big-number,
+      .big-number-number,
+      .big-number-smaller,
+      .big-number-label {
         color: $light-blue-25;
       }
 
@@ -105,7 +108,18 @@
 
     &:active,
     &:focus {
-      outline: 3px solid $yellow;
+      background: $govuk-focus-colour;
+      border: none;
+      padding: 2px; /* replace the spacing the border gave */
+      /* override the default focus style to inset the underline */
+      box-shadow: inset 0 -4px $govuk-focus-text-colour;
+
+      .big-number-number,
+      .big-number-smaller,
+      .big-number-label {
+        color: $govuk-focus-text-colour;
+        text-decoration: none;
+      }
     }
 
     .big-number,
--- a/app/assets/stylesheets/components/message.scss
+++ b/app/assets/stylesheets/components/message.scss
@@ -147,6 +147,10 @@ $message-type-bottom-spacing: govuk-spacing(4);
       background-image: file-url('folder-blue-bold-hover.svg');
     }
 
+    &:focus {
+      background-image: file-url('folder-black-bold.svg');
+    }
+
   }
 
   &-template {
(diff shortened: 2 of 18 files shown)


quis:
### Link in dashboard table

image

Think this looks crisper when the black border overlaps the grey keylines:
image

### Big numbers

image

This would look better by insetting the black border so the white space above the green box is preserved (ignore the chamfer):

image

### Service navigation links and org navigation links

image

Again, crisper if the black overlaps the grey keyline:
image

tombye:
@quis I've pushed changes for most of the things you mentioned here and on slack.

## Pill items

image

image

## Org nav links

image

## Link in dashboard table

image

## Big numbers

image

quis:
A few more things I’ve noticed…

## Org nav links

image

The org link has a `1px` thicker border than switch service 

# Template page

image

The focus state is being triggered on hover here

# Dashboard

image

The blue boxes have lost their hover state (click ‘0 emails’ and look at the sending/delivered/failed tabs for an example of what the hover should look like)

The focus state is triggered on hover

quis:
Also `placeholders.scss` should be updated to use the same yellow, like so:

image

tombye:
I've pushed a few commits that:
- make the underline of the org links an equal thickness
- return the hover effect to the big numbers and the separate pill items

I think the big number hover wasn't happening before. At least it doesn't on preview at the moment, so I made a guess at how to put it back.

tombye:
I've done the following which, I believe, makes this ready for reviewing:
1. fixed the edit links on user list items (missed in previous work)
2. rebased all fixes for the same links/buttons into single commits

Number 1. turned this:
image
...into this:
image

quis:
I found one place where the old yellow is still hiding:
image

For extra credit you could use [the monochrome version](https://github.com/alphagov/notifications-admin/blob/master/app/assets/images/folder-black-bold.svg) of the folder icon here:
image

Everything else looks excellent 👏

quis (approved):
I have weirdly enjoyed getting all these pixels lining up good. Thanks for all your work on this.

PR #3629, opened 2020-09-15
fix back links for broadcast libraries

leohemsted:
the county page links to districts. add a `prev_area_slug` query param to those URLs, so we can make the back link go back to the county page, if the user came from there.

if the user came straight from the library page, don't add the query param and don't change behaviour.

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -147,6 +147,25 @@ def choose_broadcast_area(service_id, broadcast_message_id, library_slug):
     )
 
 
+def _get_broadcast_sub_area_back_link(service_id, broadcast_message_id, library_slug):
+    prev_area_slug = request.args.get('prev_area_slug')
+    if prev_area_slug:
+        return url_for(
+            '.choose_broadcast_sub_area',
+            service_id=service_id,
+            broadcast_message_id=broadcast_message_id,
+            library_slug=library_slug,
+            area_slug=prev_area_slug,
+        )
+    else:
+        return url_for(
+            '.choose_broadcast_area',
+            service_id=service_id,
+            broadcast_message_id=broadcast_message_id,
+            library_slug=library_slug,
+        )
+
+
 @main.route(
     '/services/<uuid:service_id>/broadcast/<uuid:broadcast_message_id>/libraries/<library_slug>/<area_slug>',
     methods=['GET', 'POST'],
@@ -159,6 +178,8 @@ def choose_broadcast_sub_area(service_id, broadcast_message_id, library_slug, ar
         service_id=current_service.id,
     )
     area = BroadcastMessage.libraries.get_areas(area_slug)[0]
+
+    back_link = _get_broadcast_sub_area_back_link(service_id, broadcast_message_id, library_slug)
 
     is_county = any(sub_area.sub_areas for sub_area in area.sub_areas)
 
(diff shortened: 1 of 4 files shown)


app/main/views/broadcast.py
```diff
@@ -147,6 +147,25 @@ def choose_broadcast_area(service_id, broadcast_message_id, library_slug):
     )
 
 
+def _get_broadcast_sub_area_back_link(service_id, broadcast_message_id, library_slug, area):
```
idavidmcdonald, line 150: We don't appear to be using the 'area' param anywhere in the method?

tests/app/main/views/test_broadcast.py
```diff
@@ -520,6 +522,51 @@ def test_choose_broadcast_sub_area_page_for_district_shows_checkboxes_for_wards(
...
+    expected_back_link_extra_kwargs
+):
+    service_one['permissions'] += ['broadcast']
+    page = client_request.get(
+        'main.choose_broadcast_sub_area',
+        service_id=SERVICE_ONE_ID,
+        broadcast_message_id=str(uuid.UUID(int=0)),
+        library_slug='wd20-lad20-ctyua19',
+        area_slug='lad20-E07000105',  # Ashford
+        prev_area_slug=prev_area_slug,  # Kent
```
idavidmcdonald, line 555: Optional: Remove the Kent comment as it's not true in the second param test

tests/app/main/views/test_broadcast.py
```diff
@@ -469,6 +469,8 @@ def test_choose_broadcast_area_page_for_area_with_sub_areas(
             'York',
         ),
     ]
+    # note: doesn't populate prev_area_slug query param for ashford, so the back link will come here rather than to kent
+    assert choices[12] == (partial_url_for(area_slug='lad20-E07000105'), 'Ashford')
```
idavidmcdonald, line 473: Did you add these lines because you wanted some symmetry with the other tests which were testing that if you visit Ashford from the Kent page then it goes back?

I feel like the rest of the urls you are asserting in the lines above in this test are also testing that we aren't adding a prev_area_slug to them so just wanted to check if I was missing something or if this test has just been added for further clarity on behaviour?

idavidmcdonald (changes requested):
Looks good and runs good locally. One small thing to fix please

tests/app/main/views/test_broadcast.py
leohemsted: yeah, thats right, and you're right that this new line doesn't add anything really - i've removed the extra assert but kept the comment.

app/main/views/broadcast.py
```diff
@@ -160,6 +179,8 @@ def choose_broadcast_sub_area(service_id, broadcast_message_id, library_slug, ar
     )
     area = BroadcastMessage.libraries.get_areas(area_slug)[0]
 
+    back_link = _get_broadcast_sub_area_back_link(service_id, broadcast_message_id, library_slug, area)
```
idavidmcdonald, line 182: Need to remove the area arg

tests/app/main/views/test_broadcast.py
```diff
@@ -453,6 +453,7 @@ def test_choose_broadcast_area_page_for_area_with_sub_areas(
         for choice in page.select('.file-list-item')
     ]
     assert len(choices) == 394
+    # note: we don't populate prev_area_slug query param, so the back link will come here rather than to kent
```
idavidmcdonald, line 456: Does 'Kent' make sense here? When the assertion below is talking about Aberdeen

idavidmcdonald (approved):

PR #3630, opened 2020-09-15
Fix copy to clipboard selection

tombye:
🐛 spotted by @klssmith: since I added a hidden prefix to the key element on some variants of the `api_key` component, this got included in what was copied.

In other words, you got

`ID:ca6482bb-2ccf-4008-873d-ecf53db12ed5` 

...in your clipboard, not

`ca6482bb-2ccf-4008-873d-ecf53db12ed5`

This fixes it by moving the start of the selected [range](https://developer.mozilla.org/en-US/docs/Web/API/Range) so it starts after the prefix, if it exists.

## Notes

This also includes a change to the tests that makes some mocks renew before each test instead of just at the start. This is due to some state carrying across between tests.

## How to test this

Try copying an ID on `/service-settings/email-reply-to` (they have prefixes) and one on a template page (they don't).

### DIFF
--- a/app/assets/javascripts/apiKey.js
+++ b/app/assets/javascripts/apiKey.js
@@ -27,11 +27,27 @@
       `
     };
 
+    this.getRangeFromElement = function (keyElement) {
+      const range = document.createRange();
+      let prefixIndex = -1;
+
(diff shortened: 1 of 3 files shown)


klssmith (approved):
Tested this locally and it all works well

PR #3632, opened 2020-09-17
Add a tour screen once a broadcast is approved

quis:
For a training broadcast the user doesn’t get that immediate feedback that something has happened, like they would with a real alert, or even sending themselves a text message.

This commit adds another tour-style page which will interrupt their journey and hopefully reinforce the message we’ve given them earlier in the tour.

We’re adding this because we’ve found in research that users don’t have a good grasp of the consequences and severity of emergency alerts, versus regular text messages.

Before | After 
---|---
image | image
↓ |↓
image | image

### DIFF
--- a/app/templates/views/broadcast/tour/6.html
+++ b/app/templates/views/broadcast/tour/6.html
@@ -1,0 +1,41 @@
+{% from "components/banner.html" import banner_wrapper %}
+
+{% extends "admin_template.html" %}
+
+{% block per_page_title %}
+  You’re still in training mode. Notify has not broadcast your alert.
+{% endblock %}
+
+{% set mainClasses = "govuk-!-padding-top-0 govuk-!-padding-bottom-0" %}
+
+{% block content %}
+
+  <div class="navigation-service">
+    <div class="navigation-service-name govuk-!-font-weight-bold">
+      {{ current_service.name }} <span class="navigation-service-type--training">Training</span>
+    </div>
+  </div>
+
+  <div class="banner-tour banner-tour-no-fixed-height banner-tour-with-service-name">
+    <div class="govuk-grid-row">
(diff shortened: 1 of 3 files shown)


quis:
@karlchillmaid I’ve swapped the order of the first paragraph so it joins into one sentence, and leads with ‘Notify has not broadcast your alert’ which is the response to what the user has just done. I don’t think we need to say ‘still’ because the idea is users will swap back and forth between training and real depending on the situation.

If you’re happy with this now then I’m happy with this too.

image

karlchillmaid:
@quis Yeah, definitely an improvement.

app/templates/views/broadcast/tour/6.html
```diff
@@ -0,0 +1,41 @@
...
+        <h1 class="heading-medium">
+          Notify has not broadcast your alert because you’re still in training mode.
+        </h1>
+        <p class="govuk-body heading-medium">
+          In a real emergency, every mobile phone in the area
+          you chose would make a loud alarm noise.
+        </p>
+        <p class="govuk-body heading-medium">
+          <a class="govuk-link govuk-link--no-visited-state" href='{{ url_for(".service_dashboard", service_id=current_service.id) }}'>
+            Continue to dashboard
```
idavidmcdonald, line 31: Surprised me that you'd take them to the dashboard, I would have assumed that we would have sent them to view the broadcast they sent so they learn what they should expect to see in the real case.

idavidmcdonald (approved):

app/templates/views/broadcast/tour/6.html
quis: I think once we make the dashboard better it will be the right place to take them. Appreciate it’s different from the ‘live‘ journey at the moment.

PR #3636, opened 2020-09-18
Fix the display of sms statuses

klssmith:
The styling of SMS statuses had broken - all statuses except for `sent` had the error styling.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -480,9 +480,6 @@ def format_notification_status_as_field_status(status, notification_type):
             'returned-letter': None,
             'cancelled': 'error',
         },
-        'sms': {
-            'sent': 'sent-international'
-        },
     }.get(
         notification_type,
         {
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Great work Katie!

Two optional improvements but I don't think I'd suggest doing them at the moment. Let's just fix the bug I think
- Change `format_notification_status_as_field_status` so it's not doing this hard to read three `get` with default. I wonder if just having a key for sms with all possible statuses and same for emails would be a better approach
- Add unit test cases for letters (I didn't spot them unless they are tested in a different test)

klssmith:
I did also wonder if changing `format_notification_status_as_field_status` so that it has separate email and sms keys in the dict would be clearer.

It wasn't possible to add letters as an extra parameter to the test that I added (since the status displays in a different way). I will deploy this just to fix the current issue, then take another look at tests for letters.

idavidmcdonald:
Thanks Katie. I wouldn't worry too much about the letter tests unless you are looking for extra work ;)

PR #3638, opened 2020-09-21
Step 1 of renaming cache keys for templates

idavidmcdonald:
We want to change cache keys for templates and broadcasts to include
their service ID. So cache keys should change from
`template-{template_id}-versions` to
`service-{service_id}-template-{template_id}-versions`.

The first step of this which needs to be deployed as a change first is
to delete both keys when updating service templates (even if they key is
not yet set). This means that when we release code in the next PR to
start setting the new key, we won't run into a case where either the old
or the new key can remain set with stale data.


quis (approved):
Stacking the decorators is a neat solution 👏

PR #3641, opened 2020-09-21
Suggest previously-used areas when adding new area

quis:
If you’re adding another area to your broadcast it’s likely to be close to one of the areas you’ve already added.

But we make you start by choosing a library, then you have to find the local authority again from the long list. This is clunky, and it interrupts the task the user is trying to complete.

We thought about redirecting you somewhere deep into the hierarchy, perhaps by sending you to either:
- the parent of the last area you’d chosen
- the common ancestor of all the areas you’d chosen

This approach would however mean you’d need a way to navigate back up the hierarchy if we’d dropped you in the wrong place. And we don’t have a pattern for that at the moment.

So instead this commit adds some ‘shortcuts’ to the chose library page, giving you a choice of all the parents of the areas you’ve currently selected. In most cases this will be one (unitary authority) or two (county and district) choices, but it will scale to adding areas from multiple different authorities.

It does mean an extra click compared to the redirect approach, but this is still fewer, easier clicks compared to now.

# Before

The ‘Choose where to broadcast to’ page always looks like this, no matter where you’ve come from:

image

# After 

If you click ‘Add another area’ on this page:
image

You get some extra choices:
image

# Implementation details 

This meant a couple of under-the-hood changes:
- making `BroadcastArea`s hashable so it’s possible to do `set([BroadcastArea(…), BroadcastArea(…), BroadcastArea(…)])`
- making `BroadcastArea`s aware of which library they live in, so we can link to the correct _Choose area_ page

### DIFF
--- a/app/broadcast_areas/__init__.py
+++ b/app/broadcast_areas/__init__.py
@@ -17,6 +17,12 @@ class SortableMixin:
         # method are sortable
         return self.name < other.name
 
+    def __eq__(self, other):
+        return self.id == other.id
+
+    def __hash__(self):
+        return hash(self.id)
+
 
 class GetItemByIdMixin:
     def get(self, id):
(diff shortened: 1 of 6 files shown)


leohemsted (approved):
looks good, tested locally and works fine. i like the new test, all makes sense

one thing that this has brought to my attention though is we don't prepopulate the area pages based on what's already on the broadcast. that'd be good for a future PR i think

quis:
Yeah, that’d be a nice detail to have.

PR #3643, opened 2020-09-22
move service to broadcast org when broadcasting is enabled

leohemsted:
we want to keep track of all broadcast services across govt easily. As such, when broadcasting is enabled for a service, we've decided we're going to add the service to a special broadcasting organisation.

This organisation is defined in the config file. It's hard coded based on the db migration in the api pr referenced.

This organisation already exists on production(https://www.notifications.service.gov.uk/organisations/38e4bf69-93b0-445d-acee-53ea53fe02df).

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -84,6 +84,8 @@ class Config(object):
     ASSET_DOMAIN = ''
     ASSET_PATH = '/static/'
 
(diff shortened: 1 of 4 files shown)


quis (approved):
I think we should run a migration to seed this organisation in other environments and add any existing broadcast services to it.

PR #3645, opened 2020-09-23
Make only the scrollable table focusable

tombye:
The JS clones the scrollable table so was passing its attributes across to the fixed one (which provides the row headings).

This bug was pushed in:

https://github.com/alphagov/notifications-admin/pull/3637

### DIFF
--- a/app/assets/javascripts/fullscreenTable.js
+++ b/app/assets/javascripts/fullscreenTable.js
@@ -48,6 +48,7 @@
             .clone()
             .addClass('fullscreen-fixed-table')
             .removeClass('fullscreen-scrollable-table')
+            .removeAttr('role aria-labelledby tabindex')
             .attr('aria-hidden', true)
         )
         .append(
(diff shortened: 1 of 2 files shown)


leohemsted (commented):
i think this seems sensible, my concerns are:

* Is this a sustainable approach? If there are more attributes we start using, we'll need to decide whether to move them across. I don't really know enough context about the various attributes to know if that's likely though, and if a list of attrs we explicitly allow to be cloned is better or a list of attrs we explicitly remove from the cloned element
* should we add (or modify an existing) test?

quis:
Tested locally and it fixes the bug. I’ll let you address @leohemsted’s comments before giving my 👍

tombye:
@leohemsted I see what you mean. I think at this stage, putting in some code to check the attributes seems overkill but I've tried to name the ones being removed more obviously and put in a test to check the difference between the two table frames (which I think was sorely needed).

What do you think?

leohemsted (approved):
test looks good - if you're happy with the elements in that list being fairly stable then 👍, i'll defer to your expertise

PR #3646, opened 2020-09-23
Replace uses of $yellow for focus

quis:
There are a few places where we missed updating to the new focus styles because they were using the `$yellow` SASS variable and not the `$focus-colour` variable.

This commit updates them to the new colour, and where needed adds the black lower border to match.


tombye (approved):
This is nice work, thanks for finding all the things and fixing them 👍

PR #3648, opened 2020-09-23
Fix styling of ‘Training’ tag on tour step 6

quis:
It was missing the base class so didn’t get all the styling (like the spacing and uppercase text) on this specific page.

# Before 

image

# After 

image

### DIFF
--- a/app/templates/views/broadcast/tour/6.html
+++ b/app/templates/views/broadcast/tour/6.html
@@ -12,7 +12,7 @@
 
   <div class="navigation-service">
     <div class="navigation-service-name govuk-!-font-weight-bold">
-      {{ current_service.name }} <span class="navigation-service-type--training">Training</span>
+      {{ current_service.name }} <span class="navigation-service-type navigation-service-type--training">Training</span>
     </div>
   </div>
 


idavidmcdonald (approved):

app/templates/views/broadcast/tour/6.html
```diff
@@ -12,7 +12,7 @@
 
   <div class="navigation-service">
     <div class="navigation-service-name govuk-!-font-weight-bold">
-      {{ current_service.name }} <span class="navigation-service-type--training">Training</span>
+      {{ current_service.name }} <span class="navigation-service-type navigation-service-type--training">Training</span>
```
tombye, line 15: I don't actually see what styles adding `navigation-service-type` brings in.

I'm looking at https://github.com/alphagov/notifications-admin/blob/master/app/assets/stylesheets/components/navigation.scss#L21 so please correct me if I'm missing something.

quis: `navigation-service-type` brings in: https://github.com/alphagov/notifications-admin/blob/cc456fa718a88c9a88d4f36ee144530e3b14332f/app/assets/stylesheets/components/navigation.scss#L23-L29

`navigation-service-type--training` brings in: https://github.com/alphagov/notifications-admin/blob/cc456fa718a88c9a88d4f36ee144530e3b14332f/app/assets/stylesheets/components/navigation.scss#L32-L34

PR #3650, opened 2020-09-24
Broadcast org always on

leohemsted:
depends on:

- [x] alphagov/notifications-api#2977
- [x] https://github.com/alphagov/notifications-admin/pull/3643


quis (approved):
Might wanna rebase just to ensure it’s only changes from the second commit that are getting applied. Functionally good though.

PR #3651, opened 2020-09-24
Add instructions for converting Shapefiles

quis:


idavidmcdonald (approved):
I'll take your word for it

PR #3660, opened 2020-09-28
Fix incorrect processing started time on jobs

quis:
We were doing the UTC to BST conversion twice, meaning the job was shown as being started 2 hours later than UTC, and 1 hour later than actual BST.

Bug in the wild, where the notifications are showing the correct time but the job is showing the wrong time:
Screenshot 2020-09-28 at 14 54 15

### DIFF
--- a/app/main/views/two_factor.py
+++ b/app/main/views/two_factor.py
@@ -16,7 +16,7 @@ from app import user_api_client
 from app.main import main
 from app.main.forms import TwoFactorForm
 from app.models.user import User
-from app.utils import is_less_than_90_days_ago, redirect_to_sign_in
+from app.utils import is_less_than_days_ago, redirect_to_sign_in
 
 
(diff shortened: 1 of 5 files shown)


idavidmcdonald:
So this looks good and does fix it.

I guess my question is, ideally we are consistent with the API either sending things in UTC or timezone aware (BST at the mo). I see other examples of the API sending things in UTC and the admin app then doing the time conversion. Should we be doing this in this case? If so, why should we be doing the conversion in the API instead?

quis:
We should be passing around everything in UTC up until the point we:
- display the date on a page
- need to compare it to a local timezone (e.g. was this letter created after 5pm)

This code was violating that principle a bit, so I’ve refactored to avoid the local timezone conversion at all in 7d604ca

idavidmcdonald (approved):
Good work, thanks for working with me on the gnarly timezone stuff too!

PR #3662, opened 2020-09-30
Give email preview page a sensible page title

quis:
It’s derived from the subject line of the email now as of https://github.com/alphagov/notifications-utils/pull/793


tombye (approved):
Looks good to me 👍

PR #3663, opened 2020-09-30
Move intro to Notify tour into separate views to reduce code complexity

idavidmcdonald:
## Problem this solves
The code for the intro tour to Notify and sending one off notifications is very complex. It's grown organically over the past few years and there are several bugs in it, for example back links often don't work and the tour mobile phone placeholder is not filled in. We plan to do a refactor of the code to remove the complexity. This is the first step in doing that by moving the intro tour into its own view. Apart from fixing some bugs as we do this and introducing a new URL schema for the new tour, the behaviour of the tour should remain pretty much exactly as it was before.

There may be some code duplication between this new view and the existing send one of notification code but the duplication is less worse than the code complexity. It will then also enable us to more easily refactor the send one off code (details in section below).

## How to review
- suggest to review commit by commit as this will mean smaller chunks of code to review at a time (although the whole PR could be reviewed in one if you wanted).
- spin this up locally
- go through the flow completely from creating a new account to receiving the tour SMS to your phone
- test all back links
- test validation and error messages are shown
- think about what test coverage is missing

## Bigger picture plan

So before this we had the following urls visited

#### When sending a notification using ‘send to my own number’, also used for the tour
Set-sender
One-off/step-0 (to choose phone number)
Test/step-0 (to set placeholder 1)
Test/step-1 (to set placeholder 2)
notification/Check

#### When sending a notification not to your own number
Set-sender
One-off/step-0 (choose phone and then choose not yourself)
One-off/step-1 (placeholder 1)
One-off/step-2 (placeholder 2)
Notification/check

You'll see the slightly nasty bit where filling in the first placeholder for the two different flows has you visiting `one-off/step-0` or `test/step-1` urls which is confusing with the difference indices, adds to the complexity and is a source of some of the bugs for the back links.

### Proposed steps

1. Introduce new tour route as part of this PR
2. Wait a day before removing the old tour start page just in case anyone was on the tour when we did the deploy of step 1 so we don't break their flow
3. Move flow for when sending a notification to your own number to use the
Set-sender
One-off/step-0 (choose phone and then choose not yourself)
One-off/step-1 (placeholder 1)
One-off/step-2 (placeholder 2)
Notification/check
pattern. This will be done by if you click on 'Send message to yourself' it will do a GET request to a page that will set your session recipient as your own phone number before redirecting you to `One-off/step-1`. 
4. Wait a day before removing the `Test/step-<>` routes

We will then end up with a set of separate routes for the tour and then a single URL scheme for sending one off notifications:
Set-sender
One-off/step-0 (choose phone and then choose not yourself)
One-off/send-to-myself (optional GET and immediate redirect to step-1 having set in the session your own phone number)
One-off/step-1 (placeholder 1)
One-off/step-2 (placeholder 2)
Notification/check


app/main/views/tour.py
```diff
@@ -0,0 +1,141 @@
...
+            ))
+        return redirect(url_for(
+            '.tour_step', service_id=current_service.id, template_id=template_id, step_index=1
+        ))
+
+    form = get_placeholder_form_instance(
+        current_placeholder,
+        dict_to_populate_from=get_normalised_placeholders_from_session(),
+        template_type=template.template_type,
+        # we allow international numbers for tours in case the user is using an international number
```
idavidmcdonald, line 75: So I think we want to do this but it does open us up to the fact that if someone wanted to send a text to an international number but didn't have permission to then they could use the tour route to bypass this protection. Worth the risk?

tests/app/main/views/test_tour.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+def test_should_404_if_no_mobile_number_for_tour_start(
+    client_request,
+    mock_get_service_template_with_multiple_placeholders,
+    service_one,
+    fake_uuid,
+):
+    # todo fix this test
+    current_user.mobile_number = None
```
idavidmcdonald, line 93: So currently `current_user.mobile_number = None` doesn't do anything. I'm not quite sure how to mock the current_user so I can set the mobile number to None. 

Note, that this edge case isn't covered in the current tour. Is it something we want to fix or is it OK to just remove this test and leave it? 

When someone does the tour they will have a valid mobile number. It's only if they got their service switched to email auth and then removed their phone number and then revisited the tour that it might cause an error.

app/main/views/tour.py
```diff
@@ -0,0 +1,141 @@
...
+    template = get_template(
+        db_template,
+        current_service,
+        show_recipient=True,
+    )
+
+    placeholders = fields_to_fill_in(template, prefill_current_user=True)
+    try:
+        # user urls are 1 indexed, so start at step-1
+        current_placeholder = placeholders[step_index - 1]
```
idavidmcdonald, line 61: If someone visits `step-0` this will evaluate to `placeholders[-1]` and show them the last step. We probably don't want this to happen. Any suggestions for how to code this up?

app/main/views/tour.py
```diff
@@ -0,0 +1,141 @@
...
+    get_placeholder_form_instance,
+    get_recipient_and_placeholders_from_session,
+)
+from app.utils import get_template, user_has_permissions
+
+
+@main.route("/services/<uuid:service_id>/tour/<uuid:template_id>")
+@user_has_permissions('send_messages')
+def begin_tour(service_id, template_id):
+    db_template = current_service.get_template_with_user_permission_or_403(template_id, current_user)
```
idavidmcdonald, line 18: We don't have any test coverage for if they are trying to get a template that doesn't belong to their service or that they don't have folder permissions for. Is this OK? I couldn't see any equivalent test coverage on the previous code either to copy

app/main/views/tour.py
quis: Every service starts with the `international_sms` permission switched on. So in retrospect I don’t think we do need to treat this as a special case.

app/main/views/tour.py
quis: I would `404` it because it will never be a valid URL
```suggestion
        if step_index < 1:
            abort(404)
        current_placeholder = placeholders[step_index - 1]
```

idavidmcdonald: So the suggestion is to 404 for `step-0` but on something with 3 placeholders, `step-4` would take them back to `step-1`.  Feels slightly inconsistent but I guess it doesn't really matter that much?

app/main/views/tour.py
idavidmcdonald: Thanks. I've changed this line in a commit and rebased on top

app/main/views/tour.py
idavidmcdonald: Part of me feels like we are trying to be too smart by if they try and visit a url for a step doesn't exist, send them to a more appropriate page (like the first step or the page to check the notification). I guess it could be useful but is the code worth it compared to we could just be 404ing always those types of requests.

tests/app/main/views/test_tour.py
quis: You’d need to make a new fixture, like this:
https://github.com/alphagov/notifications-admin/blob/3182fc7668dca795586d5acd5634ab31dc71eb90/tests/conftest.py#L1438-L1447

app/main/views/tour.py
quis: If you have 3 placeholders then `…/step-3` will always redirect to `…/step4`. Then this code notices that `4` is out of bounds, and tries to send them to the `…/check` page. If the some of the placeholders are missing then it catches that, and sends them back to the start, rather than the check page.

I think it’s sensible behaviour to keep for the one-off journey, because someone could edit a template and change the number of placeholders while it’s being sent. So I guess my argument would be to keep it working the same way here for consistency.

app/main/views/tour.py
quis: Would be good to have test coverage for this, along with a check for whether the template is an `sms` template (if we have that already).

app/main/views/tour.py
idavidmcdonald: It appears that that isn't correct about the redirect to step 4. When you do a POST on step-3 it will validate the form, set your new value and then as you have all the needed placeholders in the session, it will redirect you to the check notification page. `step-4` is never accessed.

I'll keep the behaviour anyway for consistency though.

tests/app/main/views/test_tour.py
idavidmcdonald: Some slight tweaking to this needed but this put us on the right tracks. Have rebased this change into the first commit

app/main/views/tour.py
idavidmcdonald: Have put in check for `== 0`. Note, flask already rejects negative numbers so `== 0` was fine rather than `<= 0`

app/main/views/tour.py
idavidmcdonald: I've got test coverage already for only allowing sms templates.

The only thing I can think of doing for testing that they can't get a template from the API that they don't have permissions to do see is just to check if `get_template_with_user_permission_or_403` is called as that is actually the bit responsible for calling the api, checking the permissions and aborting if it doesn't have them. Doesn't feel like the most valuable of tests though, we are essentially just writing some code and testing we wrote that code rather than giving it an input and testing the result.

quis (approved):
Checked out locally, looked at each commits, works nice 👍

PR #3666, opened 2020-10-05
Remove old tour and help arguments from sending flow

idavidmcdonald:
## What

Follows on from https://github.com/alphagov/notifications-admin/pull/3663

We no longer need the `start_tour` page as this has been replaced with
the new `begin_tour` page.

We also no longer need to handle the `help` argument in the
`send_test_step` or `send_one_off_step` as these no longer are
responsible for the tour and don't need to show the help text.

Worth pointing out, the new tour joins into the send one off flow. When
doing a GET `check_tour_notification`, and submitting the form shown on
this page you are POSTed to `send_notification` with `help=3`. Also for
general sending of one off notifications, the POST to
`send_notification` is done with `help=0` which is a bit of a hack to
make sure that we don't show a back link on the `view_notification` page
for when someone gets there having just sent a one off notification.
This use of `help=0` may be a candidate for a refactor in the future as
it feels like a bit of a hacky way of doing things and is therefore not
as clear to developers what is going on.

Also removes the help argument from the csv routes used here. There is
no reason that we need to ever show help for CSVs and this is leftover
code from when we used to do the tour that way.

## How to review
Spin it up locally and test the following scenarios
- doing the full tour works as expected
- sending a one off notification to a TV number
- sending a one off notification to your own mobile number (by entering it in the text box manually)
- sending a one off notification to your own mobile number (by using the link to send it to yourself)


leohemsted (approved):
looks nice, tested locally, seems fine.

we could possibly use `request.referrer` to ascertain what page we've come from when loading the post page. Not convinced that's better than a query parameter, feels brittle to me though. But I agree we could call the query param something more descriptive, and get rid of all the other help stuff.

Also we could see if we can get rid of `help=3` from the tour post

Also also, I don't think we need `send_test` vs `send_one_off` really. There's a lot of leg work going on there and the only thing that is different is the `prefill_current_user` flag - I definitely think we could get rid of all the other logic, and just have that prefill_current_user flag based on some state of the form with whether you pressed `continue` or `use my phone number`. 🤔 

But those are all thoughts for another day I think, lets get this in without it ballooning

PR #3668, opened 2020-10-05
When sending to yourself use 'one-off' rather than 'test' routes

idavidmcdonald:
## What

Follows on from https://github.com/alphagov/notifications-admin/pull/3663 and https://github.com/alphagov/notifications-admin/pull/3666

There is no real reason to have to support both 'one-off' steps and also
'test' steps when sending a one off notification. It's a lot of complex
code, just to now set the one of the placeholders in the session.

We make our code much simpler but no longer using the 'test' routes but
instead adding a new endpoint to set the notification recipient when
sending to yourself before continuing on with the rest of the 'one-off'
flow.

After this is deployed for a day then we can completely remove the
'test' routes and this will help remove a lot of code complexity.

## How to review
spin up locally and test the following
- send a one off notification to a number that isn't yours (try 07700900000)
- send a one off email to yourself by clicking the link to do so
- send a one off text message to yourself by clicking the link to do so
- send a one off email to yourself by entering your email address in the text box yourself
- send a one off text message to yourself by entering your number in the text box yourself


app/main/views/send.py
```diff
@@ -947,16 +939,40 @@ def get_skip_link(step_index, template):
     if (
         request.endpoint == 'main.send_one_off_step'
         and step_index == 0
-        and template.template_type != 'letter'
+        and template.template_type in ("sms", "email")
         and not (template.template_type == 'sms' and current_user.mobile_number is None)
         and current_user.has_permissions('manage_templates', 'manage_service')
```
idavidmcdonald, line 944: @quis any idea why you have to be able to manage a template or manage a service to be able to send a notification to yourself using the helpful link? Is it because we don't want to let those with just send permissions to send one to themselves or is it just they can if they want to but we don't think they have a real use case to do and therefore don't need the short cut?

quis: We think it’s only people who are writing and testing content who need to send messages to themselves, to check how they look on a device.

And the `manage_service` permission is for the same reason, but with changing branding etc, rather than the content of the message.

So the idea is to make the range of choices smaller, and therefore simpler for those who are only sending messages to their users.

idavidmcdonald: Perfect, thank you :)

app/main/views/send.py
```diff
@@ -947,16 +939,40 @@ def get_skip_link(step_index, template):
...
+    if db_template['template_type'] not in ("sms", "email"):
+        abort(404)
+
+    # We aren't concerned with creating the exact template (for example adding recipient and sender names)
+    # we just want to create enough to use `fields_to_fill_in`
+    template = get_template(
+        db_template,
+        current_service,
+    )
+    fields_to_fill_in(template, prefill_current_user=True)
```
quis, line 966: It’s not entirely obvious here that calling `fields_to_fill_in` is the thing that populates the session. Does any other part of the codebase need to populate the session with the current user? Could the code to populate the session go directly in this view method instead?

idavidmcdonald: Good suggestion. I'd be keen to merge this as is and then in the next PR which does the removal of the test routes, to also then do this refactor and move the prefilling just into this view. 

The reason for doing it in the next PR is that until we remove the test routes, we do have other places that also populate the session with the current user.

idavidmcdonald: here it is in a later commit https://github.com/alphagov/notifications-admin/pull/3674/commits/8b4bd734b5071edcb36879ac7d14e816f2f9f610

idavidmcdonald: Actually that commit I just referenced breaks the tour routes. The tour is also using the prefilling. Would require some slightly larger changes will I'll likely do in a separate PR to keep them small. But it's still possible.

quis (approved):

PR #3669, opened 2020-10-05
Restyle the list of areas and emergency alert message preview to be less busy  

quis:
This PR lays some groundwork for making some design changes to the dashboard and individual alert views which currently aren’t possible because the individual components are too busy or visually inconsistent. The individual changes are detailed in the commits.

Before | After 
---|---
image | image
image | image
image | image


quis:
Update: loosened up the spacing between the items a tiny bit

image

quis:
The longer-term aim of this work is to make the list of areas something that can be reused on the dashboard, like this:

image

idavidmcdonald (approved):

PR #3670, opened 2020-10-05
Take user to page they are meant to visit in various sign-in flow scenarios

CrystalPea:
This PR makes sure user is redirected to the page they initially were meant to visit after they sign in. It makes sure redirects are passed through in various scenarios, like code or email not being received, password being reset etc.

This work started as making sure that users who reset their password see the broadcast tour, but the scope was widened to include all redirects and all or at least most sign-in scenarios.

Pivotal ticket: https://www.pivotaltracker.com/story/show/174488590
Goes together with: https://github.com/alphagov/notifications-api/pull/2991

### DIFF
--- a/tests/app/main/views/test_forgot_password.py
+++ b/tests/app/main/views/test_forgot_password.py
@@ -4,6 +4,7 @@ from notifications_python_client.errors import HTTPError
 
 import app
 from tests import user_json
+from tests.conftest import SERVICE_ONE_ID
 
 
 def test_should_render_forgot_password(client):
@@ -42,11 +43,11 @@ def test_forgot_password_sends_next_link_with_reset_password_email_request(
     sample_user = user_json(email_address='test@user.gov.uk')
     mocker.patch('app.user_api_client.send_reset_password_url', return_value=None)
     response = client.post(
-        url_for('.forgot_password') + "?next=blob",
+        url_for('.forgot_password') + f"?next=/services/{SERVICE_ONE_ID}/templates",
         data={'email_address': sample_user['email_address']})
     assert response.status_code == 200
     app.user_api_client.send_reset_password_url.assert_called_once_with(
-        sample_user['email_address'], next_string="blob"
+        sample_user['email_address'], next_string=f'/services/{SERVICE_ONE_ID}/templates'
     )
 
 
--- a/tests/app/main/views/test_forgot_password.py
+++ b/tests/app/main/views/test_forgot_password.py
@@ -31,7 +31,23 @@ def test_should_redirect_to_password_reset_sent_for_valid_email(
     assert response.status_code == 200
     assert 'Click the link in the email to reset your password.' \
            in response.get_data(as_text=True)
-    app.user_api_client.send_reset_password_url.assert_called_once_with(sample_user['email_address'])
+    app.user_api_client.send_reset_password_url.assert_called_once_with(sample_user['email_address'], next_string=None)
+
+
+def test_forgot_password_sends_next_link_with_reset_password_email_request(
+    client,
+    fake_uuid,
+    mocker,
+):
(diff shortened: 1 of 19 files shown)


quis:
This works for users with `email_auth`.

I think you also need to make sure that the `next` parameter is passed through to the two factor page if they’re using `sms_auth`:
https://github.com/alphagov/notifications-admin/blob/fad63117d6fa728db7b1c60bf71b3c03728961f2/app/main/views/new_password.py#L49

Otherwise they will reset their password, type in the code, and lose where they were supposed to be going.

CrystalPea:
@quis I made this PR much more versatile now, check it out :)

tests/app/main/views/test_forgot_password.py
```diff
@@ -31,7 +31,23 @@ def test_should_redirect_to_password_reset_sent_for_valid_email(
...
+):
+    sample_user = user_json(email_address='test@user.gov.uk')
+    mocker.patch('app.user_api_client.send_reset_password_url', return_value=None)
+    response = client.post(
+        url_for('.forgot_password') + "?next=blob",
+        data={'email_address': sample_user['email_address']})
+    assert response.status_code == 200
+    app.user_api_client.send_reset_password_url.assert_called_once_with(
+        sample_user['email_address'], next_string="blob"
+    )
```
quis, line 50: Can we test this with a proper URL instead of `blob`? I’d like to be confident that if that are unusual characters in the URL that they get encoded/decoded properly, for example
```
https://example.com/foo?a=1&b=2
```

tests/app/main/views/test_new_password.py
```diff
@@ -36,21 +37,26 @@ def test_should_return_404_when_email_address_does_not_exist(
     assert response.status_code == 404
 
 
+@pytest.mark.parametrize('redirect_url', [
+    None,
+    'blob',
```
quis, line 42: Same comment here, I’d like to be sure that when we put a real URL here that it doesn’t double-encode special characters or anything like that.

tests/app/main/views/test_forgot_password.py
CrystalPea: Hmm good suggestion that we should test whether standard flask component `url_for` works properly, but since it's the same component used everywhere, I don't feel we have to test it many times.

We are already testing it this way in may places, for example:
`test_valid_two_factor_email_link_shows_interstitial` or `test_should_send_email_and_redirect_to_info_page_if_user_needs_to_revalidate_email`

If you want we could have a dedicated test testing various kinds of urls this way, but I feel this is something that Flask should really take care of. What do you think?

tests/app/main/views/test_new_password.py
CrystalPea: same reply as above.

tests/app/main/views/test_forgot_password.py
CrystalPea: ok, I will change it to the one you suggested then.

CrystalPea: or maybe actually better something like "/services/{some_id}/templates", since this is a much more likely URL?
(the one you posted is actually two args, and would not be all passed in as `next`, building url for it would be `url_for('main.some_endpoint', a='1', b='2')`, which I feel is different)

quis: Yeah I think that will be OK because `/` is a [reserved character](https://tools.ietf.org/html/rfc3986#section-2.2) so it needs to be encoded when it’s passed into `next` (it becomes `%2F`). So as long as `send_reset_password_url` is being called with `/` and not `%2F` we can be confident that the encoding/unencoding dance is working properly.

quis (approved):

PR #3671, opened 2020-10-05
Clear new template folder cache

leohemsted:
couple of places we missed: when a platform admin clicks the delete template cache button, and when a user moves a folder containing templates


idavidmcdonald (approved):
Thanks Leo! Good spot.

idavidmcdonald:
So I've merged this in Leo but just thought I'd check what the bug was that it was impacting?

So the user reporting moving some folders and templates around and then getting a 404 when trying to view a template.

So I'm guessing that it was something like the template was cached in redis, including the folder it belonged to. Then when it got moved but the cache wasn't invalidated, it was still saying it belonged in an old folder. Then when the user tried to view the template they got a 404 because of part of viewing the template, we try and get the folder and the folder no longer existed so we 404d?

PR #3674, opened 2020-10-07
Remove `test/step-` routes

idavidmcdonald:
We don't need these anymore as all users will use the `one-off/step`
routes.

This has mostly involved tidying up the tests which are still a little
disorganised and not as good as I'd like but it's a step in the right
direction.

More refactoring is still possible to the routes, it may come in a later
PR if I have time.


tests/app/main/views/test_send.py
```diff
@@ -1227,129 +1227,38 @@ def test_404_for_previewing_a_row_out_of_range(
     )
 
 
-@pytest.mark.parametrize('user', (
-    create_active_user_with_permissions(),
-    create_active_caseworking_user(),
-))
-def test_send_test_doesnt_show_file_contents(
```
idavidmcdonald, line 1234: This test doesn't seem relevant anymore so I have removed it. When visiting the `send_test` routes, there was no calling of `s3download`. I think this test is leftover from when the one off flow used to do a gnarly process of creating a CSV with a single row in to be used

tests/app/main/views/test_send.py
```diff
@@ -1227,129 +1227,38 @@ def test_404_for_previewing_a_row_out_of_range(
...
+        _expected_redirect=url_for(
+            'main.send_one_off',
+            service_id=SERVICE_ONE_ID,
+            template_id=fake_uuid,
+            _external=True,
+        )
     )
 
-    with client_request.session_transaction() as session:
-        assert session['recipient'] == expected_recipient
```
idavidmcdonald, line 1351: This test was a bit gnarly. It was visiting test/step-0, being redirected back to `send_test` and then being redirected back to `test/step-0` where it would then check the the expected recipient had been put in the session. This is way to much to test in a single test so I've made it much simpler to do what the test name says. If you haven't got a session set up, we redirect you back to the route that is responsible for doing that (nothing more, nothing less)

tests/app/main/views/test_send.py
```diff
@@ -1381,106 +1290,60 @@ def test_send_one_off_does_not_send_without_the_correct_permissions(
...
-    ),
-    (
-        'sms',
-        partial(url_for, 'main.send_one_off'),
-        'Send ‘Two week reminder’',
-    ),
-    (
-        'email',
-        partial(url_for, 'main.send_test'),
-        'Personalise this message',
```
idavidmcdonald, line 1398: It looks like `Personalise this message` title has been lost somewhere in the refactoring (it's likely no longer relevant). This test brings that up to date

tests/app/main/views/test_send.py
```diff
@@ -1639,15 +1502,19 @@ def test_skip_link_will_not_show_on_sms_one_off_if_service_has_no_mobile_number(
...
     page = client_request.get(
         'main.send_one_off_step',
         service_id=SERVICE_ONE_ID,
         template_id=fake_uuid,
         step_index=0,
-        _follow_redirects=True,
     )
-    skip_links = page.select('a.top-gutter-4-3')
-    assert not skip_links
+    skip_link = page.findAll('a', text='Use my phone number')
```
idavidmcdonald, line 1516: Make this a tighter assertion based on the content. This test was passing incorrectly as we no longer use the `top-gutter-4-3` style in our HTML so it could never know if the link was there or not there.

CrystalPea (approved):
noice 👍

PR #3675, opened 2020-10-07
Delete template cache if adding a new reply to email address

idavidmcdonald:
As if this is chosen as default, then the template data will be stale as
they do not have the latest reply to address.

We are already doing this for create/update/delete for sms senders and
contact lists for letters, it was just the create we were missing for
email reply to addresses.

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -428,13 +428,15 @@ class ServiceAPIClient(NotifyAdminAPIClient):
 
     @cache.delete('service-{service_id}')
     def add_reply_to_email_address(self, service_id, email_address, is_default=False):
-        return self.post(
+        ret = self.post(
             "/service/{}/email-reply-to".format(service_id),
             data={
                 "email_address": email_address,
                 "is_default": is_default
             }
         )
+        self._delete_template_cache_for_service(service_id)
+        return ret
 
     @cache.delete('service-{service_id}')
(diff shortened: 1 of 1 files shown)


app/notify_client/service_api_client.py
```diff
@@ -428,13 +428,15 @@ def verify_reply_to_email_address(self, service_id, email_address):
 
     @cache.delete('service-{service_id}')
     def add_reply_to_email_address(self, service_id, email_address, is_default=False):
-        return self.post(
+        ret = self.post(
```
CrystalPea, line 431: I'd call that variable `response` for readability ☺️

CrystalPea (approved):
One small comment, otherwise this is good to go. Hopefully it will solve at least part of our problem :)

app/notify_client/service_api_client.py
idavidmcdonald: I'd agree if writing the code from scratch my self however it looks like the 8 other places we are doing the post and then clearing the cache, all of them are calling it `ret` so I thought I'd stick with consistency for the quickest thing at the moment!

CrystalPea: I agree that generally consistency is good, but I don't agree that making something consistently bad is good. By having a better version of variable name you are at least showing that change is underway. 

I think best thing here might have been `Select all 'ret '`, `Replace all 'ret ' to 'response '`, it would have taken 1 minute probably.

Still, it's just a variable name, it's inconsequential I guess, it's nice that the bug will be fixed now ☺️

PR #3676, opened 2020-10-08
Add content property to broadcast message model

quis:
The API is returning this new property as of https://github.com/alphagov/notifications-api/pull/2990

So that we don’t still have any objects cached without this property, this pull request also adds broadcast messages to the list of things that platform admin users can purge from the cache.


idavidmcdonald (approved):
Looks good.

One small fix to the tests but as soon as that is correct then this is good.

As discussed on the slack call, potential benefit of adding some more comments to the clear_cache tests

tests/__init__.py
```diff
@@ -659,6 +660,7 @@ def broadcast_message_json(
         'template_id': template_id,
         'template_version': 123,
         'template_name': 'Example template',
+        'content': 'This is a test',
```
quis, line 663: Good spot
```suggestion
        'content': content or 'This is a test',
```

PR #3677, opened 2020-10-09
Remove issue related to focus shifting onload

tombye:
The Digital Accessibility Centre (DAC) reported
that it was confusing that focus was moved to the
templates and folders search textbox when the page
loaded on the templates page.

We noted that this issue wasn't reported on the
page where you enter your SMS 2-factor code.

From this, we determined that pages with a single
task could benefit from focus moving to the part
of the page that lets you do that task.

Because of this we:
- found all the pages we shift focus on when they
  load
- kept it for pages where the focus shift is
  important for what users do on the page
- removed it for all other pages (including
  templates)

The first iteration of this work was pushed in https://github.com/alphagov/notifications-admin/pull/3622

That branch also changed the label of the focused
element as a hack to get the page title read out.

This normally happens on load to indicate a
navigation has happened but didn't in Voiceover
when we focused an element on load.

After testing with JAWS and NVDA we found only
Voiceover does this. As Voiceover has the smallest
market share, and navigation can be inferred by
other means, we removed it in https://github.com/alphagov/notifications-admin/pull/3673

That was deployed today.

### DIFF
--- a/app/templates/views/accessibility_statement.html
+++ b/app/templates/views/accessibility_statement.html
@@ -41,7 +41,6 @@
     <li>low vision users cannot access the table we use to show CSV data when viewed at high zoom</li>
     <li>screen reader users may find the process of moving templates and folders confusing</li>
     <li>one page links to a PDF document that is not fully accessible</li>
-    <li>screen readers users may be disorientated when pages load with the focus on a specific area</li>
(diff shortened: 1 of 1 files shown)


quis (approved):
Nice summary of the history in the commit message 👍

PR #3680, opened 2020-10-13
Show preview of content for broadcast messages and templates

quis:
Before | After
---|---
image | image
image | image

***

Depends on:
- [x] alphagov/notifications-api#2996

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -409,7 +409,7 @@ def valid_phone_number(phone_number):
 def format_notification_type(notification_type):
     return {
         'email': 'Email',
-        'sms': 'SMS',
+        'sms': 'Text message',
         'letter': 'Letter'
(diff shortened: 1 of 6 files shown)


app/__init__.py
```diff
@@ -409,7 +409,7 @@ def valid_phone_number(phone_number):
 def format_notification_type(notification_type):
     return {
         'email': 'Email',
-        'sms': 'SMS',
+        'sms': 'Text message',
```
tlwr, line 412: This will also impact the data retention service settings page, was this intentional?

tlwr (approved):
Approved pending comment

app/__init__.py
quis: Yeah. UI should always say ‘text message‘ not ‘SMS’.

PR #3681, opened 2020-10-14
Re-order hierarchy of information when viewing a single broadcast

quis:
Before | After
---|---
image | image


tlwr (approved):
:shipit: looks ace

PR #3682, opened 2020-10-15
Make sure the correct navigation item is highlighted when viewing broadcasts

quis:
Once a broadcast has been submitted for approval it either lives on the ‘Current alerts’ or ‘Previous alerts’ page, depending on where it is in its lifecycle.

Therefore when clicking into a broadcast from one of those pages the selected navigation item should match the page you’ve come from, and the status of the broadcast.

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -292,17 +292,39 @@ def view_broadcast(service_id, broadcast_message_id):
     )
     if broadcast_message.status == 'draft':
         abort(404)
+
+    if (
+        broadcast_message.status in {'completed', 'cancelled', 'rejected'}
+        and request.endpoint != 'main.view_previous_broadcast'
+    ):
+        return redirect(url_for(
+            '.view_previous_broadcast',
+            service_id=current_service.id,
+            broadcast_message_id=broadcast_message.id,
+        ))
+
+    if (
+        broadcast_message.status in {'broadcasting', 'pending-approval'}
+        and request.endpoint != 'main.view_current_broadcast'
+    ):
+        return redirect(url_for(
+            '.view_current_broadcast',
+            service_id=current_service.id,
+            broadcast_message_id=broadcast_message.id,
+        ))
+
+    back_link_endpoint = {
+        'main.view_current_broadcast': '.broadcast_dashboard',
+        'main.view_previous_broadcast': '.broadcast_dashboard_previous',
+    }[request.endpoint]
+
     return render_template(
         'views/broadcast/view-message.html',
         broadcast_message=broadcast_message,
-        back_link={
-            'main.view_current_broadcast': url_for(
-                '.broadcast_dashboard', service_id=current_service.id
-            ),
-            'main.view_previous_broadcast': url_for(
-                '.broadcast_dashboard_previous', service_id=current_service.id
-            ),
-        }[request.endpoint],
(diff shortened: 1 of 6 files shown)


tests/app/main/views/test_broadcast.py
```diff
@@ -1109,6 +1114,74 @@ def test_view_broadcast_message_page(
...
+            status=status,
+        ),
+    )
+    service_one['permissions'] += ['broadcast']
+
+    page = client_request.get(
+        endpoint,
+        service_id=SERVICE_ONE_ID,
+        broadcast_message_id=fake_uuid,
+        _follow_redirects=True
```
idavidmcdonald, line 1175: Optional: I might suggest removing the _follow_redirects=True as I don't think we need it and only increases the chance of us writing a mistake in our test itself

app/main/views/broadcast.py
```diff
@@ -272,23 +275,60 @@ def preview_broadcast_message(service_id, broadcast_message_id):
...
+        return redirect(url_for(
+            '.view_current_broadcast',
+            service_id=current_service.id,
+            broadcast_message_id=broadcast_message.id,
+        ))
+
+    back_link_endpoint = {
+        'main.view_current_broadcast': '.broadcast_dashboard',
+        'main.view_previous_broadcast': '.broadcast_dashboard_previous',
+    }[request.endpoint]
```
idavidmcdonald, line 319: Is this bunch of logic untested? If so would be nice to add some test coverage please

app/main/views/broadcast.py
```diff
@@ -65,11 +66,13 @@ def get_broadcast_dashboard_partials(service_id):
             'views/broadcast/partials/dashboard-table.html',
             broadcasts=broadcast_messages.with_status('pending-approval'),
             empty_message='You do not have any alerts waiting for approval',
+            single_endpoint='.view_current_broadcast',
```
idavidmcdonald, line 69: Optional: 'single_endpoint' could be a more precise variable name. Maybe `view_broadcast_endpoint`?

idavidmcdonald (commented):
I think there could be some more test coverage please but apart from that it looks good

app/main/views/broadcast.py
quis: It’s tested by [the use of `_follow_redirects`](https://github.com/alphagov/notifications-admin/pull/3682/files/d1283ba44a5b177cbe6deef7b09794b644aa2457#r510219449) in combination with [parametrizing the test with both endpoints](https://github.com/alphagov/notifications-admin/pull/3682/files/d1283ba44a5b177cbe6deef7b09794b644aa2457#diff-c1dd0f43b852bb0c4a074bdebe772a845e3c25ebffb3f69d72be0831fb98e8f9R1117-R1120).

idavidmcdonald: Oh nice, I see what you've done there.

I think there is still missing coverage for the back link though?

app/main/views/broadcast.py
quis: Changed in f3cf080a5

app/main/views/broadcast.py
quis: Test for back link added in 725df2e7f

idavidmcdonald (approved):
Nice, cheers for the adaptions

PR #3684, opened 2020-10-16
Sort broadcasts by start time

quis:
For emails and text messages we sort by the time the user (or API) sent them.

This makes sense for broadcasts too, since most users will receive the alert within seconds of it being broadcast.

For alerts that haven’t started yet we can sort by `updated_at`, which is when the user preparing the broadcast submitted it for approval.


quis:
Closing in favour of incorporating into https://github.com/alphagov/notifications-admin/pull/3702

PR #3685, opened 2020-10-16
Remove the styleguide

quis:
The idea was that this would be a place to document all the design patterns used in Notify. However it hasn’t been kept up to date, and, looking at the [`git blame`](https://github.com/alphagov/notifications-admin/blame/64aa0d359c029dddc49f82491756a7a61bd22ded/app/templates/views/styleguide.html) no new patterns have been added in almost 5 years.

I think it’s better to get rid of it than have to keep maintaining something which is inaccurate.

Screenshot for posterity:

localhost_6012__styleguide(iPad)


leohemsted (approved):

klssmith:
I think you can get rid of `SHOW_STYLEGUIDE` from the config now too

quis:
Nice one @klssmith, removed in 13b63e1a2

PR #3687, opened 2020-10-19
Remove custom `getattr` implementation from JSONModel

quis:
We wrote custom `__getattr__` and `__getitem__` implementations to make it easier to find code that was accessing fields in the dictionary that we weren’t aware of (see https://github.com/alphagov/notifications-admin/commit/b48305c50d0d6b1674a39e0fd5a3c4768d281a66 for details and https://github.com/alphagov/notifications-admin/pull/2423 for more context).

Now that no view-layer code accesses the service dictionary directly we don’t need to support this behaviour any more. By removing it we can make the model code simpler, and closer to the `SerialisedModel` it inherits from.

It still needs some custom implementation because:
- a lot of our test fixtures are lazy and don’t set up all the expected fields, so we need to account for fields sometimes being present in the underlying dictionary and sometimes not
- we often implement a property that has the same name as one of the fields in the JSON, so we have to be careful not to try to override this property with the value from the underlying JSON

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -31,11 +31,9 @@ class Service(JSONModel):
     ALLOWED_PROPERTIES = {
         'active',
         'contact_link',
-        'email_branding',
         'email_from',
         'id',
         'inbound_api',
-        'letter_branding',
         'message_limit',
         'name',
         'prefix_sms',
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -39,7 +39,6 @@ class User(JSONModel, UserMixin):
         'mobile_number',
         'password_changed_at',
         'permissions',
-        'platform_admin',
         'state',
     }
 
--- a/tests/app/models/test_base_model.py
+++ b/tests/app/models/test_base_model.py
@@ -11,7 +11,7 @@ def test_looks_up_from_dict():
     assert Custom({'foo': 'bar'}).foo == 'bar'
 
 
-def test_prefers_property_to_dict():
+def test_raises_when_overriding_custom_properties():
 
     class Custom(JSONModel):
 
@@ -19,9 +19,12 @@ def test_prefers_property_to_dict():
 
         @property
         def foo(self):
-            return 'bar'
+            pass
 
-    assert Custom({'foo': 'NOPE'}).foo == 'bar'
+    with pytest.raises(AttributeError) as e:
+        Custom({'foo': 'NOPE'})
+
+    assert str(e.value) == "can't set attribute"
 
 
 @pytest.mark.parametrize('json_response', (
--- a/tests/app/models/test_base_model.py
+++ b/tests/app/models/test_base_model.py
@@ -39,10 +39,7 @@ def test_model_raises_for_unknown_attributes(json_response):
     with pytest.raises(AttributeError) as e:
         model.foo
 
-    assert str(e.value) == (
-        "'Custom' object has no attribute 'foo' and 'foo' is not a "
-        "field in the underlying JSON"
-    )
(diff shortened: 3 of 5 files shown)


app/models/user.py
```diff
@@ -39,7 +39,6 @@ class User(JSONModel, UserMixin):
         'mobile_number',
         'password_changed_at',
         'permissions',
-        'platform_admin',
```
idavidmcdonald, line 42: How come we get rid of this property?

app/models/service.py
```diff
@@ -31,11 +31,9 @@ class Service(JSONModel):
     ALLOWED_PROPERTIES = {
         'active',
         'contact_link',
-        'email_branding',
```
idavidmcdonald, line 34: Just to check I've understood, we remove `email_branding` and `letter_branding` for the `service` JSON model because they are properties that do not exist on the JSON the admin app receives for a service? Same for `template` in the `Job` JSON model above?

tests/app/models/test_base_model.py
```diff
@@ -39,21 +39,18 @@ class Custom(JSONModel):
...
         model.foo
 
-    assert str(e.value) == (
-        "'Custom' object has no attribute 'foo' and 'foo' is not a "
-        "field in the underlying JSON"
-    )
+    assert str(e.value) == "'Custom' object has no attribute 'foo'"
 
 
 def test_model_raises_keyerror_if_item_missing_from_dict():
```
idavidmcdonald, line 45: Test name needs updating

idavidmcdonald (commented):
Few small questions but generally looks good

app/models/user.py
quis: Properties shouldn’t be in this list if they’re [already implemented on the class](https://github.com/alphagov/notifications-admin/blob/92b735f6e471473c85372b5e4fa4e0949a806ccb/app/models/user.py#L190-L192). Trying to override a property that’s already implemented on the class using `setattr` will raise an exception.

app/models/service.py
quis: It’s because we already have custom implementations of these properties. Previously `JSONModel` would prefer these custom properties. Now it prevents having a name in the `ALLOWED_PROPERTIES` if a custom property with that name exists already.

For example, `email_branding` does exist in the JSON, but returns the ID, not the details of the email branding (which is usually what we want).

So we implement `email_branding_id`, for when that’s specifically all you want:
https://github.com/alphagov/notifications-admin/blob/92b735f6e471473c85372b5e4fa4e0949a806ccb/app/models/service.py#L452-L454

…which is then referenced by a custom `email_branding` property, for when you want details of the branding:
https://github.com/alphagov/notifications-admin/blob/92b735f6e471473c85372b5e4fa4e0949a806ccb/app/models/service.py#L456-L460

tests/app/models/test_base_model.py
quis:
```suggestion
def test_model_raises_attribute_error_if_item_missing_from_dict():
```

app/models/user.py
idavidmcdonald: does the `not hasattr(self, property)` make sure it doesn't raise the exception?

Although either way it's fine to remove this line as you are right that it's already implemented.

quis: I’ve updated it so it now raises an exception where there’s this conflict, rather than resolving silently. This will help us keep the `ALLOWED_PROPERTIES` list tidy. Changes in 38e9e84f7

idavidmcdonald (approved):

PR #3688, opened 2020-10-19
Make custom data retention look a bit nicer

quis:
This should make it easier to see what’s set without having to click into the table.

# Before 

image

# After 

image

***

image


idavidmcdonald (approved):
Nice

PR #3689, opened 2020-10-19
Allow platform admins to change daily message limit

quis:
One less thing we have to go into the database to do, and remember to manually clear the cache for after.

image


idavidmcdonald (approved):
Smart work

PR #3691, opened 2020-10-21
Fix incorrect redis key deleting

idavidmcdonald:
We were trying to delete the old 'template-{template-id}' keys but
should have been deleting the new keys which have the service id as part
of the key name. This was causing the cache to not be correctly purged
when we did things like update sender names or set defaults. This should
fix it.

### DIFF
--- a/tests/app/notify_client/test_service_api_client.py
+++ b/tests/app/notify_client/test_service_api_client.py
@@ -550,5 +550,5 @@ def test_client_deletes_service_template_cache_when_service_is_updated(
     templates_to_delete = mock_redis_delete.call_args_list[0][0]
     assert len(templates_to_delete) == 6
     for template_key in templates_to_delete:
-        assert template_key.startswith('template-')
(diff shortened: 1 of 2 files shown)


tests/app/notify_client/test_service_api_client.py
```diff
@@ -550,5 +550,5 @@ def test_client_deletes_service_template_cache_when_service_is_updated(
     templates_to_delete = mock_redis_delete.call_args_list[0][0]
     assert len(templates_to_delete) == 6
     for template_key in templates_to_delete:
-        assert template_key.startswith('template-')
+        assert template_key.startswith('service-{}-template'.format(SERVICE_ONE_ID))
```
quis, line 553: This would be nicer as an f-string
```suggestion
        assert template_key.startswith(f'service-{SERVICE_ONE_ID}-template')
```

quis (approved):
Approved with an optional style change

PR #3697, opened 2020-10-22
Remove July to September section from roadmap

karlchillmaid:
We're not ready to publish a detailed roadmap for 2021 yet – but we do need to:

* remove the July to September section so the page doesn't look too out of date
* reflect the work that we're actually doing in October to December
* move everything else into next year

### DIFF
--- a/app/templates/views/roadmap.html
+++ b/app/templates/views/roadmap.html
@@ -16,9 +16,7 @@
   <h2 class="heading-medium">October to December 2020</h2>
 
   <ul class="list list-bullet">
-    <li>Let services attach forms to letter templates</li>
-    <li>Distribute the delivery of a batch of messages over several hours</li>
-    <li>Make it easy to see the replies from staff on an emergency contact list</li>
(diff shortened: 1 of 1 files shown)


yahoopete (approved):

app/templates/views/roadmap.html
```diff
@@ -41,6 +33,9 @@ <h2 class="heading-medium">January 2021 onwards</h2>
     <li>Add a link shortening service</li>
     <li>Specify an expiry period for undeliverable text messages</li>
     <li>Let users save draft templates</li>
+    <li>Start sending emails from NHS and Parliament email addresses</li>
+    <li>Make it easier for software suppliers set up Notify for their customers</li>
```
idavidmcdonald, line 37: missing 'to' in this sentence I believe (it was this way before too)

quis:
```suggestion
    <li>Make it easier for software suppliers to set up Notify for their customers</li>
```

tombye (approved):
Reviewing again as @karlchillmaid asked someone to check the latest updates. Looks good to me 👍

PR #3701, opened 2020-10-23
Store filenames in metadata, not query strings

klssmith:
When uploading an emergency contact and sending a batch of notifications (either from a stored contact list or a CSV file) we were passing the filename in the query string. It's possible for the filename to get lost and to default to an empty string - this tends to happen if someone gets redirected to sign in while they were in the middle of sending. This PR starts using the metadata to transfer the filename between view functions.

It will still be possible for the filename to be in a query string for a while - people may have left a window open before sending, so this checks the query string if the filename is not in the metadata. This check can be removed after the change has been deployed for a few days.


CrystalPea (approved):
Good work! 👏

PR #3708, opened 2020-10-30
Remove old line from send file by email feature

idavidmcdonald:
You do not need to contact us to use it anymore. You can set it up
yourself and is documented in our API docs.

Questionable whether this content should be more specific saying 'We do not currently offer this feature through our user interface' or saying 'it is only available for API users' but that's maybe an improvement for another day.


karlchillmaid (approved):
Looks good – thanks for doing this, David.

Users should infer that this is API-only from the line about reading the API docs. Let's keep an eye on support and see if we get a lot of comments like the one today. In which case, we can add something more explicit.

PR #3711, opened 2020-11-05
Add content about the extended character set

karlchillmaid:
We need to update the pricing page to explain the 9 characters that are charged as 2 characters each.

See the [Pivotal ticket for details](https://www.pivotaltracker.com/n/projects/1443052/stories/175457294).

### DIFF
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -43,7 +43,9 @@
   <p class="govuk-body">See <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.how_to_pay') }}">how to pay</a>.
 
   <h3 class="heading-small" id="long-text-messages">Long text messages</h3>
-  <p class="govuk-body">If a text message is longer than 160 characters (including spaces), it’ll be charged as more than one message:</p>
+  <p class="govuk-body">If a text message is longer than 160 characters (including spaces), it’ll be charged as more than one message.</p>
+  <p class="govuk-body">The following characters are charged as 2 characters each: [ ] { } ^ \ | ~ €</p>
+
   <div class="bottom-gutter-3-2">
     {% call mapping_table(
       caption='Text message pricing',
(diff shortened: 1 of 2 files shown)


app/templates/views/pricing/index.html
```diff
@@ -43,7 +43,9 @@ <h2 class="heading-medium" id="text-messages">Text messages</h2>
   <p class="govuk-body">See <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.how_to_pay') }}">how to pay</a>.
 
   <h3 class="heading-small" id="long-text-messages">Long text messages</h3>
-  <p class="govuk-body">If a text message is longer than 160 characters (including spaces), it’ll be charged as more than one message:</p>
+  <p class="govuk-body">If a text message is longer than 160 characters (including spaces), it’ll be charged as more than one message.</p>
+  <p class="govuk-body">The following characters are charged as 2 characters each: [ ] { } ^ \ | ~ €</p>
```
karlchillmaid, line 47: @quis @tombye I’ve put a space between each character to make it easier to read them.

I haven’t separated them with commas (as we do for the accented characters) even though it’s grammatically correct because:

a) I thought `[, ], {, }, ^, \, |, ~, €` might look a bit weird
b) there’s a chance you could read the comma as one of the double-character characters

What do you think? Could using spaces to separate the characters cause any (accessibility) problems?

Screenshot 2020-11-09 at 16 28 18
Screenshot 2020-11-09 at 16 28 45

tombye: I personally prefer it without commas but only because they make it look a bit busy.

Spacing-size I think they might not need the extra spaces if we changed the font to a monospace one. I've done a bit of research and here's what they look like with the 'tabular numbers' version of New Transport (in Chrome):
osx_tabular_chrome
By contrast, here's what they look like just using the system monospace fonts:
### Monospace on OSX
osx_monospace_chrome

### Monospace on Windows 10
windows_10_monospace_chrome

I also had a look at what it would look like using the commas with the "tabular numbers" New Transport.
osx_tabular_with_commas_chrome

karlchillmaid: Thanks @tombye – that’s really helpful.

Looking at those examples, my preference is still without commas.

quis: Monospace is what I was going to suggest. At least on Windows it makes it easier to distinguish the tilde `~` from a dash `-`.

I like the spaces, but think adding some letter spacing might look more even than using space characters.

quis: 1. Here’s letter spacing of `0.5em`, with our regular font:
image

2. Letter spacing of `0.4em` with monospace font:
image

3. Spaces:
image

4. Spaces with monospace font:
image

I think 1. looks nicest but maybe the individual characters are more distinctive with the monospace font in 2.…

karlchillmaid: I like 1 best too. You’re probably right about the individual characters in version 2, but I don’t like € sign* as much.

That’s a personal preference – I don’t have any evidence to back it up.

tombye: I've added a commit that adds some HTML and CSS to make it like option 1.

klssmith (approved):

PR #3712, opened 2020-11-05
Stop live services requesting to go live

klssmith:
Live services shouldn't be able to request to go live again. Once a service is live we remove the option to go live from the Settings page, but we still link to the page to request to go live from other places e.g. the 'Get started' page. As a result, we've seen some services make a request to go live when their service has already been live for months - this change will stop that from happening.

@karlchillmaid is going to look at the content for this

_Screenshot now updated with the agreed content_

Screenshot 2020-11-06 at 17 54 11

### DIFF
--- a/app/templates/views/service-settings/service-already-live.html
+++ b/app/templates/views/service-settings/service-already-live.html
@@ -1,0 +1,28 @@
+{% extends "withnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+
+{% block service_page_title %}
+  Your service is already live
+{% endblock %}
+
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      {{ page_header('Your service is already live') }}
+
+      <p class="govuk-body">
+        {% if current_service.go_live_at %}
+          ‘{{ current_service.name }}’ went live on {{ current_service.go_live_at | format_date_normal }}.
+        {% else %}
+          ‘{{ current_service.name }}’ is already live.
+        {% endif %}
+        </p>
+
+      <p class="govuk-body">
+        <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.choose_account') }}">Switch service</a>
+         if you want to make a different service live.
+      </p>
+
+    </div>
+  </div>
+{% endblock %}
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -187,6 +187,9 @@ def estimate_usage(service_id):
 @main.route("/services/<uuid:service_id>/service-settings/request-to-go-live", methods=['GET'])
 @user_has_permissions('manage_service')
 def request_to_go_live(service_id):
+    if current_service.live:
+        return render_template('views/service-settings/service-already-live.html')
+
     return render_template(
         'views/service-settings/request-to-go-live.html'
     )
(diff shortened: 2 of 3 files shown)


app/templates/views/service-settings/service-already-live.html
```diff
@@ -0,0 +1,24 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      {{ page_header('Your service is already live') }}
+
+      <p class="govuk-body">
+        {{ current_service.name }} is already live.
+      </p>
```
karlchillmaid, line 15: The h1 and first line of body content are very similar. We should avoid repetition if we can.

**Solution 1** – make the service name part of the h1

```suggestion
      {{ page_header('‘{{ current_service.name }}’ is already live') }}
```

Issue: we know that some service names can be very long

**Option 2** – remove the service name altogether

```suggestion
      {{ page_header('Your service is already live') }}
```

Issue: while this looks neater, it’s less helpful – especially if the user doesn’t notice the current service name in the breadcrumb.

karlchillmaid: We discussed another option, which was to make the first line of the body more useful. For example:

```‘Service name’ went live on Thursday 5 November 2020.```

We didn‘t go with this because it would make the code more complicated for a page which is very rarely seen (@klssmith can only remember seeing this happen twice).

Comment (klssmith) [reply app/templates/views/service-settings/service-already-live.html:15]: Actually, I've just taken another look and it is a very easy change to add the date and doesn't complicate things after all.

Comment (karlchillmaid) [reply app/templates/views/service-settings/service-already-live.html:15]: Great, let’s go with:

```suggestion
      {{ page_header('Your service is already live') }}

      <p class="govuk-body">
        ‘{{ current_service.name }}’ went live on [day] [date] [month] [year].
      </p>
```

app/templates/views/service-settings/service-already-live.html
```diff
@@ -0,0 +1,28 @@
...
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      {{ page_header('Your service is already live') }}
+
+      <p class="govuk-body">
+        {% if current_service.go_live_at %}
+          ‘{{ current_service.name }}’ went live on {{ current_service.go_live_at | format_date_normal }}.
+        {% else %}
+          ‘{{ current_service.name }}’ is already live.
```
klssmith, line 17: There are a few services in the database which are live (i.e. `restricted` is `False`) but that don't have a `go_live_at` timestamp. These are almost all services created by people in Notify, so it is very unlikely that anyone will end up in this `else` clause and see a message without the date.

app/templates/views/service-settings/service-already-live.html
klssmith: Ok, that's updated now - I've updated the screenshot to show the new message.

karlchillmaid (approved):
Content looks good – thank you!

app/templates/views/service-settings/service-already-live.html
CrystalPea: ok, should we still test that this edge case renders ok? (as in there is no error)

klssmith: I tried to cover that in `test_request_to_go_live_redirects_if_service_already_live` - it's parameterized to test the page when a service does have `go_live_at` and also when `go_live_at` is `None`

CrystalPea: Perfect, I see it now, sorry!

CrystalPea (approved):

PR #3717, opened 2020-11-18
Fix nesting in content page sub nav

tombye:
Nested lists in the side navigation of content pages have incorrect HTML.

image

They currently have `<ol>` tags as direct children of their parent lists when only `<li>`s are allowed:

## Currently

```
ol
- li
- li
- ol
  - li
- li
```

## After changes

```
ol
- li
- li
  - ol
    - li
- li
```

This fixes that and updates the CSS so there are no visual differences.

##


idavidmcdonald (approved):
Nice, html appears correct when I run it locally

PR #3721, opened 2020-11-24
Correct field_label attribute to 'users' on  manage folder page

tombye:
The checkbox component on the manage folder page, used for user permissions, generates a summary of what is selected. This is using the wrong noun to describe the type of permission.

image

This fixes that by changing the `field_label`attribute that field uses to set the noun.

image

Note: we don't use an icon for users so this doesn't use one, like the version for folder permissions does.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2126,7 +2126,7 @@ class TemplateFolderForm(StripWhitespaceForm):
 
     users_with_permission = GovukCollapsibleCheckboxesField(
         'Team members who can see this folder',
-        field_label='folder')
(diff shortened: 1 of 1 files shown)


klssmith (approved):

tombye:
Note: The changes I [force-pushed](https://github.com/alphagov/notifications-admin/compare/b7a3a70185a042d9fad09cea12305f70e635cebc..2f5f836dda72d459edb21c94a664ddf956664cd1) have been checked with @klssmith offline, where she confirmed the approval still stands.

PR #3724, opened 2020-11-30
Validate maximum service name length

klssmith:
**Fix use of `client_request.post` in tests**
The `post` method of the `client_request` fixture has an argument called `_data`. There were a few places where we had used an argument of `data` instead by mistake.

We could change the code to raise an error if people pass `data` as kwarg to `client_request`, but I didn't do this in case we do need a `data` kwarg in some scenarios. 

**Add form validation for max service and org name length**
There was a recent error in the logs because a service tried to change its name to one exceeding 255 characters (which is a limit on the database field). We can easily catch these errors on the form, so that the user doesn't see an error page.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1106,7 +1106,8 @@ class RenameServiceForm(StripWhitespaceForm):
         u'Service name',
         validators=[
             DataRequired(message='Cannot be empty'),
-            MustContainAlphanumericCharacters()
+            MustContainAlphanumericCharacters(),
+            Length(max=255, message='Service name must be 255 characters or fewer')
         ])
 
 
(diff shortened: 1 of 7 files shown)


quis (changes requested):
Nice.

At the moment this uses the default error message, which looks like this:
image

Following [the Design System example for errors with character counts](https://design-system.service.gov.uk/components/character-count/#error-messages) I think a better message would be
> Service name must be 255 characters or fewer

https://design-system.service.gov.uk/components/character-count/#error-messages

klssmith:
I finally got round to updating the validation messages. I also snuck in a tiny third commit to add the `govuk-link--no-visited-state` class to two links.

quis (approved):

PR #3729, opened 2020-12-03
Tell browser to connect to asset domain earlier

quis:
When a browser loads a Notify page it does the following:
- DNS and TLS handshake for notifications.service.gov.uk
- download some HTML
- sees that the HTML needs to load some CSS
- DNS and TLS handshake for static.notifications.service.gov.uk
- downloads the CSS

We can speed things up a bit in modern browsers by parallelizing this process a bit. Modern browsers support some HTTP headers<sup>1</sup> that allow them to connect to other origins sooner.

After this change the steps are:
- DNS and TLS handshake for notifications.service.gov.uk
- receive response headers and simultaneously:
  - download some HTML
  - DNS and TLS handshake for static.notifications.service.gov.uk
- sees that the HTML needs to load some CSS
- downloads the CSS

1. https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -630,6 +630,11 @@ def useful_headers_after_request(response):
             logo_domain=get_logo_cdn_domain(),
         )
     ))
+    response.headers.add('Link', (
+        '<{asset_url}>; rel=dns-prefetch, <{asset_url}>; rel=preconnect'.format(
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Nice one. Looks good according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link

PR #3730, opened 2020-12-03
Add GOVUK radios [part 4]

tombye:
Part 4 of migrating our radio buttons to use the [GOV.UK Frontend radios component](https://design-system.service.gov.uk/components/radios/).

https://www.pivotaltracker.com/story/show/170030262

This pull request makes `OrganisationTypeField` inherit from `GovukRadiosField` and so converts all form fields using `OrganisationTypeField` to use the GOVUK radios component.

It also makes `OrganisationCrownStatusForm.crown_status` use `GovukRadiosField`, with the same effect.

### Notes

Part 1 of this work was done in https://github.com/alphagov/notifications-admin/pull/3715.
Pull requests for the other parts are:
- part 2: https://github.com/alphagov/notifications-admin/pull/3727
- part 3: https://github.com/alphagov/notifications-admin/pull/3728
- part 5: https://github.com/alphagov/notifications-admin/pull/3731

Pull requests in this stream of work don't need to be completed in any order.

PR #3715, opened 2020-11-11
Add GOVUK radios [part 1]

tombye:
Initial piece of work for migrating Notify's radio buttons to use the [GOV.UK Frontend radios component](https://design-system.service.gov.uk/components/radios/).

This pull request:
- refactors the existing classes in `app/main/forms.py` that render GOV.UK Frontend components so they do that in a consistent way
- adds a basic `GovukRadiosField` class
- plugs it into a few pages

Part 1 of https://www.pivotaltracker.com/story/show/170030262.

## How to review

This pull request is written to be reviewed commit-by-commit.

PR #3731, opened 2020-12-03
Add GOVUK radios [part 5]

tombye:
Part 5 of migrating our radio buttons to use the [GOV.UK Frontend radios component](https://design-system.service.gov.uk/components/radios/).

https://www.pivotaltracker.com/story/show/170030262

This pull request adds a `GovukRadiosFieldWithNoneOption` class, sets `SetEmailBranding.branding_style` to use it and updates all the pages/routes that use the `SetEmailBranding` and `SetLetterBranding` forms.

### Notes

There is an existing version of `GovukRadiosFieldWithNoneOption` called `RadioFieldWithNoneOption` which could have been converted instead. `RadioFieldWithNoneOption` is used by some fields I intend to change in future so I'm leaving that work for then.

Part 1 of this work was done in https://github.com/alphagov/notifications-admin/pull/3715.
Pull requests for the other parts are:
- part 2: https://github.com/alphagov/notifications-admin/pull/3727
- part 3: https://github.com/alphagov/notifications-admin/pull/3728
- part 4: https://github.com/alphagov/notifications-admin/pull/3730

Pull requests in this stream of work don't need to be completed in any order.

### DIFF
--- a/app/templates/views/service-settings/set-letter-branding.html
+++ b/app/templates/views/service-settings/set-letter-branding.html
@@ -24,7 +24,7 @@
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-full">
         {{ live_search(
-          target_selector='.multiple-choice',
+          target_selector='.govuk-radios__item',
           show=True,
           form=search_form,
           label='Search by name',
--- a/app/templates/views/service-settings/set-letter-branding.html
+++ b/app/templates/views/service-settings/set-letter-branding.html
@@ -1,5 +1,4 @@
 {% extends "withnav_template.html" %}
-{% from "components/radios.html" import radios %}
 {% from "components/live-search.html" import live_search %}
 {% from "components/page-header.html" import page_header %}
 {% from "components/page-footer.html" import page_footer %}
@@ -31,7 +30,7 @@
           label='Search by name',
           autofocus=True
         ) }}
-        {{ radios(form.branding_style, hide_legend=True) }}
+        {{ form.branding_style }}
       </div>
     </div>
     <div class="js-stick-at-bottom-when-scrolling">
--- a/app/assets/javascripts/previewPane.js
+++ b/app/assets/javascripts/previewPane.js
@@ -4,7 +4,7 @@
 
   $ = global.jQuery;
 
-  let branding_style = $('.multiple-choice input[name="branding_style"]:checked');
+  let branding_style = $('.govuk-radios__item input[name="branding_style"]:checked');
 
   if (!branding_style.length) { return; }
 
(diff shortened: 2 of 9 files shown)


klssmith:
This change seems to have stopped the preview page (for both emails and letters) from appearing.

app/templates/views/service-settings/set-letter-branding.html
```diff
@@ -31,7 +30,7 @@
           label='Search by name',
           autofocus=True
         ) }}
-        {{ radios(form.branding_style, hide_legend=True) }}
```
klssmith, line 34: Not having `hide_legend` here or on the other letter branding form means that we now show the text "Branding style" above the radio buttons. That might be fine though - I'm not sure if there was a reason we were showing that for emails but not for letters.

tombye: I looked into this and both the org and service settings pages for email branding show the legend but those for letter branding don't. I can't see a reason why they should differ as they're basically the same interface.

It seems sensible to make them consistent and, speaking personally, I'd choose to hide the legend as that information is available in the page title and h1 any hiding it adds vertical space.

What do you think?

klssmith: I think the page looks better with the legend hidden since it duplicates existing information on the page. I can't think of a reason why we'd hide it for letters but show it for emails, so I'd probably go with just removing it for all branding types.

tombye:
> This change seems to have stopped the preview page (for both emails and letters) from appearing.

Great spot! 🥇 I totally missed that changing the radios HTML in the branding pages broke a CSS selector in the JS for the preview iframe. Looks like it failed silently so no tests to say this happened.

I've updated that selector and updated the bits of the JS tests that render radios HTML to match the new style.

app/templates/views/service-settings/set-letter-branding.html
tombye: Great. I've made a change in the form that serves all these pages to hide the legend so marking this as resolved.

klssmith:
The preview pane is back 🎉. One more thing I spotted is that as well as changing the preview pane so that it looks for `.govuk-radios__item` instead of `.multiple-choice input`, you'll also have to make a similar change for the `target_selector` of the `live_search` macro.

tombye:
> The preview pane is back 🎉. One more thing I spotted is that as well as changing the preview pane so that it looks for `.govuk-radios__item` instead of `.multiple-choice input`, you'll also have to make a similar change for the `target_selector` of the `live_search` macro.

Good point! I've updated that selector on all the pages changed. I also updated the tests for live-search so any radios HTML they use is the new version.

klssmith (approved):

PR #3733, opened 2020-12-04
Pre-compress static files 

quis:
At the moment we store uncompressed files in S3 and ask Cloudfront to compress them before caching and serving them.

This works great, except sometimes when Cloudfront is busy it just decides not to apply any compression<sup>1 2</sup>. As a bonus it caches this uncompressed response. So we serve uncompressed assets until we deploy a change that modifies one of the assets. This is totally the behaviour of a well-designed service that’s been built with developer ergonomics in mind, and is exactly the kind of technical decision you’d expect from a company with such clear and readable documentation as AWS.

This commit applies some blunt bash scripting to compress the files (preserving the original file name) before uploading them to S3, so that we always serve compressed assets. Using gzip with `-9` (the highest) level of compression will save us a few bytes, since Cloudfront seems to use a lower level of compression to save CPU time on their servers<sup>3</sup>.

All browsers down to Netscape 4 and Internet Explorer 4 support gzip-compressed content<sup>4</sup>.

We only need to compress CSS and Javascript, since fonts and images are already compressed.

1. https://www.debugbear.com/blog/intermittent-compression
2. https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html
3. Try entering `https://static.notifications.service.gov.uk/stylesheets/main.css` at https://tools.paulcalvano.com/compression.php
4. http://schroepl.net/projekte/mod_gzip/browser.htm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,6 +173,10 @@ generate-manifest:
 
 .PHONY: upload-static ## Upload the static files to be served from S3
 upload-static:
+	gzip -9 ./app/static/stylesheets/*.css
+	for file in app/static/stylesheets/*.gz; do mv "$file" "${file%%.gz}"; done;
+	gzip -9 ./app/static/javascripts/*.js
+	for file in app/static/javascripts/*.gz; do mv "$file" "${file%%.gz}"; done;
 	aws s3 cp --region eu-west-1 --recursive --cache-control max-age=315360000,immutable ./app/static s3://${DNS_NAME}-static
 
 .PHONY: cf-deploy


tombye:
This all looks good though I have a question about the implementation, excuse if obvious.

I'm guessing that these changes should be combined with turning compression off on the cloudfront cache behaviour (I think that's in [line 32 in the AWS config](https://github.com/alphagov/notifications-aws/blob/master/terraform/modules/admin-cloudfront/main.tf#L32) and [line 58 in the AWS config](https://github.com/alphagov/notifications-aws/blob/master/terraform/modules/admin-cloudfront/main.tf#L58)). Assuming these changes go first, that would leave a small bit of time where we serve compressed files while telling cloudfront to compress them too*.

If that's right, do you know if AWS has checks to prevent re-compression or can we do something else to mitigate this?

*I tested this by trying to run `gzip -9` on a gzipped file with the `.gz` suffix removed and it did compress it again.

quis:
I think Cloudfront will look at the mime type of the file, not the extension when deciding to compress it. We should figure out how to test this though, you’re right we want to avoid accidentally double-zipping.

tombye:
We set up a small test for this (see https://github.com/alphagov/notifications-admin/pull/3738). I've written up the results, which I think prove that a gzipped asset served by Cloudfront with compression on will be double-compressed: https://drive.google.com/drive/folders/1vJdOGTNVOltgdWXxxjQGzkQ6De4sNruI?usp=sharing

quis:
Closing this. We think it’s not worth investing more time on now because:
- it’s happening fairly intermittently 
- the fix is not straightforward

PR #3735, opened 2020-12-04
show allowed broadcast provider

leohemsted:
in the "training"/"live" window

image

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -46,7 +46,8 @@ class Service(JSONModel):
         'consent_to_research',
         'count_as_live',
         'go_live_user',
-        'go_live_at'
+        'go_live_at',
+        'allowed_broadcast_provider',
     }
 
     TEMPLATE_TYPES = (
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
Do we want the premise of live/training to be next to the premise of only sending to one provider? They feel like different things. One tells you if you actually make a call to a CBC or not. The other tells you which CBC the account is hooked up to. So technically a training Vodaphone CBC won't send messages to the Vodafone CBC even though it sounds like it might be a playground to send messages to the CBC.

I'm not actually that fussed but design guru @quis might have an opinion?

leohemsted:
that's a good point - they're independent things so i decided to make it flexible, but a restriction only makes sense for a live service, so we don't need to show it even if its set on the service

quis:
Only showing it for live services looks good to me 👍

idavidmcdonald (approved):
Looks good technically. Probably good to go unless Chris has any objections on the design

leohemsted:
After discussion on slack, removed commit https://github.com/alphagov/notifications-admin/commit/fa9240e98dbf94877b5919f17ccac0dc00a4ef04. This means we _do_ show the flag in training mode - we're still sending real messages in training mode for now so that we can self-approve, and as such it's still relevant to show at that point

PR #3737, opened 2020-12-08
Use cf7 commands for the cf CLI commands

idavidmcdonald:
As we want to upgrade from cf6 to cf7 CLI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ CF_ORG ?= govuk-notify
 CF_SPACE ?= ${DEPLOY_ENV}
 CF_HOME ?= ${HOME}
 CF_APP ?= notify-admin
+CF_MANIFEST_PATH ?= /tmp/manifest.yml
 $(eval export CF_HOME)
 
 NOTIFY_CREDENTIALS ?= ~/.notify-credentials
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -180,22 +181,31 @@ cf-deploy: ## Deploys the app to Cloud Foundry
...
-	cf push -f <(CF_APP=notify-admin-prototype-2 make -s generate-manifest)
+	make -s CF_APP=notify-admin-prototype-2 generate-manifest > ${CF_MANIFEST_PATH}
+	cf push notify-admin-prototype-2 --strategy=rolling -f ${CF_MANIFEST_PATH}
+	rm ${CF_MANIFEST_PATH}
 
 .PHONY: cf-rollback
 cf-rollback: cf-target ## Rollbacks the app to the previous release
-	cf v3-cancel-zdt-push ${CF_APP}
+	cf cancel-deployment ${CF_APP}
+	rm ${CF_MANIFEST_PATH}
```
leohemsted, line 208: should have realised in the API pull request, my bad. i think this rm command should have the `-f` arg. That way, if the file doesn't exist, there's no output and the makefile doesn't fail. It's most relevant to this rollback command, but can't hurt in the other commands I think. If you agree I can update the api makefile

```suggestion
	rm -f ${CF_MANIFEST_PATH}
```

leohemsted (commented):
One minor comment

Makefile
idavidmcdonald: Agreed

idavidmcdonald: I've squashed and pushed up changes

leohemsted (approved):

PR #3738, opened 2020-12-09
Generate asset to test CDN compression config

tombye:
Cloudfront, our CDN, sometimes decides not to gzip assets. Because of this, we're going to gzip them ourselves prior to upload instead.

This will involve:
1. adding gzipping to the make task that uploads
   them
2. turning compression off in Cloudfront

There is already a pull request up for number 1:

https://github.com/alphagov/notifications-admin/pull/3733

Because deploying all this will, at some point, create a state where Cloudfront is set to compress assets that are already compressed, we need to test that it doesn't re-compress them.

This adds a frontend build task that generates a test asset which is:
- a copy of app/static/stylesheets/main.css
- renamed to include a MD5 SHA of its contents
- already gzipped

Once deployed, the test will be to:
1. download the asset from the live environment
2. unzip it
3. diff it against app/static/stylesheets/main.css

Something like:
```bash
curl -o ~/Downloads/test-main.css https://static.notifications.service.gov.uk/test/main-110c571859be1151126352d4798dd2a4.gz.css
curl -o ~/Downloads/current-main.css https://static.notifications.service.gov.uk/stylesheets/main.css?110c571859be1151126352d4798dd2a4
gzip -d -c ~/Downloads/test-main.css > ~/Downloads/test-main-unzipped.css
if diff ~/Downloads/test-main.css ~/Downloads/current-main.css > /dev/null; then echo "Test file has differences"; else echo "Test file matches main.css"; fi
```
This snippet will need all paths checking beforehand to ensure the latest `main.css` is referenced.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,7 +18,9 @@ plugins.base64 = require('gulp-base64-inline');
 plugins.cleanCSS = require('gulp-clean-css');
 plugins.concat = require('gulp-concat');
 plugins.cssUrlAdjuster = require('gulp-css-url-adjuster');
+plugins.gzip = require('gulp-gzip');
 plugins.jshint = require('gulp-jshint');
+plugins.md5 = require('gulp-md5');
 plugins.prettyerror = require('gulp-prettyerror');
 plugins.rollup = require('gulp-better-rollup')
 plugins.sass = require('gulp-sass');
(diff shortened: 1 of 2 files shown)


quis (approved):
image

👍🏻

tombye:
I realised there's a flaw in my test. Getting the test file, and `main.css` via those curl commands misses sending the `Accept-Encoding: gzip, deflate, br` header that browsers send. Without that header, you're not simulating what a browser will receive. I've now updated the test to include it.

PR #3742, opened 2020-12-16
Fix for file input bug in safari

tombye:
Safari has a bug where it stops input[type=file] elements working if they don't specify the types of file to accept (via the [accept attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept)).

file_upload_not_working

It seems to just effect certain versions of Mojave but completely blocks this action for people on that version so worth fixing.

This changes the file_upload component, and all calls to it, so `accept` is set for the types of file allowed in each case.

## Notes on the bug

This was spotted on x-gov Slack:

https://ukgovernmentdigital.slack.com/archives/C06GCJW7R/p1607952390112800

...and StackOverflow:

https://stackoverflow.com/q/64843459/679924

## Notes to help reviewers

I've added assertions to various tests covering the pages that have file upload inputs. These are best guesses so please say if they could be in a better place.

The `accept` value for the file_upload call for jobs and contact lists comes from a `set` of extensions so its order isn't guaranteed. Because of this, I had to do some extra work in the tests to assert all values were present.

### DIFF
--- a/app/templates/components/file-upload.html
+++ b/app/templates/components/file-upload.html
@@ -27,7 +27,7 @@
     </label>
     {{ field(**{
       'class': 'file-upload-field',
-      'accept': allowed_file_extensions
+      'accept': allowed_file_extensions|format_list_items('.{item}')|join(',')|e
     }) }}
     <label class="file-upload-button" for="{{ field.name }}">
       {{ button_text }}
--- a/app/templates/components/file-upload.html
+++ b/app/templates/components/file-upload.html
@@ -2,6 +2,7 @@
 
 {% macro file_upload(
   field,
+  allowed_file_extensions,
   action=None,
   button_text="Choose file",
   alternate_link=None,
@@ -25,7 +26,8 @@
       {% endif %}
     </label>
     {{ field(**{
-      'class': 'file-upload-field'
+      'class': 'file-upload-field',
+      'accept': allowed_file_extensions
     }) }}
     <label class="file-upload-button" for="{{ field.name }}">
       {{ button_text }}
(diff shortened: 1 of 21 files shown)


app/templates/components/file-upload.html
```diff
@@ -25,7 +26,8 @@
       {% endif %}
     </label>
     {{ field(**{
-      'class': 'file-upload-field'
+      'class': 'file-upload-field',
+      'accept': allowed_file_extensions
```
quis, line 30: I’d prefer `allowed_file_extensions` was a list, and we transform it into the attribute string once here, for example
```suggestion
      'accept': allowed_file_extensions|prepend('.')|join(',')
```
(You’d need to write the `prepend` Jinja filter yourself since there isn’t a built in one)

tombye: What about using [formatted_list](https://github.com/alphagov/notifications-utils/blob/b1b9c1aa1da6ab5eef39d5b35089341f650748b4/notifications_utils/formatters.py#L195) for that instead of writing a new filter?

quis: Nice thinking 🧠

tombye: So it turns out we couldn't use it because it assumes a conjunction so even if you set that to `''`, it has the spaces around it. Instead I (kind of) took your advice and added a `format_list_items` filter which just lets you apply a [format string](https://docs.python.org/3/library/string.html#format-string-syntax) and whatever arguments you want to each item in an iterator.

tombye:
I also made `Spreadsheet.ALLOWED_FILE_EXTENSIONS` into a tuple, purely because it made checking the string version of it a lot less work in the tests. Discussed this with @quis offline before doing.

quis (approved):
Changes look good.

I also checked this against https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept

👍🏻

PR #3746, opened 2020-12-22
Tell users that postage prices will increase

karlchillmaid:
This PR follows the pattern we used for the October 2019 increases:

1. Announce the date (and value) of the rate increase in advance, using the call-out pattern.
2. Temporarily replace the ‘These prices were updated on 7 September 2020’ (we’ll reinstate this message once the new prices come into effect, and the table is updated)

## Why we’re doing this

We’re changing the cost of sending letters through Notify. This is because Royal Mail (and DVLA) are increasing their rates in 2021.

We understand that prices will change from 1 February 2021, so we need to give users a month’s notice.

## Price increases
Screenshot 2020-12-23 at 09 49 22

### DIFF
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -230,7 +230,10 @@
     {% endcall %}
   </div>
   <div class="panel panel-border-wide">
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):
Looks good

PR #3751, opened 2020-12-24
Let users create an emergency alert without a template 

quis:
We think that in some cases alerts will be composed in the moment, and therefore making people first create a template is:
- not a good use of their time
- adding some conceptual complexity which they don’t need to take onboard

This commit makes it possible to type some words and have them go straight into the `content` field in the database.

In the future we might want to progressively enhance the radio buttons so they show on the same page (like we do with the grey buttons on the templates page).

***

image

image

image

image

image

***

Depends on:
- [x] https://www.pivotaltracker.com/story/show/176186142
- [x] https://github.com/alphagov/notifications-api/pull/3095

### DIFF
--- a/app/templates/views/broadcast/new-broadcast.html
+++ b/app/templates/views/broadcast/new-broadcast.html
@@ -1,0 +1,23 @@
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/form.html" import form_wrapper %}
+
+{% extends "withnav_template.html" %}
+
+{% block service_page_title %}
+  New alert
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ page_header(
+    'New alert',
+    back_link=url_for('.broadcast_dashboard', service_id=current_service.id),
+  )}}
+
+  {% call form_wrapper() %}
+    {{ form.content }}
+    {{ page_footer('Continue') }}
+  {% endcall %}
+
+{% endblock %}
--- a/app/templates/views/broadcast/new-broadcast.html
+++ b/app/templates/views/broadcast/new-broadcast.html
@@ -1,4 +1,4 @@
-{% from "components/page-header.html" import page_header %}
+{% from "components/back-link/macro.njk" import govukBackLink %}
 {% from "components/page-footer.html" import page_footer %}
 {% from "components/form.html" import form_wrapper %}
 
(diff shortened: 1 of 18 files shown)


app/templates/views/broadcast/new-broadcast.html
```diff
@@ -0,0 +1,23 @@
...
+
+{% extends "withnav_template.html" %}
+
+{% block service_page_title %}
+  New alert
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ page_header(
```
tombye, line 13: Seeing as the radios are the only control in the form, show the legend be the h1? The radios field can be configured to do that pretty easily if so. Some examples of that here: https://github.com/alphagov/notifications-admin/pull/3771

tombye (changes requested):
I've a few comments, but overall this looks great.

I'm conscious my comments are all on the frontend of things. I have read through the rest and it seems to all make sense to me.

app/templates/views/broadcast/write-new-broadcast.html
```diff
@@ -0,0 +1,49 @@
...
+  {% call form_wrapper() %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        {{ form.name(param_extensions={
+          "classes": "govuk-!-width-full",
+          "hint": {"text": "Your recipients will not see this"},
+          "label": {"text": "Title"}
+        }) }}
+      </div>
+      <div class="govuk-grid-column-two-thirds">
```
quis, line 28: Merge the two columns:
```suggestion
```

app/templates/views/broadcast/write-new-broadcast.html
```diff
@@ -0,0 +1,49 @@
...
+{% block maincolumn_content %}
+
+  {{ page_header(
+    'New alert',
+    back_link=url_for('.new_broadcast', service_id=current_service.id),
+  )}}
+
+  {% call form_wrapper() %}
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
```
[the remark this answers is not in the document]
quis: Done in 8e7ff4c

app/templates/views/broadcast/new-broadcast.html
quis: Nice, didn’t realise we had this. I’ve added it in 066aca4

app/templates/views/broadcast/new-broadcast.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block service_page_title %}
+  New alert
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ govukBackLink({ "href": url_for('.broadcast_dashboard', service_id=current_service.id) }) }}
+
+  {% call form_wrapper() %}
```
tombye, line 15: Think this needs a bit of space between the form and the backlink, like in: https://github.com/alphagov/notifications-admin/pull/3771/files#diff-4961c7a5fefe3fac03bd7976ceb59c1b09ed8f0e2c6e7d13d424eddbefaebb26R26
```suggestion
  {% call form_wrapper(class="govuk-!-margin-top-3") %}
```

tombye (approved):
Looks good to me 👍

PR #3753, opened 2020-12-29
Use pip-tools

quis:
This commit copies the approach taken by the Digital Marketplace team in https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/280 to replace the custom `freeze-requirements` make command with [`pip-tools`](https://github.com/jazzband/pip-tools).

I‘ve tried to change as little as possible in this PR to minimise the impact and diff size. I haven’t, for example:
- renamed `requirements_for_test.txt` to something better 
- tidied up some of our bash scripts which would be better in the Makefile


leohemsted (approved):
had a poke around locally, just one small comment. happy to give this one a try

Makefile
```diff
@@ -59,23 +75,9 @@ fix-imports:
...
-	$$(pwd)/venv-freeze/bin/pip freeze -r <(sed '/^--/d' requirements-app.txt) | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
-	rm -rf venv-freeze
-
-.PHONY: test-requirements
-test-requirements:
-	@diff requirements-app.txt requirements.txt | grep '<' \
-	    && { echo "requirements.txt doesn't match requirements-app.txt."; \
-	         echo "Run 'make freeze-requirements' to update."; exit 1; } \
-|| { echo "requirements.txt is up to date"; exit 0; }
+freeze-requirements: requirements-for-test requirements.in requirements_for_test.in
```
leohemsted, line 78: actually, we can get rid of pip-tools from requirements-for-test, and then make this command much quicker and more self-contained by just installing pip-tools within the command and running something like:

```suggestion
freeze-requirements:  ## create static requirements.txt 
	${VIRTUALENV_ROOT}/bin/pip install pip-tools
	${VIRTUALENV_ROOT}/bin/pip-compile requirements.in
	${VIRTUALENV_ROOT}/bin/pip-compile requirements_for_test.in
```

leohemsted (commented):
actually one change we could make to clean this up a little

Makefile
quis: Changed in c34a605

Makefile
```diff
@@ -27,28 +27,44 @@ $(eval export CF_HOME)
...
 	npm install
 	npm rebuild node-sass
-	pip install -r requirements_for_test.txt
 
 .PHONY: generate-version-file
 generate-version-file: ## Generates the app version file
 	@echo -e "__git_commit__ = \"${GIT_COMMIT}\"\n__time__ = \"${DATE}\"" > ${APP_VERSION_FILE}
 
 .PHONY: build
-build: dependencies generate-version-file ## Build project
```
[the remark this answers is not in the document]
quis: Restored this one and added some for the new commands in 083363b

leohemsted (approved):

PR #3754, opened 2020-12-29
Tell browsers to preload fonts

quis:
While looking at [Google’s PageSpeed Insights tool](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.notifications.service.gov.uk%2Ffeatures) as part of the compression work I noticed a suggestion that we preload our font files. The tool suggests this should save 330ms on first page load time:

image

# Why this makes things faster

Our font files are referenced from our CSS. This means that the browser has to download and parse the CSS before it knows where to find the font files. This means the requests happen in sequence. 

We can make the requests happen in parallel by using a `<link>` tag with `rel=preload`. This tells the browser to start downloading the fonts before it’s even started downloading the CSS (the CSS will be the next thing to start downloading, since it’s the next `<link>` element in the head of the HTML). 

Downloading fonts before things like images is important because once the font is downloaded it causes the layout to repaint, and shift everything around. So the page doesn’t feel stable until after the fonts have loaded. 

Google call this [cumulative layout shift](https://web.dev/cls/) which is a score for how much the page moves around. A lower score means a better experience (and, less importantly for us, means the page might rank higher in search results)

# Before

Fonts don’t start to download until all the requests for images have begun.

Screenshot 2020-12-29 at 13 45 09

# After

Fonts start to download immediately before, and in parallel with the CSS.

Screenshot 2020-12-29 at 13 42 59

# Implementation notes

We’re only preloading the WOFF2 fonts because only modern browsers support preload:
image

…and these browsers also all support WOFF2:
image

We set an empty `crossorigin` attribute (which means anonymous-mode) because the preload request needs to match the origin’s CORS mode. See https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#CORS-enabled_fetches for more details.

We set `as=font` because this helps the browser use the correct content security policy, and prioritise which requests to make first.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,4 +1,5 @@
 import os
+import pathlib
 import re
 import urllib
 from datetime import datetime, timedelta, timezone
@@ -202,6 +203,11 @@ def init_app(application):
     application.before_request(load_organisation_before_request)
     application.before_request(request_helper.check_proxy_header_before_request)
 
+    font_paths = [
+        str(item)[len(asset_fingerprinter._filesystem_path):]
+        for item in pathlib.Path(asset_fingerprinter._filesystem_path).glob('fonts/*.woff2')
+    ]
+
     @application.context_processor
     def _attach_current_service():
         return {'current_service': current_service}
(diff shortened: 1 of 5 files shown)


idavidmcdonald (approved):
Nice, thank you for the thorough PR description

PR #3758, opened 2020-12-30
Add server timing headers to measure backend performance 

quis:
This is just a little spike into implementing [`Server-Timing`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) headers. We can decide if this is actually useful.

Lets us use Chrome/Firefox dev tools to see where time is spent on the backend. This is an example for the uploads page:

image


quis:
Closing this for now, can always revisit another time if we’re interested.

PR #3761, opened 2021-01-06
Extract formatters into their own module

quis:
We have lots of functions for converting various types of data into strings to be displayed to the user somewhere.

This pull request collects all these functions into their own module, rather than having them cluttering up `app/__init__.py` or buried amongst various other things that have ended up in `app/utils.py`.

It also converts some of our Jinja macros into formatters, so that they can be used in Python code as well as in Jinja templates.


app/__init__.py
```diff
@@ -821,14 +546,20 @@ def add_template_filters(application):
...
         format_thousands,
         id_safe,
         convert_to_boolean,
         format_list_items,
+        iteration_count,
+        recipient_count,
+        recipient_count_label,
+        message_count_label,
+        message_count,
+        message_count_noun,
```
leohemsted, line 562: only comment is this list is getting a bit hairy now. i note that we can decorate these functions directly

https://flask.palletsprojects.com/en/1.1.x/api/?highlight=add_template_filter#flask.Flask.template_filter

```
@app.template_filter
def message_count_noun(x):
    ....
```

might be worth investigating for being slightly cleaner (and more obvious when viewing the function that it's available in jinja templates). but we'd have to make sure imports still worked cleanly (how do we import application from formatters.py 🤔 )

leohemsted (approved):
great work, and something i'd definitely looked at and done nothing about before

app/__init__.py
quis: Agree that it’s hairy. Don’t think we can get rid of it completely because some of these functions come from `formatters.py` and some are imported from `notifications_utils.formatters`. Will have a think about whether there’s a clean way…

PR #3763, opened 2021-01-06
Add fragment or character counting when editing text message or broadcast templates

quis:
This commit adds an endpoint which counts the number of characters in some template content, and returns a snippet of useful info about how long the message is. It then adds some Javascript to make AJAX requests to this endpoint and display the results on a page.

It is similar to the [character count component from the GOV.UK Design System](https://design-system.service.gov.uk/components/character-count/) but different in that:
- it counts fragments for text messages, not characters 
- it does the count on the server side, not the client side

There’s a surprising amount of complexity in counting the length of messages. We need to do this in Python because it would be too convoluted to re-implement in client side code, let alone ensuring it had parity with its Python equivalent.

I’ve been very careful to break this up into atomic commits, so it’s best reviewed one at a time.

# Text messages

Users sending text messages are sometimes unaware that long messages will cost more, so we’re going to start displaying how much a text message will cost.

image

image

# Emergency alerts

Users sending emergency alerts need to be aware that there’s a character limit, so they can take this into account when planning and writing their messages.

cc-bc

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/831
- [x] https://github.com/alphagov/notifications-utils/pull/833

***

https://www.pivotaltracker.com/story/show/176340621


tombye:
This is really nice and obviously a lot of work but the value is obvious when you use it.

I have a few questions about the the status text:

To me at least, it feels a bit odd to have it after the submit button. Using it with text messages, where it updates (and so becomes noticeable) a lot less often makes it seem like information that should be before you confirm your input.

Is this something we could move to the hint text to the textarea perhaps? It would mean you get that information right before entering the textarea if tabbing through. The hint is also linked to the textarea by `aria-describedby` so screen reader users could hear the status text by using their help command at any time, as well as hearing it when it changes.

I also noticed the design system version has some default text ,before the JS kicks in, explaining the rules in case the JS doesn't load (see https://design-system.service.gov.uk/components/character-count/#how-it-works) I think it would be worth having with our version too. If we moved it to the hint, new templates would have that to explain the rules (even if it just links out to our guidance on this) before entering content, if tabbing through/using a screen reader.

tests/app/main/views/test_templates.py
```diff
@@ -2543,3 +2545,181 @@ def test_should_not_create_broadcast_template_with_placeholders(
...
+    assert normalize_spaces(snippet.text) == expected_message
+
+    if snippet.has_attr('class'):
+        assert snippet['class'] == [expected_class]
+    else:
+        assert expected_class is None
+
+
+@pytest.mark.parametrize('template_type', (
+    'email', 'letter', 'banana',
```
tombye, line 2766: 🍌

app/templates/views/edit-broadcast-template.html
```diff
@@ -27,6 +27,10 @@
       <div class="govuk-grid-column-full">
         <div class="js-stick-at-bottom-when-scrolling">
           {{ page_footer('Save') }}
+          <span class="template-content-count">
```
tombye, line 30: I can see why `div[data-module=update-status]` needs to be a `<div>` we shouldn't put it in a `<span>`.  Could `span.template-content-count` be a `<div>` instead?

app/templates/views/edit-sms-template.html
```diff
@@ -30,6 +30,10 @@
         <div class="govuk-grid-column-full">
           <div class="js-stick-at-bottom-when-scrolling">
             {{ page_footer('Save') }}
+            <span class="template-list-selected-counter">
```
tombye: I can see why `div[data-module=update-status]` needs to be a `<div>` we shouldn't put it in a `<span>`. Could `span.template-list-selected-counter` be a `<div>` instead?

CrystalPea (commented):
Hey, nice work, well done, and so quickly, too :).

When testing, broadcasts look well, but for text messages, the placement looks off, and I'm not sure if it's just my local, or a bigger problem. I did run the bootstrap script.

Screenshot 2021-01-08 at 12 19 16
Screenshot 2021-01-08 at 12 18 45
Screenshot 2021-01-08 at 12 17 27

quis:
@CrystalPea good spot, the text message page was using the wrong CSS class. I’ve fixed it and squashed into 97719a7f4fd8310c78aaf8f7c11b36b147548c21

tests/javascripts/updateStatus.test.js
```diff
@@ -44,7 +44,8 @@ describe('Update content', () => {
     document.body.innerHTML = `
       <form>
         <input type="hidden" name="csrf_token" value="abc123" />
-        <textarea name="template_content" id="template_content">Content of message</textarea>
+        <label for="template_content" id="template-content-label">Template content<label>
+        <textarea name="template_content" id="template_content" aria-described-by="template-content-label">Content of message</textarea>
```
tombye, line 48: This is a bit pedantic, sorry, but could we have the `<textarea>` have its existing `aria-describedby` point to a hint instead? Pointing to the label is bad practice as it would duplicate the accessible name into the accessible description so would be good not to have it here.

tests/javascripts/updateStatus.test.js
```diff
@@ -127,10 +127,44 @@ describe('Update content', () => {
...
 
     expect($.ajax.mock.calls.length).toEqual(2);
 
   });
 
+  test("It should fire only after 150ms of inactivity", () => {
+
+    let textarea = document.getElementById('template_content');
+
+    console.log("T4 start");
```
tombye, line 142: Can we remove these `console.log`s please?

tombye (changes requested):
Added a few comments, as well as [this one separately](https://github.com/alphagov/notifications-admin/pull/3763#issuecomment-756707620) but the rest looks good.

quis:
@tombye to me it doesn’t feel right having the message in the hint, above the textbox. It doesn’t look like something which is dependent on your input:

image

Also by having it in the sticky footer it’s always visible, no matter where in the message you’re scrolled to.

I’ve applied `aria-described-by` as it’s used in the Design System component, so that shouldn’t be a consideration for where we place it visually.

***

Maybe putting it above the button works though, what do you think?

image

tombye:
I really like it above the button, just because it's more like part of what you're agreeing to when you submit. I was also thinking of that type of problem left-aligning helps with where users, when really focused, often miss things that aren't at the start of the line they're on so this is great for that.

tombye:
> I also noticed the design system version has some default text ,before the JS kicks in, explaining the rules in case the JS doesn't load (see https://design-system.service.gov.uk/components/character-count/#how-it-works) I think it would be worth having with our version too. If we moved it to the hint, new templates would have that to explain the rules (even if it just links out to our guidance on this) before entering content, if tabbing through/using a screen reader.

Looking again, I think they put default text in a hint which, I guess, is removed by the JS: https://github.com/alphagov/govuk-frontend/blob/master/package/govuk/components/character-count/template.njk#L29

app/templates/views/edit-sms-template.html
quis: Changed in 55c0e566a

tests/javascripts/updateStatus.test.js
quis: Squashed out

tests/javascripts/updateStatus.test.js
quis: Changed in 6f514ef01

app/templates/views/edit-broadcast-template.html
quis: Changed in 55c0e566a

CrystalPea:
I like putting the count above the button, too, looks neat ☺️

quis:
I’ve moved the message between the textbox and button in e4042af

image

image

tombye:
> I also noticed the design system version has some default text ,before the JS kicks in, explaining the rules in case the JS doesn't load (see https://design-system.service.gov.uk/components/character-count/#how-it-works) I think it would be worth having with our version too. If we moved it to the hint, new templates would have that to explain the rules (even if it just links out to our guidance on this) before entering content, if tabbing through/using a screen reader.

Talking about this offline with @quis we decided that no default text is needed because:
- the rules are complex enough that it would have to link out to the /pricing guidance which isn't great in a hint
- this content is already in the page, at the bottom, so we're not missing out any information by not duplicating it in the hint

quis:
# Stop content length message from making page jumpy

The content length message was making the page jumpy and causing reflows in three ways. This commit addresses each of those ways:

As the user scrolled
---

The footer went from fixed to sticky and the spacing around the message changed. This change in spacing was needed so that the message looked right in both contexts.

I think the best way to resolve this is to not use the sticky footer when editing text message or broadcast templates.

On my 1440×900 screen I can fit a 5 fragment text message, plus the ‘will be charged as 5 text messages’ message, plus the save button.

Our top 10 screen resolutions according to our analytics are:

Position | Resolution | Percentage of users
---------|------------|--------------------
 1       | 1920x1080  | 27.37%
 2       | 1280×720   | 11.07%
 3       | 1366×768   |  8.88%
 4       | 1536×864   |  5.79%
 5       | 1440×900   |  4.52%
 6       | 1600×900   |  3.71%
 7       | 1280×1024  |  3.10%
 8       | 1680×1050  |  2.42%
 9       | 1920×1200  |  2.33%
10       | 2560×1440  | 1.99%

So most users are probably looking at Notify on an external screen.

When the page first loaded
---

The message is empty so takes up no space, then the javascript fires and inserts the message, taking up a line of space.

This is resolved by making the empty message take up space with a non-breaking space character.

When the user first typed
---

We previously didn’t show any message until the user started typing. This meant that, with the above fix, there was a larger than normal empty space between the `<textarea>` and the save button.

This is resolved by always showing the message, even when the user hasn’t typed anything yet.

***

These are design decisions which made sense when the message was displayed along side the button, but we’ve had to change now that the message is above the button.

tombye (approved):

CrystalPea (approved):

PR #3767, opened 2021-01-12
Clear the organisation cache when a service's organisation is changed.

servingUpAces:
The /organisations page was showing the wrong live services count after a change is made to the service organisation.


quis:
Would be good to add a test for this method to make sure it’s clearing the right caches – there are plenty of other cases you could copy in https://github.com/alphagov/notifications-admin/blob/master/tests/app/notify_client/test_organisation_client.py

servingUpAces:
Good idea. Test has been added

idavidmcdonald (approved):

PR #3770, opened 2021-01-15
Fix bug with extend_params function

tombye:
The /organisations/<org_id>/settings/edit-agreement page has a bug where the hints appear on load:

edit-agreement_on_load

but then disappear when the page refreshes.

edit-agreement_on_refresh

I'm pretty sure this is because of a side effect in the `extend_params` function in `app/main/forms.py`.

## Detail of the fix

The radios in this page are `OrganisationAgreementSignedForm.agreement_signed`which inherits from the [RadioField from the WTF library](https://wtforms.readthedocs.io/en/2.3.x/fields/#wtforms.fields.RadioField) but with its rendering rewritten to use [the Radios component from the design system](https://design-system.service.gov.uk/components/radios/) to render HTML.

When `OrganisationAgreementSignedForm.agreement_signed` renders its HTML, it passes a `dict` called `params` to the design system radios component: https://github.com/alphagov/notifications-admin/blob/master/app/main/forms.py#L812

`params` has default contents but `OrganisationAgreementSignedForm.agreement_signed` also has a `param_extensions` attribute for any updates you want to make to `params`. A function called `extend_params` is used to apply any updates from `param_extensions` to `params`, before rendering happens: https://github.com/alphagov/notifications-admin/blob/master/app/main/forms.py#L803

As part of its work, `extend_params` deletes the `'items'` property of `param_extensions`. I believe this is the cause of the bug: https://github.com/alphagov/notifications-admin/blob/master/app/main/forms.py#L647

 I think what is happening is:
1. the page is requested for the first time
2. `OrganisationAgreementSignedForm.agreement_signed.param_extensions['items']` is deleted by `extend_params`
3. `OrganisationAgreementSignedForm.agreement_signed.param_extensions is cached in memory
4. the second request (from a page refresh) happens
5. the hints, which are added to `param_extensions['items']`, don't appear

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -651,9 +651,9 @@ def merge_jsonlike(source, destination):
         return True
 
     def merge_lists(source, destination):
-        last_dest_idx = len(destination) - 1
+        last_src_idx = len(source) - 1
         for idx, item in enumerate(destination):
-            if idx <= last_dest_idx:
+            if idx <= last_src_idx:
                 # assign destination value if can't be merged into source
                 if merge_items(source[idx], destination[idx]) is False:
                     source[idx] = destination[idx]
--- a/app/utils.py
+++ b/app/utils.py
@@ -651,8 +651,13 @@ def merge_jsonlike(source, destination):
         return True
 
     def merge_lists(source, destination):
-        for item in destination:
-            if item not in source:
+        last_dest_idx = len(destination) - 1
+        for idx, item in enumerate(destination):
+            if idx <= last_dest_idx:
+                # assign destination value if can't be merged into source
+                if merge_items(source[idx], destination[idx]) is False:
+                    source[idx] = destination[idx]
+            else:
                 source.append(item)
 
     def merge_dicts(source, destination):
--- a/tests/app/test_utils.py
+++ b/tests/app/test_utils.py
@@ -616,6 +616,8 @@ def test_get_sample_template_returns_template(template_type):
     ({"a": {"b": "c"}}, {"a": {"e": "f"}}, {"a": {"b": "c", "e": "f"}}),
     # same key in both dicts, value is a string, destination supercedes source:
     ({"a": "b"}, {"a": "c"}, {"a": "c"}),
+    # nested dict added to new key of dict, additive behaviour:
+    ({"a": "b"}, {"c": {"d": "e"}}, {"a": "b", "c": {"d": "e"}}),
     # lists with same length but different items, destination supercedes source:
     (["b", "c", "d"], ["b", "e", "f"], ["b", "e", "f"]),
     # lists in dicts behave as top level lists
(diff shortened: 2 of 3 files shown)


klssmith (approved):
This all worked nicely when I tried it 🎉

You could consider adding an extra comment or two, but that's optional. At first glance, I found it a bit confusing to see that we delete items `del extensions['items']` then a couple of lines later add them back `extensions['items'] = items`

tombye:
> This all worked nicely when I tried it 🎉
> 
> You could consider adding an extra comment or two, but that's optional. At first glance, I found it a bit confusing to see that we delete items `del extensions['items']` then a couple of lines later add them back `extensions['items'] = items`

This comment made me realise it isn't clear and that, actually, the reason is that we want to use `merge_jsonlike` to merge `params` and `param_extensions` (passed in as `extensions`) but `merge_jsonlike` doesn't merge lists correctly.

I realised that if we just make `merge_jsonlike` merge lists correctly, we can use it for everything `extend_params` does.

I've added a commit to do that, done a lot of checking that it didn't break anything and expanded the tests for `merge_jsonlike`. The commit message explains those things in more detail but I think it should be ready to go.

app/utils.py
```diff
@@ -651,8 +651,13 @@ def merge_items(source_item, destination_item):
...
     def merge_lists(source, destination):
-        for item in destination:
-            if item not in source:
+        last_dest_idx = len(destination) - 1
+        for idx, item in enumerate(destination):
+            if idx <= last_dest_idx:
+                # assign destination value if can't be merged into source
+                if merge_items(source[idx], destination[idx]) is False:
+                    source[idx] = destination[idx]
+            else:
```
klssmith, line 660: This else statement won't get reached - `idx` is not ever greater than `last_dest_idx`.

tombye: Good spot! That code doesn't match my original intentions for it. I've pushed a fix and a testcase as this wasn't caught by the tests either.

tests/app/test_utils.py
```diff
@@ -614,18 +614,30 @@ def test_get_sample_template_returns_template(template_type):
...
+    ({"a": "b"}, {"c": {"d": "e"}}, {"a": "b", "c": {"d": "e"}}),
+    # lists with same length but different items, destination supercedes source:
+    (["b", "c", "d"], ["b", "e", "f"], ["b", "e", "f"]),
     # lists in dicts behave as top level lists
-    ({"a": ["b", "c", "d"]}, {"a": ["b", "e", "f"]}, {"a": ["b",  "c", "d", "e", "f"]}),
-    # lists with same string in both result in a list of unique values
-    (["a", "b", "c", "d"], ["d", "e", "f"], ["a", "b", "c", "d", "e", "f"]),
+    ({"a": ["b", "c", "d"]}, {"a": ["b", "e", "f"]}, {"a": ["b", "e", "f"]}),
+    # lists with same string in both, at different positions, result in duplicates keeping their positions
+    (["a", "b", "c", "d"], ["d", "e", "f"], ["d", "e", "f", "d"]),
```
klssmith, line 626: In this test case of `["d", "e", "f", "d"]`, if the params end up with a duplicated item does that mean that something is going to display twice on a form widget, or is that ok?

tombye: Yes, at least with the data we're sending into the GOVUK Frontend components. In that data, lists are used to configure sequences of HTML, like the radio buttons in a radios component. The data and the radio button it's for are related by position in the list rather than the value of either item.

That testcases doesn't show this too clearly because it compares primitive values, whereas lists in component data are always made up of dicts. I think it's useful to have these tests, so the basic principles are clear but they also make it a bit harder to relate to the actual usage.

klssmith: 👍 Thanks for explaining, I agree that the tests are useful.

klssmith (approved):

PR #3772, opened 2021-01-15
Add service notes

CrystalPea:
Add service notes section to Service settings

EMPTY NOTES:
Screenshot 2021-01-15 at 16 26 57

ADDING SOME NOTES:
Screenshot 2021-01-15 at 16 26 42

NOTES PRESENT:
Screenshot 2021-01-15 at 16 26 31

LONG NOTES:
Screenshot 2021-01-15 at 16 29 23


app/main/views/service_settings.py
```diff
@@ -1188,6 +1189,34 @@ def edit_data_retention(service_id, data_retention_id):
...
+
+        if form.notes.data == current_service.notes:
+            return redirect(url_for('.service_settings', service_id=service_id))
+
+        try:
+            current_service.update(
+                notes=form.notes.data
+            )
+        except HTTPError as e:
+            raise e
```
quis, line 1210: I’m not sure you need the `try`/`except` here. Won’t the default error handling catch this and show a ‘technical difficulties’ page?

tests/app/main/views/test_service_settings.py
```diff
@@ -5249,3 +5249,14 @@ def test_service_settings_links_to_edit_service_notes_page_for_platform_admins(
...
+
+def test_view_edit_service_notes(
+        platform_admin_client,
+        service_one,
+
+):
+    response = platform_admin_client.get(url_for('main.edit_service_notes', service_id=SERVICE_ONE_ID))
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+    assert page.select_one('h1').text == "Edit service notes"
+    assert page.find('label', class_="form-label").text.strip() == "Notes"
```
quis, line 5262: Would be good to add a test here to check that the textbox exists, and that it has the right `name` attribute

app/main/views/service_settings.py
```diff
@@ -1188,6 +1189,34 @@ def edit_data_retention(service_id, data_retention_id):
     )
 
 
+@main.route("/services/<uuid:service_id>/notes", methods=['GET', 'POST'])
+@user_has_permissions('manage_service')
+def edit_service_notes(service_id):
+    form = EditServiceNotesForm()
+
+    if request.method == 'GET':
+        form.notes.data = current_service.notes
```
quis, line 1198: You can achieve the same thing by passing `notes` as an argument to the form’s constructor:
```suggestion
    form = EditServiceNotesForm(notes=current_service.notes)
```
WTForms will make sure the `POST` data doesn’t get overridden with the existing value.

app/templates/components/table.html
```diff
@@ -85,8 +86,8 @@
   </td>
 {%- endmacro %}
 
-{% macro text_field(text, status='', truncate=false) -%}
-  {% call field(status=status) %}
+{% macro text_field(text, status='', truncate=false, wrap=False) -%}
```
quis, line 89: Nice attention to detail adding this option 👍🏻

CrystalPea: Thanks 😊

tests/app/main/views/test_service_settings.py
CrystalPea: Good point, I will add that ☺️

app/main/views/service_settings.py
CrystalPea: True, it should do.

quis (approved):

PR #3774, opened 2021-01-19
Basic accessibility fixes for radio select

tombye:
The Digital Accessibility Centre (DAC) raised some issues against our component for adding/moving templates and folders:

https://www.pivotaltracker.com/story/show/174440877

The work we did to fix those issues in https://github.com/alphagov/notifications-admin/pull/3664 included problems with:
- the semantics of buttons that open/close parts of the component (for which the DAC audit suggested some fixes)
- what happens to focus when the current buttons are replaced with other content

The radioSelect component also opens/closes parts of its content with buttons so will share these problems:

image

image

We're going to get them tested by DAC next week to check that but, in the meantime, this adds fixes for the button semantics we know work and a small improvement to the focus management.

This is part of this story to improve the accessibility of collapsible components:

https://www.pivotaltracker.com/story/show/175477156

## Notes for reviewers

These changes are best reviewed commit-by-commit.

It is also useful to check the branch out and try the old and new versions of this component beforehand.

### Improvements to focus management

When you select a day to send your message, ie. 'Tomorrow', radios for times in that day replace the buttons for different days:

image

We currently move focus to the default 'Now' radio. This doesn't inform non-visual users that the new radios have been added. Focusing the first one tells them the number and type of radios in the component have changed:

- [before these changes](https://drive.google.com/file/d/1zctkqc_6-x1QgGce8yJOJ-UgH8qEPX7g/view?usp=sharing)
- [after these changes](https://drive.google.com/file/d/16IwsZmVcg_mu-n3rfOXLFBtVR1awMezx/view?usp=sharing)

I think it also makes it better for keyboard-only users. They also benefit from their focus being moved to the first time rather than back to 'now'.

We could add more information but I'd like to get this, simpler version tested first to prevent adding anything unnecessary.

### DIFF
--- a/app/assets/javascripts/radioSelect.js
+++ b/app/assets/javascripts/radioSelect.js
@@ -5,6 +5,7 @@
   var Modules = global.GOVUK.Modules;
   var Hogan = global.Hogan;
 
+  // Object holding all the states for the component's HTML
   let states = {
     'initial': Hogan.compile(`
       {{#showNowAsDefault}}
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):
Ready to go when you make those changes we spoke about. Feel free to let me know when you make them so I can take a look ☺️

PR #3778, opened 2021-01-25
Fix bug with stale template cache

servingUpAces:
Update _delete_template_cache_for_service to delete all template version cache and not just the one ending in "None"

Also updated any method that was calling @cache.delete('service-{service_id}-template-None") so all template versions are deleted as well.


app/notify_client/service_api_client.py
```diff
@@ -9,10 +9,10 @@ class ServiceAPIClient(NotifyAdminAPIClient):
...
         templates_for_service = self.get_service_templates(service_id)['data']
         if templates_for_service:
-            redis_client.delete(*[
-                f"service-{service_id}-template-{x['id']}-version-None"
-                for x in templates_for_service
-            ])
+            print("len templates", len(templates_for_service))
+            for x in templates_for_service:
+                print(f"service-{service_id}-template-{x['id']}-version-*")
+                redis_client.delete_cache_keys_by_pattern(f"service-{service_id}-template-{x['id']}-version-*")
```
quis, line 15: I think you could avoid the loop and let Redis do the legwork with this instead:
```python
redis_client.delete_cache_keys_by_pattern(f"service-{service_id}-template-*")
```

See https://redis.io/commands/keys for the supported patterns.

tests/app/notify_client/test_service_api_client.py
```diff
@@ -541,14 +541,14 @@ def test_client_deletes_service_template_cache_when_service_is_updated(
...
 
-    assert len(mock_redis_delete.call_args_list) == 2
-    assert mock_redis_delete.call_args_list[1] == call('service-{}'.format(SERVICE_ONE_ID))
+    assert len(mock_redis_delete.call_args_list) == 1
+    assert mock_redis_delete.call_args_list[0] == call('service-{}'.format(SERVICE_ONE_ID))
 
-    templates_to_delete = mock_redis_delete.call_args_list[0][0]
-    assert len(templates_to_delete) == 6
-    for template_key in templates_to_delete:
+    assert len(mock_redis_delete_by_pattern) == 6
```
quis, line 551: Test is failing because you’re trying to assert the length of the mock, not its calls
```suggestion
    assert len(mock_redis_delete_by_pattern.call_args_list) == 6
```

app/notify_client/service_api_client.py
servingUpAces: Good idea - thanks

tests/app/notify_client/test_service_api_client.py
servingUpAces: doh!
Thanks

app/notify_client/service_api_client.py
```diff
@@ -9,10 +9,7 @@ class ServiceAPIClient(NotifyAdminAPIClient):
     def _delete_template_cache_for_service(self, service_id):
         templates_for_service = self.get_service_templates(service_id)['data']
         if templates_for_service:
```
quis, line 11: Sorry, should have made it clearer in my first comment that you also don’t need the API call. 

I think this is a hangover from when we cached templates without `service_id` in the key, so the only way to know which templates belonged to a service was to fetch them.

```suggestion
```

servingUpAces: I should have realised that... fixing now

quis (approved):

PR #3779, opened 2021-01-25
Show billing details for service

CrystalPea:
Allows platform admins to view and change billing details.

### No billing details yet: 
Screenshot 2021-01-25 at 15 24 18

### Change billing details page:
Screenshot 2021-01-27 at 17 37 24

### with billing details:
Screenshot 2021-01-25 at 15 23 26


app/main/forms.py
```diff
@@ -1710,6 +1710,14 @@ class EditServiceNotesForm(StripWhitespaceForm):
     notes = TextAreaField(validators=[])
 
 
+class ServiceBillingDetailsForm(StripWhitespaceForm):
+    billing_contact_email_addresses = GovukTextInputField('Billing contact email addresses')
```
servingUpAces, line 1714: Do we need the labels to include `Billing` when we are on the `Billing Details` page?

app/templates/views/service-settings/edit-service-billing-details.html
```diff
@@ -0,0 +1,35 @@
...
+  {{ page_header(
+    'Change billing details',
+    back_link=url_for('main.service_settings', service_id=current_service.id)
+  ) }}
+
+  {% call form_wrapper() %}
+    {{ form.billing_contact_names(param_extensions={"classes": "govuk-!-width-full"}) }}
+    {{ form.billing_contact_email_addresses(param_extensions={"classes": "govuk-!-width-full"}) }}
+    {{ form.billing_reference(param_extensions={"classes": "govuk-!-width-full"}) }}
+    {{ form.purchase_order_number(param_extensions={"classes": "govuk-!-width-full"}) }}
```
quis, line 24: You can make these long numbers a bit more readable with some extra letter spacing (see https://github.com/alphagov/notifications-admin/pull/1545 for screenshots)
```suggestion
  <div class="extra-tracking">
      {{ form.billing_reference(param_extensions={"classes": "govuk-!-width-full"}) }}
      {{ form.purchase_order_number(param_extensions={"classes": "govuk-!-width-full"}) }}
  </div>
```

app/templates/views/service-settings/edit-service-billing-details.html
```diff
@@ -0,0 +1,35 @@
...
+    {{ textbox(
+      form.notes,
+      rows=4,
+      width='1-1',
+      autofocus=True,
```
quis, line 29: You probably don’t want focus jumping to the last field on the page
```suggestion
      autofocus=False,
```

app/models/service.py
```diff
@@ -30,25 +30,29 @@ class Service(JSONModel):
...
         'research_mode',
         'service_callback_api',
         'volume_email',
         'volume_sms',
         'volume_letter',
-        'consent_to_research',
-        'count_as_live',
-        'go_live_user',
-        'go_live_at',
-        'allowed_broadcast_provider',
```
quis: Not something you need to change in this PR, but if you’re going to re-sort and add lines of code it’s better to do that in two separate commits.

app/main/forms.py
CrystalPea: Hmm not sure, maybe one for @karlchillmaid ?

app/templates/views/service-settings/edit-service-billing-details.html
CrystalPea: Ohh, cool! Why is it call tracking, not spacing, tho? 🤔

app/models/service.py
CrystalPea: ok, will do so in the future 👍🏼

app/main/forms.py
CrystalPea: Karl confirms we should lose them 👍🏼

app/templates/views/service-settings/edit-service-billing-details.html
quis: They both mean the same thing. Maybe we should change it to `extra-letter-spacing` to be more self explanatory for people not familiar with typography terminology.

quis (approved):

app/templates/views/service-settings/edit-service-billing-details.html
CrystalPea: Hmmm... that's a good idea, `extra-letter-spacing` sounds self-explanatory :3

CrystalPea: I can add it in my org billing details PR as a separate commit if you want?

quis: Let’s do a separate PR for that

CrystalPea: Sounds good 👍🏼

PR #3780, opened 2021-01-26
Display areas that aren’t in the library

quis:
For broadcasts that aren’t using our library of data we can get the names of the areas and the simplified polygon from the `BroadcastMessage` instead.

image

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -45,6 +45,10 @@ class BroadcastMessage(JSONModel):
             return True
         if not self.starts_at and other.starts_at:
             return False
+        if self.updated_at and not other.updated_at:
+            return self.updated_at < other.created_at
+        if not self.updated_at and other.updated_at:
+            return self.created_at < other.updated_at
         return self.updated_at < other.updated_at
 
     @classmethod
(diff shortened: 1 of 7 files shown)


quis:
Now works for broadcast messages created from the API too:

image

app/models/broadcast_message.py
```diff
@@ -74,7 +78,12 @@ def from_id(cls, broadcast_message_id, *, service_id):
 
     @property
     def areas(self):
-        return self.get_areas(areas=self._dict['areas'])
+        return self.get_areas(
+            areas=self._dict['areas']
+        ) or CustomBroadcastAreas(
+            areas=self._dict['areas'],
+            polygons=self._dict['simple_polygons'],
+        )
```
leohemsted, line 86: can we do a sanity check at this point that the length of the list returned is the same as the size of the `self._dict`? that'll catch the (admittedly, very edge casey) scenario where someone's included one place we recognise and one place we don't recognise. not sure what we should do in that situation, but i'd rather we raise an exception than just silently discard the unrecognised areas

quis: Good idea, added a check in 926ada2f2

leohemsted (approved):

PR #3781, opened 2021-01-26
Add GOVUK radios [part 7]

tombye:
Part 6 of migrating our radio buttons to use the [GOV.UK Frontend radios component](https://design-system.service.gov.uk/components/radios/).

https://www.pivotaltracker.com/story/show/170030262

This pull request updates the radioSelect component (in its base HTML and its JavaScript) to use the GOVUK radios component.

radio_select

Includes changing the code so that the radios aren't split into two columns in the base HTML. This layout is now added by the JS.


quis (approved):
Tested locally 👍🏻

PR #3785, opened 2021-02-02
Add service history for platform admins

idavidmcdonald:
We have a service history page that exists but people may not know
about. The only way to get to it is by knowing the URL as we don't
appear to link to it anywhere in the admin app.

This adds an easy link just for platform admins as this page may be of
some use I believe. There is a chance that the service history is not
super up to date and could see further improvements but it might still
be helpful when doing support type things.

@quis I imagine you might have opinions on whether this is a good addition to the admin app for us?

image

### DIFF
--- a/app/navigation.py
+++ b/app/navigation.py
@@ -532,6 +532,9 @@ class MainNavigation(Navigation):
             'guest_list',
             'old_guest_list',
         },
+        'history': {
+            'history',
+        },
     }
 
     exclude = {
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
Closing as @quis suggested an alternative approach. He was cautious about adding an extra item to the navigation like this and would prefer a link from the settings page.

Will one day get back to working on this again...

PR #3786, opened 2021-02-04
Remove a link from using-notify page

klssmith:
This stops the 'inset text' linking to the design system on the
'edit-and-format-messages' page. The link has been removed to avoid
confusion - someone thought they needed to use the design system
code in order to create inset text in templates.


quis (approved):
Nice work tracking down the source of the confusion 👍🏻

PR #3791, opened 2021-02-09
Allow broadcasts which have no created_at to be compared

klssmith:
This adds to the `__le__` method on the `BroadcastMessage` class to
allow two BroadcastMessages with no `updated_at` to be compared.
Previously, the method expected at least one message to have
`updated_at` set, but this is not necessarily the case.


quis (approved):
This fixes the issue so let’s get it merged 👍🏻 

Maybe let’s think about how we could refactor and test this better so it’s less confusing to understand what it’s doing.

PR #3796, opened 2021-02-11
Handle exception is org name already exists

benthorner:
https://www.pivotaltracker.com/story/show/176804067

Please see the commit for more details on the approach. Note 
the message matches that shown when changing the name of 
an existing organisation.

## After

Screenshot 2021-02-11 at 12 17 49

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -62,10 +62,17 @@ def add_organisation():
     form = NewOrganisationForm()
 
     if form.validate_on_submit():
-        return redirect(url_for(
-            '.organisation_settings',
-            org_id=Organisation.create_from_form(form).id,
-        ))
+        try:
+            return redirect(url_for(
+                '.organisation_settings',
+                org_id=Organisation.create_from_form(form).id,
+            ))
+        except HTTPError as e:
+            msg = 'Organisation name already exists'
+            if e.status_code == 400 and msg in e.message:
+                form.name.errors.append("This organisation name is already in use")
+            else:
+                raise e
 
     return render_template(
         'views/organisations/add-organisation.html',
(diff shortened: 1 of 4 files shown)


CrystalPea (approved):
Looks good.

Optional: since this looks like a fixture we will not re-use, we may consider simplifying it and putting it in the test? I think this test does that pretty well: https://github.com/alphagov/notifications-admin/blob/master/tests/app/main/views/test_send.py#L4018

benthorner:
> since this looks like a fixture we will not re-use, we may consider simplifying it and putting it in the test?

Sure, I've done that. My original reasoning was in the commit (see below), but I've amended it to do the mock inline. Personally I prefer keeping all the test code inside the test, as it's easier to follow that way 👍.

> The new fixture is based on one for services [7]. Unlike services, organisation tests call the mocker directly for the "create" API [8], so there wasn't an obvious place to put the new fixture. Given the complexity of the mock, it seems better for it to be a fixture.

I've added an extra commit to make this consistent with Services. Could you take a quick look?

CrystalPea:
@benthorner Looks good, I think you will have to update some dependencies to merge though.

I'm curious, do you find the way of mocking error in this PR better than here: https://github.com/alphagov/notifications-admin/blob/master/tests/app/main/views/test_send.py#L4018 or is it a preference? (asking to learn)

benthorner:
> I'm curious, do you find the way of mocking error in this PR better than here

I'm new to Python, so I can't say with much confidence. From the docs of unittest.mock: "Mock is a flexible mock object intended to replace the use of stubs and test doubles throughout your code."

I had a go changing this to use a Mock, but it looks like it's hard (see [here](https://stackoverflow.com/questions/14771039/how-to-mock-an-exception-call-in-python)), which isn't great for a framework that's supposed to mock things! The commit / PR ([here](https://github.com/alphagov/notifications-admin/pull/1391)) doesn't give the reasoning, but I suspect it's that.

PR #3800, opened 2021-02-12
Reduce default broadcast expiry time

quis:
We don’t think we need to broadcast longer than 4 hours to validate that the system is working.

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -220,7 +220,7 @@ class BroadcastMessage(JSONModel):
         self._update(
             starts_at=datetime.utcnow().isoformat(),
             finishes_at=(
-                datetime.utcnow() + timedelta(hours=23, minutes=59)
+                datetime.utcnow() + timedelta(hours=4, minutes=0)
             ).isoformat(),
         )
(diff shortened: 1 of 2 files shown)


app/models/broadcast_message.py
```diff
@@ -220,7 +220,7 @@ def approve_broadcast(self):
         self._update(
             starts_at=datetime.utcnow().isoformat(),
             finishes_at=(
-                datetime.utcnow() + timedelta(hours=23, minutes=59)
+                datetime.utcnow() + timedelta(hours=4, minutes=0)
```
idavidmcdonald, line 223: Is this a temporary thing to reduce it to 4 hours for our demos? And then we will change it back to 24 hours after? Or is this a permanent thing and if so, why?

quis: 4 hours is enough to cover the demos, but no more. What we change it to after this will depend on what we learn from the demos, and what other countries have learned.

idavidmcdonald (approved):
Seems fine.

Note, please mention this change in the broadcast channel or in the tech catch up meeting as worth people knowing about this change I think.

PR #3801, opened 2021-02-12
Show rejected broadcasts on ‘Previous alerts’ page

quis:
Two reasons to not hide rejected broadcasts:
- if a broadcast was rejected by mistake then it’s useful to have an audit of who did that 
- it means you can still see old broadcasts without having to leave them in `pending-approval`, which is dangerous because they might accidentally be approved

Before | After 
---|---
image | image

### DIFF
--- a/app/templates/views/broadcast/partials/dashboard-table.html
+++ b/app/templates/views/broadcast/partials/dashboard-table.html
@@ -18,6 +18,10 @@
             <p class="govuk-body govuk-!-margin-bottom-0 govuk-hint">
               Waiting for approval
             </p>
+          {% elif item.status == 'rejected' %}
+            <p class="govuk-body govuk-!-margin-bottom-0 govuk-hint">
+              Rejected {{ item.updated_at|format_datetime_relative }}
+            </p>
           {% elif item.status == 'broadcasting' %}
             <p class="govuk-body govuk-!-margin-bottom-0 live-broadcast">
               Live since {{ item.starts_at|format_datetime_relative }}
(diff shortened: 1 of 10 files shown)


app/templates/views/broadcast/partials/dashboard-table.html
```diff
@@ -18,6 +18,10 @@ <h2>
             <p class="govuk-body govuk-!-margin-bottom-0 govuk-hint">
               Waiting for approval
             </p>
+          {% elif item.status == 'rejected' %}
+            <p class="govuk-body govuk-!-margin-bottom-0 govuk-hint">
+              Rejected {{ item.updated_at|format_datetime_relative }}
+            </p>
```
leohemsted, line 24: i reckon we might want to distinguish these more visually from alerts that actually went out

benthorner (approved):
Nice one, makes sense to show the rejected ones 👍.

A couple of tangential things I noticed that may be relevant for follow-up work:

- I don't think we show `technical-failure` broadcasts anywhere. This could cause a similar visibility problem. Is it worth adding that status [here](https://github.com/alphagov/notifications-admin/blob/856a667dcf5dc3246536cab33cfac577ef6c90d4/app/main/views/broadcast.py#L69), perhaps?

- (Very tangential) The "previous alerts" view is getting pretty lengthy for test services on preview. Since testing will be an ongoing use case for broadcast services, is it worth looking at data retention and/or pagination for these pages?

app/templates/views/broadcast/partials/dashboard-table.html
CrystalPea: I agree with @leohemsted here, it would be good to have a separate section for them maybe?

quis:
I’ve taken @CrystalPea’s suggestion and split the page into…

# Previous alerts

Mostly as it was before this change:

image


# Rejected alerts 

Only shows rejected alerts, could be limited in the future to only show the last _n_ days:

image

CrystalPea:
that's very neat @quis ✨

CrystalPea (approved):

PR #3805, opened 2021-02-16
Fix issues with dependencies and upgrade them

benthorner:
Please see the commit messages for more info.


.pyup.yml
```diff
@@ -5,4 +5,4 @@ schedule: "every week on monday"
 search: False
 requirements:
   - requirements.in
-  - requirements_for_test.in
+  - requirements_for_text.txt
```
CrystalPea, line 8: shouldn't it be `requirements_for_test.txt` ?

benthorner: Oops (fixed). Nothing to see here 😇.

Makefile
```diff
@@ -78,7 +78,6 @@ fix-imports:
 freeze-requirements: ## create static requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install pip-tools
 	${VIRTUALENV_ROOT}/bin/pip-compile requirements.in
-	${VIRTUALENV_ROOT}/bin/pip-compile requirements_for_test.in
```
CrystalPea, line 81: why are we getting rid of this? 

As far as I'm aware we list basic dependencies in `.in` files and create more complex `.txt` files with subdependencies through `make freeze-requirements`. Are we changing that flow? :o

benthorner: Yes, I think we need to. Just checking: have you seen the detailed notes I've written in the commits? They explain why we need to do this, and that it shouldn't be a problem for this specific file.

servingUpAces (approved):
I like this change, it means we are unit testing against the same version of dependencies as the deployed version.

Makefile
CrystalPea: Ahh, ok, makes sense, sorry I did not see the messages before.

benthorner:
Tested on Staging, then Preview (with functional tests). I'm going to merge now and then monitor the traffic in production, in case there's an unexpected edge case. The only major bump now is greenlet.

PR #3806, opened 2021-02-17
✨ Add validation to prevent use of blocked terms within SMS Sender IDs

dvjones89:
To help avoid abuse of the CAST Notify service, as well as protect our
accounts with backend providers like `firetext`, we want to apply some
restrictions on the Sender IDs that users can apply to our outbound SMS
and emails.

This PR adds support to fetch a blocklist from a (newly added) API
endpoint, as well as introducing a new `WTForms` validator that'll raise
a validation error if/when a blocked phrase features within the SMS
sender that has been submitted by the user.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -52,6 +52,7 @@ from app.main.validators import (
     OnlySMSCharacters,
     ValidEmail,
     ValidGovEmail,
+    SenderBlocklistValidator,
 )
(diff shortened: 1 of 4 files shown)


dvjones89:
Oops, I intended to raise this against our own fork, rather than the official alphagov repository. Apologies for the confusion!

PR #3811, opened 2021-02-22
Simplify README and consolidate scripts

benthorner:
Partly addresses: alphagov/notifications-manuals#9

Please see the commit messages for more details. Many of the
commits are replicas of the equivalent PR in the API app [1].
I've tried to push anything new into the last commits, so it's
easier to tweak if needed.

[1]: https://github.com/alphagov/notifications-api/pull/3150


idavidmcdonald (approved):
Thanks, seems good and more succinct.

I'm not as sure about the removing of the commands to do with the node/frontend but I think as long as Concourse is happy, we can add them back if we find out we do need them to be separate

benthorner:
> I'm not as sure about the removing of the commands to do with the node/frontend

Yeah, I was also wondering if someone might be using these and not said. I'll hold off merging for a bit and ask in Slack.

benthorner:
Update: seems like no one does, so let's try without them...

PR #3812, opened 2021-02-23
Use the GOVUK Frontend inset text component instead of <aside>

klssmith:
We use the `<aside>` tag to wrap our indented text in the 'formatting' guidance you see when creating a template. This is invalid HTML. We should use the GOVUK Frontend inset text component instead.

Note, there are more places that will need to be updated to use the inset-text macro - this PR only changes the template formatting pages since these were flagged by the accessibility audit.

[Pivotal story](https://www.pivotaltracker.com/story/show/176905338)

### DIFF
--- a/app/templates/partials/templates/guidance-formatting-letters.html
+++ b/app/templates/partials/templates/guidance-formatting-letters.html
@@ -4,33 +4,27 @@
 <p class="govuk-body">
   To put a heading in your template, use a hash:
 </p>
-<div>
-  {{ govukInsetText({
-    "text": "# This is a heading",
-    "classes": "govuk-!-margin-top-0 govuk-!-margin-bottom-3"})
-  }}
-</div>
+{{ govukInsetText({
+  "text": "# This is a heading",
+  "classes": "govuk-!-margin-top-0 govuk-!-margin-bottom-3"})
+}}
 <p class="govuk-body">
   To make bullet points, use asterisks:
 </p>
-<div>
-  {{ govukInsetText({
-    "html": "* point 1<br/>
-             * point 2<br/>
-             * point 3<br/>",
-    "classes": "govuk-!-margin-top-0 govuk-!-margin-bottom-3"})
-  }}
-</div>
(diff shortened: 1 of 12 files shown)


app/templates/partials/templates/guidance-formatting-letters.html
```diff
@@ -1,31 +1,36 @@
...
+
 <h2 class="heading-medium">Formatting</h2>
 <p class="govuk-body">
   To put a heading in your template, use a hash:
 </p>
-<div class="panel panel-border-wide">
-  <p class="govuk-body">
-    # This is a heading
-  </p>
+<div>
```
tombye, line 7: I don't think we need these extra `<div>`s here or on the other ones as the component adds its own.

app/templates/partials/templates/guidance-formatting.html
```diff
@@ -1,41 +1,37 @@
...
+
 <h2 class="heading-medium">Formatting</h2>
 <p class="bottom-gutter-1-3">
   To put a title in your template, use a hash:
 </p>
-<div class="panel panel-border-wide bottom-gutter">
-  <p class="govuk-body">
-    # This is a title
-  </p>
+<div class="bottom-gutter">
```
tombye, line 7: Can we replace `bottom-gutter` with `govuk-!-margin-bottom-6`? We'll replace all instances at some point so it would be good to do these ones now when we're making changes anyway.

I checked [the CSS for `bottom-gutter`](https://github.com/alphagov/notifications-admin/blob/master/app/assets/stylesheets/_grids.scss#L54) and both of the things it adds (margin-bottom, clearfix) are also in the inset text CSS so we should be good.

tombye (changes requested):
I've added a few comments but this looks good overall. Happy to approve once they've been addressed.

klssmith:
@tombye thanks for the comments, I've made both those changes in 7928600

tombye (approved):
Looks good now 👍

PR #3813, opened 2021-02-23
Rerun freeze-requirements

idavidmcdonald:
This importantly upgrades pyyaml which has a security bug in 5.3.1

Note, strangely, I had to delete the requirements.txt file and rerun to
get these requirements to upgrade, otherwise it kept them in place
(maybe some piptools caching stuff not calculating things if it doesn't
think it has been asked to change them).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ ago==0.0.93
     # via -r requirements.in
 awscli-cwlogs==1.4.6
     # via -r requirements.in
-awscli==1.19.7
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Looks good to me - I was waiting for PyUp to come along, but thinking on it some more, it won't because PyYAML is not a top-level requirement.

I didn't get the caching behaviour you mentioned when I tried this last, so let's put it down to a one-off until it happens again 🤞.

PR #3815, opened 2021-02-23
allow caseworkers to view letter previews

leohemsted:
they can already view notifications page, but the png and pdf letter previews just 403 for them currently

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -197,7 +197,7 @@ def get_preview_error_image():
 
 
 @main.route("/services/<uuid:service_id>/notification/<uuid:notification_id>.<letter_file_extension:filetype>")
-@user_has_permissions('view_activity')
+@user_has_permissions('view_activity', 'send_messages')
 def view_letter_notification_as_preview(
     service_id, notification_id, filetype, with_metadata=False
 ):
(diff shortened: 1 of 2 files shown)


app/main/views/notifications.py
```diff
@@ -197,7 +197,7 @@ def get_preview_error_image():
 
 
 @main.route("/services/<uuid:service_id>/notification/<uuid:notification_id>.<letter_file_extension:filetype>")
-@user_has_permissions('view_activity')
+@user_has_permissions('view_activity', 'send_messages')
 def view_letter_notification_as_preview(
```
benthorner, line 201: Note to self: this corresponds with the "Download as PDF" link on the final page of sending a notification ([here](https://github.com/alphagov/notifications-admin/blob/0776a959420842e19d6dbcd0862d98bd1c991da5/app/templates/views/notifications/notification.html#L93)). I couldn't find where the `.png` link might be shown, but that's not relevant here.

benthorner (approved):
Makes sense to me 👍.

app/main/views/notifications.py
leohemsted: the png isn't shown as a link - rather, it's included as the src in an img tag so we can show it on the screen.

we generate the url here: https://github.com/alphagov/notifications-admin/blob/master/app/main/views/notifications.py#L89-L94

and it gets passed through to utils, which renders the html shown here: https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/jinja_templates/letter_image_template.jinja2#L8

benthorner: Ah yes I see, thanks for showing me (I didn't think to search by endpoint).

PR #3816, opened 2021-02-25
Fix minor accessibility bugs from recent audit

klssmith:
This fixes the incorrect use of aria, non-descriptive links and mandatory fields issues that were identified by the recent Digital Accessibility Centre Audit.

See individual commits and [Pivotal story](https://www.pivotaltracker.com/story/show/176905338) for details.

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -351,7 +351,7 @@ def _get_job_counts(job):
         ) for label, query_param, count in [
             [
                 f'''total<span class="govuk-visually-hidden">
-                 {message_count_noun(job.notification_count, job_type)}</span>''',
+                 {"text message" if job_type == "sms" else job_type}s</span>''',
                 '',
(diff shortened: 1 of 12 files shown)


app/main/views/jobs.py
```diff
@@ -349,19 +350,27 @@ def _get_job_counts(job):
             count
         ) for label, query_param, count in [
             [
-                'total', '',
+                f'''total<span class="govuk-visually-hidden">
+                 {message_count_noun(job.notification_count, job_type)}</span>''',
```
tombye, line 354: Having listened to this, I think that the job.type needs to always for plural to follow 'total'.

I mean that '1 total email' sounds a bit odd compared to '1 total emails', I think because of 'total'.

tombye (approved):
I put one, very minor, comment in there but this looks good to me otherwise 👍

PR #3819, opened 2021-02-26
Simplify tests for excluded navigation endpoints

benthorner:
Previously each navigation class had a list of endpoint to "exclude",
which was only used in tests to ensure that all endpoints in the app
were covered: either they are present in navigation, or excluded.

However, over time the "exclude" lists have grown long and repetitive,
and maintaining each of them individually adds extra work [1][2]. This
switches to a more DRY approach, where the list of excluded endpoints
is defined once, close to the single point of use in the test.

Note the resulting test is _slightly_ less prescriptive, as it will now
pass if an endpoint exist one in navigation, even if it should also
exist in another. This seems a reasonable compromise.

[1]: https://github.com/alphagov/notifications-admin/pull/3788/files#r572809972
[2]: https://github.com/alphagov/notifications-admin/pull/3794/files#diff-39387df3a9f89b313976957e7b5457be20deab1017b2d895541b142b957f1972


app/navigation.py
```diff
@@ -24,19 +23,13 @@ def endpoints_with_navigation(self):
...
-        ) + ('static', 'status.show_status', 'metrics')
-
     def is_selected(self, navigation_item):
         if request.endpoint in self.mapping[navigation_item]:
             return " " + self.selected_class
         return ''
 
     @staticmethod
-    def _get_endpoint_with_blueprint(endpoint):
+    def get_endpoint_with_blueprint(endpoint):
```
quis, line 32: I don’t feel very precious about this change, but it’s an informal Python convention to prefix method names with an underscore if they’re not intended to be used by consumers of a class.

tests/app/test_navigation.py
```diff
@@ -42,13 +367,6 @@ def test_navigation_items_are_properly_defined(navigation_instance):
             endpoint,
             type(navigation_instance).__name__
         )
-        assert (
-            endpoint not in navigation_instance.endpoints_without_navigation
-        ), '{} is listed in {}.mapping and {}.exclude'.format(
-            endpoint,
-            type(navigation_instance).__name__,
-            type(navigation_instance).__name__,
-        )
```
quis, line 51: This assertion still applies doesn’t it? Endpoints that belong to at least one navigation should not also be in the exclude list.

tests/app/test_navigation.py
```diff
@@ -57,41 +375,16 @@ def test_navigation_items_are_properly_defined(navigation_instance):
...
-        assert (
-            navigation_instance.endpoints_without_navigation.count(endpoint) == 1
-        ), '{} found more than once in {}.exclude'.format(
-            endpoint,
-            type(navigation_instance).__name__
-        )
```
quis, line 81: Again, I think all these assertions would be useful to test against your new mono-list of exclusions

tests/app/test_navigation.py
```diff
@@ -10,6 +10,331 @@
 )
 from tests.conftest import ORGANISATION_ID, SERVICE_ONE_ID, normalize_spaces
 
+EXCLUDED_ENDPOINTS = [
```
quis, line 13: This would be better as a `tuple` (if you care about order) or a `set` (if you don’t). Lists are mutable, which I don’t think is behaviour you need.

quis (commented):
Think the approach here is a good, just a few questions about the implementation

app/navigation.py
benthorner: Yes, I was unsure about this one: it's now used in the test, which I'm considering as a consumer here. That's why I removed the underscore.

Would you say tests are an exception to the convention?

benthorner (commented):
Thanks for the review @quis. I found your comments helpful and think I've addressed them. 

I've just amended the commit with the changes. Could you take another look?

app/navigation.py
quis: That sounds like solid enough reasoning for the change to me.

quis (approved):
Thanks for making those changes, looks good to go now 👍🏻

PR #3821, opened 2021-03-02
Support service IDs in ‘find services by name’

quis:
Sometimes we get a service ID from a support ticket or a Slack discussion. Rather than having to hack the URL, this PR augments the ‘Find services by name’ page to support service IDs. If a UUID is entered, it assumes that it’s been given a service ID, and redirects straight to the dashboard for that service, without showing any search results (a complete UUID would never match multiple services). If the UUID is not a service ID, the user will get a 404.

### DIFF
--- a/app/main/views/find_services.py
+++ b/app/main/views/find_services.py
@@ -1,4 +1,7 @@
-from flask import render_template
+import uuid
+from contextlib import suppress
+
+from flask import redirect, render_template, url_for
 
 from app import service_api_client
 from app.main import main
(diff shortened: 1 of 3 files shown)


app/main/views/find_services.py
```diff
@@ -12,6 +15,11 @@ def find_services_by_name():
     form = SearchByNameForm()
     services_found = None
     if form.validate_on_submit():
+        with suppress(ValueError):
```
idavidmcdonald, line 18: I had to go understand what `suppress` does which added an extra minute to be me understanding how this code works. Just pointing out that by using some of these lesser used python features in our code can have a small impact on the readability of it (although they also come with obvious benefits). I don't think you should change it but just something to be aware of.

idavidmcdonald (approved):
Nothing wrong with the code. Slightly apprehensive if the code is worth the value, I've never had any problem with editing the URL, but given the few lines of code it takes, why not!

PR #3823, opened 2021-03-05
post to get user by email

leohemsted:
rather than get. this endpoint is used every time a user signs in, accepts an invite, resets their password, etc.

- [x] requires new API endpoint https://github.com/alphagov/notifications-api/pull/3169

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -40,7 +40,7 @@ class UserApiClient(NotifyAdminAPIClient):
         return self.get("/user/{}".format(user_id))
 
     def get_user_by_email(self, email_address):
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):

app/notify_client/user_api_client.py
```diff
@@ -40,7 +40,7 @@ def _get_user(self, user_id):
         return self.get("/user/{}".format(user_id))
 
     def get_user_by_email(self, email_address):
-        user_data = self.get('/user/email', params={'email': email_address})
+        user_data = self.post('/user/email', params={'email': email_address})
```
idavidmcdonald, line 43: Do we want/need the `params` keyword here? I see other `post`s just passing it as an object and others passing it as `data=data`?

PR #3824, opened 2021-03-05
Invite user audit event

klssmith:
This adds an audit event for when a user is added to a service. The event records
* the ID of the user who was added
* the ID of the inviting user
* the service ID

Part of [Pivotal story](https://www.pivotaltracker.com/story/show/176698377)

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -34,6 +34,15 @@ def create_remove_user_from_service_event(user_id, removed_by_id, service_id):
     )
 
 
+def create_add_user_to_service_event(user_id, invited_by_id, service_id):
+    _send_event(
+        'add_user_to_service',
+        user_id=user_id,
+        invited_by_id=invited_by_id,
+        service_id=service_id
+    )
+
+
 def create_archive_user_event(user_id, archived_by_id):
     _send_event(
         'archive_user',
--- a/app/main/views/verify.py
+++ b/app/main/views/verify.py
@@ -97,5 +97,10 @@ def _add_invited_user_to_service(invited_user):
     invitation = InvitedUser(invited_user)
     user = User.from_id(session['user_id'])
     service_id = invited_user['service']
-    user.add_to_service(service_id, invitation.permissions, invitation.folder_permissions)
+    user.add_to_service(
+        service_id,
+        invitation.permissions,
+        invitation.folder_permissions,
+        invitation.from_user.id,
+    )
     return service_id
--- a/app/main/views/invites.py
+++ b/app/main/views/invites.py
@@ -68,6 +68,7 @@ def accept_invite(token):
                 service_id=invited_user.service,
                 permissions=invited_user.permissions,
                 folder_permissions=invited_user.folder_permissions,
+                invited_by_id=invited_user.from_user.id,
             )
             if service.has_permission('broadcast'):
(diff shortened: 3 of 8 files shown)


app/event_handlers.py
```diff
@@ -34,6 +34,15 @@ def create_remove_user_from_service_event(user_id, removed_by_id, service_id):
     )
 
 
+def create_invite_user_to_service_event(user_id, invited_by_id, service_id):
```
idavidmcdonald, line 37: I think we should rename all the `invite_user` to be `add_user` in our function names and also the event stored in the events table as this is more accurate

app/main/views/verify.py
```diff
@@ -97,5 +97,10 @@ def _add_invited_user_to_service(invited_user):
     invitation = InvitedUser(invited_user)
     user = User.from_id(session['user_id'])
     service_id = invited_user['service']
-    user.add_to_service(service_id, invitation.permissions, invitation.folder_permissions)
+    user.add_to_service(
+        service_id,
+        invitation.permissions,
+        invitation.folder_permissions,
+        invitation._from_user,
```
idavidmcdonald, line 104: So it took me a while to figure out what was going on here.

`_from_user` is a property on the `InvitedUser` model. I was confused because I thought it would be a user object, and therefore we would need to access it by calling `_from_user.id` but that was incorrect but in fact `_from_user` is a string of the ID from the API.

I think we should at the very least, given we are now using this outside of the class, add a new property to the `InvitedUser` class that is public (denoted by the lack of starting `_`) that we use to access this property.

That property could either be
`from_user` which returns a user object (we can then access `from_user.id`)
or a property of `from_user_id` which is then very explicit about what type of object we are getting back

What do you think?

idavidmcdonald (commented):
Looks pretty good. 2 small comments worth having a think about please

app/event_handlers.py
klssmith: Thanks, I hadn't spotted that the name was leftover from when the event was going to work slightly differently

app/main/views/verify.py
klssmith: There is actually already a `.from_user` method on `InvitedUser`, which means I was able to change `invited_user._from_user` to `invited_user.from_user.id` without adding any extra properties. I think I had used the private method originally to avoid an API call to get the User, but it shouldn't be using a private method and we cache the User anyway I don't think that's a concern.

app/event_handlers.py
```diff
@@ -34,6 +34,15 @@ def create_remove_user_from_service_event(user_id, removed_by_id, service_id):
     )
 
 
+def create_add_user_to_service_event(user_id, added_by_id, service_id):
```
idavidmcdonald, line 37: Optional:
I think I'd actually suggest `added_by_id` stay as `invited_by_id` as technically they are `invited` rather than `added` by that person I think. But I don't feel too strongly about it. Up to you :)

idavidmcdonald (approved):

app/event_handlers.py
klssmith: Good suggestion - I've changed the arguments back to `invited_by_id`, but kept the function name as `create_add_user_to_service_event` and kept the event description as `add_user_to_service`.

PR #3826, opened 2021-03-05
Add new free allowance rates for 2021 financial year

quis:
These are the new rates from https://docs.google.com/document/d/1aObNQNBw3ayPMl3b_Qc5kZBdaXIAMDnBGtV-xPtFBsc/edit

We’re changing the free allowance so we can continue to support all the teams that use Notify. The new allowance means over 90% of teams can still send all the text messages they need to without paying.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -35,27 +35,35 @@ class Config(object):
     DEFAULT_FREE_SMS_FRAGMENT_LIMITS = {
         'central': {
             2020: 250_000,
+            2021: 150_000,
         },
         'local': {
             2020: 25_000,
+            2021: 25_000,
         },
         'nhs_central': {
             2020: 250_000,
+            2021: 150_000,
         },
         'nhs_local': {
             2020: 25_000,
+            2021: 25_000,
         },
         'nhs_gp': {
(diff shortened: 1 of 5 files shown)


idavidmcdonald (commented):
So this is a tentative approval. So this code all works and is great and well tested.

My only question is this requires us to make a code change every year to keep this working without raising some exceptions. Is that OK? Or after the new financial year do we change this code back to just use a default and not look at the date? Or do we change this code to fall back to something if the key is not found?

quis:
I think we change it back after 1 April, assuming these new allowances continue indefinitely. It’s always possible we change the allowance structure again, but it might change in a way that this logic doesn’t account for (what if we did a per-organisation allowance for example). For me this a quick fix, not a foundation to build on.

idavidmcdonald:
In that case, this is good.

Is changing it back on the 1st of april you will be responsible for or something we need to write a pivotal ticket for?

idavidmcdonald (approved):

quis:
I’ve added a holding PR for the 1 April change: https://github.com/alphagov/notifications-admin/pull/3830

PR #3832, opened 2021-03-11
Add a page to show performance data 

quis:
Performance Platform is going away. This pull requests adds a page in our app that show most of the same information.

It looks a bit janky on my local machine because I can’t get my local data right. But this should be a good starting point we can tweak later.

localhost_6012_performance(iPad Pro)

Lighthouse report for @tombye:

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/3173


app/templates/views/performance.html
```diff
@@ -0,0 +1,151 @@
...
+            smallest=True,
+          ) }}
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <div class="dashboard-table">
+    {% call(item, row_number) list_table(
+      notifications_by_type,
```
tombye, line 57:
```suggestion
      notifications_by_type,
      caption='Messages sent since May 2016',
```

app/templates/views/performance.html
```diff
@@ -0,0 +1,151 @@
...
+      <h3 class="govuk-visually-hidden">Average</h3>
+      {{ big_number(
+        '{:.1f}%'.format(average_percentage_under_10_seconds),
+        label='on average',
+      ) }}
+    </div>
+  </div>
+  <div class="dashboard-table">
+    {% call(item, row_number) list_table(
+      processing_time | reverse,
```
tombye, line 99:
```suggestion
      processing_time | reverse,
      caption='Messages sent within 10 seconds',
```

app/templates/views/performance.html
```diff
@@ -0,0 +1,151 @@
...
+    </div>
+    <div class="govuk-grid-column-one-half">
+      <h3 class="govuk-visually-hidden">Services</h3>
+      <div class="product-page-big-number">{{ count_of_live_services_and_organisations.services|format_thousands }}</div>
+      services
+    </div>
+  </div>
+  <div class="dashboard-table">
+    {% call(item, row_number) list_table(
+      organisations_using_notify,
```
tombye, line 135:
```suggestion
      organisations_using_notify,
      caption='Organisations using Notify',
```

app/templates/views/performance.html
```diff
@@ -0,0 +1,151 @@
...
+      Only showing the last {{ processing_time|length }} days
+    </p>
+  </div>
+
+  <h2 class="govuk-heading-m">
+    Organisations using Notify
+  </h2>
+  <div class="govuk-grid-row bottom-gutter">
+    <div class="govuk-grid-column-one-half">
+      <h3 class="govuk-visually-hidden">Organisations</h3>
```
tombye, line 123: I'm not sure we need these `<h3>`s and they come across a bit odd, on Voiceover at least, after the `<h2>`. I'd suggest losing them and putting a bit of non-visual text around the content for the same context.

I mean to move from:

h2: Organisations using Notify
h3: Organistions
Text: 2250 organisations
h3: Services
Text: 4003 services

...to:

h2: Organisations using Notify
Text: There are 2250 organisations using Notify
Text: There are 4003 services using Notify

This kind of content also exists on the signedout page and solves the same problem by changing the `<h2>` so that might be an easier alternative.

tombye (commented):
I don't have great data locally so I've just added some comments around the HTML, based on a very reduced set of content compared to what it will host on live.

app/templates/views/performance.html
quis: Changed in 7583c8d

app/templates/views/performance.html
```diff
@@ -0,0 +1,153 @@
...
+      {% endcall %}
+      {% call field() %}
+        {{ item.sms|format_thousands }}
+      {% endcall %}
+      {% call field() %}
+        {{ item.letters|format_thousands }}
+      {% endcall %}
+    {% endcall %}
+    <p class="table-show-more-link">
+      Only showing the last {{ notifications_by_type|length }} days
```
idavidmcdonald, line 81: Optional: We aren't ever expecting our data only show 1 day, but if it was then this would want to be non plural. Probably not worth adding though

app/notify_client/performance_dashboard_api_client.py
```diff
@@ -0,0 +1,22 @@
+from app.notify_client import NotifyAdminAPIClient, cache
+
+
+class PerformanceDashboardAPIClient(NotifyAdminAPIClient):
+
+    @cache.set('performance-stats-{start_date}-to-{end_date}')
```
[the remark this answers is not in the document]
idavidmcdonald: Oh I also checked and given there is a fair bit of data to store for this key, it is well within the max supported by redis (512MB) - https://redis.io/topics/data-types

app/formatters.py
```diff
@@ -515,3 +515,7 @@ def format_mobile_network(network):
     if network in ('three', 'vodafone', 'o2'):
         return network.capitalize()
     return 'EE'
+
+
+def format_billions(count):
```
idavidmcdonald, line 520: Optional: Could add a couple of unit tests for this, mostly as self documentation so we can see what this function does to various counts (0, in the hundreds, in the thousands, in the millions, in the billions).

idavidmcdonald (approved):
Looks good.

Only one thing needs changing to allow this to go out (the missing word I called out), the rest is all optional. We can improve on things like the css alignment at a later point.

app/notify_client/performance_dashboard_api_client.py
```diff
@@ -0,0 +1,22 @@
...
+    def get_performance_dashboard_stats(
+        self,
+        *,
```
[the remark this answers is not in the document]
quis: Forces the subsequent arguments to be named, so that you don’t get `start_date` and `end_date` mixed up

app/templates/views/features.html
```diff
@@ -58,7 +58,7 @@ <h2 class="heading-medium" id="performance">Performance</h2>
 
 <p class="govuk-body">We send messages through several different providers. If one provider fails, Notify switches to another so that your messages are not affected.</p>
 
-<p class="govuk-body">Visit GOV.UK Performance to <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/performance/govuk-notify">see how Notify is performing</a>.</p>
+<p class="govuk-body">Visit our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.performance') }}">performance</a> page see how Notify is performing.</p>
```
quis, line 61:
```suggestion
<p class="govuk-body">Visit our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.performance') }}">performance data</a> page to see how Notify is performing.</p>
```

PR #3839, opened 2021-03-15
Transform usage report to billing report

CrystalPea:
by adding billing details:

purchase order number
billing contact names
billing contact email addresses
billing reference

and changing naming.

Needs to go in after this PR (which makes the API endpoint return billing details for each service it returns, and makes the endpoint only return services with bills to pay):
https://github.com/alphagov/notifications-api/pull/3163/files

Pivotal ticket: https://www.pivotaltracker.com/story/show/176521236


app/main/views/platform_admin.py
```diff
@@ -301,33 +301,45 @@ def notifications_sent_by_service():
...
         end_date = form.end_date.data
-        headers = ["organisation_id", "organisation_name", "service_id", "service_name",
-                   "sms_cost", "sms_fragments", "letter_cost", "letter_breakdown"]
-
-        result = billing_api_client.get_usage_for_all_services(start_date, end_date)
+        headers = [
+            "organisation_id", "organisation_name", "service_id", "service_name",
+            "sms_cost", "sms_fragments", "letter_cost", "letter_breakdown", "purchase_order_number",
+            "contact_names", "contact_email_addresses", "billing_reference"
+        ]
```
servingUpAces, line 314: Have you checked if Caley likes the ordering of the columns?

CrystalPea: I did send an example report to her some time ago, but I think she missed it, I will try again now :)

CrystalPea: Caley is ok with the report, she would like to add one more column that is not in the story: letter totals. 

This would require us to do a small change to API and then to Admin, I feel like it's a separate small ticket though?

servingUpAces: It's alright to add that as a separate PR. 
Letter totals are probably already included in the query since you need them to calculate the cost.

servingUpAces (approved):

app/main/views/platform_admin.py
CrystalPea: yeah, they are in the query, but not in what we send back to admin app.

PR #3850, opened 2021-03-31
Create additional non-UK broadcast test polygons

rjbaker:
This allows MNOs to test delivery to multiple non-adjacent cells without risk of sending a broadcast on the public network.

Additional polygons will also support the testing of multiple polygon geometries in a single message.

Test polygons are all non-UK (northern Finland).

Screenshot 2021-03-31 at 09 41 04

### DIFF
--- a/app/broadcast_areas/source_files/Test.geojson
+++ b/app/broadcast_areas/source_files/Test.geojson
@@ -1,20 +1,73 @@
 {
-  "type":"FeatureCollection",
-  "features":[{
-    "type": "Feature",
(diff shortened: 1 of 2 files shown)


rjbaker:
I think these will get treated as separate polygons. When the CAP is
created each polygon is treated as a separate area. I don't believe we send
any areas containing more than one polygon. This is the issue with the CBC
ignoring multiple areas which needs to be tested.

leohemsted (approved):

PR #3853, opened 2021-04-06
Don't show user as unknown for service history

idavidmcdonald:
A commit was added:
https://github.com/alphagov/notifications-admin/commit/600e3affc12947d5a7b957b71e3d2864467d6bc2

In it, it falls back to the string 'Unknown' for actions done by those
not belonging to the service.

This commit changes the behaviour such that if the user is not in the
list of active users for a service, it will go get the user from the DB
(or redis). This should be fine to do as redis will protect us from most
calls as most of these cases are for platform admins.

This will mean we can now see which user platform admin put a service
live rather than seeing 'Unknown'.

Note, tests are a bit hacky, I couldn't quite get the mocking figured out as I would have liked

Before:
image

After:
image

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -652,6 +652,12 @@ class Users(ModelList):
         for user in self:
             if user.id == id:
                 return user.name
+        # The user may not exist in the list of users for this service if they are
+        # a platform admin or if they have since left the team. In this case, we fall
+        # back to getting the user from the API (or Redis if it is in the cache)
+        user = User.from_id(id)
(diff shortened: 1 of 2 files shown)


quis (approved):
There’s a risk with the `Users.name_from_id` approach that some other code starts using that without realising the potential performance implications. 

But we’ve decided for now that getting the API to return the data is hard because of Marshmallow. The ServiceHistory class isn't easily defined; it's all done dynamically.

So this is feels like a good improvement for now.

PR #3855, opened 2021-04-07
Let the API handle setting the default free allowance.

servingUpAces:
The API has a method to handle setting the default SMS free allowance. This will save a call to the API and remove some code duplication between the two apps.

Needs to be merged after https://github.com/alphagov/notifications-api/pull/3197


quis (approved):
Nice 👍🏻

PR #3856, opened 2021-04-08
Fix sign in link on 401 page

klssmith:
We've seen someone getting a `500` when trying to visit the `/metrics`
endpoint. This is because the metrics endpoint is not in the `main`
blueprint, so we can't use `url_for` with a relative endpoint.

Although only bots visit '/metrics' without authorisation, this stops
the odd `500`.

### DIFF
--- a/app/templates/error/401.html
+++ b/app/templates/error/401.html
@@ -4,6 +4,6 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
     <h1>You’re not authorised to see this page</h1>
-      <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.sign_in' )}}">Sign in</a> to GOV.UK Notify and try again.</p>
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
Good detective work

benthorner (approved):
Nice fix 👍.

It's been a good prompt for me to look again at `url_for`. Previously I was only familiar with the `blueprint.endpoint` and `endpoint` forms. Apparently the `.` is implicit in the latter. We seem to have a mixture of [both forms](https://github.com/alphagov/notifications-admin/blob/2c8fc4d5230632879ebd55953a1d8d0958439e77/app/main/views/sign_in.py#L37) in this app for some reason.

klssmith:
@benthorner It mostly doesn't matter which form we use, it was a bit of an edgecase that it did matter here (only because we were linking to the sign in page from outside the main blueprint). I found [this explanation](https://flask.palletsprojects.com/en/1.1.x/api/#flask.url_for) useful

PR #3857, opened 2021-04-08
pin prometheus-client to 0.9.0

leohemsted:
we saw exceptions on prod that I think might have caused a worker that is being terminated to die ungracefully. While I'm not sure if this is an actual problem that changed behaviour (app instances crashing and restarting), at the very least it definitely polluted the logs and obscured any actual issues we were having.

https://github.com/prometheus/client_python/releases/tag/v0.10.0

see the pending fix for this problem here: prometheus/client_python#644

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ fix-imports:
 
 .PHONY: freeze-requirements
 freeze-requirements: ## create static requirements.txt
-	${VIRTUALENV_ROOT}/bin/pip install pip-tools
+	${VIRTUALENV_ROOT}/bin/pip install --upgrade pip-tools
 	${VIRTUALENV_ROOT}/bin/pip-compile requirements.in
 
 .PHONY: clean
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -57,7 +57,7 @@ fix-imports:
 
 .PHONY: freeze-requirements
 freeze-requirements: ## create static requirements.txt
-	${VIRTUALENV_ROOT}/bin/pip install pip-tools
+	${VIRTUALENV_ROOT}/bin/pip install --upgrade pip-tools
```
benthorner, line 60: Is this necessary here? It looks like we're only installing one package, outside of requirements.txt, so I'm confused how the flag will affect that file.

benthorner (approved):
I've got a more minor comment about the first commit. We don't seem to have had any more crashes, but the main change here is worth putting in anyway.

Makefile
leohemsted: without this flag, it won't upgrade old versions. i found that i had an older version of pip tools and it caused a massive whitespace diff cos the formatting changed at some point between v5.4.0 (my prev version) and 6.0.1 (current latest). it's not a huge deal and i dont mind dropping this, but since that's what i had to run to get freeze-requirements to work cleanly, i thought it might benefit everyone

benthorner: Ahh, that's OK. I understand what you mean now: it's the affect of running the tool itself, not the command to install it 👍. Makes sense to have people use the latest version by default.

PR #3860, opened 2021-04-13
Allow users of the API to search templates by ID

quis:
image

***

For someone who has retrieved a template ID from their system the only way to find it in Notify is:
- hack the URL
- click through every template, visually inspecting the ID shown on the page until you find the right one

Neither of these is ideal.

This commit adds searching by ID, for those services who have an API integration. This means we don’t need to confuse teams who aren’t using the API by talking about IDs.

This is similar to how we let these teams [search for notifications by reference](https://github.com/alphagov/notifications-admin/pull/3223).

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1924,6 +1924,17 @@ class SearchNotificationsForm(StripWhitespaceForm):
         )
 
 
+class SearchTemplatesForm(StripWhitespaceForm):
+
+    search = GovukSearchField()
+
+    def __init__(self, api_keys, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.search.label.text = (
+            'Search by name or ID' if api_keys else 'Search by name'
+        )
+
+
 class PlaceholderForm(StripWhitespaceForm):
 
     pass
--- a/tests/app/main/views/test_letters.py
+++ b/tests/app/main/views/test_letters.py
@@ -70,6 +70,7 @@ def test_given_option_to_add_letters_if_allowed(
     mock_get_service_templates,
     mock_get_template_folders,
     mock_get_organisations_and_services_for_user,
+    mock_get_api_keys,
     permissions,
     choices,
 ):
--- a/tests/app/main/views/test_template_folders.py
+++ b/tests/app/main/views/test_template_folders.py
@@ -375,6 +375,7 @@ def test_should_show_templates_folder_page(
     client_request,
     mock_get_template_folders,
     mock_has_no_jobs,
+    mock_get_no_api_keys,
     service_one,
     mocker,
     expected_title_tag,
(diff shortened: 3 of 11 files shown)


app/main/forms.py
```diff
@@ -1924,6 +1924,17 @@ def __init__(self, message_type, *args, **kwargs):
...
 
 
+class SearchTemplatesForm(StripWhitespaceForm):
+
+    search = GovukSearchField()
+
+    def __init__(self, api_keys, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.search.label.text = (
+            'Search by name or ID' if api_keys else 'Search by name'
```
benthorner, line 1934: Do we need this conditional? I notice we show the Template ID on the template page itself, so the risk of confusion seems pretty minimal. 

In the PR you linked to, we show "or reference", even though this can only be set by the API.

tests/app/main/views/test_template_folders.py
```diff
@@ -477,6 +478,71 @@ def test_should_show_templates_folder_page(
...
+    ] == [
+        'folder one folder one 1 template',
+        f'template one {template_1["id"]} template one Text message template',
+    ]
+
+    assert [
+        normalize_spaces(item.text)
+        for item in page.select(
+            # Text which should be hidden from all users
+            r'#template-list .govuk-\!-display-none'
```
benthorner, line 536: Minor: for consistency with the other assertions, we could use `template-list-item` here.

tests/app/main/views/test_template_folders.py
```diff
@@ -477,6 +478,71 @@ def test_should_show_templates_folder_page(
...
+        )
+    ] == [
+        template_2["id"],
+        template_1["id"],
+    ]
+
+    mock_get_api_keys.assert_called_once_with(SERVICE_ONE_ID)
```
benthorner, line 543: Do we need this assertion?

tests/app/main/views/test_letters.py
```diff
@@ -70,6 +70,7 @@ def test_given_option_to_add_letters_if_allowed(
     mock_get_service_templates,
     mock_get_template_folders,
     mock_get_organisations_and_services_for_user,
+    mock_get_api_keys,
```
benthorner, line 73: Is there a reason for choosing this over `_no_api_keys` here (and elsewhere)?

benthorner (approved):
Oh so agile 🎖️.

I've got a question about the need to hide the search box label, but it's nothing major.

It's a shame we need _so many_ extra fixtures, but that seems to be the way of things with Pytest. I did ponder suggesting the "no API keys" fixture becoming part of the service or "get templates" one, but there's a risk of future confusion there too.

tests/app/main/views/test_template_folders.py
quis: It’s a good safety check, because it would be bad for performance if the code resulted in the API being called for every template in the list.

app/main/forms.py
quis: It feels like a nice detail to keep the UI clean for those who don’t need to care about IDs. In the linked PR ‘or reference’ only appears for services which have API keys.

I think we still need template ID on the templates page for all users, because when building your integration you might grab the template ID before you’ve created any keys.

tests/app/main/views/test_template_folders.py
quis:
```suggestion
            '.template-list-item .govuk-\!-display-none'
```

tests/app/main/views/test_letters.py
quis: Used this by default until I needed to use `_no_api_keys` to not break existing tests

tests/app/main/views/test_template_folders.py
benthorner: Ah, OK, thanks for explaining 👍.

app/main/forms.py
benthorner: Hmm...I understand where you're coming from, which is different to my perspective: less complexity, more consistency, less surprising. I don't feel strongly either way, though.

>  ‘or reference’ only appears for services which have API keys

Oops, I missed that, which is a clear precedent 👍. I like that we're being consistent (although apparently not/having a "Search" button isn't a thing we do consistently 🙈.)

PR #3863, opened 2021-04-19
Show service suspension in breadcrumb

benthorner:
https://www.pivotaltracker.com/story/show/177758598

Previously there was no indication that a service was suspended.
While this could also be shown for archived/deleted services, the
meaning is similar enough that it makes sense there too - the name
of the archived service should distinguish it as being archived.

## Screenshots

| Before | After |
| ------- | ----- |
| Screenshot 2021-04-27 at 10 56 38 | Screenshot 2021-04-27 at 10 52 00 |
| Screenshot 2021-04-27 at 10 56 24 | Screenshot 2021-04-27 at 10 52 49 |

### DIFF
--- a/app/templates/withnav_template.html
+++ b/app/templates/withnav_template.html
@@ -17,16 +17,18 @@
       {% endif %}
       <div class="navigation-service-name govuk-!-font-weight-bold">
         {{ current_service.name }}
-        {% if current_service.has_permission('broadcast') %}
-          {% if current_service.trial_mode %}
-            <span class="navigation-service-type navigation-service-type--training">Training
-          {% else %}
-            <span class="navigation-service-type navigation-service-type--live">Live
-          {% endif %}
-          {% if current_service.allowed_broadcast_provider %}
-            ({{ current_service.allowed_broadcast_provider }})
-          {% endif %}
-          </span>
+          {% if not current_service.active %}
+            <span class="navigation-service-type navigation-service-type--suspended">Suspended</span>
+          {% elif current_service.has_permission('broadcast') %}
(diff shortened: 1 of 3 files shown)


app/templates/withnav_template.html
```diff
@@ -16,6 +16,9 @@
         {% endif %}
       {% endif %}
       <div class="navigation-service-name govuk-!-font-weight-bold">
+        {% if not current_service.active %}
+        [SUSPENDED]
+        {% endif %}
         {{ current_service.name }}
         {% if current_service.has_permission('broadcast') %}
```
quis, line 23: You could use the same styling as the training mode label here. It has a similar meaning – the service is prevented from doing any live stuff, be that broadcasts or emails etc.

It also removes the slightly incongruous combination of _[SUSPENDED] Service name LIVE_ for broadcast services.
```suggestion
        {{ current_service.name }}
        {% if not current_service.active %}
        <span class="navigation-service-type navigation-service-type--training">Suspended</span>
        {% elif current_service.has_permission('broadcast') %}
```

benthorner: I was thinking that a service being suspended is a separate dimension from whether it's live (and which broadcast provider was chosen). That said, I agree the practical impact overlaps with the concept of "training" and "live" in terms of what the service can do.

I've made the change you suggested, with a tweak to apply it to _all_ services (not just broadcast ones), since it makes sense in all cases. I've added a couple of screenshots to show the before/after too.

What do you think?

quis (approved):
This looks good now, thanks for making those changes.

PR #3864, opened 2021-04-20
Update email status description

karlchillmaid:
This PR updates the ‘Inbox not accepting messages right now’ email status description.

We’re making this change because a full inbox is not the only reason a user might get this status.

We’ve added the link to Service manual content because we should help users to understand *why* their content might be triggering a recipient’s spam filter.

The Service manual content could probably do with updating, but I’ll tackle this as a separate piece of work.

### DIFF
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -26,7 +26,7 @@
         ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
         ('Delivered', 'The message was successfully delivered. Notify will not tell you if a user has opened or read a message.'),
         ('Email address does not exist', 'The provider could not deliver the message because the email address was wrong. You should remove these email addresses from your database.'),
-        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full. You can try to send the message again.'),
+        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full or their anti-spam filter rejects your email. <a class="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#protect-your-users-from-spam-and-phishing">Make sure your content does not look like spam</a>. You can try to send the message again.'),
         ('Technical failure', 'Your message was not sent because there was a problem between Notify and the provider. You’ll have to try sending your messages again.'),
       ] %}
         {% call row() %}
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -31,7 +31,7 @@
       ] %}
         {% call row() %}
           {{ text_field(message_length) }}
-          {{ text_field(charge) }}
+          {{ text_field(charge | safe) }}
         {% endcall %}
       {% endfor %}
     {% endcall %}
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):

app/templates/views/message-status.html
```diff
@@ -26,7 +26,7 @@ <h2 id="email-statuses" class="heading-medium">Emails</h2>
         ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
         ('Delivered', 'The message was successfully delivered. Notify will not tell you if a user has opened or read a message.'),
         ('Email address does not exist', 'The provider could not deliver the message because the email address was wrong. You should remove these email addresses from your database.'),
-        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full. You can try to send the message again.'),
+        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full or their anti-spam filter rejects your email. <a class="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#protect-your-users-from-spam-and-phishing">Check your content does not look like spam</a> before you try to send the message again.'),
```
klssmith, line 29: This link isn't quite right (it's got the bit for the `href` in the `class`):

Screenshot 2021-04-20 at 16 23 36

karlchillmaid: @klssmith whoops! good catch

karlchillmaid:
```suggestion
        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full or their anti-spam filter rejects your email. <a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#protect-your-users-from-spam-and-phishing">Check your content does not look like spam</a> before you try to send the message again.'),
```

app/templates/views/message-status.html
```diff
@@ -26,7 +26,7 @@ <h2 id="email-statuses" class="heading-medium">Emails</h2>
         ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
         ('Delivered', 'The message was successfully delivered. Notify will not tell you if a user has opened or read a message.'),
         ('Email address does not exist', 'The provider could not deliver the message because the email address was wrong. You should remove these email addresses from your database.'),
-        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full. You can try to send the message again.'),
+        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full or their anti-spam filter rejects your email. <a href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#protect-your-users-from-spam-and-phishing">Check your content does not look like spam</a> before you try to send the message again.'),
```
klssmith, line 29: Just needs the standard link classes we use elsewhere:

```suggestion
        ('Inbox not accepting messages right now', 'The provider could not deliver the message. This can happen when the recipient’s inbox is full or their anti-spam filter rejects your email. <a class="govuk-link govuk-link--no-visited-state" href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#protect-your-users-from-spam-and-phishing">Check your content does not look like spam</a> before you try to send the message again.'),
```

Plus the `{{ text_field(charge) }}` line just below should be changed to `{{ text_field(charge | safe) }}` to make the link render as a link (I wasn't able to add that one as a suggestion).

klssmith (approved):

CrystalPea:
Good spot @klssmith I can't believe I missed it, sorry @karlchillmaid  😳

PR #3870, opened 2021-05-04
Remove some of the option we have for broadcast services.

servingUpAces:
This is the first step in getting to an easier UI for setting the
options for a broadcast service. Here we remove the options for "live"
test channels. And change "All networks" to the "test" channel rather
than "severe".

This PR could go out now. However, there is further improvement to
come.

https://www.pivotaltracker.com/story/show/177759100


idavidmcdonald (changes requested):
Bit confused here Becca. 

The pivotal ticket says:

we can remove these four options

Training mode - EE network - Test channel only
Training mode - O2 network - Test channel only
Training mode - Three network - Test channel only
Training mode - Vodafone network - Test channel only

However this change currently removes the Live mode versions of this?

We do need these different live mode accounts, the MNOs currently use them for testing things on Preview for example

servingUpAces:
That contradicts the document that CHS has added to the story. We'll need a huddle for this then.

PR #3871, opened 2021-05-05
Remove some of the option we have for broadcast services.

servingUpAces:
This is the first step in making the UI easier for setting the
options for a broadcast service. Here we remove the options for
"Training mode" test channels. When we create a broadcast message for a trail mode service it is marked as stubbed and does not create a broadcast event that is sent to a provider.

The label for the form and setting page have been updated to reflect the
change.

https://www.pivotaltracker.com/story/show/177759100


app/templates/views/service-settings.html
```diff
@@ -434,7 +434,9 @@ <h2 class="heading-medium">Platform admin settings</h2>
               {% if not current_service.broadcast_channel %}
               Off
               {% else %}
-                {% if current_service.live %}Live{% else %}Training mode{% endif %} - {% if current_service.allowed_broadcast_provider %}{{ current_service.allowed_broadcast_provider|format_mobile_network }} network{% else %}All networks{% endif %} - {% if current_service.broadcast_channel == "test" %}Test channel only{% else %}Public channel{% endif %}
+                {% if current_service.live and current_service.broadcast_channel == "severe" %}Live - Public channel{% endif %}
```
quis, line 437: to follow my sketches this should be just _Live_
```suggestion
                {% if current_service.live and current_service.broadcast_channel == "severe" %}Live{% endif %}
```

app/templates/views/service-settings.html
```diff
@@ -434,7 +434,9 @@ <h2 class="heading-medium">Platform admin settings</h2>
...
+                {% if current_service.live and current_service.broadcast_channel == "test" %}Test channel {% if current_service.allowed_broadcast_provider %}({{ current_service.allowed_broadcast_provider|format_mobile_network }} network){% else %} (All networks){% endif %}{%endif%}
```
quis, line 438: And this should be _Test_, without the word _Network_ (it’s implicit that EE is a network)
```suggestion
                {% if current_service.live and current_service.broadcast_channel == "test" %}Test{% if current_service.allowed_broadcast_provider %}({{ current_service.allowed_broadcast_provider|format_mobile_network }}){% else %} (All networks){% endif %}{%endif%}
```

app/templates/views/service-settings.html
```diff
@@ -434,7 +434,9 @@ <h2 class="heading-medium">Platform admin settings</h2>
...
+                {% if not current_service.live%}Training mode {% endif%}
```
quis, line 439: aaaand this one should just be _Training_
```suggestion
                {% if not current_service.live%}Training{% endif%}
```

quis (changes requested):
Some comments on the label next to the service name. Trying to keep it very short to fit in the minimal space we have, and also unambiguous (for example _Live - public channel_ sounds like there are other kinds of _Live_; there aren’t)

servingUpAces:
> Some comments on the label next to the service name. Trying to keep it very short to fit in the minimal space we have, and also unambiguous (for example _Live - public channel_ sounds like there are other kinds of _Live_; there aren’t)

Updates have been made to the labels.

quis (approved):

idavidmcdonald:
We should also check, definitely in production at the least, that we have no services that currently have these settings we are removing. Otherwise they will need to have their settings changed

PR #3878, opened 2021-05-11
Allow registering WebAuthn authenticators in memory

benthorner:
https://www.pivotaltracker.com/story/show/178066084

This adds basic functionality and infrastructure for registering
WebAuthn authenticators, using an in-memory store to keep the size
of the PR down. We can build on the patterns established here in
future PRs, and add more features:

- Hooking up with the API app
- Adding a graphic when there are no keys
- Managing existing keys
- Dealing with server-side errors
- Authentication

**This PR is complicated.** I've tried to be as helpful as possible
in the code and commit messages, so have a look at those if you're
not sure about something. Most of the complexity is in the second
commit, where all the pieces can be seen working together.

## Screenshots

| No key (need to add graphic) | After adding a key |
| Screenshot 2021-05-11 at 14 56 12  | Screenshot 2021-05-11 at 14 56 24 |

### Chrome

| New key | Existing key (also alert afterwards) |
| Screenshot 2021-05-11 at 14 58 16 |  Screenshot 2021-05-11 at 14 57 56 |

### Safari

| New key | Existing key (silently errors, so alert) |
| Screenshot 2021-05-11 at 14 59 42 | Screenshot 2021-05-11 at 14 59 51 |

### Firefox

| New key | Existing key (silently errors, so alert) |
| Screenshot 2021-05-11 at 15 35 24 | Screenshot 2021-05-11 at 15 35 38 |

### MS Edge

| New key | Existing key (TODO) |
| edge_yubikey_1st_prompt | edge_yubikey_error |

### DIFF
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -7,7 +7,10 @@
         .on('click', function(event) {
           event.preventDefault();
 
-          fetchWebAuthnCreateOptions()
+          fetch('/webauthn/register')
+            .then((response) => {
+              return response.arrayBuffer();
+            })
             .then((data) => {
               var options = window.CBOR.decode(data);
               // triggers browser dialogue to select authenticator
@@ -22,45 +25,21 @@
               window.location.reload();
             })
             .catch((error) => {
-              // there may be other kinds of error we should catch here
-              // https://github.com/w3c/webauthn/issues/876
-              if (error instanceof DOMException) {
-                console.error(error);
-                // not all browsers show an error dialogue, so to be safe
-                // we manually pop one open here (to be improved in future!)
-                alert('Error communicating with device.\n\n' + error.message);
-              } else {
-                // for web requests we need to manually alert the user
-                // $.ajax seems to log by itself, but that's not visible
-                alert('Error during registration. Please try again.');
-              }
+              console.error(error);
+              // some browsers will show an error dialogue for some
+              // errors; to be safe we always pop up an alert
+              var message = error.message || error;
+              alert('Error during registration.\n\n' + message);
             });
         });
     };
   };
 
-  function fetchWebAuthnCreateOptions() {
-    var xhrOverride = new XMLHttpRequest();
-    xhrOverride.responseType = 'arraybuffer';
-
-    return $.ajax({
-      url: '/webauthn/register',
-      xhr: () => xhrOverride,
-      dataType: 'x-binary',
-      converters: { '* x-binary': (value) => value }
-    });
-  }
-
   function postWebAuthnCreateResponse(response, csrf_token) {
-    return $.ajax({
-      url: '/webauthn/register',
+    return fetch('/webauthn/register', {
       method: 'POST',
-      headers: {
-        'X-CSRFToken': csrf_token
-      },
-      processData: false,
-      contentType: 'application/cbor',
-      data: window.CBOR.encode({
+      headers: { 'X-CSRFToken': csrf_token },
+      body: window.CBOR.encode({
         attestationObject: new Uint8Array(response.attestationObject),
         clientDataJSON: new Uint8Array(response.clientDataJSON),
       })
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -3,12 +3,67 @@
 
   window.GOVUK.Modules.RegisterSecurityKey = function() {
     this.start = function(component) {
-
       $(component)
         .on('click', function(event) {
           event.preventDefault();
-          alert('not implemented');
+
+          fetchWebAuthnCreateOptions()
+            .then((data) => {
+              var options = window.CBOR.decode(data);
+              // triggers browser dialogue to select authenticator
+              return window.navigator.credentials.create(options);
+            })
+            .then((credential) => {
+              return postWebAuthnCreateResponse(
+                credential.response, component.data('csrfToken')
+              );
+            })
+            .then(() => {
+              window.location.reload();
+            })
+            .catch((error) => {
+              // there may be other kinds of error we should catch here
+              // https://github.com/w3c/webauthn/issues/876
+              if (error instanceof DOMException) {
+                console.error(error);
+                // not all browsers show an error dialogue, so to be safe
+                // we manually pop one open here (to be improved in future!)
+                alert('Error communicating with device.\n\n' + error.message);
+              } else {
+                // for web requests we need to manually alert the user
+                // $.ajax seems to log by itself, but that's not visible
+                alert('Error during registration. Please try again.');
+              }
+            });
         });
     };
   };
+
+  function fetchWebAuthnCreateOptions() {
+    var xhrOverride = new XMLHttpRequest();
+    xhrOverride.responseType = 'arraybuffer';
+
+    return $.ajax({
+      url: '/webauthn/register',
+      xhr: () => xhrOverride,
+      dataType: 'x-binary',
+      converters: { '* x-binary': (value) => value }
+    });
+  }
+
+  function postWebAuthnCreateResponse(response, csrf_token) {
+    return $.ajax({
+      url: '/webauthn/register',
+      method: 'POST',
+      headers: {
+        'X-CSRFToken': csrf_token
+      },
+      processData: false,
+      contentType: 'application/cbor',
+      data: window.CBOR.encode({
+        attestationObject: new Uint8Array(response.attestationObject),
+        clientDataJSON: new Uint8Array(response.clientDataJSON),
+      })
+    });
+  }
 })(window);
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -1,0 +1,14 @@
+(function(window) {
+  "use strict";
+
+  window.GOVUK.Modules.RegisterSecurityKey = function() {
+    this.start = function(component) {
+
+      $(component)
+        .on('click', function(event) {
+          event.preventDefault();
+          alert('not implemented');
+        });
+    };
+  };
+})(window);
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -9,6 +9,10 @@
 
           fetch('/webauthn/register')
             .then((response) => {
+              if (!response.ok) {
+                throw Error(response.statusText);
+              }
+
               return response.arrayBuffer();
             })
             .then((data) => {
@@ -21,7 +25,11 @@
                 credential.response, component.data('csrfToken')
               );
             })
-            .then(() => {
+            .then((response) => {
+              if (!response.ok) {
+                throw Error(response.statusText);
+              }
+
               window.location.reload();
             })
             .catch((error) => {
--- a/app/main/views/webauthn_credentials.py
+++ b/app/main/views/webauthn_credentials.py
@@ -1,0 +1,43 @@
+from fido2 import cbor
+from flask import current_app, request, session
+from flask_login import current_user
+
+from app.main import main
+from app.models.webauthn_credential import WebAuthnCredential
+from app.notify_client.user_api_client import user_api_client
+from app.utils import user_is_platform_admin
+
+
+@main.route('/webauthn/register')
+@user_is_platform_admin
+def webauthn_begin_register():
+    server = current_app.webauthn_server
+
+    registration_data, state = server.register_begin(
+        {
+            "id": bytes(current_user.id, 'utf-8'),
+            "name": current_user.email_address,
+            "displayName": current_user.name,
+        },
+        credentials=[],  # TODO: get from user
+        user_verification="discouraged",  # don't ask for PIN
+        authenticator_attachment="cross-platform",
+    )
+
+    session["webauthn_registration_state"] = state
+    return cbor.encode(registration_data)
+
+
+@main.route('/webauthn/register', methods=['POST'])
+@user_is_platform_admin
+def webauthn_complete_register():
+    credential = WebAuthnCredential.from_registration(
+        session.pop("webauthn_registration_state"),
+        cbor.decode(request.get_data()),
+    )
+
+    user_api_client.create_webauthn_credential_for_user(
+        current_user.id, credential
+    )
+
+    return ''
--- a/app/main/views/webauthn_credentials.py
+++ b/app/main/views/webauthn_credentials.py
@@ -19,7 +19,10 @@ def webauthn_begin_register():
             "name": current_user.email_address,
             "displayName": current_user.name,
         },
-        credentials=[],  # TODO: get from user
+        credentials=[
+            credential.to_credential_data()
+            for credential in current_user.webauthn_credentials
+        ],
         user_verification="discouraged",  # don't ask for PIN
         authenticator_attachment="cross-platform",
     )
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -1,0 +1,58 @@
+{% extends "withoutnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/button/macro.njk" import govukButton %}
+{% from "components/back-link/macro.njk" import govukBackLink %}
+{% from "components/table.html" import mapping_table, row, field, row_heading %}
+
+{% set page_title = 'Security keys' %}
+{% set credentials = current_user.webauthn_credentials %}
+
+{% block per_page_title %}
+  {{ page_title }}
+{% endblock %}
+
+{% block maincolumn_content %}
+  {{ page_header(
+    page_title,
+    back_link=url_for('.user_profile')
+  ) }}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-five-sixths">
+      {% if credentials %}
+
+        {% call mapping_table(
+          caption=page_title,
+          field_headings=['Security key'],
+          field_headings_visible=False,
+          caption_visible=False,
+        ) %}
+          {% for credential in credentials %}
+            {% call row() %}
+              {% call field() %}
+                <div class="govuk-body">{{ credential.name }}</div>
+                <div class="govuk-body hint">Registered {{ credential.created_at|format_delta }}</div>
+              {% endcall %}
+            {% endcall %}
+          {% endfor %}
+        {% endcall %}
+
+      {% else %}
+
+        <p class="govuk-body">
+          Security keys are an alternative way of signing in to Notify.
+        </p>
+
+      {% endif %}
+
+      {{ govukButton({
+        "element": "button",
+        "text": "Register a key",
+        "classes": "govuk-button--secondary",
+        "attributes": {
+          "data-module": "register-security-key",
+        }
+      }) }}
+    </div>
+  </div>
+{% endblock %}
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -51,6 +51,7 @@
         "classes": "govuk-button--secondary",
         "attributes": {
           "data-module": "register-security-key",
+          "data-csrf-token": csrf_token(),
         }
       }) }}
     </div>
--- a/app/webauthn_server.py
+++ b/app/webauthn_server.py
@@ -1,0 +1,32 @@
+from urllib.parse import urlparse
+
+from fido2.server import Fido2Server
+from fido2.webauthn import PublicKeyCredentialRpEntity
+
+
+def init_app(app):
+    base_url = urlparse(app.config["ADMIN_BASE_URL"])
+    verify_origin_callback = None
+
+    # stub verification in dev (to avoid need for HTTPS)
+    if app.config["NOTIFY_ENVIRONMENT"] == "development":
+        verify_origin_callback = stub_origin_checker
+
+    relying_party = PublicKeyCredentialRpEntity(
+        id=base_url.hostname,
+        name="GOV.UK Notify",
+    )
+
+    app.webauthn_server = Fido2Server(
+        relying_party,
+        attestation="direct",
+        verify_origin=verify_origin_callback,
+    )
+
+    # some browsers don't seem to have a default timeout
+    # 30 seconds seems like a generous amount of time
+    app.webauthn_server.timeout = 30_000
+
+
+def stub_origin_checker(*args):
+    return True
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,12 +29,18 @@ cachetools==4.2.1
     # via notifications-utils
 certifi==2020.12.5
     # via requests
+cffi==1.14.5
+    # via cryptography
 chardet==4.0.0
     # via requests
 click==7.1.2
     # via flask
 colorama==0.4.3
     # via awscli
+cryptography==3.3.2
+    # via
+    #   -r requirements.in
+    #   fido2
 dnspython==1.16.0
     # via eventlet
 docopt==0.6.2
@@ -44,6 +50,8 @@ docutils==0.15.2
 et-xmlfile==1.0.1
     # via openpyxl
 eventlet==0.30.2
+    # via -r requirements.in
+fido2==0.9.1
     # via -r requirements.in
 flask-login==0.5.0
     # via -r requirements.in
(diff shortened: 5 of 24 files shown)


app/webauthn_server.py
```diff
@@ -0,0 +1,32 @@
...
+
+    app.webauthn_server = Fido2Server(
+        relying_party,
+        attestation="direct",
+        verify_origin=verify_origin_callback,
+    )
+
+    # some browsers don't seem to have a default timeout
+    # 10 seconds seems like a reasonable amount of time
+    app.webauthn_server.timeout = 10_000
```
CrystalPea, line 28: when in the flow this timeout could happen?

If it's the time that user has to touch the key / make a decision, we could make it a bit longer, maybe 30 seconds? Otherwise it's ok 👍🏼

tests/app/main/views/test_webauthn_credentials.py
```diff
@@ -0,0 +1,109 @@
...
+    mocker,
+    platform_admin_user,
+    platform_admin_client,
+):
+    mocker.patch.dict(
+        app_.config,
+        values={'ADMIN_BASE_URL': 'http://localhost:6012'}
+    )
+
+    # override base URL so it's consistent on CI and locally
```
CrystalPea, line 29: should this comment go here, or above the mocking of the base URL?

tests/app/models/test_webauthn_credential.py
```diff
@@ -0,0 +1,39 @@
...
...R0U3RtdKJjc2lnWEgwRgIhAI1qbvWibQos/t3zsTU05IXw1Ek3SDApATok09uc4UBwAiEAv0fB/lgb5Ot3zJ691Vje6iQLAtLhJDiA8zDxaGjcE3hjeDVjgVkCUzCCAk8wggE3oAMCAQICBDxoKU0wDQYJKoZIhvcNAQELBQAwLjEsMCoGA1UEAxMjWXViaWNvIFUyRiBSb290IENBIFNlcmlhbCA0NTcyMDA2MzEwIBcNMTQwODAxMDAwMDAwWhgPMjA1MDA5MDQwMDAwMDBaMDExLzAtBgNVBAMMJll1YmljbyBVMkYgRUUgU2VyaWFsIDIzOTI1NzM0ODExMTE3OTAxMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEvd9nk9t3lMNQMXHtLE1FStlzZnUaSLql2fm1ajoggXlrTt8rzXuSehSTEPvEaEdv/FeSqX22L6Aoa8ajIAIOY6M7MDkwIgYJKwYBBAGCxAoCBBUxLjMuNi4xLjQuMS40MTQ4Mi4xLjUwEwYLKwYBBAGC5RwCAQEEBAMCBSAwDQYJKoZIhvcNAQELBQADggEBAKrADVEJfuwVpIazebzEg0D4Z9OXLs5qZ/ukcONgxkRZ8K04QtP/CB5x6olTlxsj+SXArQDCRzEYUgbws6kZKfuRt2a1P+EzUiqDWLjRILSr+3/o7yR7ZP/GpiFKwdm+czb94POoGD+TS1IYdfXj94mAr5cKWx4EKjh210uovu/pLdLjc8xkQciUrXzZpPR9rT2k/q9HkZhHU+NaCJzky+PTyDbq0KKnzqVhWtfkSBCGw3ezZkTS+5lrvOKbIa24lfeTgu7FST5OwTPCFn8HcfWZMXMSD/KNU+iBqJdAwTLPPDRoLLvPTl29weCAIh+HUpmBQd0UltcPOrA/LFvAf61oYXV0aERhdGFYwnSm6pITyZwvdLIkkrMgz0AmKpTBqVCgOX8pJQtghB7wQQAAAAAAAAAAAAAAAAAAAAAAAAAAAECKU1ppjl9gmhHWyDkgHsUvZmhr6oF3/lD3llzLE2SaOSgOGIsIuAQqgp8JQSUu3r/oOaP8RS44dlQjrH+ALfYtpAECAyYhWCAxnqAfESXOYjKUc2WACuXZ3ch0JHxV0VFrrTyjyjIHXCJYIFnx8H87L4bApR4M+hPcV+fHehEOeW+KCyd0H+WGY8s6')  # noqa
```
CrystalPea, line 29: do we want to have this registration_response defined here, or would we want to hide it away since it's this long encoded string?

CrystalPea: Also do we want to present a decoded, readable version, to be able to visually compare with assertions?

tests/app/test_webauthn_server.py
```diff
@@ -0,0 +1,35 @@
...
+    mocker.patch.dict(
+        app_.config,
+        values={'NOTIFY_ENVIRONMENT': environment}
+    )
+
+    webauthn_server.init_app(app_)
+    assert app_.webauthn_server._verify('fake-domain') == allowed
+
+
+def test_server_rp_id(
```
CrystalPea, line 25: what is `rp`? Maybe we could have a bit more descriptive test name here?

CrystalPea (commented):
Great work! 💖 

I left a few comments. Some are suggestions, some are just questions ☺️ 

Do we want to integrate with API in this PR? I think we could?


Also, H1 header seems to be a bit low here:
Screenshot 2021-05-12 at 11 11 55

app/templates/views/user-profile/security-keys.html
```diff
@@ -0,0 +1,58 @@
...
+                <div class="govuk-body hint">Registered {{ credential.created_at|format_delta }}</div>
+              {% endcall %}
+            {% endcall %}
+          {% endfor %}
+        {% endcall %}
+
+      {% else %}
+
+        <p class="govuk-body">
+          Security keys are an alternative way of signing in to Notify.
```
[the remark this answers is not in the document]
benthorner: I'm not sure, I just used the content from @quis. 

Since this is only for Platform Admins, I think we're OK without a content review for now.

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -3,12 +3,67 @@
...
+              window.location.reload();
+            })
+            .catch((error) => {
+              // there may be other kinds of error we should catch here
+              // https://github.com/w3c/webauthn/issues/876
+              if (error instanceof DOMException) {
+                console.error(error);
+                // not all browsers show an error dialogue, so to be safe
+                // we manually pop one open here (to be improved in future!)
+                alert('Error communicating with device.\n\n' + error.message);
```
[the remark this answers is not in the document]
benthorner: No, this is all happening on the client side. We'd need to do another server request to get it into our logs, which would be unusual for JS code. While it's not ideal for remote debugging, we do log the full error object in the console and the alert, so we can always ask the user for it (and for now we're the only users anyway).

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -3,12 +3,67 @@
...
+
+  function fetchWebAuthnCreateOptions() {
+    var xhrOverride = new XMLHttpRequest();
+    xhrOverride.responseType = 'arraybuffer';
+
+    return $.ajax({
+      url: '/webauthn/register',
+      xhr: () => xhrOverride,
+      dataType: 'x-binary',
+      converters: { '* x-binary': (value) => value }
```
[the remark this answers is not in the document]
benthorner: I did mention this in the commit message - it's [this StackOverflow article](https://stackoverflow.com/questions/12394622/does-jquery-ajax-or-load-allow-for-responsetype-arraybuffer) - but the comment is a bit hidden 😬.

- `dataType: 'x-binary'` is overriding jQuery's auto-detection.
- `converters: ...` is telling jQuery what to do with that data type.

We could probably use `'* text': ...` without the `dataType`, as I suspect that's what jQuery infers it as. I think it's better to be explicit, as the default jQuery behaviour is so "magical" (and not in a good way).

app/webauthn_server.py
benthorner: Yeah, this is [the timeout for the presence check](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions/timeout).

Keeping it short makes testing the timeout easier, but we can always change it locally for testing. I originally thought 10 seconds would be enough, even for someone with a motor or visual impairment. Maybe not though if it's the first time they're seeing the dialogue. I've amended the second commit to use 30 seconds 👍.

app/main/views/webauthn_credentials.py
```diff
@@ -0,0 +1,43 @@
...
+    server = current_app.webauthn_server
+
+    registration_data, state = server.register_begin(
+        {
+            "id": bytes(current_user.id, 'utf-8'),
+            "name": current_user.email_address,
+            "displayName": current_user.name,
+        },
+        credentials=[],  # TODO: get from user
+        user_verification="discouraged",
```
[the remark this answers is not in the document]
benthorner: This is referring to [one of the authenticatorSelection settings](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions/authenticatorSelection).

It means we don't require the user to setup a PIN, or other form of verification, on their authenticator. I've amended the second commit with a quick commit to hint at the meaning.

tests/app/main/views/test_webauthn_credentials.py
benthorner: Sort of, it applies to both: `init_app` needs to be called in order to pick up the change.

I've amended the second commit to move the comment and group the code together a bit more.

tests/app/models/test_webauthn_credential.py
benthorner: The decoded version wouldn't be very useful, I'm afraid, as it's also CBOR-encoded. Even the CBOR-decoded version is pretty unreadable: it is a dict, but most of the values are still long byte sequences.

> would we want to hide it away

I was wondering if we had a place to put this sort of thing. It's basically a file fixture. I don't think it belongs in `conftest.py`, as it's only used once. On the other hand, I think it's OK in this file too.

Where do you think it could go instead?

tests/app/test_webauthn_server.py
benthorner: `rp` stands for "relying party", which isn't very intuitive but unfortunately that's the terminology used by the standard. Most of the WebAuthn docs I've seen use "rp" instead of "relying party". Since it relates to `rp.id` in the code, I don't think we should stray from this name; we just need to be familiar with WebAuthn. 

I've amended the second commit to expand it to `_relying_party_id(` in the tests.

tests/app/main/views/test_user_profile.py
```diff
@@ -11,22 +11,35 @@
...
+def test_overview_page_shows_security_keys_for_platform_admin(
+    mocker,
+    client_request,
+    platform_admin_user
+):
+    client_request.login(platform_admin_user)
+    mocker.patch('app.user_api_client.get_webauthn_credentials_for_user', return_value=[])
+    page = client_request.get('main.user_profile')
+    security_keys_row = page.select_one('#security-keys')
+    assert ' '.join(security_keys_row.text.split()) == 'Security keys 0 Change'
```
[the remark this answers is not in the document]
benthorner: Sure, that's fair. I've amended the first commit for test for no keys and one key in the same test.

benthorner:
Thanks @CrystalPea! I've responded to all your comments; just a few left where I think there might be more to discuss.

> Do we want to integrate with API in this PR? I think we could?

We certainly could, but I decided not to because:

- I wanted to get this up for review quickly, and integrating the API will take more time.
- Integrating with that PR kind of blocks this one, or at least makes it more complicated to try out.
- We should be able to do the integration in parallel with other work, but only once this PR is merged.

> Also, H1 header seems to be a bit low here

I see what you mean. However, I checked against the other settings pages, and it's consistent with those. The one page it's _not_ consistent with is the platform admin one, where the heading is closer to the top because it uses [a different technique](https://github.com/alphagov/notifications-admin/blob/78824f54fdf7288172387f8ed3b332d26d74e083/app/templates/views/user-profile/disable-platform-admin-view.html#L22) to set the heading. We seem to have at least three different ways of putting a heading on a page in the Admin app, which is just bonkers! So I think we should leave it as-is, but maybe fix the "disable platform admin" page to be consistent.

app/templates/views/user-profile/security-keys.html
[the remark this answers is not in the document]
CrystalPea: ok 👍🏼

app/assets/javascripts/registerSecurityKey.js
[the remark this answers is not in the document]
CrystalPea: thanks!

tests/app/models/test_webauthn_credential.py
CrystalPea: Hmmm you are right we don't have an optimal place to put it. Maybe at the top of the file as a constant? I think we have put some bulky things at top of file like that in other places.

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -0,0 +1,69 @@
...
+                // $.ajax seems to log by itself, but that's not visible
+                alert('Error during registration. Please try again.');
+              }
+            });
+        });
+    };
+  };
+
+  function fetchWebAuthnCreateOptions() {
+    var xhrOverride = new XMLHttpRequest();
```
tombye, line 43: Looking at the support for [window.fetch]() and the [web auth API](https://caniuse.com/webauthn), we could assume fetch is available if web auth is. You mentioned in your commit adding this code that fetch would be preferred so this might be an opportunity to use it instead.

In my opinion, it would be useful to:
- not have to have a hack to make `$.ajax` work
- not use a 'promise-like' thing like `$.ajax` with actual Promises if we can help it

tests/app/test_webauthn_server.py
CrystalPea: I think that expansion helps, because we have it as "relying_party" elsewhere in the code, thanks!

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -0,0 +1,69 @@
...
+  function postWebAuthnCreateResponse(response, csrf_token) {
+    return $.ajax({
+      url: '/webauthn/register',
+      method: 'POST',
+      headers: {
+        'X-CSRFToken': csrf_token
+      },
+      processData: false,
+      contentType: 'application/cbor',
+      data: CBOR.encode({
```
tombye, line 63: Would be nice to use `window.CBOR` here, to match your use on line 12
```suggestion
      data: window.CBOR.encode({
```

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -0,0 +1,146 @@
...
+      </a>`
+
+    button = document.querySelector('[data-module="register-security-key"]')
+    window.GOVUK.modules.start()
+  })
+
+  test('creates a new credential and reloads', (done) => {
+    // pretend window.navigator.credentials exists in test env
+    // defineProperty is used as window.navigator is read-only
+    Object.defineProperty(window.navigator, 'credentials', {
```
tombye, line 29: This and the other blocks that redefine DOM APIs should really have their original values reset after the test that uses them. Just to avoid any unexpected side effects in tests that run after these.

CrystalPea:
> I wanted to get this up for review quickly, and integrating the API will take more time.
Integrating with that PR kind of blocks this one, or at least makes it more complicated to try out.
We should be able to do the integration in parallel with other work, but only once this PR is merged.
Also, H1 header seems to be a bit low here

Fair, we can do it separately that's true.

> I see what you mean. However, I checked against the other settings pages, and it's consistent with those. The one page it's not consistent with is the platform admin one, where the heading is closer to the top because it uses a different technique to set the heading. We seem to have at least three different ways of putting a heading on a page in the Admin app, which is just bonkers! So I think we should leave it as-is, but maybe fix the "disable platform admin" page to be consistent.

I think it would be up to @quis to decide which way the consistency should go, from visual standpoint, no?

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -0,0 +1,146 @@
...
+    // pretend window.location exists in test env
+    // defineProperty is used as window.location is read-only
+    Object.defineProperty(window, 'location', {
+      // signal that the async promise chain was called
+      value: { reload: () => done() },
+      // allow global property to be redefined in other tests
+      writable: true,
+    })
+
+    jest.spyOn(window.$, 'ajax').mockImplementation((options) => {
```
tombye, line 57: This and other mocked spies should be reset after their test is run, to avoid any unexpected cases in future tests.

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -0,0 +1,146 @@
...
+    button.click()
+  })
+
+  test('alerts if fetching WebAuthn options fails', (done) => {
+    jest.spyOn(window.$, 'ajax').mockImplementation((options) => {
+      return Promise.reject('error')
+    })
+
+    jest.spyOn(window, 'alert').mockImplementation((msg) => {
+      done()
```
tombye, line 83: Jest's docs have `done` as the last line of code run, after any assertions, so we should probably do the same.

See: https://jestjs.io/docs/asynchronous#callbacks

tombye (changes requested):
I read most of the code, to understand what was happening, but kept my review to the frontend bits.

I've left some comments on a few things but this looks good.

The only largish thing I noticed was the lack of states for users without JS and those whose devices/browsers don't support web auth. We've chatted about how to message users in those scenarios out of this pull request so the work is planned but worth mentioning it here as a record of that.

app/assets/javascripts/registerSecurityKey.js
benthorner: Definitely, great idea! It's quite a big change, so I've added a couple more commits to switch to `fetch` (the 4th and 5th commits). How does it look now?

app/assets/javascripts/registerSecurityKey.js
benthorner: Sure, I've amended the second commit to do that 👍.

tests/javascripts/registerSecurityKey.test.js
benthorner: That makes sense. It kind of slipped my mind (/we don't seem to do it [in other tests](https://github.com/alphagov/notifications-admin/blob/78824f54fdf7288172387f8ed3b332d26d74e083/tests/javascripts/enhancedTextbox.test.js#L282)). In the 4th commit and onwards, I've shifted the test code to restore original values (if there were any).

How does it look now?

tests/javascripts/registerSecurityKey.test.js
benthorner: Yup, I've addressed this as part of https://github.com/alphagov/notifications-admin/pull/3878#discussion_r631222693.

I'll resolve this one and we can continue any further discussion there.

tests/javascripts/registerSecurityKey.test.js
benthorner: Good spot, that makes sense. I accidentally put the fix in the 4th commit, but it's not a big change so I'll leave it there and just mention it in the commit message.

benthorner:
Thanks for the review @tombye 💯. I think I've addressed all your comments.

> the lack of states for users without JS and those whose devices/browsers don't support web auth

We do indeed have a plan to address this: I've made a note to do it as part of the card for this PR. It should be a quick, best practice thing to implement, so we should just do it (vs. saying it's not MVP and then forgetting).

> I think it would be up to @quis to decide which way the consistency should go, from visual standpoint, no?

@CrystalPea I guess it is. However, I think that's out-of-scope for this PR: once we make a decision, we'll need to change at least one other HTML file to make them consistent. The new HTML file is consistent with four other HTML files, so I'm not introducing anything new here. I'm happy to make all of them consistent in a follow-up PR, so we have fewer ways of doing headings 😄. But I don't think it should block this one going in.

tests/app/models/test_webauthn_credential.py
benthorner: Sure, I've amended the second commit with the data in constant at the top. I realised the `clientDataJSON` is quite readable when decoded, so I've swapped it for that. I've moved the session data up there two, since the three pieces of data need to match / go together.

I guess having it at the top is easier to read if you have wrapping turned on in your editor.

How does it look now?

requirements.txt
```diff
@@ -29,12 +29,16 @@ cachetools==4.2.1
...
     # via requests
+cffi==1.14.5
+    # via cryptography
 chardet==4.0.0
     # via requests
 click==7.1.2
     # via flask
 colorama==0.4.3
     # via awscli
+cryptography==3.4.7
```
klssmith, line 40: If we need a version of cryptography that's > 3.4 we'll need to work out how to get round the build failing on the PaaS due to not having Rust installed

tests/app/models/test_webauthn_credential.py
CrystalPea: Thank you, looks good 👍🏼

CrystalPea (approved):

app/templates/views/user-profile/security-keys.html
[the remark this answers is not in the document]
karlchillmaid: Agree. I'll review another time

requirements.txt
benthorner: Nice spot / save! I've amended the second commit to pin it below 3.4 and explained why in comments. Deploying to Preview shows it works 🎉  (with a bit of variability, since we're using an in-memory store that's not shared between the two instances running in Preview).

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -77,16 +77,23 @@ describe('Register security key', () => {
...
-        return Promise.resolve()
+        return Promise.resolve({ ok: true })
       }
     })
 
     button.click()
   })
 
-  test('alerts if fetching WebAuthn options fails', (done) => {
+  test.each([
```
tombye, line 80: This is a really nice way of describing the different error states 👍🏻

app/assets/javascripts/registerSecurityKey.js
tombye: Looks great, and the code and tests seem to describe the various scenarios a lot better I think 👍🏻

tests/javascripts/registerSecurityKey.test.js
tombye: That test without them you spotted should also have them. I'll put a fix up soon but do say if you know of any others.

This looks good now.

benthorner: Thanks!

> That test without them you spotted should also have them.

I think it's done already: that test [also calls `teardown.js`](https://github.com/alphagov/notifications-admin/blob/78824f54fdf7288172387f8ed3b332d26d74e083/tests/javascripts/enhancedTextbox.test.js#L8), which is where I added the hook.

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -0,0 +1,56 @@
+(function(window) {
+  "use strict";
+
+  window.GOVUK.Modules.RegisterSecurityKey = function() {
+    this.start = function(component) {
+      $(component)
```
tombye, line 6: I should have mentioned this before but can we put in a guard at the top to make it return if the needed web APIs aren't available?
```suggestion
      if (!('fetch' in window) || !('credentials' in window.navigator)) { return }
      $(component)
```

benthorner: We did [discuss it before](https://github.com/alphagov/notifications-admin/pull/3878#pullrequestreview-657903585), but not from a JS point of view. I'll leave it as-is for now, and make it part of the follow-up PR so it's all in one place (I've made a not of your suggested change on the card).

tombye:
@benthorner I'm happy all my comments have been addressed. I added [an extra one](https://github.com/alphagov/notifications-admin/pull/3878#discussion_r631712899), sorry 😬, which the chat with you about browser support/fallbacks made me realise was needed. Everything else is 💯 so thanks for all the work.

tombye (approved):

klssmith:
I was happy with this too. I just had one question - when Chrome gives you the "Allow this site to see your security key?" window are we saving exactly the same data when someone clicks "Skip" vs when they click "Allow"?

benthorner:
Thanks all, it's time merge! 🎉

> I just had one question - when Chrome gives you the "Allow this site to see your security key?" window are we saving exactly the same data when someone clicks "Skip" vs when they click "Allow"?

Good spot @klssmith! I ran a few tests locally and clicking "Skip" means (referring to [the API docs](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAttestationResponse)):

- `authData.attestedCredentialData.aaguid` is set to `00000000000000000000000000000000` (masking the make/model).
- `fmt` is `none`, which goes with...
- `attStmt` is `{}`

If we verified attestation in future, then this wouldn't be allowed through. And it would be a red flag if we ever audit the credentials. For now I think it's OK to request `direct` attestation, and allow people to opt-out.

PR #3879, opened 2021-05-11
Add an option to set a service to the government channel for emergency alerts

quis:
We have been asked to support the government channel so that:
- it can be tested
- the option to use it is available for the most severe of emergencies, where the public’s choice to opt-out is outweighed by the widespread risk to life

In this PR I have also split the choices into multiple steps, with a confirmation page, so that whoever is using this is really clear about the implications of the choice they are selecting.

# Before

image

# After 

image
image

***

image
image
image

***

image
image

***

image
image

***

image
image
image
image
image


servingUpAces:
Looking good. I have a minor concern, do the labels convey the full meaning of what happens when selected?
"Test channel" is a live service that can send messages on all or selected channel. 
"Live channel" is a live service that can send messages on the severe broadcast channel.
"Government channel" is a live service that can send messages on the government channel.

klssmith:
The api part is ready now if you want to try both parts together https://github.com/alphagov/notifications-api/pull/3241

quis:
@servingUpAces Hopefully the confirmation page makes this clear by spelling out the consequences of what you’re about to do. I‘ve also added _(all networks)_ to the _Test_ label in you haven’t specified a network, to make it clear that alerts will go out to the networks.

I want most users to just worry about the difference between ‘Training’ and ‘Live’, since that’s generally what they’ll be given, and not have to know what the severe channel is, for example.

quis:
I also wanna do some work to reiterate these messages at the time of approving an alert, since that just looks like this at the moment: 
image

servingUpAces (approved):

PR #3881, opened 2021-05-12
Add a confirmation checkbox for live broadcasts

quis:
We want people to be really sure before sending a live broadcast, not just clicking through the green buttons.

This commit adds a checkbox which explains exactly the consequences of what they’re about to do, tailored to:
- the channel they’re on
- the area chosen by the person creating the alert.

***

image

***

image

***

image

***

image

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -22,6 +22,15 @@ from app.models.broadcast_message import BroadcastMessage, BroadcastMessages
 from app.utils import service_has_permission, user_has_permissions
 
 
+def get_back_link_endpoint():
+    return {
+        'main.view_current_broadcast': '.broadcast_dashboard',
+        'main.view_previous_broadcast': '.broadcast_dashboard_previous',
+        'main.view_rejected_broadcast': '.broadcast_dashboard_rejected',
+        'main.approve_broadcast_message': '.broadcast_dashboard',
+    }[request.endpoint]
+
+
 @main.route('/services/<uuid:service_id>/broadcast-tour/<int:step_index>')
 @user_has_permissions()
 @service_has_permission('broadcast')
@@ -389,17 +398,11 @@ def view_broadcast(service_id, broadcast_message_id):
                 broadcast_message_id=broadcast_message.id,
             ))
 
-    back_link_endpoint = {
-        'main.view_current_broadcast': '.broadcast_dashboard',
-        'main.view_previous_broadcast': '.broadcast_dashboard_previous',
-        'main.view_rejected_broadcast': '.broadcast_dashboard_rejected',
-    }[request.endpoint]
-
     return render_template(
         'views/broadcast/view-message.html',
         broadcast_message=broadcast_message,
         back_link=url_for(
-            back_link_endpoint,
+            get_back_link_endpoint(),
             service_id=current_service.id,
         ),
         form=ConfirmBroadcastForm(
@@ -442,6 +445,16 @@ def approve_broadcast_message(service_id, broadcast_message_id):
         ))
     elif form.validate_on_submit():
         broadcast_message.approve_broadcast()
+    else:
+        return render_template(
+            'views/broadcast/view-message.html',
+            broadcast_message=broadcast_message,
+            back_link=url_for(
+                get_back_link_endpoint(),
+                service_id=current_service.id,
+            ),
+            form=form,
+        )
 
     return redirect(url_for(
         '.view_current_broadcast',
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -22,7 +22,7 @@ from app.models.broadcast_message import BroadcastMessage, BroadcastMessages
 from app.utils import service_has_permission, user_has_permissions
 
 
-def get_back_link_endpoint():
+def _get_back_link_from_view_broadcast_endpoint():
     return {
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_broadcast.py
```diff
@@ -1817,6 +1821,116 @@ def test_view_pending_broadcast_from_api_call(
...
+        'app.broadcast_message_api_client.get_broadcast_message',
+        return_value=broadcast_message_json(
+            id_=fake_uuid,
+            service_id=SERVICE_ONE_ID,
+            template_id=None,
+            created_by_id=None,
+            status='pending-approval',
+        ),
+    )
+    service_one['permissions'] += ['broadcast']
```
benthorner, line 1855: Minor: we could DRY this up with a `broadcast_service` fixture.

tests/app/main/views/test_broadcast.py
```diff
@@ -1817,6 +1821,116 @@ def test_view_pending_broadcast_from_api_call(
...
+    'test',
+    'severe',
+    'government',
+))
+@pytest.mark.parametrize('post_data', (
+    pytest.param({}, marks=pytest.mark.xfail),
+    {'confirm': 'y'},
+))
+@freeze_time('2020-02-22T22:22:22.000000')
+def test_confirm_approve_non_training_broadcasts(
```
benthorner, line 1882: Do we need this test? The second commit moves the fail scenario to a separate test, and what remains looks the same as one of the `test_request_approval` scenarios.

tests/app/main/views/test_broadcast.py
```diff
@@ -1931,6 +1926,58 @@ def test_confirm_approve_non_training_broadcasts(
     )
 
 
+@pytest.mark.parametrize('channel', (
```
benthorner, line 1929: Do we need repeat it over the channel? I can see that proves the behaviour isn't dependent on the channel, but there are other variables we could equally prove that for, which we're not testing here.

app/main/views/broadcast.py
```diff
@@ -22,6 +22,15 @@
 from app.utils import service_has_permission, user_has_permissions
 
 
+def get_back_link_endpoint():
```
benthorner, line 25: This is ambiguous, as it's not used in a lot of the views. We've also got [a bit of a precedent for naming this sort of method](https://github.com/alphagov/notifications-admin/blob/06cc5b58c7367b6e29dc71a5851492f20ef53a55/app/main/views/broadcast.py#L238). How about `_get_back_link_from_view_broadcast_endpoint`?

tests/app/main/views/test_broadcast.py
```diff
@@ -1931,6 +1926,58 @@ def test_confirm_approve_non_training_broadcasts(
...
+        _data={},
+        _expected_status=200,
+    )
+
+    error_message = page.select_one('form.banner .govuk-error-message')
+    assert error_message['id'] == 'confirm-error'
+    assert normalize_spaces(error_message.text) == (
+        'Error: You need to confirm that you understand'
+    )
+    assert page.select_one('form.banner input#confirm')['aria-describedby'] == (
```
benthorner, line 1973: Minor: I can' find any other tests for this, and it's not specific to this form. Do we need to test it?

tests/app/main/views/test_broadcast.py
```diff
@@ -2111,39 +2158,44 @@ def test_view_only_user_cant_approve_broadcast(
...
+            '.view_current_broadcast',
+            broadcast_message_id=sample_uuid,
+        ), 302),
+        (True, 'pending-approval', {}, True, partial(
+            url_for,
+            '.broadcast_tour',
+            step_index=6,
+        ), 302),
+        (False, 'pending-approval', {}, False, (
+            lambda service_id, _external: None
```
benthorner, line 2175: Do we need the expected status? It's kind of implicit in the lack of redirect.

In fact, do we need this scenario here at all? I think it's covered by `test_confirm_approve_non_training_broadcasts_errors_if_not_ticked`.

benthorner (commented):
Much safer 🔒.

I've got a couple of questions about the tests, as I'm not sure I understand what's being tested where, and why. My comments err on the side of removing stuff, but I may have missed something.

tests/app/main/views/test_broadcast.py
quis: Good point, removed in bf609c7a4

tests/app/main/views/test_broadcast.py
quis: Tested in 00d4798af

app/main/views/broadcast.py
quis: Renamed per your suggestion in 822e83fa8

tests/app/main/views/test_broadcast.py
quis: Removed in d9311bc91

tests/app/main/views/test_broadcast.py
quis: For a separate PR maybe

tests/app/main/views/test_broadcast.py
quis: Removed in 748f81628

benthorner (approved):
Thanks for working through my comments! You've got a few test failures, for which the fixes look pretty straightforward, so I'll approve this now.

PR #3882, opened 2021-05-12
Rename "app_" fixture to just "app"

benthorner:
This naming was introduced in 2016 without explanation [1]. I find it
confusing because:

- It's reminiscent of "_app", which is a Python convention indicating
the variable is internal, so maybe avoid using it.

- It suggests there's some other "app" fixture I should be using (there
isn't, though).

Our other repos use the app name (e.g. "notify_api") or just "app" for
the equivalent fixture. I'm not aware suffixing an underscore like this
has any meaning in Python, or in this codebase.

Let's fix the weird naming!

[1]: https://github.com/alphagov/notifications-admin/commit/3b1d521c10a515dab85700a1103912cf47ed414c

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -559,7 +559,7 @@ def single_notification_json(
 
 
 def validate_route_permission(mocker,
-                              app_,
+                              app,
                               method,
                               response_code,
                               route,
@@ -582,8 +582,8 @@ def validate_route_permission(mocker,
     mocker.patch('app.service_api_client.get_service', return_value={'data': service})
     mocker.patch('app.models.user.Users.client_method', return_value=[usr])
     mocker.patch('app.job_api_client.has_jobs', return_value=False)
-    with app_.test_request_context():
-        with app_.test_client() as client:
+    with app.test_request_context():
+        with app.test_client() as client:
             client.login(usr)
             if session:
                 with client.session_transaction() as session_:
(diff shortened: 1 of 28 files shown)


quis:
It’s a Python convention for giving a variable, method or property the same name as a reserved keyword. You can see an example of this with SQLAlchemy’s `all_`, `and_` and `any_` methods here: https://docs.sqlalchemy.org/en/13/core/sqlelement.html

I don’t think actually applies in this case though, because `app_` is not a reserved keyword.

benthorner:
Ooh, thanks @quis. I put my effort into doing the PR rather than checking my own knowledge. An easy Google search lead me to [PEP 8](https://pep8.org/#descriptive-naming-styles), which shows more formally that it is "a thing".

Would you like me to change anything about this PR?

quis (approved):

leohemsted:
One thing of note is that in a few (five) test files, we call `import app` - not sure if we called it app_ to avoid conflicts there. But if things work they work and it looks better like this 🤷‍♂️

benthorner:
Aha, great spot @leohemsted. I think it doesn't affect the tests because they use one or the other (not both), and I'm guessing flake isn't smart enough to spot potentially redundant imports.

Still, there's definitely a potential for clashes, so I'm closing this in favour of https://github.com/alphagov/notifications-admin/pull/3883.

PR #3883, opened 2021-05-12
Rename "app_" fixture to "notify_admin"

benthorner:
This naming was introduced in 2016 without explanation [1]. I find it
confusing because:

- It's reminiscent of "_app", which is a Python convention indicating
the variable is internal, so maybe avoid using it.

- It suggests there's some other "app" fixture I should be using (there
isn't, though).

The Python style guide describes using an underscore suffix to avoid 
clashes with inbuilt names [1], which is sort of applicable if we need 
to import the "app" module [2]. However, we can also avoid clashes by 
choosing a different name, without the strange underscore.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -559,7 +559,7 @@ def single_notification_json(
 
 
 def validate_route_permission(mocker,
-                              app_,
+                              notify_admin,
                               method,
                               response_code,
                               route,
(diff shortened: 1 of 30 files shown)


CrystalPea (approved):

CrystalPea:
I think you could change PR and commit messages, as you are not renaming it to `app`, you are renaming it to `notify_admin`, right?

benthorner:
Good spot, thanks @CrystalPea! I've updated the commit/PR 👍.

PR #3885, opened 2021-05-13
Add a version of the tour for live services

quis:
At the moment if you’re invited to a live broadcast service you get the training mode tour. This is misleading, and could make people think they weren’t in danger of sending a real alert.

This commit adds a short, 2 step tour for users invited to a live broadcast service.

***

image

***

image

***

We should add some way of labelling/differentiating broadcast services on the ‘choose service’ page in the future.


benthorner:
I had a quick look at this, and have a more fundamental question: most of the tour pages look helpful irrespective of the service status, except the last one; so could we just skip that one if the service is live?

tests/app/main/views/test_accept_invite.py
```diff
@@ -55,6 +55,44 @@ def test_existing_user_accept_invite_calls_api_and_redirects_to_dashboard(
...
+    mock_get_users_by_service,
+    mock_accept_invite,
+    mock_add_user_to_service,
+    mocker,
+    mock_events,
+    mock_get_user,
+    trial_mode,
+    expected_endpoint,
+):
+    service_one['permissions'] = ['broadcast']
```
benthorner, line 76: Minor: this does [keep cropping up](https://github.com/alphagov/notifications-admin/pull/3881#discussion_r633557198), doesn't it ;-).

tests/app/main/views/test_broadcast.py
```diff
@@ -286,6 +289,28 @@ def test_broadcast_tour_page_4_shows_service_name(
     )
 
 
+@pytest.mark.parametrize('step_index', (
+    1,
+    2,
+))
+def test_live_broadcast_tour_shows_service_name_and_switch(
```
benthorner, line 296: Slightly out-of-scope: the previous test title is wrong ("page 4").

Minor: based on [my other comment on the navigation](https://github.com/alphagov/notifications-admin/pull/3885/files#diff-52f6655a83670dbb811bb81ad5b5a4dc575a59b964b26508ff17c1b764fc47c9), does it make sense to combine this with the previous test, using a `trial_mode` parameter?

tests/app/main/views/test_accept_invite.py
```diff
@@ -55,6 +55,44 @@ def test_existing_user_accept_invite_calls_api_and_redirects_to_dashboard(
     )
 
 
+@pytest.mark.parametrize('trial_mode, expected_endpoint', (
+    (True, '.broadcast_tour'),
+    (False, '.broadcast_tour_live'),
+))
+def test_broadcast_service_shows_tour(
```
benthorner, line 62: Good spot adding this 👍.

benthorner (commented):
Thanks for explaining the reasoning, that makes sense 👍.

I found one omission of a conditional to fix/test, and otherwise just a few, mostly minor, comments.

app/templates/service_navigation.html
```diff
@@ -0,0 +1,25 @@
+<div class="navigation-service">
```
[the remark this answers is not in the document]
quis: Refactoring and rationale in e4eddccb7

tests/app/main/views/test_accept_invite.py
quis: ;-)

tests/app/main/views/test_broadcast.py
quis: Combined and renamed in a34a00c33

app/main/views/invites.py
```diff
@@ -37,12 +37,14 @@ def accept_invite(token):
         return render_template('views/cancelled-invitation.html',
                                from_user=invited_user.from_user.name,
                                service_name=service.name)
-
     if invited_user.status == 'accepted':
         session.pop('invited_user_id', None)
         service = Service.from_id(invited_user.service)
         if service.has_permission('broadcast'):
-            return redirect(url_for('main.broadcast_tour', service_id=service.id, step_index=1))
+            if service.live:
```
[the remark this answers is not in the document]
quis: Added in 8805280bb

app/templates/service_navigation.html
```diff
@@ -1,31 +1,44 @@
+{% macro navigation_service_name(service) %}
+  <div class="navigation-service-name govuk-!-font-weight-bold">
+    {{ service.name }}
+    {% if not service.active %}
+      <span class="navigation-service-type navigation-service-type--suspended">Suspended</span>
+    {% elif service.has_permission('broadcast') %}
+      {{ broadcast_service_name_tag(
+        service.trial_mode,
```
benthorner, line 8: Minor: could we just pass the service, instead of splitting up its data? I find that helpful because it retain knowledge of the origin of the data, confirming its not be pre-processed in some way.

app/templates/views/service-settings/service-confirm-broadcast-account-type.html
```diff
@@ -15,30 +16,19 @@
         'Confirm emergency alert settings',
         back_link=url_for('.service_set_broadcast_channel', service_id=current_service.id)
       ) }}
+      <p class="govuk-body">
```
benthorner, line 19: Nice!

app/main/views/invites.py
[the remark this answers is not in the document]
benthorner: Thanks 👍. I've left a follow-up comment on the new code.

app/templates/service_navigation.html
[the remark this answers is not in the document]
benthorner: Much better!

tests/app/main/views/test_broadcast.py
benthorner: We've lost the check for switching service now, though. I thought that was important as it's functionality that's referred to in the graphic.

I was thinking you could use `trial_mode` to predicate the endpoint and the extended assertion. But maybe a separate test is just easier.

(Hence the minor suggestion 😄.)

tests/app/main/views/test_accept_invite.py
```diff
@@ -611,6 +612,54 @@ def test_new_invited_user_verifies_and_added_to_service(
     assert page.find('h1').text == 'Dashboard'
 
 
+@pytest.mark.parametrize('service_permissions, trial_mode, expected_endpoint, extra_args', (
+    ([], True, 'main.service_dashboard', {}),
+    ([], False, 'main.service_dashboard', {}),
+    (['broadcast'], True, 'main.broadcast_tour', {'step_index': 1}),
+    (['broadcast'], False, 'main.broadcast_tour_live', {'step_index': 1}),
+))
+def test_new_invited_user_is_redirected_to_correct_place(
```
benthorner, line 621: Could this test be simplified, and perhaps relocated to [the ones in test_verify.py](https://github.com/alphagov/notifications-admin/blob/5de1c4f6ca8d6a0210700a5bdae4b28071a982e8/tests/app/main/views/test_verify.py#L49)? Those tests are for the same endpoint, but require less setup.

I can see there's [one other test for this endpoint in here](https://github.com/alphagov/notifications-admin/blob/5de1c4f6ca8d6a0210700a5bdae4b28071a982e8/tests/app/main/views/test_accept_invite.py#L591), which seems to be testing the combination of endpoints; perhaps that's why it's in this test file.

benthorner (approved):
Nice refactoring of the broadcast tag 👌. I think this is good to go now, albeit with a few minor comments on the new commits; of these, the one about the `main.verify` test interests me most, as it could start an awkward trend.

tests/app/main/views/test_accept_invite.py
quis: The one in `test_verify.py` is testing with an existing user. This is testing with a user who doesn’t yet have an account.

app/templates/service_navigation.html
quis: This doesn’t always take these parameters from the service, on the confirm page it takes them from the submitted form data instead:
https://github.com/alphagov/notifications-admin/blob/658e40b6c5ce5ad41ee1799796a499c5432c37dc/app/main/views/service_settings.py#L399-L401

…and because the form uses slightly different property names we can’t use polymorphism here.

benthorner: Ah yes, thanks for pointing it out. I also discovered it while doing https://github.com/alphagov/notifications-admin/pull/3893. Nothing to change here then 👍.

tests/app/main/views/test_accept_invite.py
benthorner: Ah, I see: the new redirect behaviour only happens if [`InvitedUser.from_session` is truthy](https://github.com/alphagov/notifications-admin/blob/658e40b6c5ce5ad41ee1799796a499c5432c37dc/app/main/views/verify.py#L79), which requires an `invited_user_id`, which needs to be set by [the `accept_invite` endpoint](https://github.com/alphagov/notifications-admin/blob/658e40b6c5ce5ad41ee1799796a499c5432c37dc/app/main/views/invites.py#L48). It looks like [there's still a leaner way to test the redirect, by setting the session manually](https://github.com/alphagov/notifications-admin/blob/658e40b6c5ce5ad41ee1799796a499c5432c37dc/tests/app/main/views/test_verify.py#L193). Ideally we might do that, although it's also refreshing to see some integration-level tests in the repo, so I don't feel strongly either way.

PR #3892, opened 2021-05-18
Let admin user update and delete their security key

CrystalPea:
Pivotal ticket: https://www.pivotaltracker.com/story/show/178066154

### "Manage" link on the page listing the keys:
Screenshot 2021-05-18 at 15 34 13

### Manage key page
Screenshot 2021-05-18 at 15 34 31

### Confirm delete key dialog
Screenshot 2021-05-18 at 15 35 01

### "Can't delete last credential" message
Screenshot 2021-05-18 at 15 35 08

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2506,8 +2506,8 @@ class BroadcastAreaFormWithSelectAll(BroadcastAreaForm):
         return self.areas.data
 
 
-class ChangeNameOfSecurityKey(StripWhitespaceForm):
-    name_of_key = GovukTextInputField(
+class ChangeSecurityKeyNameForm(StripWhitespaceForm):
+    security_key_name = GovukTextInputField(
         'Name of key',
         validators=[
             DataRequired(message='Cannot be empty'),
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2504,3 +2504,13 @@ class BroadcastAreaFormWithSelectAll(BroadcastAreaForm):
         if self.select_all.data:
             return [self.select_all.area_slug]
         return self.areas.data
+
+
+class ChangeNameOfSecurityKey(StripWhitespaceForm):
+    name_of_key = GovukTextInputField(
+        'Name of key',
+        validators=[
+            DataRequired(message='Cannot be empty'),
+            MustContainAlphanumericCharacters(),
+            Length(max=255, message='Name of key must be 255 characters or fewer')
+        ])
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -237,3 +237,11 @@ def user_profile_security_keys():
     return render_template(
         'views/user-profile/security-keys.html',
     )
+
+
+@main.route("/user-profile/security-keys/<uuid:key_id>/manage", methods=['GET'])
+@user_is_platform_admin
+def user_profile_manage_security_key():
+    return render_template(
+        'views/user-profile/manage-security-key.html',
+    )
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -11,6 +11,7 @@ from flask import (
     url_for,
 )
 from flask_login import current_user
+from notifications_python_client.errors import HTTPError
 from notifications_utils.url_safe_token import check_token
 
 from app import user_api_client
@@ -283,8 +284,18 @@ def user_profile_manage_security_key(key_id):
 @main.route("/user-profile/security-keys/<uuid:key_id>/delete", methods=['POST'])
 @user_is_platform_admin
 def user_profile_delete_security_key(key_id):
-    user_api_client.delete_webauthn_credential_for_user(
-        user_id=current_user.id,
-        credential_id=key_id
-    )
+
+    try:
+        user_api_client.delete_webauthn_credential_for_user(
+            user_id=current_user.id,
+            credential_id=key_id
+        )
+    except HTTPError as e:
+        message = "Cannot delete last remaining webauthn credential for user"
+        if e.message == message:
+            flash("You cannot delete your last security key.")
+            return redirect(url_for('.user_profile_manage_security_key', key_id=key_id))
+        else:
+            raise e
+
     return redirect(url_for('.user_profile_security_keys'))
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -264,11 +264,12 @@ def user_profile_manage_security_key(key_id):
     form = ChangeNameOfSecurityKey(name_of_key=security_key["name"])
 
     if form.validate_on_submit():
-        user_api_client.update_webauthn_credential_for_user(
-            user_id=current_user.id,
-            credential_id=key_id,
-            new_name_for_credential=form.name_of_key.data
-        )
+        if form.name_of_key.data != security_key["name"]:
+            user_api_client.update_webauthn_credential_for_user(
+                user_id=current_user.id,
+                credential_id=key_id,
+                new_name_for_credential=form.name_of_key.data
+            )
         return redirect(url_for('.user_profile_security_keys'))
 
     if (request.endpoint == "main.user_profile_confirm_delete_security_key"):
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -3,8 +3,10 @@ import json
 from flask import (
     abort,
     current_app,
+    flash,
     redirect,
     render_template,
+    request,
     session,
     url_for,
 )
@@ -240,7 +242,16 @@ def user_profile_security_keys():
     )
 
 
-@main.route("/user-profile/security-keys/<uuid:key_id>/manage", methods=['GET', 'POST'])
+@main.route(
+    "/user-profile/security-keys/<uuid:key_id>/manage",
+    methods=['GET', 'POST'],
+    endpoint="user_profile_manage_security_key"
+)
+@main.route(
+    "/user-profile/security-keys/<uuid:key_id>/delete",
+    methods=['GET'],
+    endpoint="user_profile_confirm_delete_security_key"
+)
 @user_is_platform_admin
 def user_profile_manage_security_key(key_id):
     security_keys = user_api_client.get_webauthn_credentials_for_user(current_user.id)
@@ -259,8 +270,21 @@ def user_profile_manage_security_key(key_id):
         )
         return redirect(url_for('.user_profile_security_keys'))
 
+    if (request.endpoint == "main.user_profile_confirm_delete_security_key"):
+        flash("Are you sure you want to delete this security key?", 'delete')
+
     return render_template(
         'views/user-profile/manage-security-key.html',
         security_key=security_key,
         form=form
     )
+
+
+@main.route("/user-profile/security-keys/<uuid:key_id>/delete", methods=['POST'])
+@user_is_platform_admin
+def user_profile_delete_security_key(key_id):
+    user_api_client.delete_webauthn_credential_for_user(
+        user_id=current_user.id,
+        credential_id=key_id
+    )
+    return redirect(url_for('.user_profile_security_keys'))
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -20,8 +20,8 @@ from app.main.forms import (
     ChangeEmailForm,
     ChangeMobileNumberForm,
     ChangeNameForm,
-    ChangeNameOfSecurityKey,
     ChangePasswordForm,
+    ChangeSecurityKeyNameForm,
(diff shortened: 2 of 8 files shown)


app/main/forms.py
```diff
@@ -2504,3 +2504,13 @@ def selected_areas(self):
         if self.select_all.data:
             return [self.select_all.area_slug]
         return self.areas.data
+
+
+class ChangeNameOfSecurityKey(StripWhitespaceForm):
```
servingUpAces, line 2509: minor: naming conventions for Forms would suggest this should be named `ChangeSecurityKeyNameForm`

app/main/forms.py
```diff
@@ -2504,3 +2504,13 @@ def selected_areas(self):
...
+    name_of_key = GovukTextInputField(
```
servingUpAces, line 2510: minor: this field could be `security_key_name`

app/main/views/user_profile.py
```diff
@@ -237,3 +241,62 @@ def user_profile_security_keys():
...
+    endpoint="user_profile_manage_security_key"
+)
+@main.route(
+    "/user-profile/security-keys/<uuid:key_id>/delete",
+    methods=['GET'],
+    endpoint="user_profile_confirm_delete_security_key"
+)
+@user_is_platform_admin
+def user_profile_manage_security_key(key_id):
+    security_keys = user_api_client.get_webauthn_credentials_for_user(current_user.id)
```
servingUpAces, line 258: Would it be useful to have a method to return a single key?

app/notify_client/user_api_client.py
```diff
@@ -201,5 +201,15 @@ def create_webauthn_credential_for_user(self, user_id, credential):
 
         return self.post(endpoint, data=credential.serialize())
 
+    def update_webauthn_credential_for_user(self, *, user_id, credential_id, new_name_for_credential):
```
servingUpAces, line 204: We could be more specific about this method. If we can only change the name of a security key, then we can call this method `update_security_key_name`

servingUpAces (commented):
A couple of minor comments. 
I'm not able to test deleting a key since I only have one yubikey, but I do know I get an error message if I delete the last key, which is good.

app/templates/views/user-profile/security-keys.html
```diff
@@ -29,7 +29,7 @@
 
         {% call mapping_table(
           caption=page_title,
-          field_headings=['Security key'],
+          field_headings=['Label', 'Action'],
```
benthorner, line 32: Minor: the first column isn't really a label. How about "Security key details"?

app/templates/views/user-profile/security-keys.html
```diff
@@ -39,6 +39,7 @@
                 <div class="govuk-body">{{ credential.name }}</div>
                 <div class="govuk-body hint">Registered {{ credential.created_at|format_delta }}</div>
               {% endcall %}
+              {{ edit_field('Manage', url_for('.user_profile_manage_security_key', key_id=credential.id)) }}
```
benthorner, line 42: I think the "Manage" link is meant to be the same size as the key name (from [the design](https://docs.google.com/document/d/1dZlxBzaT0mPYgjQ18q8YunDFEVBX_NEaV7JohR2WJP4/edit#heading=h.roph9nirx3h7)). Currently it is a bit obscure, could we enlarge it?

tests/app/main/views/test_user_profile.py
```diff
@@ -425,3 +425,40 @@ def test_non_platform_admin_user_doesnt_see_manage_security_key_page(client_requ
...
+    platform_admin_user,
+    webauthn_credential,
+    webauthn_credential_2,
+    mocker
+):
+    client_request.login(platform_admin_user)
+
+    mocker.patch(
+        'app.user_api_client.get_webauthn_credentials_for_user',
+        return_value=[webauthn_credential, webauthn_credential_2],
```
benthorner, line 446: Minor: we don't need two credentials here (similarly for some of the following tests).

tests/app/main/views/test_user_profile.py
```diff
@@ -374,3 +374,54 @@ def test_should_show_security_keys_page(
...
+        return_value=[webauthn_credential_2],
+    )
+    client_request.get(
+        '.user_profile_manage_security_key',
+        key_id=webauthn_credential['id'],
+        _expected_status=404,
+    )
+
+
+def test_non_platform_admin_user_doesnt_see_manage_security_key_page(client_request, webauthn_credential,):
```
benthorner, line 427: Can we check the post/delete endpoints as well for this restriction? I think it would be worth having a test that one user can't delete another's key.

We should be able to parametrize it, if we allow for optional POST data.

app/main/views/user_profile.py
```diff
@@ -283,8 +284,18 @@ def user_profile_manage_security_key(key_id):
...
-        credential_id=key_id
-    )
+
+    try:
+        user_api_client.delete_webauthn_credential_for_user(
+            user_id=current_user.id,
+            credential_id=key_id
+        )
+    except HTTPError as e:
+        message = "Cannot delete last remaining webauthn credential for user"
```
benthorner, line 295: We use "security keys" in user-facing stuff, and WebAuthn (the technical term) in the backend code (since security key is just a specific type of authenticator).

How about this:

```
You cannot delete your last security key.
```

tests/app/main/views/test_user_profile.py
```diff
@@ -469,6 +469,39 @@ def test_should_redirect_after_change_of_security_key_name(
...
+    mock_update = mocker.patch(
+        'app.user_api_client.update_webauthn_credential_for_user',
+        return_value=[webauthn_credential],
+    )
+
+    client_request.post(
+        'main.user_profile_manage_security_key',
+        key_id=webauthn_credential['id'],
+        _data={'name_of_key': webauthn_credential['name']},
+        _expected_status=302,
```
benthorner, line 495: Minor: we don't need to re-test the redirect behaviour here.

benthorner (commented):
Great work ✨. I'm looking forward to changing the default title.

@servingUpAces's comments sound sensible to me. I've tested with a spare key, and can confirm the removal works (you could also temporarily remove the duplicate check and add your single key twice).

I found a couple of things to look at before this goes in, but it's close. We should sort the "Manage" link so it's consistent with other pages; and I think we're missing a couple of tests.

app/main/views/user_profile.py
CrystalPea: I was thinking about this, that it would be more straight-forward. But then also
- I remember chatting with @benthorner that we want to return credentials per user, not per id
- one user will usually only have a couple credentials at most, so the loop would be very short anyway

Happy to change that if @benthorner is ok with that and we think it's worth it ☺️

app/templates/views/user-profile/security-keys.html
CrystalPea: I'm happy for it to be bigger, but I'm not sure how to make it so. I have no idea why it's small in the first place 🤔 . @quis any suggestions before I dig in deep?

tests/app/main/views/test_user_profile.py
CrystalPea: good spot, I removed the second creds from tests for `post` type routes.

tests/app/main/views/test_user_profile.py
CrystalPea: good idea, I added a parametrised test. Data was not even needed, as we should get a 403 before we are checking for that.

app/main/views/user_profile.py
CrystalPea: Good point, I tried to pay attention to this distinction, but it slipped here, probably because just passing message we got from API was so easy. I changed this now ☺️

tests/app/main/views/test_user_profile.py
CrystalPea: I think it's still good to see that all the other behaviour stays the same while we don't call the API. For example we could have a bug that view would only redirect when API was called.

app/main/views/user_profile.py
benthorner: Sounds like I good idea to me. Not an API, but a helper method somewhere to encapsulate the loop. (If we feel the need for an API in future, we can always add one.)

By the way, we can use [the method in the User model](https://github.com/alphagov/notifications-admin/blob/2d2c82ca8703cafde81c3e8a689c4dcba3a6278b/app/models/user.py#L352) instead of calling the API client directly.

app/templates/views/user-profile/security-keys.html
benthorner: From a quick scan of the User Profile page, I reckon you need [this class](https://github.com/alphagov/notifications-admin/blob/2d2c82ca8703cafde81c3e8a689c4dcba3a6278b/app/templates/views/user-profile.html#L13).

tests/app/main/views/test_user_profile.py
benthorner: Thanks. I had a look to check which ones those were. Can we do it for `patch` type routes as well?

tests/app/main/views/test_user_profile.py
benthorner: That's true. It's debatable: one thing I'm always cautious of is writing too many tests, as that's more maintenance work for the future; it's tricky to find a good balance between acceptable risk and maintainability.

It's only a line here, so definitely nothing major 👍.

app/templates/views/user-profile/security-keys.html
CrystalPea: Sweet, thanks a lot!

app/main/views/user_profile.py
CrystalPea: Good suggestions, done!

tests/app/main/views/test_user_profile.py
CrystalPea: Hey, I have two creds in those endpoints to make sure we get the right one out of the list of credentials, since we get a list back from API, not a single cred.

CrystalPea: Although since I encapsulated that function now, I could have it as a separate test always.

CrystalPea: done!

tests/app/main/views/test_user_profile.py
```diff
@@ -366,6 +373,252 @@ def test_should_show_security_keys_page(
...
+    assert credential_row.find('a')["href"] == url_for(
+        '.user_profile_manage_security_key',
+        key_id=webauthn_credential['id']
+    )
 
     register_button = page.select_one("[data-module='register-security-key']")
     assert register_button.text.strip() == 'Register a key'
+
+
+def test_get_key_from_list_of_keys(webauthn_credential, webauthn_credential_2):
```
benthorner, line 386: Minor: we don't need a test for this, as it's an internal method that's tested implicitly by the view - we're checking the happy path, and the unhappy (404) path, which is enough to cover what this test is checking.

tests/app/main/views/test_user_profile.py
```diff
@@ -366,6 +373,252 @@ def test_should_show_security_keys_page(
...
+    client_request.login(platform_admin_user)
+
+    mocker.patch(
+        'app.user_api_client.get_webauthn_credentials_for_user',
+        return_value=[webauthn_credential],
+    )
+
+    mock_update = mocker.patch(
+        'app.user_api_client.update_webauthn_credential_name_for_user',
+        return_value=[webauthn_credential],
```
benthorner, line 473: Minor: the return value is inaccurate/irrelevant here (similarly below).

benthorner (approved):
Just one parting comment about tests, but we can also just get this in now if we want to 🚀.

tests/app/main/views/test_user_profile.py
CrystalPea: ahh, you are right

tests/app/main/views/test_user_profile.py
CrystalPea: I don't know - we are not checking anywhere else that from list of two credentials, it will choose the right one.

benthorner: That's true. However, we're not checking anywhere for a list of 3, 4, 5, etc. credentials either. At some point we have to stop testing, so I always ask myself the question: is it likely this will break, and what's the risk if it does?

The danger of trying to test everything is we end up with a lot of tests, which then become a big maintenance burden - for one thing, it takes _ages_ to run all of them. Some repos on GOV.UK are very hard to work with due to this.

CrystalPea: True in principle, still, I decided a list with two credentials is a pretty basic case, so I left it in.

PR #3893, opened 2021-05-20
Allow setting provider for any channel

benthorner:
https://www.pivotaltracker.com/story/show/178228905

Previously we could only select a provider when using the test
channel, but this is also required for others channels when we
do tests on the production network with individual MNOs.

In order to reduce duplication and improve consistency, I've reused
the new broadcast_service_name_tag macro to show the setting.

## Before

| Severe channel  | Government channel |
| Screenshot 2021-05-20 at 14 46 43 | Screenshot 2021-05-20 at 14 46 29 |

## After (different selections)

| Severe channel  | Government channel |
| Screenshot 2021-05-20 at 14 45 41 | Screenshot 2021-05-20 at 14 45 59 |


tests/app/main/views/test_broadcast.py
```diff
@@ -317,23 +317,23 @@ def test_some_broadcast_tour_pages_show_service_name(
             False,
             'severe',
             'all',
-            '.navigation-service-type.navigation-service-type--live',
-            'service one Live Switch service',
-            'Live',
+            '.navigation-service-type.navigation-service-type--severe',
+            'service one Severe (all networks) Switch service',
+            'Severe (all networks)',
```
quis, line 322: We have onboarded several users now who have been taught to look for ‘Training’ or ‘Live’ at the top of the page. I think I’m OK with the change to add _(all networks)_ but I’d prefer we didn’t use _Severe_. This is also consistent with how the tour talks about being invited to send ‘live emergency alerts’.

benthorner: Sure, I've amended the commit to revert that part of the change. Shame we can't be totally generic. If we keep the Government channel, perhaps we'll need to revisit this, to disambiguate what the banner is referring to.

quis:
I also wonder if we need to say (all networks) in the header. The real benefit of this change is forcing the platform admin person making the change to explicitly give their choice of network.

idavidmcdonald (commented):
So I think this is generally good.

I like the change of url structure to `/services/<uuid:service_id>/service-settings/broadcasts/<broadcast_channel>`.

I think I spotted one or two things that need fixing that I will include in a commit shortly.

There are a few small additional comments that it might be worth considering for a subsequent PR @benthorner but I'm not going to do them now as a matter of getting this merged so we can do some testing for the MNOs in prod.

One thing worth us thinking about is the `Form` objects we use were started when we had a form that was a single page. Over time the options have changed and it is now multi page. I wonder if it is worth us relooking at what our `Form` objects might be if we were to build this from scratch knowing what the current design is. I guess I'm saying, it feels like we've probably taken on some tech debt with this form by adapting it quickly to our changes and maybe it needs a fresh start. This doesn't need actioning now but one to consider

idavidmcdonald (approved):

app/main/forms.py
```diff
@@ -2407,24 +2407,29 @@ class ServiceBroadcastChannelForm(StripWhitespaceForm):
 
 
 class ServiceBroadcastNetworkForm(StripWhitespaceForm):
+    def __init__(self, broadcast_channel, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.broadcast_channel = broadcast_channel
+
+        self.network_variant.choices = [
```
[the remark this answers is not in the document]
benthorner: Agreed, I found it added unnecessarily to my learning curve when I first came across this. In attempting to address this on it's own, I ended up refactoring the forms more generally, as you suggested in [your overall comment](https://github.com/alphagov/notifications-admin/pull/3893#pullrequestreview-674404574).

benthorner (commented):
Thanks @quis and @idavidmcdonald for nudging this through. I'm pleased it was useful in the end!

@idavidmcdonald I agree the account type field is quite difficult to understand, through a combination of naming and liberal reuse on each page of the form. I've raised a PR to try and refactor it, if you'd like to take a look: https://github.com/alphagov/notifications-admin/pull/3919.

PR #3895, opened 2021-05-21
Disable the remaining messages check for uploads

pixeltrix:
The HTTP request for the statistics is taking more 30 seconds which leads to 504 errors from CloudFront.

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -42,9 +42,20 @@ class ServiceAPIClient(NotifyAdminAPIClient):
         return self.get('/service/{0}'.format(service_id))
 
     def get_service_statistics(self, service_id, today_only, limit_days=None):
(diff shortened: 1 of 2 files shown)


leohemsted (approved):

app/notify_client/service_api_client.py
```diff
@@ -42,9 +42,17 @@ def get_service(self, service_id):
         return self.get('/service/{0}'.format(service_id))
 
     def get_service_statistics(self, service_id, today_only, limit_days=None):
```
idavidmcdonald, line 44: Worth checking both places this function is used to make sure there are not knock on effects to other users of this function

pixeltrix: Returning the empty dict for all calls would lead to zeros in the headers of the view notifications pages. This however uses `today_only=False` so have updated the PR to only return the empty dict when `today_only=True`.

NeilvB (approved):

PR #3898, opened 2021-05-24
Revert "Disable the remaining messages check for uploads"

servingUpAces:
Reverts alphagov/notifications-admin#3895

We've updated the query in API to use an existing query. The query is still slow for large volume services but I saw a marked difference and show mean that the CSV upload will not time out. However we should revisit how we handle checking for daily limits.


idavidmcdonald (approved):
Good to go when the API has been deployed with the query change

PR #3899, opened 2021-05-25
Bump utils version for new invalid address character

servingUpAces:
Added a new invalid address character `~`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -116,7 +116,7 @@ mistune==0.8.4
     # via notifications-utils
 notifications-python-client==6.0.2
     # via -r requirements.in
-git+https://github.com/alphagov/notifications-utils.git@44.2.0#egg=notifications-utils==44.2.0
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -49,7 +49,7 @@ docutils==0.15.2
     # via awscli
 et-xmlfile==1.0.1
     # via openpyxl
-eventlet==0.30.2 # pyup: ignore
+eventlet==0.30.2
```
idavidmcdonald, line 52: Optional: Not sure if removing the pyup line is intentional here?

Opens a can of worms about pyup and eventlet but the easiest thing to do would be to hardcode it back in and deal with that problem another day...

idavidmcdonald (approved):
Nice!

requirements.txt
servingUpAces: No I'll add it back - thanks

PR #3900, opened 2021-05-25
Add help text for new make steps

idavidmcdonald:
Helpful for when you run `make` as you've forgotten what the command is
and you don't want to go find the readme

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,12 +23,12 @@ VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$V
 ## DEVELOPMENT
 
 .PHONY: bootstrap
-bootstrap: generate-version-file
+bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 	npm install && npm rebuild node-sass && npm run build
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,7 +52,7 @@ test: ## Run tests
 	./scripts/run_tests.sh
 
 .PHONY: fix-imports
-fix-imports:
+fix-imports: ## Fix imports using isort
```
benthorner, line 55: Minor / out-of-scope: I was wondering if we should delete this one, as it's inconsistent with other apps, and normally it's just as easy to manually type the command (versus knowing and remembering this one is here).

Do we really need it?

benthorner (approved):

Makefile
idavidmcdonald: I'm happy to keep it. Mostly because I don't remember this command and so this might save me a few seconds. Also if it got added to other repos I wouldn't have a problem with that.

I don't feel too strongly though

benthorner: I wonder if it would help to know the flags aren't required: 

```
UserWarning: W0501: The following deprecated CLI flags were used and ignored: -rc!
```

If you have a file X whose imports you need to fix, all you need to do is run `isort X`, which takes less time than running `isort app tests`.

PR #3902, opened 2021-05-27
Designerise the pages for adding a security key

quis:
Before | After
---|---
image | image
image | image

### DIFF
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -43,7 +43,7 @@
 
 
   {{ govukBackLink({ "href": url_for('.user_profile') }) }}
-  <div class="govuk-grid-row govuk-clearfix">
+  <div class="govuk-grid-row">
     {% if credentials %}
       <div class="govuk-grid-column-five-sixths">
         {{ page_header(
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -46,10 +46,7 @@
   <div class="govuk-grid-row">
     {% if credentials %}
       <div class="govuk-grid-column-five-sixths">
-        {{ page_header(
-          page_title,
-          back_link=url_for('.user_profile')
-        ) }}
+        {{ page_header(page_title) }}
         <div class="body-copy-table">
           {% call mapping_table(
             caption=page_title,
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -80,7 +80,7 @@
         }) }}
       </div>
       <div class="govuk-grid-column-one-quarter">
-        <img src="{{ asset_url('images/security-key.svg') }}" alt="" class="govuk-!-margin-top-9">
+        <img src="{{ asset_url('images/security-key.svg') }}" alt="" class="webauthn-illustration">
       </div>
       {% endif %}
     </div>
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -82,11 +82,10 @@
       <div class="govuk-grid-column-one-quarter">
         <img src="{{ asset_url('images/security-key.svg') }}" alt="" class="govuk-!-margin-top-9">
       </div>
+      {% endif %}
     </div>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-five-sixths">
-
-      {% endif %}
 
       {{ govukErrorMessage({
         "classes": "webauthn__api-missing",
--- a/app/templates/views/user-profile/security-keys.html
+++ b/app/templates/views/user-profile/security-keys.html
@@ -18,14 +18,16 @@
 {% endblock %}
 
 {% block maincolumn_content %}
-  {{ page_header(
-    page_title,
-    back_link=url_for('.user_profile')
-  ) }}
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-five-sixths">
-      {% if credentials %}
+
+  {{ govukBackLink({ "href": url_for('.user_profile') }) }}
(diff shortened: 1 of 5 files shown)


app/templates/views/user-profile/security-keys.html
```diff
@@ -45,11 +47,44 @@
           {% endcall %}
         </div>
 
-      {% else %}
+        {{ govukButton({
```
benthorner, line 50: Minor: we could DRY this up into a variable.

app/templates/views/user-profile/security-keys.html
```diff
@@ -45,11 +47,44 @@
...
+            "data-module": "register-security-key",
+            "data-csrf-token": csrf_token(),
+          }
+        }) }}
+      </div>
+      <div class="govuk-grid-column-one-quarter">
+        <img src="{{ asset_url('images/security-key.svg') }}" alt="" class="govuk-!-margin-top-9">
+      </div>
+    </div>
+    <div class="govuk-grid-row">
```
benthorner, line 86: Should this be after the `endif`?

The end result looks like it's out of alignment...

Screenshot 2021-05-28 at 10 27 26

app/templates/views/user-profile/security-keys.html
```diff
@@ -18,14 +18,16 @@
...
-    <div class="govuk-grid-column-five-sixths">
-      {% if credentials %}
+
+  {{ govukBackLink({ "href": url_for('.user_profile') }) }}
+  <div class="govuk-grid-row govuk-clearfix">
+    {% if credentials %}
+      <div class="govuk-grid-column-five-sixths">
+        {{ page_header(
+          page_title,
+          back_link=url_for('.user_profile')
```
benthorner, line 29: We should delete this, otherwise we'll get two back links.

app/templates/views/user-profile/security-keys.html
```diff
@@ -18,14 +18,16 @@
 {% endblock %}
 
 {% block maincolumn_content %}
-  {{ page_header(
```
benthorner, line 21: Moving this is inconsistent with [other pages](https://github.com/alphagov/notifications-admin/blob/24f4b3f3eb1b645744d77d639f8d83964e38b6cb/app/templates/views/service-settings/name-local.html#L13). Could you help me understand why we're making this change?

app/templates/views/user-profile/security-keys.html
```diff
@@ -18,14 +18,16 @@
...
-    page_title,
-    back_link=url_for('.user_profile')
-  ) }}
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-five-sixths">
-      {% if credentials %}
+
+  {{ govukBackLink({ "href": url_for('.user_profile') }) }}
+  <div class="govuk-grid-row govuk-clearfix">
```
benthorner, line 24: What does `clearfix` do? 

I had a quick play with the UI, and can't see the effect it has.

app/templates/views/user-profile/security-keys.html
```diff
@@ -45,11 +47,44 @@
...
+          "element": "button",
+          "text": "Register a key",
+          "classes": "govuk-button--secondary webauthn__api-required",
+          "attributes": {
+            "data-module": "register-security-key",
+            "data-csrf-token": csrf_token(),
+          }
+        }) }}
+      </div>
+      <div class="govuk-grid-column-one-quarter">
```
benthorner, line 82: The narrow width version of this is quite large, is that OK?

Screenshot 2021-05-28 at 10 28 41

app/templates/views/user-profile/security-keys.html
quis: It’s so we can put the page header into a column, and align the illustration beside it

app/templates/views/user-profile/security-keys.html
quis: It was a hangover from when I was trying to achieve this layout in a different way. It’s documented at https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-clearfix and I’ve removed it in f74db8ad5

app/templates/views/user-profile/security-keys.html
quis: Refactored in f7fb5646c

quis: Also means the error messages show up in the left column now, which looks nicer:

image

benthorner: Nice!

app/templates/views/user-profile/security-keys.html
benthorner: Ah yes, I see. Makes sense now 👍.

benthorner:
Thanks for addressing some of the issues 👍. I noticed there's a linting error on the new CSS, and there's still the issue of the two back buttons to fix (maybe you didn't get time to do that yet, but I thought I'd review again anyhow).

app/templates/views/user-profile/security-keys.html
quis: Removed in 77ea7af

leohemsted (approved):

app/templates/views/user-profile/security-keys.html
```diff
@@ -44,34 +65,25 @@
...
-      {% else %}
-
+        {{ webauthn_button }}
+      </div>
+    {% else %}
+      <div class="govuk-grid-column-one-half">
+        {{ page_header(page_title) }}
         <p class="govuk-body">
-          Security keys are an alternative way of signing in to Notify.
+          Security keys are an alternative way of signing in to Notify,
```
idavidmcdonald, line 74: Don't know if this content has gone through @karlchillmaid , it may have already but wondering if we'd considered putting something here to mention that using a security key is a more secure form of MFA. So they don't just see it as oh there is an alternative but actually see it as more secure too so may be more tempted to change to it

PR #3907, opened 2021-06-01
Remove ‘manage settings’ permission for alerts

quis:
We shouldn’t have any real users on emergency alert services inviting other users or renaming the service.

If we need to do these things then we can do them as platform admins.

This reduce a potential risk of people having permissions they shouldn’t on production, while we are rolling out the service.

### DIFF
--- a/app/models/roles_and_permissions.py
+++ b/app/models/roles_and_permissions.py
@@ -28,7 +28,6 @@ permissions = (
 broadcast_permissions = (
     ('send_messages', 'Prepare and approve broadcasts'),
     ('manage_templates', 'Add and edit templates'),
-    ('manage_service', 'Manage settings and team'),
 )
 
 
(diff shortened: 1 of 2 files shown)


quis:
Do we also need to do a migration to remove this permission from existing users?

benthorner (approved):
Makes sense to me 👍.

One downside of this is that we'll lose visibility of anyone who _currently_ has these permissions. I did a quick check of our production services and confirmed that only people in our team have the permission to manage services at the moment.

For consistency with the UI, you might want to consider a data migration to remove the redundant service permissions. (They'll also get removed automatically when someone tries to update the permissions for a team member.)

quis:
Even with a migration the problem would be re-introduced when we create a new service, then transform it into a broadcast service. @klssmith do you want to think about how this fits in with the work you’re doing, and whether you’d like me to hold off on merging this?

PR #3912, opened 2021-06-03
Get webauthn login to honour next queryparam url

leohemsted:
unfortunately i couldn't get mocking location for a single js test to work, but by changing the global config i was able to add some query params that i can expect to be passed through. Don't love this at all but not quite sure of a good way round this. I think we're not practicing very good hygiene and best practices with our mocking (i'm sure that the assigning an empty lambda and then deleting it later isn't the proper way to do things, plus doing that ) and it's really confounding me here.

A few other changes snuck in here so I suggest reviewing commit by commit, and also reviewing the following:

- [x] alphagov/notifications-api#3260

### DIFF
--- a/tests/javascripts/jest.config.js
+++ b/tests/javascripts/jest.config.js
@@ -1,4 +1,4 @@
 module.exports = {
   setupFiles: ['./support/setup.js'],
-  testURL: 'https://www.notifications.service.gov.uk'
+  testURL: 'https://www.notifications.service.gov.uk/?next=%2Ffoo%3Fbar%3Dbaz'
 }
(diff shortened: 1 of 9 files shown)


CrystalPea (approved):
Code looks good, useful docstrings and all works fine on my local 👍🏼

tests/javascripts/jest.config.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
   setupFiles: ['./support/setup.js'],
-  testURL: 'https://www.notifications.service.gov.uk'
+  testURL: 'https://www.notifications.service.gov.uk/?next=%2Ffoo%3Fbar%3Dbaz'
```
benthorner, line 3: Yuk (nice find, though!). How about this instead: https://github.com/alphagov/notifications-admin/pull/3922

benthorner (commented):
Appreciate this has been merged already. Given @leohemsted's comments about the approach (in the commit messages), I thought it would be useful if I left a few comments, some of which I've raised PRs to illustrate.

PR #3913, opened 2021-06-03
Add new allowed email domains

CrystalPea:
`gp.hscni.net` for GPs in Northern Ireland
and
`senedd.wales` is the new name for assembly.wales


CrystalPea:
User's problem was not with a domain after all

PR #3915, opened 2021-06-07
Tweak content for incorrect flash message

benthorner:
Currently someone needs to manually switch a user's account to sign
in with WebAuthn, so this text was misleading.


quis:
Could we make the code match the content instead? e.g. https://github.com/alphagov/notifications-admin/pull/3917

quis:
Closing in favour of #3917

PR #3917, opened 2021-06-07
Set user to WebAuthn login after adding token

quis:


servingUpAces (approved):
Tested this locally and it works as expected.

app/main/views/webauthn_credentials.py
```diff
@@ -52,9 +52,8 @@ def webauthn_complete_register():
         current_app.logger.info(f'User {current_user.id} could not register a new webauthn token - {e}')
         return cbor.encode(str(e)), 400
 
-    user_api_client.create_webauthn_credential_for_user(
-        current_user.id, credential
-    )
+    current_user.create_webauthn_credential(credential)
+    current_user.update(auth_type='webauthn_auth')
```
benthorner, line 56: We only need to do this if this is their first credential (to avoid an unnecessary API call). Rather than doing it here, I'd suggest doing it in the API app, since it's effectively a single DB transaction - if the second change fails, we should roll back the first, so the user can try again.

quis: I’m going to merge this now since Becca has approved it and it delivers value. We can chat about whether the extra work of making it transactional in the API is worth it vs. other priorities this afternoon.

PR #3920, opened 2021-06-08
Refactor User.webauthn_credentials into a ModelList

quis:
This saves a bit of repetition, and lets us attach other methods to the collection, rather than having multiple methods on the user object prefixed with the same name, or random functions floating about.


benthorner (approved):
Much nicer - definitely agree with keeping the User model slim if possible 🧘.

I can see you've got an import order test failing, but otherwise it looks good to me.

PR #3923, opened 2021-06-09
Split out utils code into separate modules

benthorner:
Prerequisite to addressing: https://github.com/alphagov/notifications-admin/pull/3894#discussion_r640513356

Previously we had limited places to put code:

- Pretend it's part of a model. Mixing business logic with models isn't
scalable, as it tends to bloat models like User.

- Dump it in utils.py. This makes it hard to find relevant code, due to the
mixture of responsibilities and sheer length of the file.

- Add a top-level file under app/. This is confusing because it mixes
low-level utility code with high-level files like config.py and version.py.

In other apps we have more scalable options:

- In API we have a folder for each concept e.g. app/templates/. We also
have app/dao/, which is another place to put certain types of code.

- In Download API we have an app/utils/ folder.

This takes a step away from putting code in places it doesn't easily belong.
While it's not necessary for the change I'd like to make, the more code we
add, the harder the change will be when we do decide to refactor.

I've left some code in the top-level package, as I'm not trying to solve the
entire problem here - we can always extract more of it in future.

If we have other thoughts on where utility code should go, then I'm open to
hearing those to - it's definitely not a solved problem in software.


app/utils/__init__.py
```diff
@@ -0,0 +1,211 @@
...
+from werkzeug.datastructures import MultiDict
+from werkzeug.routing import RequestRedirect
+
+SENDING_STATUSES = ['created', 'pending', 'sending', 'pending-virus-check']
+DELIVERED_STATUSES = ['delivered', 'sent', 'returned-letter']
+FAILURE_STATUSES = ['failed', 'temporary-failure', 'permanent-failure',
+                    'technical-failure', 'virus-scan-failed', 'validation-failed']
+REQUESTED_STATUSES = SENDING_STATUSES + DELIVERED_STATUSES + FAILURE_STATUSES
+
+NOTIFICATION_TYPES = ["sms", "email", "letter", "broadcast"]
```
quis, line 31: This points towards this repo not having a `Notification` model, but that’s a bigger piece of work

benthorner: Oh yes, I hadn't noticed that before. This rabbit hole is deep.

leohemsted (approved):
Went through commit by commit, nice sensible changes 👍

PR #3924, opened 2021-06-10
Permanent failure message for letters

servingUpAces:
When a letter has passed our validation but the postal provider is unable to print the letter we need to mark the letter as failed.

If we mark the letter as a technical-failure, we say that we will fix the issue, which is wrong because we cannot fix the issue.

If we mark the letter as validation-failed, the letter is in the wrong bucket, so the letter is not viewable/downloadable by the client.

This PR updates the message for a letter marked as permanent-failure to better reflect what has actually happened.

### DIFF
--- a/app/templates/views/message-status.html
+++ b/app/templates/views/message-status.html
@@ -80,6 +80,7 @@
         ('Printed', 'The provider has printed the letter. Letters are printed at 5:30pm and dispatched the next working day.'), 
         ('Cancelled', 'Sending cancelled. Your letter will not be printed or dispatched.'),
         ('Technical failure', 'Notify had an unexpected error while sending the letter to our printing provider.'),
+        ('Permenent failure', 'The provider cannot print the letter. Your letter will not be dispatched.')
       ] %}
         {% call row() %}
           {{ text_field(message_length) }}
--- a/app/templates/views/notifications/notification.html
+++ b/app/templates/views/notifications/notification.html
@@ -41,7 +41,7 @@
     </p>
 
     {% if template.template_type == 'letter' %}
-      {% if notification_status in ('permanent-failure', 'cancelled') %}
+      {% if notification_status == 'cancelled' %}
         <p class="notification-status-cancelled">
           Cancelled {{ updated_at|format_datetime_short }}
         </p>
@@ -49,6 +49,10 @@
         <p class="notification-status-cancelled">
           {{ message.summary | safe }}
         </p>
+      {% elif notification_status == 'permanent-failure' %}
+         <p class="notification-status-cancelled">
+          Permanent failure – The postal provider is unable to print the letter. Your letter has not been sent.
+         </p>
       {% elif notification_status == 'technical-failure' %}
         <p class="notification-status-cancelled">
           Technical failure – Notify will resend once the team have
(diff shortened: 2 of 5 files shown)


app/templates/views/notifications/notification.html
```diff
@@ -41,14 +41,18 @@
...
         <p class="notification-status-cancelled">
           Cancelled {{ updated_at|format_datetime_short }}
         </p>
       {% elif notification_status == 'validation-failed' %}
         <p class="notification-status-cancelled">
           {{ message.summary | safe }}
         </p>
+      {% elif notification_status == 'permanent-failure' %}
+         <p class="notification-status-cancelled">
+          Permanent failure – The postal provider is unable to print the letter. Your letter has not been sent.
```
karlchillmaid, line 54:
```suggestion
          Permanent failure – The provider cannot print the letter. Your letter will not be dispatched.
```

tests/app/main/views/test_notifications.py
```diff
@@ -442,7 +442,7 @@ def test_notification_page_shows_validation_failed_precompiled_letter(
 @pytest.mark.parametrize('notification_status, expected_message', (
     (
         'permanent-failure',
-        'Cancelled 1 January at 1:02am',
+        'Permanent failure – The postal provider is unable to print the letter. Your letter has not been sent.',
```
karlchillmaid, line 445:
```suggestion
        'Permanent failure – The provider cannot print the letter. Your letter will not be dispatched.',
```

app/templates/views/notifications/notification.html
karlchillmaid: The way we currently describe it, we use:
* ‘sent’ to mean the user sends the letter using Notify
* ‘dispatched’ to mean the provider posts the letter

I just want to check I’ve got the usage right here.

servingUpAces: right - I've updated this message and the status doc page and tech docs

app/templates/views/message-status.html
```diff
@@ -80,6 +80,7 @@ <h2 id="letter-statuses" class="heading-medium">Letters</h2>
         ('Printed', 'The provider has printed the letter. Letters are printed at 5:30pm and dispatched the next working day.'), 
         ('Cancelled', 'Sending cancelled. Your letter will not be printed or dispatched.'),
         ('Technical failure', 'Notify had an unexpected error while sending the letter to our printing provider.'),
+        ('Permenent failre', 'The provider cannot print the letter. Your letter will not be dispatched.')
```
karlchillmaid, line 83:
```suggestion
        ('Permanent failure', 'The provider cannot print the letter. Your letter will not be dispatched.')
```

app/templates/views/message-status.html
```diff
@@ -80,6 +80,7 @@ <h2 id="letter-statuses" class="heading-medium">Letters</h2>
         ('Printed', 'The provider has printed the letter. Letters are printed at 5:30pm and dispatched the next working day.'), 
         ('Cancelled', 'Sending cancelled. Your letter will not be printed or dispatched.'),
         ('Technical failure', 'Notify had an unexpected error while sending the letter to our printing provider.'),
+        ('Permenent failure', 'The provider cannot print the letter. Your letter will not be dispatched.')
```
karlchillmaid, line 83:
```suggestion
        ('Permanent failure', 'The provider cannot print the letter. Your letter will not be dispatched.')
```

leohemsted (approved):

klssmith:
Might need to update this page too (the bottom letter has permanent-failure status)

Screenshot 2021-06-11 at 15 04 18

klssmith (approved):

PR #3930, opened 2021-06-14
Move the '- archived' label for organisations to be next to the org name

klssmith:
Before:
Screenshot 2021-06-14 at 12 18 50

After:
Screenshot 2021-06-14 at 12 33 26


benthorner (approved):
Much nicer 👍.

(It's strange how we archive orgs and services differently - with a rename 🤷 .)

PR #3931, opened 2021-06-14
Refactor email revalidation check

benthorner:
In response to: https://github.com/alphagov/notifications-admin/pull/3894#discussion_r640513356

This also includes further refactoring of the "utils" module,
continuing the work started in [1]. If the first commit makes
this PR too big, I'm happy to move it elsewhere. Please see the
commit messages for more details.

[1]: https://github.com/alphagov/notifications-admin/pull/3923

### DIFF
--- a/app/main/views/two_factor.py
+++ b/app/main/views/two_factor.py
@@ -17,7 +17,11 @@ from app.main import main
 from app.main.forms import TwoFactorForm
 from app.models.user import User
 from app.utils import is_less_than_days_ago
-from app.utils.login import redirect_to_sign_in
+from app.utils.login import (
+    log_in_user,
+    redirect_to_sign_in,
+    redirect_when_logged_in,
+)
 
 
 @main.route('/two-factor-email-sent', methods=['GET'])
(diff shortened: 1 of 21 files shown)


tests/app/utils/test_pagination.py
```diff
@@ -0,0 +1,20 @@
+from app.utils.pagination import generate_next_dict, generate_previous_dict
+
+
+def test_generate_previous_dict(client):
+    ret = generate_previous_dict('main.view_jobs', 'foo', 2, {})
```
CrystalPea, line 5: Minor: 
in those tests, could we change `ret` to `result`, or even better something descriptive, like `previous_page_dict`/`next_page_dict`? I find it makes it easier to follow. This is because then I don't have to translate the shorthand in my head to a full word or remember what it was.

app/main/views/two_factor.py
```diff
@@ -79,11 +79,11 @@ def _check_code(code):
     redirect_url = request.args.get('next')
 
     if form.validate_on_submit():
-        if is_less_than_days_ago(user.email_access_validated_at, 90):
-            return log_in_user(user_id)
-        else:
+        if email_needs_revalidating(user):
```
CrystalPea, line 82: Nice one ⭐

CrystalPea (approved):
Good work, one minor comment 👍🏼

tests/app/utils/test_pagination.py
benthorner: Sure, I don't mind doing that as part of this (the code is unchanged in this PR).

I've added a commit to change `ret` to `result`.

CrystalPea: sweet, thanks a lot, and sorry, because the code was moved it looked like new code, but you are right it was just moved 😅

PR #3933, opened 2021-06-15
Show an error if user cannot login with WebAuthn

benthorner:
https://www.pivotaltracker.com/n/projects/1443052

This follows the same approach as for registration [1]. None of
this code is tested, as:

- We don't have a way to test inline JS.
- The risk of this code not working is low.
- We might change the approach in future [2].

[1]: https://github.com/alphagov/notifications-admin/pull/3886
[2]: https://github.com/alphagov/notifications-admin/pull/3886#issuecomment-841128380

## Screenshots

| Error (no JS)  | Error (JS, but no WebAuthn) |
| Screenshot 2021-06-15 at 14 03 42 | Screenshot 2021-06-15 at 14 04 13 |

[1]: https://github.com/alphagov/notifications-admin/pull/3886

### DIFF
--- a/app/templates/views/two-factor-webauthn.html
+++ b/app/templates/views/two-factor-webauthn.html
@@ -2,8 +2,14 @@
 {% from "components/page-header.html" import page_header %}
 {% from "components/button/macro.njk" import govukButton %}
 {% from "components/back-link/macro.njk" import govukBackLink %}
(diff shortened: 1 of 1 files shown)


tombye (approved):
Looks good to me 👍🏻

PR #3934, opened 2021-06-16
Add a line about retention for files sent by email

karlchillmaid:


app/templates/views/security.html
```diff
@@ -21,7 +21,8 @@ <h2 class="heading-medium">Data</h2>
     <li>when it passes through the service</li>
     <li>when it’s stored on the service</li>
   </ul>
-  <p class="govuk-body">Any user data you upload is only held for 7 days.</p>
+  <p class="govuk-body">Any recipient data you upload is only held for 7 days.</p>
```
karlchillmaid, line 24: Do we need to mention that this is just the default retention period, and that services can keep data for longer?

idavidmcdonald: A good question. I lean towards we don't mention it.

- it will probably increase our requests for custom data retention which generally I think aren't really needed
- it will make this more complex for the user to understand without really providing any benefit
- we won't change their retention period without them knowing, therefore they don't need to be warned about having longer than 7 days because they only need to worry if they ask for it, at which point we will tell them

app/templates/views/security.html
```diff
@@ -21,7 +21,8 @@ <h2 class="heading-medium">Data</h2>
...
+  <p class="govuk-body">If you <a href="/features/email#send-files">send a file by email</a>, it will be available for the recipient to download for 18 months.</p>
```
idavidmcdonald, line 25: Wondering if 'it' should be 'the file', so it is really explicit that it is just the file and not the recipient data mentioned in the paragraph above. But might be overthinking that and will leave you to decide on that one Karl

app/templates/views/security.html
karlchillmaid: Thanks David. I agree. Less is more – so let’s keep the content short and simple for the reasons you described.

app/templates/views/security.html
karlchillmaid: We say ‘the file’ in the docs. I changed it because our style is a little less formal on Notify compared to the tech docs – but in this instance I think being specific is helpful. The content on the Security page is a little different to some of our other pages anyway.

karlchillmaid, line 25:
```suggestion
  <p class="govuk-body">If you <a href="/features/email#send-files">send a file by email</a>, the file will be available for the recipient to download for 18 months.</p>
```

idavidmcdonald (approved):

app/templates/views/security.html
```diff
@@ -21,7 +21,8 @@ <h2 class="heading-medium">Data</h2>
     <li>when it passes through the service</li>
     <li>when it’s stored on the service</li>
   </ul>
-  <p class="govuk-body">Any user data you upload is only held for 7 days.</p>
+  <p class="govuk-body">Any recipient data you upload is only held for 7 days.</p>
+  <p class="govuk-body">If you <a href="/features/email#send-files">send a file by email</a>, the file will be available for the recipient to download for 18 months.</p>
```
idavidmcdonald, line 25:
```suggestion
  <p class="govuk-body">If you <a class="govuk-link govuk-link--no-visited-state" href="/features/email#send-files">send a file by email</a>, the file will be available for the recipient to download for 18 months.</p>
```

PR #3938, opened 2021-06-22
Backup original precompiled uploads to S3

benthorner:
https://www.pivotaltracker.com/story/show/178542772

This continues the work from Template Preview [1], so that we have
a complete store of original PDFs to use for testing changes to it.

Previously we did store some originals, but these were only invalid
PDFs that had failed sanitisation; for valid PDFs, the "transient"
bucket only contains the sanitised versions, which the API deletes
/ moves when the notification is sent [2].

Since the notification is only created at a later stage [3], there's
no easy way to get the final name of the PDF we send to DVLA. Instead,
we use the "upload_id", which eventually becomes the notification ID
[4]. This should be enough to trace the file for specific debugging.

Note that we only want to store original PDFs if they're valid (and
virus free!), since there's no point testing changes with bad data.

[1]: https://github.com/alphagov/notifications-template-preview/pull/545
[2]: https://github.com/alphagov/notifications-api/blob/c44ec57c1784084d92ab54903034241c9612253d/app/service/send_notification.py#L212
[3]: https://github.com/alphagov/notifications-admin/blob/7930a53a58bd86f6fe4f979db5253b889fd10c06/app/main/views/uploads.py#L362
[4]: https://github.com/alphagov/notifications-admin/blob/7930a53a58bd86f6fe4f979db5253b889fd10c06/app/main/views/uploads.py#L373

### DIFF
--- a/app/s3_client/s3_letter_upload_client.py
+++ b/app/s3_client/s3_letter_upload_client.py
@@ -10,6 +10,18 @@ def get_transient_letter_file_location(service_id, upload_id):
     return 'service-{}/{}.pdf'.format(service_id, upload_id)
 
 
+def backup_original_letter_to_s3(
+    data,
+    upload_id,
+):
+    utils_s3upload(
+        filedata=data,
+        region=current_app.config['AWS_REGION'],
+        bucket_name=current_app.config['PRECOMPILED_ORIGINALS_BACKUP_LETTERS'],
+        file_location=f'{upload_id}.pdf',
+    )
+
+
 def upload_letter_to_s3(
     data,
     *,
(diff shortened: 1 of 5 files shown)


benthorner:
Have checked to make sure I can download a valid PDF as expected from S3.

app/s3_client/s3_letter_upload_client.py
```diff
@@ -10,6 +10,18 @@ def get_transient_letter_file_location(service_id, upload_id):
...
 
+def backup_original_letter_to_s3(
+    data,
+    upload_id,
+):
+    utils_s3upload(
+        filedata=data,
+        region=current_app.config['AWS_REGION'],
+        bucket_name=current_app.config['PRECOMPILED_ORIGINALS_BACKUP_LETTERS'],
+        file_location=f'{str(upload_id)}.pdf',
```
klssmith, line 21: You could drop the `str()` here if you wanted, since it's not necessary (but that's really minor, so no need to change)

klssmith (approved):

app/s3_client/s3_letter_upload_client.py
benthorner: Thanks for pointing it out. I forgot that `str` is implicit when interpolating. I've changed it for the sake of me remembering this better in future.

PR #3939, opened 2021-06-23
Check the daily limit get the daily notification_count from redis.

servingUpAces:
The daily limit cache is set by the api when a notification is created. There is one cache key per service per day and it expires after 24 hours.

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -1,5 +1,7 @@
 from datetime import datetime
 
+from notifications_utils.clients.redis import daily_limit_cache_key
+
 from app.extensions import redis_client
 from app.notify_client import NotifyAdminAPIClient, _attach_current_user, cache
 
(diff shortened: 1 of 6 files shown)


tests/app/notify_client/test_service_api_client.py
```diff
@@ -40,7 +40,6 @@ def test_client_gets_service(mocker):
 @pytest.mark.parametrize('today_only, limit_days', [
     (False, None),
     (False, 30),
-    pytest.param(True, None, marks=pytest.mark.xfail(raises=AssertionError)),
```
benthorner, line 43: Can we remove the whole `today_only` param now?

benthorner (commented):
Nice, just one question to check I've understood the change...

app/notify_client/service_api_client.py
```diff
@@ -43,9 +43,9 @@ def get_service(self, service_id):
         """
         return self.get('/service/{0}'.format(service_id))
 
-    def get_service_statistics(self, service_id, today_only, limit_days=None):
+    def get_service_statistics(self, service_id, limit_days=None):
         return self.get('/service/{0}/statistics'.format(service_id), params={
-                'today_only': today_only, 'limit_days': limit_days
+                'today_only': False, 'limit_days': limit_days
```
benthorner, line 48: We could remove this now, as [the API uses `MultiDict.get` to retrieve it](https://github.com/alphagov/notifications-api/blob/master/app/service/rest.py#L235), which returns `None` by default.

benthorner (approved):
Thanks for doing the extra cleanup 🧹

PR #3940, opened 2021-06-23
Update data format for platform admin report

klssmith:
The format of dates returned from API for the "Monthly notification statuses for live services" report has changed, so we no longer need to format the dates here.

https://github.com/alphagov/notifications-api/pull/3279


benthorner (approved):
Matches the test data in the API PR 👍.

PR #3941, opened 2021-06-24
Use `CustomBroadcastArea` to estimate number of phones in bleed area

quis:
Our current assumption is that the bleed area has the same population density as the broadcast area.

This is particularly naïve when:
- the bleed area overlaps the sea – no-one lives in the sea
- the broadcast area is a village or town and the bleed area is the surrounding countryside
- the broadcast area is adjacent to a densely populated area like a city

We can be smarter about this now that we have a way of determining the number of phones in an arbitrary area based on the known areas that we have population data about (see https://github.com/alphagov/notifications-admin/pull/3844).

Calculating the population in an overlap is a slightly more intensive calculation. So we only do it for areas which are small enough that it doesn’t slow things down too much. For larger areas (bigger than any single county) we still use the more naïve algorithm.

Before | After
---|---
Screenshot 2021-07-02 at 14 08 38 | Screenshot 2021-07-02 at 14 06 59
Screenshot 2021-07-02 at 14 10 34 | Screenshot 2021-07-02 at 14 10 58
Screenshot 2021-07-02 at 14 13 02 | Screenshot 2021-07-02 at 14 13 31
Screenshot 2021-07-02 at 14 17 25 | Screenshot 2021-07-02 at 14 17 39

***

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/3943

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -7,6 +7,7 @@ from orderedset import OrderedSet
 from werkzeug.utils import cached_property
 
 from app.broadcast_areas.models import (
+    CustomBroadcastArea,
     CustomBroadcastAreas,
     broadcast_area_libraries,
 )
@@ -16,6 +17,10 @@ from app.models.user import User
 from app.notify_client.broadcast_message_api_client import (
     broadcast_message_api_client,
 )
+
+ESTIMATED_AREA_OF_LARGEST_UK_COUNTY = broadcast_area_libraries.get_areas(
+    'ctyua19-E10000023'  # North Yorkshire
+)[0].polygons.estimated_area
 
 
 class BroadcastMessage(JSONModel):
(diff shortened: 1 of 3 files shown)


app/models/broadcast_message.py
```diff
@@ -176,12 +181,18 @@ def count_of_phones(self):
...
     def count_of_phones_likely(self):
-        area_estimate = self.simple_polygons.estimated_area
-        bleed_area_estimate = self.simple_polygons_with_bleed.estimated_area - area_estimate
-        return round_to_significant_figures(
-            self.count_of_phones + (self.count_of_phones * bleed_area_estimate / area_estimate),
-            1
-        )
+        estimated_area = self.simple_polygons.estimated_area
+
+        if estimated_area > ESTIMATED_AREA_OF_LARGEST_UK_COUNTY:
```
idavidmcdonald, line 186: I'd encourage an inline comment here essentially saying

If the area is big, we use a naive and less computationally expensive way to calculate the number of phones in the area. If the area is small though (smaller than the largest county), we can use a more precise but more computationally expensive way.

idavidmcdonald (approved):
So looks good although I think a code comment here explaining why the code is doing what it is doing will be worth it please

I'd be interested to hear how the size of the largest county was decided to be used as the boundary of using the smarter way or not please. Why the largest county rather than say something 10 times smaller or ten times larger?

app/models/broadcast_message.py
quis: Comment added in a0d6c153b

idavidmcdonald (approved):

PR #3942, opened 2021-06-24
Don’t remove very small areas from polygons unless they are likely to contain errors

quis:
As part of importing the geographic data we remove very small areas<sup>1</sup>, which are likely to be artefacts introduced by subtracting the sea from the land using datasets of two separate resolutions.

We think this problem only happens with the wards and local authority data. So we should specify that the fix should only be applied there.

This has the advantage of not removing very small areas from other datasets, for example the Farne Islands<sup>2</sup> in the England polygon.

***

1. #3627
2. https://en.wikipedia.org/wiki/Farne_Islands


quis:
Closing this because the problem with the Farne Islands seems to be in the source data.

PR #3943, opened 2021-06-24
Fix invalid polygons while importing geographic data

quis:
The work in this PR is based on the excellent investigation by @klssmith documented here: https://gist.github.com/klssmith/a487e12cbf4f3c604606b45a37e42585

***

Some of the polygons in our source data are invalid. An invalid polygon is one that self intersects, in other words has a point which causes the boundary of the shape to cross itself.

This doesn’t cause an exception until we try to perform certain operations on one of these polygons, like intersecting them with another polygon. This is why we haven’t spotted that they are invalid until now.

This commit adds checks so that as we import the polygons we make sure they are valid.

If they are not valid, we can automatically fix them by just looking at the exterior boundary of the shape, and ignore any holes created by self intersection.

# Examples of invalid polygons

This one doubles back to the same point, so self-intersects at the base of the ‘spike’:

121551814-df9b2280-ca07-11eb-85dd-235b1238bf68

This one goes inside itself, causing a triangular-shaped void:

image (4)

This one has all the points in a line, so has zero area (we just remove this):

image (3)

This one needs splitting into two polygons, rather than one polygon that converges to a point of zero width:

image (2)

This one self intersects because it’s genuine donut 🍩 

image (5)

The only way I could think to fix this without filling the hole is manually remove one of the points to leave a little gap (about 50m so not significant):

image


leohemsted (commented):
looks good, great comments in the commits and code.

i'm a little concerned about the manual doughnut change to the electoral wards file though - if a new version comes out and we re-import, we'll need to fix the file again - will that get flagged by the asserts, or will we need to remember each time?

🍩 🍩 🍩

quis:
Without the gap the donut gets filled in. Like a jam donut. This changes its area enough that the `assert` on this line will trip: https://github.com/alphagov/notifications-admin/pull/3943/files#diff-8c8fde3505da9c209b3b71805e3c8b5ba985c9673e29cf4dbc5ee7ef94a7bee9R98

leohemsted (approved):

PR #3944, opened 2021-06-24
Bump utils to bring in new polygon simplification

quis:
We’ve changed our simplification a bit so:
- polygons have slightly more points (https://github.com/alphagov/notifications-utils/pull/873)
- the individual points have less precision (https://github.com/alphagov/notifications-utils/pull/872)
- and slightly more precision again 😜 https://github.com/alphagov/notifications-utils/pull/881

Overall this reduces the size of the data we’re storing from 74MB to 63MB, and should make any pages where we are rendering lots of coordinates load a bit quicker.

Full changes: https://github.com/alphagov/notifications-utils/compare/44.2.2...44.4.2


quis:
Closing in favour of rolling into https://github.com/alphagov/notifications-admin/pull/3943

quis:
Decided to split them out into separate PRs again…

app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -125,7 +125,7 @@ def polygons_and_simplified_polygons(feature):
 
     point_counts.append(simplified.point_count)
 
-    if simplified.point_count >= 200:
+    if simplified.point_count >= 250:
```
idavidmcdonald, line 128: Could you enlighten me to what the number 250 represents please and why it has been chosen? I don't see anything in this commit message to indicate it, nor anything in the prs that the commit message mentions

tests/app/main/views/test_broadcast.py
```diff
@@ -811,8 +811,8 @@ def test_broadcast_page(
         'England remove',
         'Scotland remove',
     ], [
-        'An area of 177,439.8 square miles Will get the alert',
-        'An extra area of 6,392.3 square miles is Likely to get the alert',
+        'An area of 166,426.2 square miles Will get the alert',
```
idavidmcdonald: Am I correct in thinking that all the changes in estimated areas are due to
- slightly different polygons due to differing behaviour with the simplifying
- slightly different polygons due to their coordinates being 5 decimal places so therefore are less accurate (although would expect to see less difference caused by this)

Or is it something to do with the 250 number above?

tests/app/main/views/test_broadcast.py
```diff
@@ -889,24 +889,24 @@ def test_preview_broadcast_areas_page(
             [[7, 8], [9, 10], [11, 12]],
         ],
         [
-            'An area of 543.3 square miles Will get the alert',
-            'An extra area of 1,595.9 square miles is Likely to get the alert',
+            'An area of 700 square miles Will get the alert',
+            'An extra area of 1,000 square miles is Likely to get the alert',
```
idavidmcdonald: I'm assuming something else has interfered with the area in this test because rounding 543.3 to 1 sig fig should go to 500, not 700. And the same for 1595.9 -> 2000.

idavidmcdonald (commented):
Generally looks good but few small comments

app/broadcast_areas/create-broadcast-areas-db.py
quis: Because of https://github.com/alphagov/notifications-utils/pull/873 polygons will have slightly more points after simplification. The hard limit is 6,000 points but we don’t need anywhere near this detail. 250 is just a number that’s slightly above the point count of the most detailed polygon in out collection. This is line is a check to ensure that we don’t accidentally introduce a change which significantly increases the number of points.

tests/app/main/views/test_broadcast.py
quis: Slightly different polygons because they have a bit more detail as of alphagov/notifications-utils#873, which is the same reason for increasing the max point count to 250, yes.

tests/app/main/views/test_broadcast.py
quis: Looking at the overall diff it’s 722 > 700 and 1498 > 1000.

Might not be reflected in the individual commits because I don’t always remember to regenerate the database for each commit.

app/broadcast_areas/create-broadcast-areas-db.py
idavidmcdonald: Thanks.

Optional improvement then is to move this 250 into a constant that indicates what is represents, something like

`MAX_ALLOWED_POLYGON_POINTS = 250` with a one liner for why this is a thing

idavidmcdonald (approved):

PR #3949, opened 2021-07-02
Clarify service title included in SMS charges

benthorner:
We've just had a couple of tickets from different users about this
issue, so it's worth explaining it in the UI. The service name is
only shown as part of the message when viewing the preview on the
next page, but we can still explain it in words while they edit.

## Screenshots (without personalisation)

| Before  | After |
| Screenshot 2021-07-02 at 14 19 47 | Screenshot 2021-07-02 at 14 19 25 |

## Screenshot (with personalisation)

| Before  | After |
| Screenshot 2021-07-02 at 14 19 58 | Screenshot 2021-07-02 at 14 19 34 |

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -686,10 +686,11 @@ def _get_content_count_error_and_message_for_template(template):
         if template.placeholders:
             return False, (
                 f'Will be charged as {message_count(template.fragment_count, template.template_type)} '
-                f'(not including personalisation)'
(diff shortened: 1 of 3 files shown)


benthorner:
Moving back to draft as @leohemsted pointed out this isn't accurate if the service switch off "prefix sms sender" in their settings. We'll need to discuss a reasonable approach before this goes in.

benthorner:
Closing as this needs some more thought. I'll write a card so we can discuss the approach first.

PR #3950, opened 2021-07-02
Explain about text message prefixes when editing

benthorner:
Supersedes: https://github.com/alphagov/notifications-admin/pull/3949

Previously this lead to 2 support tickets because the user didn't
understand why their messages was being split into 2 fragments. We
tried modifying the message about charges, but that made it more
complicated. Adding a hint should hopefully be enough.

## Screenshots (with SMS prefix turned on)

| Before  | After |
| Screenshot 2021-07-02 at 16 30 35 | Screenshot 2021-07-02 at 16 30 17 |

### DIFF
--- a/app/templates/views/edit-sms-template.html
+++ b/app/templates/views/edit-sms-template.html
@@ -16,7 +16,7 @@
     ) }}
 
     {% if current_service.prefix_sms %}
-      {% set content_hint = 'All messages will start with your service name' %}
+      {% set content_hint = 'Your message will start with your service name' %}
     {% endif %}
 
     {% call form_wrapper() %}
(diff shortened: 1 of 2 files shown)


CrystalPea (commented):
The technical side looks great, I feel the content might be better, worth checking with @karlchillmaid ☺️

benthorner:
@CrystalPea @karlchillmaid has reviewed the content and I've made the changes in b4bd97828927ba22dccdbe6e5f332d1735a9d9ca.

Karl did also suggest trying to show what the prefix will look like, but we agreed that's a bigger design-type change for later and this PR is good to go by itself.

CrystalPea (approved):

PR #3951, opened 2021-07-05
Hide details of go live request ticket from the user 

quis:
We put some content in the go live ticket which is for our benefit, for example notes about the organisation.

It’s hard for us to be able to say what we want here if we know that the person making a go live request is going to see those notes.

This commit changes go live requests so that the initial content of the ticket is hidden from the person raising it (in Zendesk it will appear as an ‘internal note’, rather than a ‘public reply’).

---

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/877

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -243,6 +243,7 @@ def submit_request_to_go_live(service_id):
         user_email=current_user.email_address,
         user_name=current_user.name,
         tags=current_service.request_to_go_live_tags,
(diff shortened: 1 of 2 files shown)


benthorner (approved):
I was thinking about this and wondered if we actually need the content we have currently: aside from the volume responses, I tend to look at everything else online.

Appreciate that's a bit out-of-scope and the change here is a net improvement to hide the internal content we currently include, so happy to approve now.

PR #3952, opened 2021-07-05
Emergency alerts content review

karlchillmaid:
This PR updates some of the UI content for emergency alerts.

Specifically:

- replacing the word 'broadcast' where it makes sense to
- replacing 'previous alerts' with 'past alerts' to be consistent with gov.uk/alerts
- making the 4 destructive actions consistent (discard, reject, stop and delete)
- updating some of the progress/action buttons to make their meaning clearer

## Why we’re doing this
We know that some of the buttons, links, page titles and other content:

- use the term `broadcast` as a noun – we want to move away from this
- ‘broadcast’ as a verb can sound a bit weird in some contexts – we want to make the UI as clear as possible for new users, even if this means using ‘send’ and ‘broadcast’
- are not consistent with the public facing content about emergency alerts
- are not consistent within the emergency alerts sender UI

We want to fix these things before onboarding more users – so any user research we do highlights the things we’ve missed, not the things we (think) we already know about.

## To discuss
Should we be changing the code as part of this PR to remove/reduce the use of 'broadcast' and make page URLs more consistent with page titles?

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -64,7 +64,7 @@ def broadcast_dashboard(service_id):
     )
 
 
-@main.route('/services/<uuid:service_id>/previous-alerts')
+@main.route('/services/<uuid:service_id>/past-alerts')
 @user_has_permissions()
 @service_has_permission('broadcast')
 def broadcast_dashboard_previous(service_id):
@@ -74,8 +74,8 @@ def broadcast_dashboard_previous(service_id):
             'cancelled',
             'completed',
         ),
-        page_title='Previous alerts',
-        empty_message='You do not have any previous alerts',
+        page_title='Past alerts',
+        empty_message='You do not have any past alerts',
         view_broadcast_endpoint='.view_previous_broadcast',
     )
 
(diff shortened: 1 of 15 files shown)


klssmith:
Is the idea to replace the term `broadcasting` everywhere in the content now? It's still in a few places, but I'm not sure if that is intentional.

For example, on the tour we still have "There’s no risk of broadcasting a real alert.", there still a few "Start broadcasting now" buttons, and a couple of places where we say things like "Finished broadcasting at..."

karlchillmaid:
Thanks @klssmith 

I think we may need to keep `broadcasting` as a verb in certain specific circumstances.

For example:

* Start broadcasting now
* Broadcasting (what you see when an alert is being sent out)
* Stop broadcasting

The main reason for this is that showing an alert as ‘Sending’ during a broadcast _might_ make users think it’s stuck somewhere (as they sometimes do with email and SMS on Notify). And if that label stays as ‘broadcasting’ then the actions that start and stop it should probably be consistent.

The main thing to avoid is using ‘broadcast’ as a noun – it’s not a broadcast, it’s an alert.

We may be able to use ‘send’ as a verb wherever it sounds more natural than ‘broadcast’, for example:
* Get ready to send
* Choose where to send this alert 
* Sent at [time]
* Sent by [team member]
* Stopped sending at [time]

This **is** inconsistent, but it means we confine the idea of ‘broadcasting’ to the part of the journey where it’s actually happening. We can test to see if this makes sense to users.

All of this is still up for debate.

@quis is going to get this PR merged while I’m on leave.

quis (approved):

PR #3953, opened 2021-07-05
Show status when job has exceeded the daily sending limit

servingUpAces:
If a job exceeds the daily sending limit, show that on the job page. The job is only created if the sending limit has been reached when the delivery app is processing the job, usually this error is caught at the time the CSV is uploaded and the job is not created.
exceeded_daily_limit

### DIFF
--- a/app/templates/partials/jobs/status.html
+++ b/app/templates/partials/jobs/status.html
@@ -21,8 +21,11 @@
     {% endif %}
   </p>
   {% if job.status == 'sending limits exceeded'%}
-    <p class="govuk-error-message" >
-    You have exceeded your daily sending limits. You can upload your CSV tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a>
+    <p class="govuk-error-message">
+        Notify cannot send these messages because you have reached your daily limit. You can only send {{ current_service.message_limit|format_thousands }} messages per day.
+    </p>
+    <p class="govuk-error-message govuk-!-margin-bottom-6">
+        Upload this spreadsheet again tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a> to raise the limit.
     </p>
     {% endif %}
 </div>
--- a/app/templates/partials/jobs/status.html
+++ b/app/templates/partials/jobs/status.html
@@ -11,7 +11,7 @@
       {% else %}
         Uploaded by {{ job.created_by.name }} on {{ job.created_at|format_datetime_short }}
       {% endif %}
-    {% else %}
+   {% else %}
       Sent by {{ job.created_by.name }} on {{ job.created_at|format_datetime_short }}
       {% if job.template.template_type == "letter" %}
         <p class="govuk-body" id="printing-info">
@@ -20,4 +20,9 @@
       {% endif %}
     {% endif %}
   </p>
+  {% if job.status == 'sending limits exceeded'%}
+    <p class="govuk-error-message" >
+    You have exceeded your daily sending limits. You can upload your CSV tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a>
+    </p>
+    {% endif %}
 </div>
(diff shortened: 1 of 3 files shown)


app/templates/partials/jobs/status.html
```diff
@@ -20,4 +20,9 @@
       {% endif %}
     {% endif %}
   </p>
+  {% if job.status == 'sending limits exceeded'%}
+    <p class="govuk-error-message" >
+    You have exceeded your daily sending limits. You can upload your CSV tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a>
```
benthorner, line 25: Minor: I think we should avoid encouraging users to contact us about this, as we only want to be increasing the limit in special cases (where they can't wait until tomorrow). Would be good to get a second opinion from @karlchillmaid.

tests/app/main/views/test_jobs.py
```diff
@@ -221,6 +221,26 @@ def test_should_show_job_without_notifications(
...
+    mock_get_notifications_with_no_notifications,
+    mock_get_service_data_retention,
+    fake_uuid,
+):
+    page = client_request.get(
+        'main.view_job',
+        service_id=service_one['id'],
+        job_id=fake_uuid,
+    )
+    assert normalize_spaces(page.select('main p')[1].text) == (
```
benthorner, line 239: Minor: this is brittle if we add or change the surrounding content on the page in future. Could we just do:

```
assert "You have exceeded..." in page.text
```

app/templates/partials/jobs/status.html
benthorner: We should also explain that the job will never be processed: it's ambiguous because just below the notifications are shown as "sending".

benthorner (commented):
Code looks good - much clearer experience for users 🏅 . 

I'd like to get a second opinion on the content before this goes in.

tests/app/main/views/test_jobs.py
quis: I prefer this approach because it’s much harder to see what’s changed. `assert … in page.text` gives gnarly assertions that are hard to unpick.

app/templates/partials/jobs/status.html
quis: My suggestion, adapting https://github.com/alphagov/notifications-admin/blob/cfe022bc7f673df28799e75422fff211460a194a/app/templates/partials/check/too-many-messages.html#L1-L23
```suggestion
    Notify cannot send these messages because you have reached your daily limit. You can only send {{ current_service.message_limit|format_thousands }} messages per day. Upload this spreadsheet again tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a> to raise the limit.
```

tests/app/main/views/test_jobs.py
benthorner: That's fair. My main concern was grabbing hold of a paragraph at a specific index, which could easily break (even though the paragraph is still there). How about just having `page.select('main').text`?

app/templates/partials/jobs/status.html
servingUpAces: @quis Would you like me to make the message for exceeding the daily limits the same? This is the message from the upload page, I had changed the message from the /jobs/{id} page. 
Current upload page error message: 
too-many-recipients

New error message on jobs page: 
process_job_exceeded

quis: I think that looks OK. Could be slightly improved by splitting into two paragraphs and adding some spacing, eg
```html
<p class="govuk-error-message">
    Notify cannot send these messages because you have reached your daily limit. You can only send {{ current_service.message_limit|format_thousands }} messages per day.
</p>
<p class="govuk-error-message govuk-!-margin-bottom-6">
    Upload this spreadsheet again tomorrow or <a class="govuk-link" href="https://www.notifications.service.gov.uk/support">contact the GOV.UK Notify team</a> to raise the limit.
</p>
```

app/templates/partials/jobs/status.html
```diff
@@ -20,4 +20,12 @@
       {% endif %}
     {% endif %}
   </p>
+  {% if job.status == 'sending limits exceeded'%}
```
quis, line 23: Confirmed this matches https://github.com/alphagov/notifications-api/blob/d781da2981c9d9fbc5627f05dc4658b53575fbc9/app/models.py#L1191

quis (approved):

PR #3956, opened 2021-07-05
Remove Sizewell and the ‘demo areas’ library

quis:
We added this for the 25 May demo. We’re unlikely to need it again.

We’re also unlikely to need this library again, so this commit removes the source and the code that creates it.

If we want to do a custom area again we can copy and paste from the code that creates the ‘test areas’ library.


idavidmcdonald (approved):
Looks good but I'll just mention something to double check if you haven't given it any thought.

Will the removal of these areas have any implications on viewing the demo alert sent to Sizewell in the services pages?

quis:
Urgh, yes it will. Need to think about how we can archive a library in that case. Not one for now so I’m going to close this.

PR #3957, opened 2021-07-05
Make calculating overlapping areas faster

quis:
At the moment, when looking at how much an arbitrary polygon overlaps known areas, we intersect the arbitrary polygon with the full resolution polygon of each potential known area.

By using the simplified polygons instead of the full resolutions ones we:
- query less data from SQLite
- pass less data around
- give Shapely a less complicated shape to do its intersection and calculations on

This makes it faster to calculate how much of each electoral ward a custom area overlaps.

We trade off a small amount of accuracy by doing this, but nothing significant considering that we round the numbers before displaying them to the user.

For larger polygons, which overlap many electoral wards we can also improve performance by using a `JOIN` to reduce the number of SQLite queries we are doing from _n_ to 1 (where _n_ is the number of potential overlapping areas).

It also turns out we were doing all of the above multiple times because `BroadcastMessage.count_of_phones_likely` appeared multiple times in the Jinja template. So we get some extra benefit by using [`werkzeug.cached_property`](https://tedboy.github.io/flask/generated/generated/werkzeug.cached_property.html) to make sure this is only calculated once per page view.

# Results

For just the part of the code which calculates the count of likely phones:

Place represented by custom area | Time to calculate number of phones | Time to calculate number of phones using `simple_polygons` | Time to calculate number of phones using `simple_polygons` and with query optimisation
---------------------------------|--------|--------| ----
Bristol                          | 0.07s  | 0.02s | 0.01s
Skye                             | 0.02s  | 0.01s | 0.01s 

To render a whole page with a large polygon from our library (which uses the overlap of the bleed area to count the number of likely phones):

Library polygon | Time to render preview page before | Time to render preview page after | Time to render preview page with caching
---|---|---|---
Lincolnshire | 1.01s | 0.76s | 0.32s

### DIFF
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -171,7 +171,7 @@ class CustomBroadcastArea(BaseBroadcastArea):
     @cached_property
     def count_of_phones(self):
         return sum(
-            area.polygons.ratio_of_intersection_with(self.polygons) * area.count_of_phones
+            area.simple_polygons.ratio_of_intersection_with(self.polygons) * area.count_of_phones
             for area in self.overlapping_areas
         )
 
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -86,6 +86,12 @@ class BroadcastArea(BaseBroadcastArea, SortableMixin):
     def __init__(self, row):
         self.id, self.name, self._count_of_phones, self.library_id = row
 
+    @classmethod
+    def from_row_with_simple_polygons(cls, row):
+        instance = cls(row[:4])
+        instance.simple_polygons = Polygons(row[4])
+        return instance
+
     @cached_property
     def polygons(self):
         return Polygons(
@@ -162,7 +168,7 @@ class CustomBroadcastArea(BaseBroadcastArea):
     def overlapping_areas(self):
         if not self.polygons:
             return []
-        return broadcast_area_libraries.get_areas([
+        return broadcast_area_libraries.get_areas_with_simple_polygons([
             overlap.data for overlap in rtree_index.query(
                 Rect(*self.polygons.bounds)
             )
@@ -228,6 +234,13 @@ class BroadcastAreaLibraries(SerialisedModelCollection, GetItemByIdMixin):
 
         areas = BroadcastAreasRepository().get_areas(area_ids)
         return [BroadcastArea(area) for area in areas]
+
+    def get_areas_with_simple_polygons(self, *area_ids):
+        if len(area_ids) == 1 and isinstance(area_ids[0], list):
+            area_ids = area_ids[0]
+
+        areas = BroadcastAreasRepository().get_areas_with_simple_polygons(area_ids)
+        return [BroadcastArea.from_row_with_simple_polygons(area) for area in areas]
 
 
 broadcast_area_libraries = BroadcastAreaLibraries()
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -164,7 +164,7 @@ class CustomBroadcastArea(BaseBroadcastArea):
 
     simple_polygons = polygons
 
-    @property
+    @cached_property
     def overlapping_areas(self):
         if not self.polygons:
             return []
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -227,18 +227,11 @@ class BroadcastAreaLibraries(SerialisedModelCollection, GetItemByIdMixin):
     def __init__(self):
         self.items = BroadcastAreasRepository().get_libraries()
 
-    def get_areas(self, *area_ids):
-        # allow people to call `get_areas('a', 'b') or get_areas(['a', 'b'])`
-        if len(area_ids) == 1 and isinstance(area_ids[0], list):
-            area_ids = area_ids[0]
-
+    def get_areas(self, area_ids):
         areas = BroadcastAreasRepository().get_areas(area_ids)
         return [BroadcastArea(area) for area in areas]
 
-    def get_areas_with_simple_polygons(self, *area_ids):
-        if len(area_ids) == 1 and isinstance(area_ids[0], list):
-            area_ids = area_ids[0]
-
+    def get_areas_with_simple_polygons(self, area_ids):
         areas = BroadcastAreasRepository().get_areas_with_simple_polygons(area_ids)
         return [BroadcastArea.from_row_with_simple_polygons(area) for area in areas]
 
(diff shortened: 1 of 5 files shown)


app/broadcast_areas/models.py
```diff
@@ -229,5 +235,12 @@ def get_areas(self, *area_ids):
         areas = BroadcastAreasRepository().get_areas(area_ids)
         return [BroadcastArea(area) for area in areas]
 
+    def get_areas_with_simple_polygons(self, *area_ids):
+        if len(area_ids) == 1 and isinstance(area_ids[0], list):
```
benthorner, line 239: Could we just make `area_ids` a simple array, to reflect how it's used? It looks like we only use the varargs behaviour for the tests, but it's not much more work to put `[]` around the test input.

app/broadcast_areas/models.py
```diff
@@ -229,5 +235,12 @@ def get_areas(self, *area_ids):
...
+            area_ids = area_ids[0]
+
+        areas = BroadcastAreasRepository().get_areas_with_simple_polygons(area_ids)
+        return [BroadcastArea.from_row_with_simple_polygons(area) for area in areas]
```
benthorner, line 243: Very optional, would it be more conventional to use a `map` here?

app/broadcast_areas/models.py
```diff
@@ -86,6 +86,12 @@ class BroadcastArea(BaseBroadcastArea, SortableMixin):
     def __init__(self, row):
         self.id, self.name, self._count_of_phones, self.library_id = row
 
+    @classmethod
+    def from_row_with_simple_polygons(cls, row):
```
benthorner, line 90: It's a shame we can't pass this directly to the constructor, so that it can optionally handle simple polygons. Relying on numeric indices it's quite confusing in general; did we ever consider [using a dict instead](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.row_factory)?

app/models/broadcast_message.py
```diff
@@ -172,14 +172,14 @@ def approved_by(self):
     def cancelled_by(self):
         return User.from_id(self.cancelled_by_id)
 
-    @property
+    @cached_property
```
benthorner, line 175: Should all of these properties be cached? My understanding is that this is readonly model, so it seems easier to just compute everything once, even if it (currently) doesn't take very long.

app/broadcast_areas/repo.py
```diff
@@ -141,6 +141,23 @@ def get_areas(self, area_ids):
 
         return areas
 
+    def get_areas_with_simple_polygons(self, area_ids):
+        q = """
+        SELECT broadcast_areas.id, name, count_of_phones, broadcast_area_library_id, simple_polygons
+        FROM broadcast_areas
+        JOIN broadcast_area_polygons on broadcast_area_polygons.id = broadcast_areas.id
+        WHERE broadcast_areas.id IN ({})
+        """.format(("?," * len(area_ids))[:-1])
```
benthorner, line 150: Optional: would it be more conventional to do `",".join("?" * len(area_ids))`?

tests/app/main/views/test_broadcast.py
```diff
@@ -999,6 +1002,11 @@ def test_preview_broadcast_areas_page_with_custom_polygons(
         for item in page.select('ul li.area-list-key')
     ] == expected_list_items
 
+    # Make sure we haven’t run a SQLite query per overlapping area.
+    # We will have called BroadcastAreasRepository.get_areas_with_simple_polygons
+    # instead which does a `JOIN` in SQL instead of a loop in Python
+    assert mock_get_simple_polygons.called is False
```
benthorner, line 1008: I don't think we need to test this: it's effectively a performance test, and we don't generally write tests for performance in the main test suite.

Personally, I'm comfortable just seeing we're calling the right method, and I can't imagine a scenario where someone would change it back.

app/broadcast_areas/models.py
```diff
@@ -162,7 +168,7 @@ def polygons(self):
     def overlapping_areas(self):
```
benthorner, line 168: Optional, could this be a cached property as well?

benthorner (commented):
Very nice explanation of the changes 💯.

I've got a couple of comments about bits of code where I think we're doing more than we to.

app/broadcast_areas/models.py
quis: Decided not to use a `map` because we like the consistency of comprehension: it covers all scenarios, whereas `map` doesn’t always work if you want to call `len` on it, for example.

app/models/broadcast_message.py
quis: We don’t access all the properties on every page, so there’s a speed saving in only computing them when they are needed (but still not recomputing them if they’re needed more than once).

The overhead of `cached_property` is very small.

app/broadcast_areas/models.py
quis: Changed in e0e893f – that’s both you and @leohemsted who don’t like it.

app/broadcast_areas/models.py
quis: Didn’t know that was a thing, looks nice, seems a bit too big a change to include in this PR, will consider in the future.

tests/app/main/views/test_broadcast.py
quis: Would you just remove all these lines and the associated mock?

app/broadcast_areas/models.py
quis: Can’t hurt can it?

```suggestion
    @cached_property
    def overlapping_areas(self):
```

app/broadcast_areas/repo.py
quis: Yup, that is nicer, changed in d481b6ea7

app/models/broadcast_message.py
benthorner: Ahh, I see my comment can be interpreted in two ways: what I meant was "Should all of the other properties be cached as well?". It's a minor concern, though, as it's more for consistency / simplicity than a definite performance gain.

app/broadcast_areas/models.py
benthorner: Can you remove the redundant `@property` in 6cf03a6?

tests/app/main/views/test_broadcast.py
benthorner: Yes, the existing test is worth keeping, but we don't need [the new mock / assertions](https://github.com/alphagov/notifications-admin/pull/3957/files/429bbbaf321b3bdcd5c136fd3d45443c9ada8dbc..27d8a502289a1c7d1e2df13f12afaeb78ba03da5#diff-c1dd0f43b852bb0c4a074bdebe772a845e3c25ebffb3f69d72be0831fb98e8f9).

quis: I’ve squashed this out

app/broadcast_areas/models.py
quis: I’ve squashed that out in de364bb

benthorner (approved):
Thanks for addressing my comments ⭐

PR #3958, opened 2021-07-06
Remove March to June section from roadmap page

karlchillmaid:
While we work on what our roadmap should look like in the future, we need to:

* remove the March to June section
* check that the rest of the roadmap is accurate

This content has been sense checked by Katie M.


benthorner (approved):
🚗 🗺️

PR #3959, opened 2021-07-08
Audit when services are suspended or archived

benthorner:
https://www.pivotaltracker.com/story/show/178770416

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -67,6 +67,17 @@ def create_broadcast_account_type_change_event(
     )
 
 
+def create_suspend_service_event(
+    service_id,
+    suspended_by_id,
+):
+    _send_event(
+        'suspend_service',
+        service_id=service_id,
+        suspended_by_id=suspended_by_id,
+    )
+
+
 def _send_event(event_type, **kwargs):
     event_data = _construct_event_data(request)
     event_data.update(kwargs)
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -78,6 +78,17 @@ def create_suspend_service_event(
     )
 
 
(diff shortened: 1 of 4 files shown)


servingUpAces (approved):
This looks alright. 
A couple of comments not related to merging this PR. 
1) A user that is not a platform admin can only see the archive or suspend buttons for a trial mode service, however, if I know the url as a regular user can suspend/archive a live service they can perform the action. Should we add extra protection for that?  
2) Do we want to add an audit event for resume service?

idavidmcdonald (approved):
Lovely jubbly

benthorner:
Thanks for the reviews @idavidmcdonald and @servingUpAces 👍.

> Do we want to add an audit event for resume service?

Sure, that should be quick and concerning if it's a broadcast service. I've added it as a stretch goal on the card.

>  if I know the url as a regular user can suspend/archive a live service they can perform the action

Great detective work! I noticed the tests didn't check this (concerning, but not critical) and now I see [the conditionals in the HTML](https://github.com/alphagov/notifications-admin/blob/61e52faaeebf1056ad7a3e5154bfa5b8e32e3f54/app/templates/views/service-settings.html#L455-L470) are inconsistent with the backend view. I agree we should fix that, and it's reasonable for it to be part of this card: the principle behind this work is to protect these features better more thoroughly.

Do we want it to be like:

- A service admin can delete a trial mode service.
- A Platform Admin can delete any service.
- A Platform Admin can suspend (/resume) any service.

That's what the HTML currently does/shows. Do we want to keep it like that? If we're comfortable with the HTML side I can add a task on my card to fix the backend views to enforce it there as well (with tests).

PR #3965, opened 2021-07-13
Refactor organisation invite form for reuse

quis:
It’s exactly the same code as `BaseInviteUserForm` so there’s really no need to duplicate it (and means that changes we make to `BaseInviteUserForm` in the future will get inherited).


benthorner (approved):
Nice 🚀 

(I was toying with a comment about this on your main PR.)

PR #3967, opened 2021-07-13
Clear user cache when broadcast service settings form is submitted

klssmith:
When the broadcast service settings form is submitted it now [removes all permissions for users](https://github.com/alphagov/notifications-api/pull/3277) in notifications-api. This means we need to clear the user cache.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -396,11 +396,14 @@ def service_confirm_broadcast_account_type(service_id, account_type):
         abort(404)
 
     if form.validate_on_submit():
+        cached_service_user_ids = [user.id for user in current_service.active_users]
+
         service_api_client.set_service_broadcast_settings(
             current_service.id,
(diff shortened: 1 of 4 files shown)


benthorner (approved):
Nice one 🇦🇺.

I find the caching in Admin is quite subtle / tangential, which is probably why this slipped through in the first place. While the code looks fine, could you help me understand where this was causing an issue? I tried it without locally and it seems to be work already - the team members page updates, at least 🤔.

klssmith:
I saw something yesterday on preview which I guessed was caused by caching issues - I went to change the permissions of a team member of a broadcast service who had no permissions (which I confirmed by checking the database) and the checkboxes were prefilled with a tick.

That didn't happen when I tried to replicate this locally - I may have missed a step that had happened on preview. But I could see that there was old data in the Redis (it still had the old permissions even though they'd been removed), so this still needs doing as it may cause other issues.

PR #3970, opened 2021-07-15
Audit permissions when they change for a user

benthorner:
https://www.pivotaltracker.com/story/show/178770155

Depends on: https://github.com/alphagov/notifications-python-client/pull/201

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -7,8 +7,8 @@ EVENT_SCHEMAS = {
     "update_user_email": {"user_id", "updated_by_id", "original_email_address", "new_email_address"},
     "update_user_mobile_number": {"user_id", "updated_by_id", "original_mobile_number", "new_mobile_number"},
     "remove_user_from_service": {"user_id", "removed_by_id", "service_id"},
-    "add_user_to_service": {"user_id", "invited_by_id", "service_id", "admin_roles"},
-    "set_user_permissions": {"user_id", "service_id", "original_admin_roles", "new_admin_roles", "set_by_id"},
+    "add_user_to_service": {"user_id", "invited_by_id", "service_id", "ui_permissions"},
+    "set_user_permissions": {"user_id", "service_id", "original_ui_permissions", "new_ui_permissions", "set_by_id"},
     "archive_user": {"user_id", "archived_by_id"},
     "change_broadcast_account_type": {"service_id", "changed_by_id", "service_mode", "broadcast_channel", "provider_restriction"},  # noqa: E501 (length)
     "archive_service": {"service_id", "archived_by_id"},
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -1,6 +1,6 @@
 from flask import request
 
-from app import events_api_client
+from app.notify_client.events_api_client import events_api_client
 
 EVENT_SCHEMAS = {
     "sucessful_login": {"user_id"},
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -7,7 +7,7 @@ EVENT_SCHEMAS = {
     "update_user_email": {"user_id", "updated_by_id", "original_email_address", "new_email_address"},
     "update_user_mobile_number": {"user_id", "updated_by_id", "original_mobile_number", "new_mobile_number"},
     "remove_user_from_service": {"user_id", "removed_by_id", "service_id"},
-    "add_user_to_service": {"user_id", "invited_by_id", "service_id"},
+    "add_user_to_service": {"user_id", "invited_by_id", "service_id", "admin_roles"},
     "set_user_permissions": {"user_id", "service_id", "original_admin_roles", "new_admin_roles", "set_by_id"},
     "archive_user": {"user_id", "archived_by_id"},
     "change_broadcast_account_type": {"service_id", "changed_by_id", "service_mode", "broadcast_channel", "provider_restriction"},  # noqa: E501 (length)
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -8,6 +8,7 @@ EVENT_SCHEMAS = {
     "update_user_mobile_number": {"user_id", "updated_by_id", "original_mobile_number", "new_mobile_number"},
     "remove_user_from_service": {"user_id", "removed_by_id", "service_id"},
     "add_user_to_service": {"user_id", "invited_by_id", "service_id"},
+    "set_user_permissions": {"user_id", "service_id", "original_admin_roles", "new_admin_roles", "set_by_id"},
     "archive_user": {"user_id", "archived_by_id"},
     "change_broadcast_account_type": {"service_id", "changed_by_id", "service_mode", "broadcast_channel", "provider_restriction"},  # noqa: E501 (length)
     "archive_service": {"service_id", "archived_by_id"},
@@ -34,6 +35,10 @@ def create_remove_user_from_service_event(**kwargs):
 
 def create_add_user_to_service_event(**kwargs):
     _send_event('add_user_to_service', **kwargs)
+
+
+def create_set_user_permissions_event(**kwargs):
+    _send_event('set_user_permissions', **kwargs)
 
 
 def create_archive_user_event(**kwargs):
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -4,7 +4,10 @@ from notifications_python_client.errors import HTTPError
 from notifications_utils.timezones import utc_string_to_aware_gmt_datetime
 from werkzeug.utils import cached_property
 
-from app.event_handlers import create_add_user_to_service_event
+from app.event_handlers import (
+    create_add_user_to_service_event,
+    create_set_user_permissions_event,
+)
 from app.models import JSONModel, ModelList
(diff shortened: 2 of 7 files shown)


app/models/user.py
```diff
@@ -4,6 +4,7 @@
 from notifications_utils.timezones import utc_string_to_aware_gmt_datetime
 from werkzeug.utils import cached_property
 
+from app.event_handlers import create_add_user_to_service_event
```
CrystalPea: wow, well done with how you dealt with circular imports 👏🏼 . I remember struggling with them once or twice 😳

app/models/user.py
```diff
@@ -215,8 +215,7 @@ def has_permissions(self, *permissions, restrict_admin_usage=False, allow_org_us
             return True
 
         if any(
-            self.has_permission_for_service(service_id, permission)
-            for permission in permissions
+            self.permissions_for_service(service_id) & set(permissions)
```
CrystalPea, line 227: I don't think I understand this new code here.

I understand we are checking if user has permissions for a service with `self.permissions_for_service(service_id)`, but what `& set(permissions)` is doing there?

app/event_handlers.py
```diff
@@ -8,6 +8,7 @@
     "update_user_mobile_number": {"user_id", "updated_by_id", "original_mobile_number", "new_mobile_number"},
     "remove_user_from_service": {"user_id", "removed_by_id", "service_id"},
     "add_user_to_service": {"user_id", "invited_by_id", "service_id"},
+    "set_user_permissions": {"user_id", "service_id", "original_admin_roles", "new_admin_roles"},
```
CrystalPea, line 11: *looking at commit message* ahh, this is what admin roles means?

I was thinking it's roles for admin users :o. Could this be unclear for people looking at this in the future, too? Why are we not just calling them permissions or user permissions?

CrystalPea (changes requested):
Hey, good work, only calling user permissions `admin roles` just because we use names for them that are user-facing and our frontend is called `admin` feels very confusing.

app/models/user.py
benthorner: So this is doing the same thing as before: it's checking there is an overlap between some of the required permissions (`permissions`) and the permissions a user actually has (`permissions_for_service`).

The `&` is doing [a set intersection](https://stackoverflow.com/questions/3697432/how-to-find-list-intersection). We use it in a few other places to e.g. [here](https://github.com/alphagov/notifications-admin/blob/master/app/models/service.py#L237). A couple of longer and (slightly) less efficient ways to write it (which I also find harder to read) would be:

```
# one alternative (much longer)
any(permission for permission in permissions if permission in self.permissions_for_service(service_id))

# another alternative (denser)
self.permissions_for_service(service_id).intersection(set(permissions))
```

I do find the way `has_permissions` works is confusing: it does an "or" with `permissions`, but I would expect it to do an "and". But making that clearer isn't in scope for this PR (and I can't think of an obvious change).

Does that help?

app/event_handlers.py
benthorner: I found this confusing too. The problem is that we have two different concepts with the same name:

- "API permissions" e.g. `send_emails`. These are what we actually store in the DB.
- "Admin permissions" / "Admin roles" e.g. `send_messages`. These are what the user sees.

We have [some code](https://github.com/alphagov/notifications-admin/blob/d414e6d345930f9a47ed1fc2fe0faf99d3769840/app/models/roles_and_permissions.py#L30) that maps from "API permissions" to "Admin permissions" and vice versa. The problem with just saying "permissions" (or even "user permissions") is that it's unclear which one we're talking about. Both concepts are present in the Admin app. When I was reading all this code for the first time, I found it really confusing because people have used the same word "permissions" everywhere, when they mean totally different things.

I've used "admin roles" because that's how we refer to them in other places in the code e.g. see the link above. I don't think we should introduce a third name for them, but I'm open to changing all the existing places where we say `...admin_roles` (or "permissions") if we can agree on a clearer name for them. We don't actually use the term "permissions" in the UI, so I think we call it something that makes sense to us.

I'd prefer to do any big renames in a follow-up PR.

Can we keep the existing name for now?

app/models/user.py
CrystalPea: yeah, that helps, thanks a lot!

app/event_handlers.py
CrystalPea: ok, if we were already using "admin roles" before this PR, I guess it can stay for now. Still, it's confusing.

Also we do use the word "permissions" in an edit permissions page for a user.

Maybe in the separate PR we could change it to "user_permissions_ui" and "user_permissions_db" ? That would be very unambiguous I think?

CrystalPea (approved):
I still think "admin_roles" is a very confusing name, but if we were already using it before this PR, then it's ok to fix it in a separate one 👍🏼 

Maybe in the separate PR we could change it to "user_permissions_ui" and "user_permissions_db" ? That would be very unambiguous I think?

app/event_handlers.py
benthorner: Thanks. I'm thinking about how that work in code e.g.

- `roles_and_permissions.py` => `permissions.py`
- `translate_permissions_from_db_to_admin_roles` => `translate_permissions_from_db_to_ui`
- `permissions` => `ui_permissions` / `db_permissions` (depending on the function)
- `original_admin_roles` => `original_ui_permissions` (similarly for other event fields)

After looking through the codebase some more, I don't think it's worth doing any more than that: it's a losing battle to try and make everyone write `ui_permissions` or `db_permissions` everywhere. The real problem is that we've got two concepts with the same name - changing that is the only way to fully remove the confusion.

As a small improvement, I can just add another commit to remove the use of the term "(admin) role" everywhere, as illustrated above. It's small enough to do on this PR, and avoids having two different kinds of JSON for the new events in the DB (in case someone changes permissions before I do the rename).

That sound OK to you?

CrystalPea: Hey, that sounds cool, thanks a lot! Only thing is I think `user_permissions_ui` and `user_permissions_db` would be even more useful than `ui_permissions` and `db_permissions`, as that would tell us quickly we are dealing with user permissions, not with service permissions (which is also something that can be confusing).

benthorner: Ah interesting. What's the different between a user permission and a service permission?

(I actually thought they were the same thing: user permissions for a service.)

CrystalPea: Ahh, service permissions are permissions that a service has, like `international_sms`, `sms`, `email` etc.

benthorner: Aha, I see what you mean. How about a compromise: if I rename this file to `user_permissions.py`, then it should be clear everything inside relates to the user and we can just say `ui_permissions` or `db_permissions`.

Would that work for you?

klssmith:
Just curious, was it a deliberate decision not to include the user_id of the person who made the change? Most other audit events have the two user ids (the person making the change and the id of the user being changed)

app/event_handlers.py
CrystalPea: you mean to rename `event_handlers.py` to `user_permissions.py`? That seems weird, doesn't that file have handlers for all sorts of events?

benthorner: Sorry, I mean the `roles_and_permissions.py` file. That's where all the code that says "admin_roles" currently lives. The only change in `event_handlers.py` would be to use the new wording we discussed above.

So we'd rename `roles_and_permissions.py` to `user_permissions.py`, and inside there we'd refer to `ui_permissions` and `db_permissions` to make it clear which we're talking about.

Does that make more sense?

benthorner:
> was it a deliberate decision not to include the user_id of the person who made the change?

Good spot @klssmith. It was an accidental decision since the user setting the permissions wasn't one of the available bits of data in the `set_permissions` method on the User model.

I agree we should audit who set the permissions, so I've amended the third commit to add a `set_by_id` field and changed the method so that data is passed in for inclusion in the event.

app/event_handlers.py
CrystalPea: Hey, thanks, that makes sense, I still think `user_permissions_ui` and `user_permissions_db` would be better, as you don't have to look what file they came from, but what you propose is an improvement.

benthorner: Hmm...I had a go at eradicating "admin roles", but it turned out to be quite a complicated change.

For now, I've added a new commit (832422fc666d3cade6dd4c1c5c61f318732bf662) to change it in the code for this PR. I've also added a follow-up task on the Pivotal card to remove other uses of "admin roles", so we have consistent terminology everywhere.

Do you want to have a final look over before this goes in?

CrystalPea: ship it 🚢

benthorner: @CrystalPea here's a follow-up PR to do the rest of the renaming: https://github.com/alphagov/notifications-admin/pull/3976.

I did try to prefix everything with `user_`, but it got very long / repetitive and I think it's clear enough. One future option for when it's not clear would be to use the module as part of the reference e.g.

```
import app.utils.user_permissions
...
user_permissions.all_ui_permissions # instead of just all_ui_permissions
```

Anyway, do you want to take a look at the PR?

PR #3973, opened 2021-07-20
show error banner rather than alert when registering an invalid key

leohemsted:
you can test this by trying to re-register the same key twice, or by letting the prompt timeout without touching your yubikey (30 secs).

Sample what-the-user-sees, a story in three parts:

image
image
image

Depending on which browser they use, the user may not see that second error panel, that's a chrome implementation detail.
Content will need a sprinkle of @karlchillmaid magic, but I'd rather show a generic message than something that varies from browser to browser but can be as cryptic and unhelpful as something like this.

> The operation either timed out or was not allowed. See: https://www.w3.org/TR/webauthn-2/#sctn-privacy-considerations-client.

### DIFF
--- a/app/assets/javascripts/authenticateSecurityKey.js
+++ b/app/assets/javascripts/authenticateSecurityKey.js
@@ -50,22 +50,17 @@
               });
             })
             .then(response => {
-              if (response.status === 403) {
-                // flask will have `flash`ed an error message up
-                window.location.reload();
-                return;
+              if (!response.ok) {
+                throw Error(response.statusText);
               }
 
-              return response.arrayBuffer()
-                .then(cbor => {
-                  return Promise.resolve(window.CBOR.decode(cbor));
-                })
-                .catch(() => {
-                  throw Error(response.statusText);
-                })
-                .then(data => {
-                  window.location.assign(data.redirect_url);
-                });
+              return response.arrayBuffer();
+            })
+            .then(cbor => {
+              return Promise.resolve(window.CBOR.decode(cbor));
+            })
+            .then(data => {
+              window.location.assign(data.redirect_url);
             })
             .catch(error => {
               console.error(error);
--- a/app/assets/javascripts/authenticateSecurityKey.js
+++ b/app/assets/javascripts/authenticateSecurityKey.js
@@ -71,7 +71,7 @@
               console.error(error);
               // some browsers will show an error dialogue for some
               // errors; to be safe we always display an error message on the page.
-              window.GOVUK.ErrorBanner.showBanner('Something went wrong');
+              window.GOVUK.ErrorBanner.showBanner();
             });
         });
     };
--- a/app/assets/javascripts/authenticateSecurityKey.js
+++ b/app/assets/javascripts/authenticateSecurityKey.js
@@ -7,6 +7,9 @@
         .on('click', function (event) {
           event.preventDefault();
 
+          // hide any existing error prompt
+          window.GOVUK.ErrorBanner.hideBanner();
+
           fetch('/webauthn/authenticate')
             .then(response => {
               if (!response.ok) {
@@ -67,9 +70,8 @@
             .catch(error => {
               console.error(error);
               // some browsers will show an error dialogue for some
-              // errors; to be safe we always pop up an alert
-              var message = error.message || error;
-              alert('Error during authentication.\n\n' + message);
+              // errors; to be safe we always display an error message on the page.
+              window.GOVUK.ErrorBanner.showBanner('Something went wrong');
             });
         });
     };
--- a/app/assets/javascripts/errorBanner.js
+++ b/app/assets/javascripts/errorBanner.js
@@ -10,8 +10,7 @@
   */
   window.GOVUK.ErrorBanner = {
     hideBanner: () => $('.banner-dangerous').addClass('govuk-!-display-none'),
-    showBanner: (errorMessage) => $('.banner-dangerous')
-      .html(`<span class="govuk-visually-hidden">Error:</span> ${errorMessage}`)
+    showBanner: () => $('.banner-dangerous')
       .removeClass('govuk-!-display-none')
       .trigger('focus'),
   };
--- a/app/assets/javascripts/errorBanner.js
+++ b/app/assets/javascripts/errorBanner.js
@@ -1,0 +1,18 @@
+(function (window) {
+  "use strict";
+
+  /*
+  This module is intended to be used to show and hide an error banner based on a javascript trigger. You should make
+  sure the banner has an appropriate aria-live attribute, and a tabindex of -1 so that screenreaders and keyboard users
+  are alerted to the change respectively.
+
+  This may behave in unexpected ways if you have more than one element with the `banner-dangerous` class on your page.
+  */
+  window.GOVUK.ErrorBanner = {
+    hideBanner: () => $('.banner-dangerous').addClass('govuk-!-display-none'),
+    showBanner: (errorMessage) => $('.banner-dangerous')
+      .html(`<span class="govuk-visually-hidden">Error:</span> ${errorMessage}`)
+      .removeClass('govuk-!-display-none')
+      .trigger('focus'),
+  };
+})(window);
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -49,7 +49,7 @@
               // some browsers will show an error dialogue for some
               // errors; to be safe we always display an error message on the page.
               const message = error.message || error;
-              window.GOVUK.ErrorBanner.showBanner('Something went wrong');
+              window.GOVUK.ErrorBanner.showBanner();
             });
         });
     };
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -30,16 +30,7 @@
             })
             .then((response) => {
               if (!response.ok) {
-                return response.arrayBuffer()
-                  .then((cbor) => {
-                    return Promise.resolve(window.CBOR.decode(cbor));
-                  })
-                  .catch(() => {
-                    throw Error(response.statusText);
-                  })
-                  .then((text) => {
-                    throw Error(text);
-                  });
+                throw Error(response.statusText);
               }
 
               window.location.reload();
@@ -48,7 +39,6 @@
               console.error(error);
               // some browsers will show an error dialogue for some
               // errors; to be safe we always display an error message on the page.
-              const message = error.message || error;
               window.GOVUK.ErrorBanner.showBanner();
             });
         });
--- a/app/assets/javascripts/registerSecurityKey.js
+++ b/app/assets/javascripts/registerSecurityKey.js
@@ -7,6 +7,9 @@
         .on('click', function(event) {
           event.preventDefault();
 
+          // hide any existing error prompt
+          window.GOVUK.ErrorBanner.hideBanner();
+
           fetch('/webauthn/register')
             .then((response) => {
               if (!response.ok) {
@@ -44,9 +47,9 @@
             .catch((error) => {
               console.error(error);
               // some browsers will show an error dialogue for some
-              // errors; to be safe we always pop up an alert
-              var message = error.message || error;
-              alert('Error during registration.\n\n' + message);
+              // errors; to be safe we always display an error message on the page.
+              const message = error.message || error;
+              window.GOVUK.ErrorBanner.showBanner('Something went wrong');
             });
         });
     };
--- a/app/main/views/webauthn_credentials.py
+++ b/app/main/views/webauthn_credentials.py
@@ -3,7 +3,6 @@ from fido2.client import ClientData
 from fido2.ctap2 import AuthenticatorData
 from flask import abort, current_app, flash, redirect, request, session, url_for
 from flask_login import current_user
-from werkzeug.exceptions import Forbidden
 
 from app.main import main
 from app.models.user import User
@@ -50,7 +49,7 @@ def webauthn_complete_register():
         )
     except RegistrationError as e:
         current_app.logger.info(f'User {current_user.id} could not register a new webauthn token - {e}')
-        return cbor.encode(str(e)), 400
+        abort(400)
 
     current_user.create_webauthn_credential(credential)
     current_user.update(auth_type='webauthn_auth')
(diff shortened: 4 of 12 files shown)


app/assets/javascripts/registerSecurityKey.js
```diff
@@ -44,9 +48,14 @@
...
               console.error(error);
               // some browsers will show an error dialogue for some
-              // errors; to be safe we always pop up an alert
+              // errors; to be safe we always display an error message on the page.
               var message = error.message || error;
-              alert('Error during registration.\n\n' + message);
+
+              const html = `<span class="govuk-visually-hidden">Error:</span> Something went wrong. "${message}"`;
+
+              $('.banner-dangerous')
```
benthorner, line 56: Could we extract this into a separate module / utility? That would be easier to test since we would be calling synchronous code, while the tests for this module can more easily stub / assert on some simple function calls. It would also avoid duplicating the approach for the auth module, and in general allow for managing any number of JS errors on a page.

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -98,12 +98,18 @@ describe('Register security key', () => {
...
-    })
+    // REMOVED alert spy as we no longer alert at all
+    // jest.spyOn(window, 'alert').mockImplementation((msg) => {
+    //   expect(msg).toEqual('Error during registration.\n\nerror')
+    //   done()
+    // })
 
     button.click()
+
+    // TODO: Work out where to put these asserts now that there is no longer a window.alert to spy on
```
benthorner, line 109: I left a comment above with my preferred idea. Another idea would be to test this in a spy of `console.error`, but this is surprising (hence not my preference).

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -44,9 +48,14 @@
             .catch((error) => {
               console.error(error);
               // some browsers will show an error dialogue for some
-              // errors; to be safe we always pop up an alert
+              // errors; to be safe we always display an error message on the page.
               var message = error.message || error;
-              alert('Error during registration.\n\n' + message);
+
+              const html = `<span class="govuk-visually-hidden">Error:</span> Something went wrong. "${message}"`;
```
benthorner, line 54: Slightly out-of-scope: do we want to show the full message? It's pretty unhelpful (and hard to read since it's all in bold). We could just log it for now and start with something generic, which we could then iterate based on feedback.

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -44,9 +48,14 @@
...
+
+              $('.banner-dangerous')
+                .html(html)
+                .removeClass('govuk-!-display-none');
```
benthorner, line 58: We should also make sure the dialogue has focus when it shows.

benthorner (commented):
Works locally for me 👍. 

I've left a few comments - hope they're help towards the full PR on this!

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -7,6 +7,10 @@
         .on('click', function(event) {
           event.preventDefault();
 
+          // hide any existing error prompt
+          $('.banner-dangerous')
```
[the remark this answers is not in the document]
leohemsted: it'll remove that flash when the user clicks the button.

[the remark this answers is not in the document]
leohemsted: maybe we should only hide when the ajax has completed? To avoid the flash of it disappearing then reappearing if there's an error 🤔

tests/javascripts/registerSecurityKey.test.js
```diff
@@ -23,13 +23,13 @@ describe('Register security key', () => {
...
 
-    // ensure window.alert() is implemented to simplify errors
-    jest.spyOn(window, 'alert').mockImplementation(() => {})
-
     document.body.innerHTML = `
       <a href="#" role="button" draggable="false" class="govuk-button govuk-button--secondary" data-module="register-security-key">
         Register a key
-      </a>`
+      </a>
+      <span class="govuk-error-message webauthn__js-error banner-dangerous js-error-visible">
```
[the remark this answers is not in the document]
leohemsted: removed

app/assets/javascripts/registerSecurityKey.js
leohemsted: done

app/assets/javascripts/errorBanner.js
```diff
@@ -0,0 +1,11 @@
+(function (window) {
+  "use strict";
+
+  window.GOVUK.ErrorBanner = {
+    hideBanner: () => $('.banner-dangerous').addClass('govuk-!-display-none'),
+    showBanner: (errorMessage) => $('.banner-dangerous')
+      .html(`<span class="govuk-visually-hidden">Error:</span> ${errorMessage}`)
+      .removeClass('govuk-!-display-none')
+      .focus(),
```
benthorner, line 9: The banners [needs a `tabindex`](https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/templates/components/banner.html#L9) in order for this to work. Could we just use the `banner` component, or does it need to be this one?

benthorner (commented):
The small diff hides the effort that went into this 🏅.

Can we simplify [the error handling for authentication](https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/main/views/webauthn_credentials.py#L115) before this goes in? Since the JS can now show any error properly, the extra complexity on the server side and [the client side](https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/assets/javascripts/authenticateSecurityKey.js#L50-L54) isn't adding any value but the duplicate use of a flash is quite confusing. If we want to show a custom error message from the server in future, we could always upgrade the authentication JS to work like [the register JS](https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/assets/javascripts/registerSecurityKey.js#L32), so there's no risk in simplifying and doing so would make the two parts more consistent with each other from an error handling perspective.

app/assets/javascripts/errorBanner.js
leohemsted: hmmm, i dont know - is there a reason we chose `govukErrorMessage` rather than `banner` for the existing no js/no webauthn errors? happy to switch them all to `banner` if poss

benthorner: Ah yes, I forgot about those. They're displayed a bit differently (inline, don't need focus) and represent technical issue, rather than a user one. We could make them appear the same, but I think they're also fine as they are.

Screenshot 2021-08-18 at 11 53 08

The new errors are more user-focussed (and need focus 😉) and we're displaying them like a banner already, so it would make sense to cement that relationship by using the banner component directly.

app/assets/javascripts/errorBanner.js
```diff
@@ -0,0 +1,11 @@
+(function (window) {
+  "use strict";
+
+  window.GOVUK.ErrorBanner = {
+    hideBanner: () => $('.banner-dangerous').addClass('govuk-!-display-none'),
```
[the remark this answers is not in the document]
leohemsted: i was chatting to tom about this, and we decided to run on the assumption that we'll only ever have one error message visible. I think that feels like a sensible decision in keeping with service design patterns etc. that's part of the reason i'm using `.banner-dangerous` (you might remember there was a `webauthn__js-error` class in my first draft pr but after chatting with tom i removed it).

Given that these js ones only trigger due to user action, I think if you come to the page and there's an error visible (or there's an error visible from a previous attempt at auth/registration), it's fine to overwrite that old error as it is unlikely to be relevant any more.

thoughts?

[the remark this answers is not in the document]
benthorner: Hmm...yeah it's a tricky one. The downside of accidentally using an existing flash is that it wouldn't have the `aria-live` attribute on it. Interestingly, [the existing flash code is designed to show multiple banners](https://github.com/alphagov/notifications-admin/blob/ff12ba689d36eed545e0b0ce74b7e3458415375f/app/templates/flash_messages.html#L4).

One way to accommodate existing banners would be to have a two-step process: hide the existing ones, show the new one (using a distinct class the find it). Another option would be to make all flashes `aria-live`.

I think the possibility of an existing flash is unlikely, so I'd be happy with just adding a comment in the JS code to explain: how it's meant to be paired with some hidden `aria-live` HTML but might interact with an existing flash.

What would you prefer?

app/assets/javascripts/errorBanner.js
leohemsted: ah cool, didnt realise they rendered so differently to normal errors.

I'll switch over to use `banner`, cheers for flagging 👍

leohemsted: turns out `banner` was harder to add the existing class `govuk-!-display-none` and attribute `aria-live=polite` to, so i'm sticking with the `govukErrorMessage` for now. would like to revisit the difference between these at some point (and ideally standardise the banner component to be more modifyable and use govukErrorMessage under the hood) but I don't think now is the time for that

app/assets/javascripts/authenticateSecurityKey.js
```diff
@@ -67,9 +70,8 @@
             .catch(error => {
               console.error(error);
               // some browsers will show an error dialogue for some
-              // errors; to be safe we always pop up an alert
-              var message = error.message || error;
-              alert('Error during authentication.\n\n' + message);
+              // errors; to be safe we always display an error message on the page.
+              window.GOVUK.ErrorBanner.showBanner('Something went wrong');
```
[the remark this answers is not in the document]
leohemsted: image

took some wrangling to get there but i did it 👍

app/templates/views/user-profile/security-keys.html
```diff
@@ -46,6 +46,15 @@
 {% endset %}
 
   <div class="govuk-grid-row">
+
+    {{ govukErrorMessage({
```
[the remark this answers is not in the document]
leohemsted: I don't think it's worth it, as this is only used twice. see https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

app/assets/javascripts/errorBanner.js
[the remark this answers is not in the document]
leohemsted: adding a comment to js file

benthorner (commented):
Nice technique getting the formatted error message 💯. We could always tweak the module in future to include some dynamic text in there as well, but I agree we don't need to so long as we continue to log errors.

I'd like to discuss my comment ☝️ before this goes in. There's a separate issue that the register endpoint [returns CBOR errors which are logged server-side anyway](https://github.com/alphagov/notifications-admin/blob/master/app/main/views/webauthn_credentials.py#L52) (we could simplify and not return anything).

leohemsted:
with regards to server side errors (as you raised here: https://github.com/alphagov/notifications-admin/pull/3973#pullrequestreview-731838112), i think the best option is to:

* remove the server side cbor error returned by the registration flow
* remove the server side flash thrown by the authentication flow
* instead, in the js, on a 400 or 403 response from the server, showErrorBanner


this will

* make the two flows much more consistent
* remove a bunch of extra JS logic around error handling

app/main/views/webauthn_credentials.py
```diff
@@ -50,7 +49,7 @@ def webauthn_complete_register():
         )
     except RegistrationError as e:
         current_app.logger.info(f'User {current_user.id} could not register a new webauthn token - {e}')
-        return cbor.encode(str(e)), 400
+        return '', 400
```
benthorner, line 52: This can just be `abort(400)` now I think.

app/main/views/webauthn_credentials.py
```diff
@@ -142,6 +125,12 @@ def _verify_webauthn_authentication(user):
             signature=request_data['signature']
         )
     except ValueError as exc:
+        # We don't expect to reach this case in normal situations - normally errors (such as using the wrong
+        # security key) will be caught in the browser inside `window.navigator.credentials.get`, and the js will
+        # error first meaning it doesn't send the POST request to this method. If this method is called but the key
+        # couldn't be authenticated, something went wrong along the way, probably:
+        # * The browser didn't implement the webauthn standard correctly, and let something through it shouldn't have
+        # * The key itself is in some way corrupted, or of lower security standard
```
benthorner, line 133: Suggest appending: ..."(currently we don't check key compatibility on registration)". And maybe also amending "lower security standard" to "uses a signature algo we don't support".

app/assets/javascripts/registerSecurityKey.js
```diff
@@ -30,16 +30,7 @@
...
-                  .then((cbor) => {
-                    return Promise.resolve(window.CBOR.decode(cbor));
-                  })
-                  .catch(() => {
-                    throw Error(response.statusText);
-                  })
-                  .then((text) => {
-                    throw Error(text);
-                  });
+                throw Error(response.statusText);
```
benthorner, line 33: We can also remove [the "bad request" test case associated with this](https://github.com/alphagov/notifications-admin/blob/c42fc071b5c15b7ab7a03d079006412414a3763e/tests/javascripts/registerSecurityKey.test.js#L135) so it matches [the tests for the auth module](https://github.com/alphagov/notifications-admin/blob/c42fc071b5c15b7ab7a03d079006412414a3763e/tests/javascripts/authenticateSecurityKey.test.js#L217), although I think we should rephrase it to e.g. "server error" to make it more general. Also not really sure what that comment in the auth module tests means so we could lose that.

benthorner (approved):
Double checked again and this works locally for me for both registration and authentication ✅.

Thanks for addressing my comments - the end result is much simpler. Just a few bits left to tidy up...

app/assets/javascripts/authenticateSecurityKey.js
```diff
@@ -50,18 +50,16 @@
...
+                throw Error(response.statusText);
               }
 
               return response.arrayBuffer()
                 .then(cbor => {
                   return Promise.resolve(window.CBOR.decode(cbor));
                 })
                 .catch(() => {
-                  throw Error(response.statusText);
+                  throw Error('Could not decode cbor');
```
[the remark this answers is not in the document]
leohemsted: ah even better

app/main/views/webauthn_credentials.py
leohemsted: good point, more internally consistnet

app/assets/javascripts/registerSecurityKey.js
leohemsted: got rid of everything, but `authenticates a credential and redirects based on the admin app response` is raising an error. It's not failing tests but it's really irritating me, and I can't figure out how to fix it. I wonder if the console error mock is being torn down as soon as `done` is called even though there's still stuff to go on in the main tested code.

app/assets/javascripts/authenticateSecurityKey.js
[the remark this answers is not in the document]
leohemsted: also meant i could get rid of this nested promise chain and put it one upper level chain 👍

app/assets/javascripts/registerSecurityKey.js
benthorner: This is happening because [you added a `Promise.reject` after a `done()`](https://github.com/alphagov/notifications-admin/pull/3973/commits/fa4869793b49dbeb79cfdfa89f3a68c7ca91df24#diff-3c23eec76b884c81ad3a45859acb8f168d3e51c3db15566b070312afd1a2c81bR139), which means the `console.error` mock has been removed. I think what you actually want there is a `Promise.resolve(ok: true,...)` to match the previous test. Probably best to make the `done()` last as well, to ensure all mocks are still in place.

leohemsted: Yeah, that's right.

I added the promise.reject because it was throwing an error as the return value of the promise was previously undefined, and then the next step in the promise chain tries to check response.ok

so either I do `Promise.resolve({ok: true, ...})` and then put the done in a spy on window.location.assign, or I do a `Promise.reject` and put the done call in a spy on showErrorBanner

Both of these work fine and don't log out any errors, but I don't really like either, because neither gives me a guarantee that the `window.fetch` mock that contains the expect was called. The only way to guarantee that is to call `done()` from within that spy.

Or to have a slightly circuitous thing where I set a value on a variable within that mock body, then check it in the final mock before calling `done()`.

Thoughts?

benthorner: I think we can rely on the previous test for that, which implicitly verifies the data flow of `/foo` between the second `window.fetch` and the subsequent redirect. If we assume that test passes, then we can be confident the block will run here. It is a pain to have to do the extra spies. I'd prefer the `Promise.resolve...` option as that's more accurate - or add a comment explaining why `Promise.reject...` is used as a shortcut.

leohemsted: update: used `expect.assertions(1)` in the final mock - this checks that exactly one assertion was called, which we know is in the middle mock so that must have been called

PR #3974, opened 2021-07-21
dont let people get into one-off flow for deleted templates

leohemsted:
previously we'd skip the template page entirely if someone didnt have manage templates/api keys permission. however, if the template is deleted you'd then go through the flow entering placeholders and stuff before it would then crash when trying to send.

instead, just bounce the user to the template page. It has the content and says when the template was deleted.

### DIFF
--- a/tests/app/main/views/test_templates.py
+++ b/tests/app/main/views/test_templates.py
@@ -587,6 +587,32 @@ def test_caseworker_redirected_to_set_sender_for_one_off(
     )
 
 
+@freeze_time('2020-01-01 15:00')
+def test_caseworker_sees_template_page_if_template_is_deleted(
+    client_request,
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -587,6 +587,34 @@ def test_caseworker_redirected_to_set_sender_for_one_off(
     )
 
 
+@freeze_time('2020-01-01 15:00')
+def test_caseworker_sees_template_page_if_template_is_deleted(
+    client_request,
+    mock_get_service_templates,
```
klssmith, line 593: `mock_get_service_templates` and `mock_get_template_folders` aren't needed for the test

klssmith (approved):

PR #3975, opened 2021-07-21
follow sign_in redirect even if you're already signed in

leohemsted:
my user flow is this:

* click on notify link in slack, email client, etc. `https://www.notifications.service.gov.uk/foo/bar`
* accidentally open it in the wrong browser where i'm not signed in
* url in that browser redirects to `https://www.notifications.service.gov.uk/sign-in?next=%2Ffoo%2Fbar`
* i copy that into clipboard and paste it into the correct browser where i'm signed in
* THE STINKY AND RUBBISH PAST: It redirects to my accounts screen
* THE NEW AND SHINY FUTURE: it redirects to the page where i want to go

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -15,17 +15,20 @@ from app.main import main
 from app.main.forms import LoginForm
 from app.models.user import InvitedUser, User
 from app.utils import hide_from_search_engines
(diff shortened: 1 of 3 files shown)


quis (approved):
so shiny ✨

PR #3978, opened 2021-07-23
Add Pyup comment to requirements.txt

quis:
This matches what’s in requirements.in, but the Pyup comments don’t automatically get copied when generating requirements.txt. This causes Pyup’s safety CI to complain we’re using a version with a known vulnerability (a vulnerability which doesn’t affect us since we don’t use websockets).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -49,7 +49,7 @@ docutils==0.15.2
     # via awscli
 et-xmlfile==1.1.0
     # via openpyxl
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Thanks for fixing ⭐ 

Interesting to learn about this bug. Looks like [this specific issue is already well known](https://github.com/jazzband/pip-tools/issues/391).

PR #3979, opened 2021-07-23
Stop checking for `send_messages` permission for broadcast pages

klssmith:
The `send_messages` permission has been deprecated for use with broadcast services, so we can drop support for it in the code. We were supporting both the old permissions and new permissions (`create_broadcasts` and `approve_broadcasts`) while we switched people over.

This also tidies up a few tests (see individual commits for details) but nothing in this PR should affect how things currently work.

### DIFF
--- a/tests/app/main/views/test_templates.py
+++ b/tests/app/main/views/test_templates.py
@@ -22,9 +22,7 @@ from tests.conftest import (
     TEMPLATE_ONE_ID,
     ElementNotFound,
     create_active_caseworking_user,
-    create_active_user_create_broadcasts_permissions,
     create_active_user_view_permissions,
-    create_active_user_with_permissions,
     create_letter_contact_block,
     create_template,
     normalize_spaces,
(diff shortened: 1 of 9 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -922,19 +920,16 @@ def test_should_be_able_to_view_a_template_with_links(
...
     service_one,
     mock_get_broadcast_template,
     mock_get_template_folders,
     fake_uuid,
-    user,
+    active_user_create_broadcasts_permission,
 ):
-    client_request.login(user)
+    client_request.login(active_user_create_broadcasts_permission)
+    active_user_create_broadcasts_permission['permissions'][SERVICE_ONE_ID].append('manage_templates')
```
idavidmcdonald, line 932: Optional: Might be preferred to use a fixture here instead but given it's just once I don't mind either way

idavidmcdonald (approved):
Excellent work, in particular your detailed and clear commit messages

PR #3981, opened 2021-07-26
Add support for single quote in SMS sender name

idavidmcdonald:
image

https://www.pivotaltracker.com/story/show/178705080

Review commit by commit

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -144,9 +144,9 @@ class BroadcastLength:
             )
 
 
-class LettersNumbersFullStopsAndUnderscoresOnly:
+class LettersNumbersSingleQuotesFullStopsAndUnderscoresOnly:
 
-    regex = re.compile(r'^[a-zA-Z0-9\s\._]+$')
+    regex = re.compile(r"^[a-zA-Z0-9\s\._']+$")
 
     def __init__(self, message='Use letters and numbers only'):
         self.message = message
--- a/tests/app/main/test_validators.py
+++ b/tests/app/main/test_validators.py
@@ -199,6 +199,7 @@ def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
         ('00111222333', True, 'Cannot start with 00'),
         ('UK_GOV', False, None),  # Underscores are allowed
         ('UK.GOV', False, None),  # Full stops are allowed
+        ("'UC'", False, None),  # Straight single quotes are allowed
     ]
 )
 def test_sms_sender_form_validation(
--- a/tests/app/main/test_validators.py
+++ b/tests/app/main/test_validators.py
@@ -187,46 +187,34 @@ def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
     MustContainAlphanumericCharacters()(None, _gen_mock_field(string))
 
 
+@pytest.mark.parametrize(
+    "sms_sender,error_expected,error_message",
+    [
+        ('', True, 'Cannot be empty'),
+        ('22', True, 'Enter 3 characters or more'),
+        ('333', False, None),
+        ('elevenchars', False, None),  # 11 chars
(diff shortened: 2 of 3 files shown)


tests/app/main/test_validators.py
```diff
@@ -187,46 +187,35 @@ def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
...
 
+@pytest.mark.parametrize(
+    "sms_sender,error_expected,error_message",
+    [
+        ('', True, 'Cannot be empty'),
+        ('22', True, 'Enter 3 characters or more'),
+        ('333', False, None),
+        ('elevenchars', False, None),  # 11 chars
+        ('twelvecharas', True, 'Enter 11 characters or fewer'),  # 12 chars
+        ('###', True, 'Use letters and numbers only'),
```
idavidmcdonald, line 198: @karlchillmaid worth pointing out that we display the error message 'Use letters and numbers only' when there is a character in the sms sender name that is not allowed. The error message is not strictly true. We technically allow, letters, numbers, full stops, underscores and straight single quote marks.

This PR does not make this particularly any worse as we already don't specify that we support full stops and underscores so I don't think the arguably incorrect error message should block this PR going in. However, it could be something that you wish to look at addressing in the future.

tests/app/main/test_validators.py
```diff
@@ -187,46 +187,35 @@ def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
...
+        ('00111222333', True, 'Cannot start with 00'),
+        ('UK_GOV', False, None),  # Underscores are allowed
+        ('UK.GOV', False, None),  # Full stops are allowed
+        ("'UC'", False, None),  # Straight single quotes are allowed
+    ]
+)
 def test_sms_sender_form_validation(
```
klssmith, line 205: Just an optional thing to change, but it might be nice to split this into two tests - one for valid senders and one for invalid senders. It makes it extra easy to see at a glance what's allowed and what's not allowed (like how we test email addresses in utils) and also means you don't need conditional logic in the tests

tests/app/main/test_validators.py
```diff
@@ -187,46 +187,34 @@ def test_if_string_contains_alphanumeric_characters_does_not_raise(string):
...
-    assert 'Enter 3 characters or more' == form.errors['sms_sender'][0]
-
-    form.sms_sender.data = '333'
     form.validate()
-    assert not form.errors
 
-    form.sms_sender.data = '00111222333'
-    form.validate()
-    assert "Cannot start with 00" == form.errors['sms_sender'][0]
+    if error_expected:
```
benthorner, line 217: Optional: you could combine the flag with the expected message - e.g. `expected_error_message` and just check if it's truthy.

app/main/validators.py
```diff
@@ -146,7 +146,7 @@ def __call__(self, form, field):
 
 class LettersNumbersFullStopsAndUnderscoresOnly:
```
klssmith, line 147: This class name isn't accurate now, but I can't think of a good alternative `LettersNumbersFullStopsSingleQuotesAndUnderscoresOnly` is getting a bit long 😄

klssmith (approved):

benthorner (approved):
Looks good 👍. Just one optional suggestion for a test. Thanks for confirming it works with both providers. Do you think we need to worry about international SMS, or just wait until someone tries to send internationally with it?

app/main/validators.py
idavidmcdonald: I changed it to this, only because at least then it is accurate (although I agree it is a bit long)

tests/app/main/test_validators.py
idavidmcdonald: Thanks, did consider this and ended up consciously deciding to go for the more explicit for readability. Although I don't feel very strongly about it

PR #3983, opened 2021-07-27
Always show form to choose / copy template

benthorner:
Previously a user couldn't copy an existing template if their service
only had a single notification channel enabled. While we could amend
the conditional to also show the radio buttons if the service has more
than one template, it's easier to always show it.

The only downside to always showing the form is that a user will see a
single radio button in the rare case they are adding the first template
and only one channel is enabled (for some reason).

Note that the form is currently broken if all notification channels are
disabled for a service. We don't fix that here.

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -187,19 +187,12 @@
       return changed;
     };
 
-    this.$singleNotificationChannel = (document.querySelector('div[id=add_new_template_form]')).getAttribute("data-channel");
-    this.$singleChannelService = (document.querySelector('div[id=add_new_template_form]')).getAttribute("data-service");
-
     this.actionButtonClicked = function(event) {
       event.preventDefault();
       this.currentState = $(event.currentTarget).val();
 
-      if (event.currentTarget.value === 'add-new-template' && this.$singleNotificationChannel) {
-        window.location = "/services/" + this.$singleChannelService + "/templates/add-" + this.$singleNotificationChannel;
-      } else {
-        if (this.stateChanged()) {
-          this.render();
-        }
+      if (this.stateChanged()) {
+        this.render();
       }
     };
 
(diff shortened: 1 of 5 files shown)


benthorner:
Note: this is in response to a user on xGov Slack, who experienced the issue.

CrystalPea (commented):
Hmm do I understand correctly, that the data attributes were meant to ensure that user only can create templates of their allowed type, and that they don't see unnecessary radios, but for some reason in addition it prevented the user from copying an existing template? :o 

I see how proposed solution is more workable, but potentially less aesthetic. I'm ok with it, but I feel it's @quis remit to say if this acceptable from design perspective.

quis:
Looking at [the pull request which introduced this functionality](https://github.com/alphagov/notifications-admin/pull/3558) we did this partly for Emergency Alerts services. We maybe made the incorrect assumption that services only sending one type of notification would likely not have complex template management needs, and therefore it would be better for them to skip the radio buttons too.

If you look at the associated Javascript you can see this is deliberate behaviour, not a bug: https://github.com/alphagov/notifications-admin/blob/2dd48de1a77afc17c2620f22b0ef02b2e9307524/app/assets/javascripts/templateFolderForm.js#L197-L203

I’d be tempted to say we keep it as-is because:
- it’s only one user
- it would make the first-time use of Emergency Alerts weird (the one radio button problem)
- at some point we should re-engineer ‘Copy template’ to work like how people expect (select the thing you want to copy then press a button, not vice versa)

quis:
Alternatively you could change it to only set the `data-channel` attribute for Emergency Alerts services, so those with any other single channel would get the extra option. That might be better than leaving it as-is.

CrystalPea:
> Alternatively you could change it to only set the data-channel attribute for Emergency Alerts services, so those with any other single channel would get the extra option. That might be better than leaving it as-is.

That sounds good

benthorner:
Thanks for your comments @quis @CrystalPea. For now I'd like to keep it as-is to avoid getting distracted from the other work I'm doing. I've written [a bug card](https://www.pivotaltracker.com/story/show/179038517) that captures some of the discussion here.

PR #3984, opened 2021-07-27
Catch cancel_letter errors from API

CrystalPea:
When we catch such error, if the message is recognised, redirect user to view_notification page and show the error message.

Needs to go in after: https://github.com/alphagov/notifications-api/pull/3296

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -17,7 +17,7 @@ from flask import (
     stream_with_context,
     url_for,
 )
-from notifications_python_client.errors import APIError
+from notifications_python_client.errors import APIError, HTTPError
 from notifications_utils.letter_timings import (
     get_letter_timings,
     letter_can_be_cancelled,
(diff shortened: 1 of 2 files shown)


app/main/views/notifications.py
```diff
@@ -185,7 +185,13 @@ def view_notification(service_id, notification_id):
 def cancel_letter(service_id, notification_id):
 
     if request.method == 'POST':
-        notification_api_client.update_notification_to_cancelled(current_service.id, notification_id)
+        try:
+            notification_api_client.update_notification_to_cancelled(current_service.id, notification_id)
+        except HTTPError as e:
+            if e.status_code == 400 and "already been cancelled" or "too late to cancel" in e.message:
```
benthorner, line 191: This will also succeed for any error with status code `400` since `"already been called"` is always `True`. We could fix it with this:

```
e.status_code == 400 and any(msg in e.message for msg in ["already been cancelled", "too late to cancel"])
```

We could add a test case to make sure we re-raise e.g. `pytest.param("other", marks=pytest.mark.xfail())`.

CrystalPea: Ahh, good spot, thank you!

CrystalPea: Ok, I've updated the conditional and added the xfail assertion 👍🏼  (it's first time I used an xfail assertion, neat! <3 )

benthorner (approved):

PR #3985, opened 2021-07-30
Revert "Reduce default broadcast expiry time"

quis:
Reverts alphagov/notifications-admin#3800

We set it to 4 hours for the first Reading demo, so we could test expiry within the working day. Now we should change it back to the maximum possible for flexibility in real emergencies. 

We can’t go longer than 24 hours because there is a risk that phones will forget they’ve received the alert and display it again.


benthorner (approved):
A clean revert 👍

PR #3988, opened 2021-08-02
Move backlink out of `<main>`

klssmith:
This is to fix issue 47: 'Page template 3.' from the DAC audit, which is about the backlink being inside the `<main>` tag on all pages it is on. This means it's the first thing you encounter when skipping to main content which is confusing. It also doesn't match the design system.

[Pivotal story](https://www.pivotaltracker.com/story/show/176963591)


tombye (approved):
These are great changes and make a massive difference when using a screen reader. It also has seemed to have made some of the code cleaner and moving the backlink into its own area is much easier to work with 👍🏻

quis:
Would it be possible to get before/after screenshots of a few of different pages just to make sure nothing has shifted around visually?

tombye:
I've tested the following pages in OSX Chrome, IE11 on win10 and win7, Safari on IOS 14 and Chrome on Android 9.

I took a sample of the pages changed, split by the layout used. I tested:
- withnav_template.html
  - /services/<service_id>/service-settings/name
  - /services/<service_id>/api/keys
  - /services/<service_id>/send/<notification_id>/set-sender
- withoutnav_template.html
  - /provider/<provider_id>
  - /user-profile/name
  - /support/general
- org_template.html
  - /organisations/<org_id>/settings/edit-name
- views/platform-admin/_base_template.html
  - /email-branding/<branding_id>/edit

Here's some screenshots to compare (excuse the slight differences in content):

|Before|After|
|---|---|
|page_with_left_nav_before|page_with_left_nav_after|
|page_without_left_nav_before|page_without_left_nav_after|
|org_page_with_left_nav_before|org_page_with_left_nav_after|
|platform_admin_page_with_left_nav_before|platform_admin_page_with_left_nav_after|
|page_with_left_nav_on_mobile_before|page_with_left_nav_on_mobile_after|

quis (approved):
I can’t see any differences in the screenshots. Thanks for checking 👍🏻

PR #3990, opened 2021-08-05
Give user profile change links more context

tombye:
They all currently say 'Change' which makes it confusing when they are viewed out of their context (ie. when all the links in the page are listed out by a screen reader).

This gives them a suffix relating to the thing they will change, like [the links on the service settings page](https://github.com/alphagov/notifications-admin/blob/master/app/templates/views/service-settings.html#L23).

## Before

profile_before_changes

## After

profile_after_changes


idavidmcdonald (approved):
Good work :)

PR #3991, opened 2021-08-06
Update content

karlchillmaid:
After discussing with Ale, we’re going to update the Thanks page content so that it matches the commitment we make on the Support page.

### DIFF
--- a/app/templates/views/support/thanks.html
+++ b/app/templates/views/support/thanks.html
@@ -22,7 +22,7 @@
           {% if out_of_hours %}
             We’ll reply within one working day.
           {% else %}
-            We’ll aim to read your message in the next 30 minutes and reply within one
+            We’ll aim to read your message in the next 30 minutes and we’ll reply within one
             working day.
           {% endif %}
         {% else %}
--- a/app/templates/views/support/thanks.html
+++ b/app/templates/views/support/thanks.html
@@ -22,14 +22,14 @@
           {% if out_of_hours %}
             We’ll reply within one working day.
           {% else %}
-            We’ll read your message in the next 30 minutes and reply within one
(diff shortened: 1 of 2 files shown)


app/templates/views/support/thanks.html
```diff
@@ -22,14 +22,14 @@
           {% if out_of_hours %}
             We’ll reply within one working day.
           {% else %}
-            We’ll read your message in the next 30 minutes and reply within one
+            We’ll aim to read your message in the next 30 minutes and reply within one
```
karlchillmaid, line 25:
```suggestion
            We’ll aim to read your message in the next 30 minutes and we’ll reply within one
```

app/templates/views/support/thanks.html
```diff
@@ -22,14 +22,14 @@
...
-            We’ll read your message in the next 30 minutes and reply within one
+            We’ll aim to read your message in the next 30 minutes and we’ll reply within one
             working day.
           {% endif %}
         {% else %}
           {% if out_of_hours %}
             We’ll read your message when we’re back in the office.
           {% else %}
-            We’ll read your message in the next 30 minutes.
+            We’ll aim to read your message in the next 30 minutes.
```
karlchillmaid, line 32: Can the reviewer please double check the scenario in which users see this content?

benthorner: This is for [anonymous tickets](https://github.com/alphagov/notifications-admin/pull/3991/files#diff-8580ead6559100907653c52ac77a0304e95c02e412667f87700964617c6754c7R658). Do you want to proceed with the change?

karlchillmaid: If it’s anonymous tickets in office hours, then yes please.

benthorner (approved):

PR #3993, opened 2021-08-09
Expire test and operator alerts after 4 hours

quis:
While testing alerts on these channels the MNOs sometimes need to restart their CBCs to make sure everything is failing over properly.

If the CBC does not come back up, for whatever reason, then we are left in a state where the alert can’t be cancelled.

To minimise the impact to the public in this scenario we should keep the expiry time at 4 hours for alerts sent on test channels. We [recently](https://github.com/alphagov/notifications-admin/pull/3985) increased it back up to 24 hours for all channels, so this in effect is reverting that change for channels that won’t be used in a real emergency.

### DIFF
--- a/app/main/views/broadcast.py
+++ b/app/main/views/broadcast.py
@@ -449,14 +449,14 @@ def approve_broadcast_message(service_id, broadcast_message_id):
         ))
 
     if current_service.trial_mode:
-        broadcast_message.approve_broadcast()
+        broadcast_message.approve_broadcast(channel=current_service.broadcast_channel)
         return redirect(url_for(
             '.broadcast_tour',
             service_id=current_service.id,
             step_index=6,
         ))
     elif form.validate_on_submit():
-        broadcast_message.approve_broadcast()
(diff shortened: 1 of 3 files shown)


quis:
# Reduce expiry time to 22 hours 30 minutes for real alerts

Theoretically the maximum expiry time of a broadcast should be 24 hours. If it goes over 24 hours there can be problems.

However we want to make it more conservative to mitigate two potential issues:

1. The CBC has a repetition period (eg 60 seconds) and a count (eg 1,440). If these were slightly inaccurate or generous it could take us over 24 hours. For this reason we should give ourselves half an hour of buffer.
2. It’s possibly that the CBC could interpret a UTC time as BST or vice versa. Until we’re sure that it’s using UTC everywhere, we need to remove another whole hour as buffer.

In total this means we remove 1 hour 30 minutes from 24 hours, giving an expiry time of 22 hours 30 minutes.

rjbaker (approved):
Discussed and agreed with external stakeholders.

leohemsted (approved):

leohemsted (approved):

PR #3994, opened 2021-08-10
Update utils to bring in coordinate transformation

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/915

***

At the moment, we are not doing any transformation of features before applying geometric algorithms to them. This is, in effect, assuming that the earth is flat.

This new version of utils implements the transformation of our polygons to a Cartesian plane. In other words, it converts them from being defined in spherical degrees to metres.

For the admin app this means we need to convert places where the code expects things to be measured in degrees to work in metres instead.

***

Because the process of converting from metres to degrees and back again does take some computation this pull request also does some optimisations which make sure the pages still load in a reasonable time.

***

Before | After
---|---
image | image


tombye (approved):
Discussed with @quis outside of here, I still can't check it working locally but approving it because:
- we're the only users at the moment
- once it's merged we can have a look on production
- the code looks good and answers any questions I had in my review

PR #3996, opened 2021-08-10
Accessibility fixes for interactive map

tombye:
Makes the controls and links inside it match GOVUK Frontend styles and:
- gives the map container a focus style matching that for GOVUK Frontend text inputs
- makes it all work in high contrast modes (on Windows and Firefox)

Note: the focus style for the container is applied with [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) so only appears when it gets focus directly, not when it does due to child elements (like the controls or links) getting focused. Browsers without support for [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) get the same styling for all forms of focus.

This doesn't have a story because most of the code is ported over from CSS written for the interactive map we had on gov.uk/alerts pages.

## New styles

I used Firefox's high contrast mode for these screenshots as an easy way to see those styles.

### Map container focus styles

| Normal styles | High contrast mode |
|---|---|
|map_container_focused|map_container_high_contrast_mode_firefox_focused|

### Map controls with normal styles

| First control hovered | First control focused | Second control focused | First control focused and hovered | Link focused |
|---|---|---|---|---|
|map_controls_first_button_hovered|map_controls_first_button_focused|map_controls_second_button_focused|map_controls_first_button_focused_hovered|map_links_focused|

### Map controls and links high contrast

Note: following GOVUK styles, there are no hover states in high contrast mode.

| Controls | Control focused | Link focused |
|---|---|---|
|map_controls_high_contrast_mode_firefox|maps_controls_high_contrast_mode_firefox_first_button_focused|map_links_high_contrast_mode_firefox_focused|

## Added accessibility information

This also adds WAI-ARIA attributes and content to the map to make sure it has an accessible name and accessible description. This means assistive tech' can tell users what the map is when they start interacting with it (the accessible name) and give help if they need it (the accessible description).

## Notes for reviewers

I've tested in all our supported browsers ([the GDS browser matrix](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2021) but with IE down to 9).

Reviewers are best to look at it in their browser and Firefox with its high contrast mode turned on:

Settings > General > Colours > Override the colours specified by the page with your selections above > Always

They should try using it with just keyboard as well as mouse/trackpad/touch.

### DIFF
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -108,7 +108,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
       height: 2px;
       background: $govuk-focus-text-colour;
 
-      // hide for focus:hover state
+      // hide in forced-color mode, where the normal focus state the divider is part of does not show
       @media (forced-colors: active) {
         display: none;
       }
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -10,15 +10,14 @@ $govuk-focus-colour: #FFDD00;
 $zoom-button-colour: govuk-colour("white");
 $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
 
-// The focus style is a 2-colour outline, made to match the GOVUK text input focus style, using
+// The map focus style is a 2-colour outline, made to match the GOVUK text input focus style, using
 // box-shadow. When colours are overridden, for example when users have Windows high contrast mode
-// on, box-shadows disappear, so the outline which is left as a single colour (defined by the OS,
-// to replace 'transparent').
+// on, box-shadows disappear, so the outline is left as a single colour (defined by the OS, to
+// replace 'transparent') which we offset from the map to mimic the 2-colour version.
 //
 // Leaflet adds focus styles with JS, through inline styles. Because of their higher precedence
 // we need to mark our overrides with !important.
 //
-// This also uses :focus-visible to stop it showing focus when you click zoom in/out.
 // This approach, taken from MDN, allows it to fall back to :focus for browsers without support:
 // https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#selectively_showing_the_focus_indicator
 .leaflet-container {
@@ -26,8 +25,6 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
   &:focus {
     box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
     outline: solid 3px transparent !important; // sass-lint:disable-line no-important
-    // You only see the outline in forced colour mode which doesn't have enough contrast with the
-    // map so this creates some space between them to mimic the two-colour default styling
     outline-offset: 3px;
   }
 
@@ -43,10 +40,11 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
 // https://design-system.service.gov.uk/components/button/
 //
 // GOVUK buttons have an invisible outline, used for high contrast modes. Our buttons are cropped by
-// their container so we use a border instead.
+// their container so we use their border instead.
 //
-// Also introduces a black divider between the buttons, made with a pseudo element, appearing on
-// focus to mimic the 2-colour style GOVUK Frontend buttons and links have
+// Also introduces a thick black divider between the buttons, made with a pseudo element, appearing
+// on focus. Together with the yellow background, this copies the 2-colour style GOVUK Frontend
+// buttons and links have.
 .leaflet-bar {
   border-radius: 0; // Remove rounded corners
   background: #CCCCCC;
@@ -66,7 +64,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
     }
 
     &:last-child {
-      // Allow it to contain the absolutely positioned divider bar we show between buttons
+      // Allow it to contain the absolutely positioned divider we show between buttons
       // when one is focused
       position: relative;
     }
@@ -99,7 +97,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
       background-color: $govuk-focus-colour;
     }
 
-    // 3px divider between buttons as a version of the underline from the GOVUK focus style
+    // Shared styles for the divider between buttons that appears when either are focused
     &:last-child:focus:not(:active):not(:hover):before,
     &:first-child:focus:not(:active):not(:hover) + a:before {
       content: "";
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -32,7 +32,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
   // from other sources (like clicks on child elements).
   &:focus:not(:focus-visible) {
     box-shadow: none;
-    outline: none;
+    outline: none !important; // sass-lint:disable-line no-important
   }
 }
 
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -26,21 +26,16 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
   &:focus {
     box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
     outline: solid 3px transparent !important; // sass-lint:disable-line no-important
-  }
-
-  &:focus:not(:focus-visible) {
-    box-shadow: none;
-  }
-
-  &:focus-visible {
-    box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
-    outline: solid 3px transparent !important; // sass-lint:disable-line no-important
-
     // You only see the outline in forced colour mode which doesn't have enough contrast with the
     // map so this creates some space between them to mimic the two-colour default styling
-    @media (forced-colors: active) {
-      outline-offset: 3px;
-    }
+    outline-offset: 3px;
+  }
+
+  // The map should only recieve focus from being tabbed to. This stops the focus state coming
+  // from other sources (like clicks on child elements).
+  &:focus:not(:focus-visible) {
+    box-shadow: none;
+    outline: none;
   }
 }
 
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -83,7 +83,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
       // visible outline so it can be set to a colour by the OS.
       border: solid 2px $govuk-focus-colour;
       box-sizing: border-box; // make sure height includes the border
-      line-height: 26px; // subtract the border from the height (normally 30px)
+      line-height: 22px; // subtract the border from the height (normally 26px)
 
       // The inline display box for the button text overlaps the button edge.
       // This is only visible in high contrast mode because of the background colour being set.
@@ -150,6 +150,11 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
     border-bottom-right-radius: 0;
   }
 
+}
+
+// buttons are 30px for touch-capable devices/browsers
+.leaflet-touch .leaflet-bar a:focus {
+  line-height: 26px; // subtract the border from the height (normally 30px)
 }
 
 // Map attribution links
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -49,6 +49,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
 // focus to mimic the 2-colour style GOVUK Frontend buttons and links have
 .leaflet-bar {
   border-radius: 0; // Remove rounded corners
+  background: #CCCCCC;
 
   @media (forced-colors: active) {
     background: canvasText;
@@ -58,6 +59,11 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
 
     // Outlines don't work because of the buttons being so close together so turn off
     outline: none;
+
+    &:first-child {
+      border-bottom: none;
+      margin-bottom: 1px;
+    }
 
     &:last-child {
       // Allow it to contain the absolutely positioned divider bar we show between buttons
@@ -91,11 +97,6 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
     &:focus:not(:active):not(:hover) {
       color: $govuk-focus-text-colour;
       background-color: $govuk-focus-colour;
-    }
-
-    // Retain the space between buttons now we're using the border in the focus style
-    &:first-child:focus {
-      margin-bottom: 1px;
     }
 
     // 3px divider between buttons as a version of the underline from the GOVUK focus style
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -121,7 +121,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
     &:last-child:focus:not(:active):not(:hover):before {
       left: -2px; // Subtract 2px left border
       top: -3px; // Subtract 2px top border and position so it overlaps space above button
-      width: calc(100% + 4px); // Subtract borders
+      width: calc(100% + 4px); // Compensate for borders
     }
 
     // Styles for the divider, specifc to the first button being focused
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -105,7 +105,7 @@ $zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
       height: 2px;
       background: $govuk-focus-text-colour;
 
-      // hide in forced-color mode, where the normal focus state the divider is part of does not show
+      // Hide in forced-color mode, since the button outlines suffice
       @media (forced-colors: active) {
         display: none;
       }
--- a/app/assets/stylesheets/map.scss
+++ b/app/assets/stylesheets/map.scss
@@ -1,0 +1,148 @@
+@import 'leaflet/dist/leaflet';
+
+// Styles to make the leaflet map match GOVUK (accessibility-focused) styling
+
+@import "settings/all";
+@import "helpers/all";
+
+// Reset focus colour to latest version, to match the rest of the app
+$govuk-focus-colour: #FFDD00;
+$zoom-button-colour: govuk-colour("white");
+$zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
+
+// The focus style is a 2-colour outline, made to match the GOVUK text input focus style.
+// When colours are overridden, for example when users have a dark mode, box-shadows disappear,
+// so the outline which is left as a single colour (defined by the OS, to replace 'transparent').
+//
+// Leaflet adds focus styles with JS, through inline styles. Because of their higher precedence
+// we need to mark our overrides with !important.
+//
+// This also uses :focus-visible to stop it showing focus when you click zoom in/out.
+// This approach, taken from MDN, allows it to fall back to :focus for browsers without support:
+// https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#selectively_showing_the_focus_indicator
+.leaflet-container {
+
+  &:focus {
+    box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
+    outline: solid 3px transparent !important; // sass-lint:disable-line no-important
+  }
+
+  &:focus:not(:focus-visible) {
+    box-shadow: none;
+  }
+
+  &:focus-visible {
+    box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
+    outline: solid 3px transparent !important; // sass-lint:disable-line no-important
+
+    // You only see the outline in forced colour mode which doesn't have enough contrast with the
+    // map so this creates some space between them to mimic the two-colour default styling
+    @media (forced-colors: active) {
+      outline-offset: 3px;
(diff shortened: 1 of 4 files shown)


tombye:
After an off-PR chat with @quis I've changed the map buttons to be square (in https://github.com/alphagov/notifications-admin/pull/3996/commits/c099829a5b18007bca8c50f7be762ac64ca6c65e). They now look like this:

| default | hover | focus first | focus last | focus and hover | high contrast default | high contrast focused |
|---|---|---|---|---|---|---|
|square_buttons_default|square_buttons_hover_first|square_buttons_focus_first|square_buttons_focus_last|square_buttons_focus_hover_first|square_buttons_high_contrast_default|square_buttons_high_contrast_focused|

benthorner (commented):
Phew, this was quite a learning curve. ⛰️

Thanks for all the comments and pointers, they definitely helped make this easier to review. 💯

tombye:
Just to say thanks @benthorner for taking the time to dig into this and to learn about all the styling/visual modes it involves. Quite a bit of work to take on, 💯 points for effort and for the detailed and considered comments.

app/templates/views/broadcast/partials/area-map-javascripts.html
```diff
@@ -29,16 +29,51 @@
 
 
   var mapElement = document.getElementById('area-list-map');
+  var mapContainer = mapElement.parentNode;
+  var grandparent = mapContainer.parentNode;
+  var isInDetails = grandparent.className.indexOf('govuk-details') > -1;
```
benthorner, line 34: Optional: this could be, more precisely `!=`.

app/templates/views/broadcast/partials/area-map-javascripts.html
```diff
@@ -29,16 +29,51 @@
...
-  if (grandparent.className.split(/\s+/).indexOf('govuk-details') > -1) {
+  if (isInDetails) {
     details = grandparent;
     details.open = true;
   }
 
+  label = labelPrefix + " " + getStringFromAreas(areas);
   mapElement.style.height = Math.max(320, window.innerHeight - mapElement.offsetTop - 100) + 'px';
+  mapElement.setAttribute('role', 'region');
+  mapElement.setAttribute('aria-label', label);
```
benthorner, line 75: Could we encapsulate all the label code in a single function? This could include the aggregation of the area names. Having all the label code together would make it easier to follow which variables are used for what.

Similarly for the description / associating it with the map element.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,168 @@
...
+// When colours are overridden, for example when users have a dark mode, box-shadows disappear,
+// so the outline which is left as a single colour (defined by the OS, to replace 'transparent').
+//
+// Leaflet adds focus styles with JS, through inline styles. Because of their higher precedence
+// we need to mark our overrides with !important.
+//
+// This also uses :focus-visible to stop it showing focus when you click zoom in/out.
+// This approach, taken from MDN, allows it to fall back to :focus for browsers without support:
+// https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#selectively_showing_the_focus_indicator
+.leaflet-container {
```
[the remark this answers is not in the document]
benthorner: That works much better, except for the area names, where the "remove" button is being said together with each area name:

Screenshot 2021-08-31 at 12 28 23

This got me wondering about the general accessibility of the lozenges above the map, which I appreciate is out-of-scope for this PR. When I focus on the "X" all I get is "remove, link" and I'm not sure what I'm removing.

Perhaps you could partially address the issue with the lozenges and avoid the trimming in the JS by setting an `aria-label` on each list item and using that as the source of the area names in the JS.

(The `X`s could be `describedby` the labels.)

app/templates/views/broadcast/partials/area-map-javascripts.html
```diff
@@ -29,16 +29,51 @@
...
+    var idx;
+
+    if (areasLen === 1) { return areas[0].textContent.replace(trimRegExp, ''); }
+
+    for (idx = 0; idx < areasLen; idx++) {
+      areaStrings.push(areas[idx].textContent.replace(trimRegExp, ''));
+    }
+
+    // always join last 2 areas with 'and', the rest with commas
+    return areaStrings.slice(0, areasLen - 1).join(', ') + ' and ' + areaStrings[areasLen - 1];
```
benthorner, line 55: Optional: I'm not sure the value of "and" is worth the additional complexity. A comma-separated list should be clear enough.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,179 @@
...
+
+  &:focus {
+    box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
+    outline: solid 3px transparent !important; // sass-lint:disable-line no-important
+  }
+
+  &:focus:not(:focus-visible) {
+    box-shadow: none;
```
benthorner, line 32: Should we also turn off the outline here so it behaves similarly in forced colours mode? Otherwise it remains when clicking the zoom buttons.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  &:focus-visible {
+    box-shadow: 0px 0px 0 3px $govuk-focus-text-colour, 0 0 0 6px $govuk-focus-colour;
+    outline: solid 3px transparent !important; // sass-lint:disable-line no-important
+
+    // You only see the outline in forced colour mode which doesn't have enough contrast with the
+    // map so this creates some space between them to mimic the two-colour default styling
+    @media (forced-colors: active) {
```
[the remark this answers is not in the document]
benthorner: Looking at this again, I'm actually confused why we need the conditional at all. Since the outline and the box-shadow will never appear together, could we just set its offset always? This also leads me to question if we need this block at all: could we just have the two above? I had a quick go with just the `focus` and `not` blocks and it seems to work OK in all scenarios.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,168 @@
...
+@import "settings/all";
+@import "helpers/all";
+
+// Reset focus colour to latest version, to match the rest of the app
+$govuk-focus-colour: #FFDD00;
+$zoom-button-colour: govuk-colour("white");
+$zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
+
+// The focus style is a 2-colour outline, made to match the GOVUK text input focus style.
+// When colours are overridden, for example when users have a dark mode, box-shadows disappear,
```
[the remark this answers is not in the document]
benthorner: Nice, thanks @tombye! I'm pleased it wasn't me misunderstanding something.

> I copied most of this comment from similar comments in GOVUK Frontend

Do you think it's worth a quick PR to fix the comment at source as well? Might help someone else...

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,168 @@
...
+
+  @media (forced-colors: active) {
+    background: canvasText;
+  }
+
+  a {
+
+    &:first-child {
+      // Remove rounded corners
+      border-top-left-radius: 0;
```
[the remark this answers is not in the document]
benthorner: Nice, it's now clearer in general that these are also extra overrides.

benthorner (commented):
Thanks @tombye 😊.

Sorry for the sporadic comments - GitHub held some back for the re-review, and put others through straight away 🤷. Anyway, I've resolved a bunch of comments but also added a few more, one or two of which are potential functional issues. I also think it's worth a quick refactoring on the JS before this goes in.

The more I dig into that divider, the more...intricate I see it is 😄.

app/assets/stylesheets/map.scss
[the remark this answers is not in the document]
tombye: Ah, that's my sloppy JS targeting. I was looking at other variants of the areas list (without the 'remove' buttons). That's easy enough to fix though.

The issue you mention about the 'remove' links has been raised in the accessibility testing done so far:

https://docs.google.com/spreadsheets/d/1KojEKnwirUNQmUBCi2NpkCelXVIHYyFrAZVSf_x0wTg/edit#gid=2036472444

I've added a comment referencing your comment to it, for whoever picks the bug up. My general feeling is the JS should be made to work with whatever markup pattern we choose to fix them.

app/templates/views/broadcast/partials/area-map-javascripts.html
tombye: I see what you're saying but I would like to keep this one, if only for consistency with how we write lists as sentences elsewhere (see https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/formatters.py#L199).

app/templates/views/broadcast/partials/area-map-javascripts.html
tombye: Good idea. Now added in https://github.com/alphagov/notifications-admin/pull/3996/commits/e2af2f63a41e48ce51fc4be3beb5893049e5d3b6.

app/assets/stylesheets/map.scss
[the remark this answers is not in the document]
tombye: Fixed up the JS so it doesn't include the 'Remove' link text in https://github.com/alphagov/notifications-admin/pull/3996/commits/e2af2f63a41e48ce51fc4be3beb5893049e5d3b6.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,179 @@
...
+      // Button focus is shown by a change in background colour.
+      // When colours are overridden, for example when users have Windows high
+      // contrast mode on, backgrounds disappear, so we need to ensure there's a
+      // visible outline so it can be set to a colour by the OS.
+      border: solid 2px $govuk-focus-colour;
+      box-sizing: border-box; // make sure height includes the border
+      line-height: 22px; // subtract the border from the height (normally 26px)
+
+      // The inline display box for the button text overlaps the button edge.
+      // This is only visible in high contrast mode because of the background colour being set.
```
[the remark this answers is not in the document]
tombye: What zoom level where you using? With that, I can add this to [the accessibility bugs spreadsheet](https://docs.google.com/spreadsheets/d/1KojEKnwirUNQmUBCi2NpkCelXVIHYyFrAZVSf_x0wTg/edit#gid=2036472444)

app/assets/stylesheets/map.scss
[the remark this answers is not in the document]
tombye: That's a good point. I've not used this technique of applying `:focus-visible` before so didn't quite get how it worked. I've made some changes in https://github.com/alphagov/notifications-admin/pull/3996/commits/9c9e7a7c61dda066d709f594afb763671238c2fd that should address the points you made.

app/assets/stylesheets/map.scss
tombye: Good point. Should be done in https://github.com/alphagov/notifications-admin/pull/3996/commits/9c9e7a7c61dda066d709f594afb763671238c2fd.

app/assets/stylesheets/map.scss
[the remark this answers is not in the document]
benthorner: Sure, thanks. This was at 200% on my 15" Macbook.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,179 @@
...
+      color: $govuk-focus-text-colour;
+      background-color: $govuk-focus-colour;
+    }
+
+    // Retain the space between buttons now we're using the border in the focus style
+    &:first-child:focus {
+      margin-bottom: 1px;
+    }
+
+    // 3px divider between buttons as a version of the underline from the GOVUK focus style
```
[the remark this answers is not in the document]
tombye: I've had a go at [rewriting the comments](https://github.com/alphagov/notifications-admin/pull/3996/commits/3b6b7c10ee14122432a398866991e9a17a3ef4b5) to make this clearer. I also [changed the styling for the space/border between the buttons](https://github.com/alphagov/notifications-admin/pull/3996/commits/4a47870221300c58518fc5fbb4aa2c1455911038) (not the thick one that appears on focus) which I hope should also make it clearer.

app/assets/stylesheets/map.scss
benthorner: Thanks! We discussed this in person as it's not behaving as expected: the map container always seems to be `focus-visible` in Firefox's forced colours mode. We agreed we may as well leave this here as it _should_ work and until it does it makes sense to see it here as it's consistent.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,168 @@
...
+
+    // Focused style sets the button to the focus colour
+    &:focus:not(:active):not(:hover) {
+      color: $govuk-focus-text-colour;
+      background-color: $govuk-focus-colour;
+    }
+
+    // Retain the space between buttons now we're using the border in the focus style
+    &:first-child:focus {
+      margin-bottom: 1px;
```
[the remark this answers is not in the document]
benthorner: I can see you fixed this in https://github.com/alphagov/notifications-admin/pull/3996/commits/ee3e3c6c90d6b9e10957792f45a10061a941bacd.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+
+// The map focus style is a 2-colour outline, made to match the GOVUK text input focus style, using
+// box-shadow. When colours are overridden, for example when users have Windows high contrast mode
+// on, box-shadows disappear, so the outline is left as a single colour (defined by the OS, to
+// replace 'transparent') which we offset from the map to mimic the 2-colour version.
+//
+// Leaflet adds focus styles with JS, through inline styles. Because of their higher precedence
+// we need to mark our overrides with !important.
+//
+// This approach, taken from MDN, allows it to fall back to :focus for browsers without support:
```
benthorner, line 21: Optional: I'm not sure what "This approach" and "it" are referring to here. Perhaps there used to be a paragraph above that provided the context.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+
+@import "settings/all";
+@import "helpers/all";
+
+// Reset focus colour to latest version, to match the rest of the app
+$govuk-focus-colour: #FFDD00;
+$zoom-button-colour: govuk-colour("white");
+$zoom-button-hover-colour: govuk-shade($zoom-button-colour, 10%);
+
+// The map focus style is a 2-colour outline, made to match the GOVUK text input focus style, using
```
benthorner, line 13: Optional: I found it helpful to remember the two colours are "black" and "yellow", just to orient myself when reviewing this. Arguably someone should just know that's what you mean, so I don't feel strongly about specifying it, especially as they might change.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+  &:focus:not(:focus-visible) {
+    box-shadow: none;
+    outline: none;
+  }
+}
+
+// Overrides for zoom controls to make them match GOVUK buttons
+// https://design-system.service.gov.uk/components/button/
+//
+// GOVUK buttons have an invisible outline, used for high contrast modes. Our buttons are cropped by
```
benthorner, line 42:
```suggestion
// GOVUK buttons usually have an outline that shows up in high contrast modes. Our buttons are cropped by
```

"invisible" is ambiguous as it could be interpreted as "transparent".

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+// their container so we use their border instead.
+//
+// Also introduces a thick black divider between the buttons, made with a pseudo element, appearing
+// on focus. Together with the yellow background, this copies the 2-colour style GOVUK Frontend
+// buttons and links have.
+.leaflet-bar {
+  border-radius: 0; // Remove rounded corners
+  background: #CCCCCC;
```
benthorner, line 50:
```suggestion
  background: #CCCCCC; // Gray background to give a thin divider between buttons when they're not in focus
```

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  a {
+
+    // Outlines don't work because of the buttons being so close together so turn off
+    outline: none;
+
+    &:first-child {
+      border-bottom: none;
+      margin-bottom: 1px;
```
benthorner, line 63:
```suggestion
      margin-bottom: 1px; // Thin divider to distinguish buttons
```

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+
+    // Shared styles for the divider between buttons that appears when either are focused
+    &:last-child:focus:not(:active):not(:hover):before,
+    &:first-child:focus:not(:active):not(:hover) + a:before {
+      content: "";
+      position: absolute;
+      height: 2px;
+      background: $govuk-focus-text-colour;
+
+      // hide in forced-color mode, where the normal focus state the divider is part of does not show
```
benthorner, line 108:
```suggestion
      // Hide in forced-color mode, since the button outlines suffice
```

We could fix it so it does "show". I think the point here is that we just don't need it.

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+    &:last-child:focus:not(:active):not(:hover):before {
+      left: -2px; // Subtract 2px left border
+      top: -3px; // Subtract 2px top border and position so it overlaps space above button
+      width: calc(100% + 4px); // Compensate for borders
+    }
+
+    // Styles for the divider, specifc to the first button being focused
+    &:first-child:focus:not(:active):not(:hover) + a:before {
+      left: 0px; // No borders on parent of :before when not focused
+      top: -1px; // No borders so just overlap space above parent button
```
benthorner, line 124:
```suggestion
      top: -1px; // Shift divider up 1px so it overlaps buttons equally
```

app/assets/stylesheets/map.scss
```diff
@@ -0,0 +1,173 @@
...
+      // hide in forced-color mode, where the normal focus state the divider is part of does not show
+      @media (forced-colors: active) {
+        display: none;
+      }
+    }
+
+    // Styles for the divider, specifc to the second button being focused
+    &:last-child:focus:not(:active):not(:hover):before {
+      left: -2px; // Subtract 2px left border
+      top: -3px; // Subtract 2px top border and position so it overlaps space above button
```
benthorner, line 117:
```suggestion
      top: -3px; // Shift divider up 1px so it overlaps buttons equally, and another 2px to compensate for the reduces height of the last button (due to the border)
```

If you reverse the order of these two blocks you could omit the first part of the sentence here to save space, if we assume people are reading top-to-bottom.

app/assets/stylesheets/map.scss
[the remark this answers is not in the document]
benthorner: Thanks. The style change does make it a lot simpler. I think the comments could be a bit more helpful, so I've added some optional suggestions based on my experience reviewing this.

benthorner (approved):
Thanks for doing even more work on this to address my comments 🚀.

GitHub decided to dump out all my comments again 🤷, but fortunately this is now a ✅ from me with just some optional suggestions to help future devs understand the CSS.

Hurray, it's over 🏖️.

benthorner (approved):
Tested with the latest commit and the remaining issue is fixed 🎉

app/assets/stylesheets/map.scss
tombye: I think it does depend on where you're coming from when you read this. I mean that the intention of a 2-colour focus style is for it to work across all possible backgrounds - which is a relatively new WCAG requirement. The actual colours matter less than their application and so may change or be split into multiple variants in future versions of design system. Not sure that makes any difference to a comment, which could be changed along with the colour values, but I thought worth explaining the context I was in when writing it.

benthorner: Yeah, that's fair. I just remember I couldn't picture what "2-colour outline" meant in relation to things I'd seen before, since my memory only registered the yellow and not the black.

I think it's fine to leave it as-is: someone changing this could should just know.

app/assets/stylesheets/map.scss
tombye: I'm not sure this is the same thing. The forced-color focus style is intentionally different, to match the look of high contrast interfaces. My original comment was about the focus style the divider was part of not showing. This one seems to say that it's not needed as part of the focus style applied in forced-color mode. My original words are pretty bad though. How about:
`// Hide in forced-color mode, since the focus style it's part of doesn't show`?

benthorner: That's pretty much the same, right? I don't get what you mean by "focus style it's part of".

Could you help me understand how my suggestion is incorrect?

tombye: I'm not saying it's incorrect. I'm saying that there's a bit of my original comment that says the divider is part of these styles:
buttons_first_focused
...and they aren't used in forced-color modes.
My thought was that saying the focus styles we use for forced-color modes suffice to show the button is focused isn't the same thing as one style being used over the other.

I don't think my suggested change helped explain that. I also don't think losing the point discussed here makes much difference so happy to move it to your version, I just wanted to explain my original comment.

tombye:
> Thanks for doing even more work on this to address my comments 🚀.

Thanks for all the work on this, much appreciated 💯 🏅.

> GitHub decided to dump out all my comments again 🤷

I think I can see all of them myself. They are a bit all over the place, I think thanks to changes in the code, but still reachable.

app/assets/stylesheets/map.scss
benthorner: I do see more what you're saying now. I do find it requires some mental effort to grasp, which might be because I don't work in frontend all the time. I'll try and break it down so you can see:

- "Hide in forced-color mode..." - fine
- "...since the focus style/state..." - ambiguous: could refer to the style/state of the divider itself (no), or some broader collection of styles/states across multiple elements (yes)
- "...it's part of..." - paradox: we're either asserting a divider is part of the rules for how it looks / state it's in (nested paradox), or that it's both a visual object and a style/state (type paradox)
- "...doesn't show" - ambiguous: does the style not apply (no), or is the style invisible (yes)

I'd find it easier to understand if it were something like: "Hide in forced-color mode, since the surrounding focus styles this is meant to complement don't show". I can see how both our versions can be helpful: together they explain why it's pointless to keep it. I don't feel strongly about what the final text is, though.

PR #3997, opened 2021-08-11
Revert "Add URLs to download the agreement without login"

quis:
This reverts commit 771f9166305c42788c3a4863dca3d24ce2059940.


quis:
Closing this because we think we still need the link to share with people on whose behalf the agreement has been signed, who might not themselves have a Notify account.

PR #4000, opened 2021-08-17
Update `email_access_validated_at` as soon as a you click a fresh link

quis:
When someone uses a fresh password reset or invite link they have proved that they have access to their inbox.

At the moment, when resetting a user’s email address we wait until after they’ve put in the 2FA code before updating the timestamp which records when they last validated their email address<sup>1</sup>.

We can’t think of a good reason that we need the extra assurance of a valid 2FA code to assert that the user has access to their email – they’ve done that just by clicking the link. When the user clicks the link we already update their failed login count before they 2fa. Think it makes sense to handle `email_access_validated_at` then too.

As a bonus, the functional tests never go as far as getting a 2FA code after a password reset<sup>2</sup>, so the functional test user never gets its timestamp updated. This causes the functional tests start failing after 90 days. By moving the update to this point we ensure that the functional tests will keep passing indefinitely.

1. [This code in the API](https://github.com/alphagov/notifications-api/blob/91542ad33eb995c9b3e617cde30a9714846dd37b/app/dao/users_dao.py#L131) which is called by this code in the admin app: https://github.com/alphagov/notifications-admin/blob/9ba37249a47f08b1396e63a34bc591c781739686/app/utils/login.py#L26)
2. https://github.com/alphagov/notifications-functional-tests/blob/5837eb01dc5c9b5bb255d2d1c54e64f778ff4f4e/tests/functional/preview_and_dev/test_email_auth.py#L43-L46

***

https://www.pivotaltracker.com/story/show/172893003

### DIFF
--- a/app/main/views/new_password.py
+++ b/app/main/views/new_password.py
@@ -33,6 +33,9 @@ def new_password(token):
         flash('The link in the email has already been used')
         return redirect(url_for('main.index'))
 
+    if request.method == 'GET':
+        user.update_email_access_validated_at()
+
     form = NewPasswordForm()
 
     if form.validate_on_submit():
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -114,8 +114,8 @@ class User(JSONModel, UserMixin):
         response = user_api_client.update_user_attribute(self.id, **kwargs)
         self.__init__(response)
 
-    def update_password(self, password, validated_email_access=False):
-        response = user_api_client.update_password(self.id, password, validated_email_access=validated_email_access)
+    def update_password(self, password):
+        response = user_api_client.update_password(self.id, password)
         self.__init__(response)
 
     def update_email_access_validated_at(self):
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from flask import abort, request, session
 from flask_login import AnonymousUserMixin, UserMixin, login_user, logout_user
 from notifications_python_client.errors import HTTPError
@@ -115,6 +117,11 @@ class User(JSONModel, UserMixin):
     def update_password(self, password, validated_email_access=False):
         response = user_api_client.update_password(self.id, password, validated_email_access=validated_email_access)
         self.__init__(response)
+
+    def update_email_access_validated_at(self):
+        self.update(
+            email_access_validated_at=datetime.utcnow().isoformat()
+        )
 
     def password_changed_more_recently_than(self, datetime_string):
         if not self.password_changed_at:
--- a/tests/app/main/views/organisations/test_organisation_invites.py
+++ b/tests/app/main/views/organisations/test_organisation_invites.py
@@ -4,6 +4,7 @@ from unittest.mock import ANY
 import pytest
 from bs4 import BeautifulSoup
 from flask import url_for
+from freezegun import freeze_time
 
 from app.models.user import InvitedOrgUser
 from tests.conftest import ORGANISATION_ID, normalize_spaces
@@ -138,14 +139,17 @@ def test_user_invite_already_accepted(
     )
 
 
+@freeze_time('2021-12-12 12:12:12')
 def test_existing_user_invite_already_is_member_of_organisation(
     client,
     mock_check_org_invite_token,
     mock_get_user,
     mock_get_user_by_email,
+    api_user_active,
     mock_get_users_for_organisation,
     mock_accept_org_invite,
     mock_add_user_to_organisation,
+    mock_update_user_attribute,
 ):
     response = client.get(url_for('main.accept_org_invite', token='thisisnotarealtoken'))
 
@@ -160,8 +164,13 @@ def test_existing_user_invite_already_is_member_of_organisation(
     mock_accept_org_invite.assert_called_once_with(ORGANISATION_ID, ANY)
     mock_get_user_by_email.assert_called_once_with('invited_user@test.gov.uk')
     mock_get_users_for_organisation.assert_called_once_with(ORGANISATION_ID)
+    mock_update_user_attribute.assert_called_once_with(
+        api_user_active['id'],
+        email_access_validated_at='2021-12-12T12:12:12',
+    )
 
 
+@freeze_time('2021-12-12 12:12:12')
 def test_existing_user_invite_not_a_member_of_organisation(
     client,
     api_user_active,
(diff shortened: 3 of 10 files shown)


tests/app/main/views/test_new_password.py
```diff
@@ -95,6 +120,9 @@ def test_should_redirect_index_if_user_has_already_changed_password(
     mock_reset_failed_login_count
 ):
     user = mock_get_user_by_email_user_changed_password.return_value
+    # We don’t update email_validated_at if someone is trying to reuse
```
benthorner, line 123: Optional: I don't think we need this comment. I'm seldom surprised by missing mocks in tests, since we have so many. Alternatively, you could replace this with a negative assertion that we _don't_ update the attribute. That said, it's not a big deal if we do update the attribute in all scenarios, since the token only lasts for 1 hour.

tests/app/main/views/test_accept_invite.py
```diff
@@ -740,9 +769,13 @@ def test_existing_user_doesnt_get_auth_changed_by_service_without_permission(
         _expected_redirect=url_for('main.service_dashboard', service_id=service_one['id'], _external=True),
     )
 
-    assert not mock_update_user_attribute.called
+    mock_update_user_attribute.assert_called_once_with(
```
benthorner, line 772: How did you decide which tests to add these assertions for? Some of the tests just have the mock, while others have this as well.  Could we have a separate test to do the assertion for an existing user?

Optional: unlike the password reset, the token for this endpoint lasts for [several days](https://github.com/alphagov/notifications-api/blob/5cb97253acb6f527cb39cca92fbe378e4a7d200d/app/service_invite/rest.py#L106), so it might be worth adding some negative assertions in tests where we shouldn't update the attribute.

app/models/user.py
```diff
@@ -112,10 +114,15 @@ def update(self, **kwargs):
         response = user_api_client.update_user_attribute(self.id, **kwargs)
         self.__init__(response)
 
-    def update_password(self, password, validated_email_access=False):
```
benthorner, line 115: There's still one relic of this to remove [here](https://github.com/alphagov/notifications-admin/blob/ff12ba689d36eed545e0b0ce74b7e3458415375f/tests/conftest.py#L1356).

benthorner (commented):
Nice one making support a fraction easier 🪜.

Tested locally and works as expected for password resets and service invites ✅ .

Can you update / remove [the redundant guidance](https://github.com/alphagov/notifications-functional-tests#running-the-tests-against-preview-staging-or-production) when this goes in?

I left a couple of questions I'd like to double check with you 👇.

app/main/views/new_password.py
```diff
@@ -33,6 +33,7 @@ def new_password(token):
         flash('The link in the email has already been used')
         return redirect(url_for('main.index'))
 
+    user.update_email_access_validated_at()
```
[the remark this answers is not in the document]
quis: I’ve made it `GET` only because that way we only need to change one test, since most of `test_new_password.py` tests the `POST` route.

tests/app/main/views/test_new_password.py
quis: I removed this and squashed it out.

app/models/user.py
quis: Changed and squashed into 929b0ca

tests/app/main/views/test_accept_invite.py
quis: I’ve changed it so that it only makes assertions where the test had an existing assertion. I’ve also added negative assertions to `test_if_existing_user_accepts_twice_they_redirect_to_sign_in` and `test_cancelled_invited_
user_accepts_invited_redirect_to_cancelled_invitation` to make it explicit that we don’t update the attribute if the invite has already been accepted or cancelled.

benthorner: That's much clearer, thanks. Given this is a nice-to-have behaviour, I don't think we need an explicit test for the update, so happy to leave it there.

tests/app/main/views/organisations/test_organisation_invites.py
```diff
@@ -160,8 +163,13 @@ def test_existing_user_invite_already_is_member_of_organisation(
     mock_accept_org_invite.assert_called_once_with(ORGANISATION_ID, ANY)
     mock_get_user_by_email.assert_called_once_with('invited_user@test.gov.uk')
     mock_get_users_for_organisation.assert_called_once_with(ORGANISATION_ID)
+    mock_update_user_attribute.assert_called_once_with(
+        mock_get_user_by_email.side_effect(None)['id'],
```
benthorner, line 167: Optional: this seems to be a new pattern across all the tests for the Admin app. I think it would be clearer to just use the `api_user_active` fixture directly. (It's a shame the mock doesn't just have a simple `return_value`.)

benthorner (approved):
Thanks for addressing my comments 💯.

I've left a couple of optional ones I forgot to add first time round.

tests/app/main/views/test_new_password.py
```diff
@@ -3,28 +3,40 @@
...
-    mock_login,
     mock_send_verify_code,
     mock_get_user_by_email_request_password_reset,
 ):
-    data = json.dumps({'email': api_user_active['email_address'], 'created_at': str(datetime.utcnow())})
+    user = mock_get_user_by_email_request_password_reset.return_value
+    user['password_changed_at'] = '2021-01-01 00:00:00'
+    mock_update_user_attribute = mocker.patch(
+        'app.user_api_client.update_user_attribute',
+        return_value=user,
```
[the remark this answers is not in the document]
quis: We do, because otherwise we get a different user back, without the overridden value for `user['password_changed_at']`.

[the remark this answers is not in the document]
benthorner: We will get a different user back, but [nothing uses the return value of that particular method](https://github.com/alphagov/notifications-admin/pull/4000/files#diff-ddadb123ce20f6ae776e190cb082205d8c796fcd4a6497e12e98dbe472f86286R37). The change to the user attribute is still necessary, but the object is acquired via a different mock.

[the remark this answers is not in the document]
quis: Calling `User.update` replaces the current user with what it gets back from the API (or in this case the mock): https://github.com/alphagov/notifications-admin/blob/ff12ba689d36eed545e0b0ce74b7e3458415375f/app/models/user.py#L111-L113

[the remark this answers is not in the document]
benthorner: Ah, I see. I don't think it matters in this case - the test still passes without it - because the same object is being used everywhere anyway.

PR #4003, opened 2021-08-24
Add utility function to aggregate alert areas

benthorner:

Shamelessly based on #3980. @quis did all the prototyping
work, and this PR massages it into a more maintainable form, 
which should also be easier to review.

As written on the other PR, we need a means to aggregate
areas for a broadcast because:

- electoral ward names (one level of area) aren't meaningful
- the list of areas may be very long

A secondary concern is applying this to custom polygons. We're
not sure if this code will get applied to messages submitted
via the API, which can have their own area descriptions. Adding
support for it now has two benefits:

- It makes it easier to test the approach on larger regions by
using coordinates instead of listing all the wards.

- It helps ensure we're forwards compatible with applying it to
API polygons, or even custom ones created in this app.

@quis recommended [this diagram](https://www.ons.gov.uk/methodology/geography/ukgeographies/administrativegeography/england) to understand the structure 
of UK authorities.


benthorner:
I've also tested this against all historical broadcasts ([gist](https://gist.github.com/benthorner/5483e335b592d00dd3a16e1cd9560d11)):

- All UK broadcast areas were aggregated with no errors ✅.
- Overall, the aggregation cuts 741 areas down to 492 🚀.

Although most of the historical messages are non-realistic, running them through the aggregation and inspecting the output gives me more confidence it will produce meaningful results in a variety of scenarios.

app/broadcast_areas/utils.py
```diff
@@ -0,0 +1,10 @@
+def aggregate_areas(areas):
+    areas = _aggregate_wards_by_local_authority(areas)
+    return areas
+
+
+def _aggregate_wards_by_local_authority(areas):
+    return {
+        area.parent if area.id.startswith('wd20-')
```
CrystalPea: Optional: could be useful explaining in an inline comment why we look specifically for parents of areas starting with `wd20-`. I see it's explained in the tests that this signifies electoral wards, but might be also useful to have the explanation where the code is.

app/broadcast_areas/utils.py
```diff
@@ -1,8 +1,24 @@
...
     areas = _aggregate_wards_by_local_authority(areas)
     return areas
 
 
+def _convert_custom_areas_to_wards(areas):
+    results = set()
+
+    for area in areas:
+        if type(area) == CustomBroadcastArea:
+            results |= set(area.overlapping_electoral_wards)
```
CrystalPea, line 18: fun, haven't seen `ior` operator before :D

In case anybody later is reading later and wondering, here is a good explanation: https://stackoverflow.com/questions/3929278/what-does-ior-do-in-python

app/broadcast_areas/utils.py
```diff
@@ -24,3 +27,39 @@ def _aggregate_wards_by_local_authority(areas):
...
+
+    for cluster in clusters:
+        # always keep lone area as itself
+        if len(cluster) == 1:
+            results |= set(cluster)
+        # aggregate multi-area cluster
+        elif len(cluster) > 3:
+            results |= {cluster[0].parent}
+        # aggregate many small clusters
+        elif len(clusters) > 1:
```
CrystalPea, line 44: shouldn't it be `cluster` instead of `clusters` on this line?

app/broadcast_areas/utils.py
```diff
@@ -24,3 +27,39 @@ def _aggregate_wards_by_local_authority(areas):
...
+            area = cluster[0]
+            results |= {area.parent or area}
```
CrystalPea, line 46: I'm not sure what we are doing here, could use some help please.

app/broadcast_areas/utils.py
```diff
@@ -24,3 +27,39 @@ def _aggregate_wards_by_local_authority(areas):
...
+        # keep one small cluster in full
+        else:
```
CrystalPea, line 48: what condition would meet this `else`? It feels like all the cases were matched above 🤔

CrystalPea (changes requested):
Hey, good job, that was nice to read through and review 👏🏼 

I spotted something that may be a little mistake, and then there are a couple smaller comments / asks for clarification.

I also have one more general comment: when we aggregate wards into a larger area, but the wards do not represent the full area, should we not add `parts of`? Otherwise people may be wondering why they didn't get the broadcast even though they are in this area.

So for example if message was sent to `Hackney Wick` and `Hackney Marshes`, and we aggregate that to `Hackney`, wouldn't people from `Hackney Central` be potentially confused? But if we say `parts of Hackney`, that's more accurate, and together with the map would work pretty well. What do you think?

tests/app/broadcast_areas/test_models.py
```diff
@@ -0,0 +1,23 @@
...
+    expected_wards_length,
+):
+    custom_area = CustomBroadcastArea(
+        name='foo', polygons=[simple_polygon]
+    )
+
+    assert len([
+        area.name for area in
+        custom_area.overlapping_electoral_wards
+    ]) == expected_wards_length
```
quis, line 23: The list comprehension isn’t adding anything here – this assertion is equivalent:
```suggestion
    assert len(custom_area.overlapping_electoral_wards) == expected_wards_length
```

tests/app/broadcast_areas/test_utils.py
```diff
@@ -0,0 +1,184 @@
...
+        [CHELTENHAM_AND_GLOUCESTER], [
+            # Area covers more than 3 lower-tier authorities
+            # in Gloucestershire so we aggregate them
+            'Gloucestershire',
+        ]
+    ),
+    (
+        [BURFORD], [
+            # Area covers two lower-tier authorities
+            # in Gloucestershire (upper tier authority)
```
quis, line 152:
```suggestion
            # Area covers lower-tier authorities in
            # Gloucestershire and Oxfordshire (both
            # upper tier authorities)
```

tests/app/broadcast_areas/test_models.py
benthorner: Good spot! I've amended the first commit with that.

tests/app/broadcast_areas/test_utils.py
benthorner: Oops, thanks for spotting.

I've updated with [something slightly more specific](https://github.com/alphagov/notifications-admin/pull/4003/commits/bcfa21428f3161b1e3964671d1ffda720c71b358#diff-7ccef2c9a90607cf55d1b7a1bbce444014a3b8e2c1a70bde2eed43a0fe871daeR151).

PR #4008, opened 2021-09-01
Update service name advice

karlchillmaid:
This PR adds 'or initialism' to the advice about avoiding acronyms in service names.

As acronyms **and** initialisms are both problematic, it's probably worth calling them out as separate things.


tombye (approved):
Looks good to me 👍🏻

PR #4013, opened 2021-09-03
Give WebAuthn illustration a fixed size

quis:
The browser uses the `width` and `height` attributes of the `` tag to allocate space on the page for the image.

If these aren’t provided then the browser will assume the image takes up no space, until it downloads it and has a look at what the file’s dimensions are. This causes the layout of the page to jump once the image downloads. We can stop this jump from happening by providing the `width` and `height` attributes.

149 × 150 pixels is the native size of the image. But we don’t want it to display at that size, so this commit also adds some extra CSS which keeps it looking the same, namely:
- the full width (100%) of the 1/4 page column on desktop
- the full width (100%) of the column minus a 40 pixel gutter either side on mobile (by using `box-sizing: border-box` the 40 pixels of padding is subtracted from the 100% width, rather than added to it)

### DIFF
--- a/app/assets/stylesheets/views/webauthn.scss
+++ b/app/assets/stylesheets/views/webauthn.scss
@@ -26,15 +26,15 @@
 
 .webauthn-illustration {
 
-  @include govuk-media-query($from: tablet) {
-    margin: govuk-spacing(9) auto 0 auto;
-    padding: 0;
-  }
-
   box-sizing: border-box;
   width: 100%;
   height: 100%;
   margin: govuk-spacing(6) auto 0 auto;
   padding: 0 govuk-spacing(9) 0 govuk-spacing(9);
+
+  @include govuk-media-query($from: tablet) {
+    margin: govuk-spacing(9) auto 0 auto;
+    padding: 0;
+  }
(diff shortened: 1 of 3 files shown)


tombye (commented):
Looks good in my browser and nice explanation.

I'd suggest swapping the media queries to be mobile-first. 

I'm also curious about the padding on mobile. I understand the reasoning but can't see it making a difference when testing. Your having the width as 100% but the height in px seems to be doing a nice job letting the image fill the container while locking the webauth icon part to a set size and so centralising it, I think because of the SVG settings. Do you think you need it?

quis:
@tombye I think my code wasn’t actually doing what I was intending, because `width: 100%` doesn’t scale the SVG, just widens its `viewBox`. I had to also set `height: 100%` to get the desired scaling, so on desktop it looks like this (fills the one-quarter column):

image

And on mobile it looks like this (has some padding so it doesn’t fill the column):

image

I’ve made these changes and rewritten the media query to be mobile first in eee4cdd0a

tombye (approved):
Looks good to me 👍🏻

PR #4014, opened 2021-09-06
Fix areas not showing for custom broadcasts

benthorner:

This also reduces the reliance we have on the API defaulting
fields to empty arrays. Since "areas" is just an arbitrary JSON
blob, this app should be more resilient to what it contains or
doesn't contain. Best reviewed commit by commit.


benthorner:
Adding "DNM" as this breaks historical custom broadcasts, of which there are quite a few.

klssmith (approved):
This looks good to me - the commits were nice and clear, so made it easier to review 👍

PR #4017, opened 2021-09-08
Support broadcasts with unidentifiable areas

benthorner:

The original code to raise the exception was flawed: if a broadcast
only had a single area that was invalid, we would assume it was a
custom broadcast [1]. Since the recent changes [2] fixed the flaw
we're now getting exceptions for broadcasts of this kind.

It's not practical to go and manually fix the invalid broadcasts,
and the likelihood is there will be more in future as the set of
areas we support changes. This takes a pragmatic approach of simply
logging the issue and pretending such broadcasts are custom.

[1]: https://github.com/alphagov/notifications-admin/commit/926ada2f218c59b083e99dcf087ae58c1b188d16
[2]: https://github.com/alphagov/notifications-admin/pull/4014/files#diff-2dd8f77d6df281e7674b20263cdf27a3d58b839dc5930c0087ac8b9749b313e4R92

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -1,6 +1,7 @@
 import itertools
 from datetime import datetime, timedelta
 
+from flask import current_app
 from notifications_utils.polygons import Polygons
 from notifications_utils.template import BroadcastPreviewTemplate
 from orderedset import OrderedSet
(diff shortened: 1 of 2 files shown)


leohemsted (approved):

leohemsted:
Good fix :+1:

Worth pointing out that we may well expect more in the future, I'm not sure we've got a concrete plan on what to do if when electoral areas change. I think it's reasonable to expect that electoral wards may well merge or be renamed in due course.

benthorner:
For reference, one of the problematic broadcasts was "ab1b0c38-2638-45b2-aeab-3ead46488899" which had a single area of `["test-santa-claus-village-rovaniemi"]`, which looks like it pre-dates us splitting this into 4 separate areas.

PR #4022, opened 2021-09-14
fix contact list bst bug

leohemsted:
the api returns UTC timestamps, we should keep them as UTC timestamps until the very last moment, and only convert them into BST when we know we want to return to a user (ie: in contact-list.html and other places like that)

see also

- [ ] https://github.com/alphagov/notifications-api/pull/3329

### DIFF
--- a/app/models/contact_list.py
+++ b/app/models/contact_list.py
@@ -4,7 +4,6 @@ from os import path
 from flask import abort, current_app
 from notifications_utils.formatters import strip_whitespace
 from notifications_utils.recipients import RecipientCSV
-from notifications_utils.timezones import utc_string_to_aware_gmt_datetime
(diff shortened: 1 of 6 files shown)


benthorner (approved):
Works locally with and without the API change - it was funny to apply one fix and then the other, as each one knocks off an hour 😆.

I found [a couple of places where the timestamp format still needs updating](https://github.com/alphagov/notifications-admin/blob/c42fc071b5c15b7ab7a03d079006412414a3763e/tests/app/main/views/uploads/test_upload_contact_list.py#L475). The value isn't used since the tests are passing. Ideally it should be consistent. Rather than just fixing it in-place, you could DRY-up the dict and the existing ones in conftest.py with a `create_contact_list` function with some kwargs to tweak it.

PR #4023, opened 2021-09-14
Bump gulp & gulp-sass to latest versions

tombye:
Intended to deal with this security vulnerability:

https://github.blog/2021-09-08-github-security-update-vulnerabilities-tar-npmcli-arborist/

This also:
- bumps all the gulp plugin modules
- swaps the [diffDOM](https://github.com/fiduswriter/diffDOM) module for [domdiff](https://github.com/WebReflection/domdiff)
- fixes an issue with the required Node version
- sets a NPM version range, to prevent use of insecure versions

The swapping of diffDOM for domdiff should complete this story: https://www.pivotaltracker.com/story/show/165380360.


tests/javascripts/updateContent.test.js
```diff
@@ -46,52 +46,58 @@ afterAll(() => {
...
-      <div data-module="update-content" data-resource="${resourceURL}" data-key="${updateKey}" aria-live="polite">
-        <div class="bottom-gutter ajax-block-container">
-          <ul role="tablist" class="pill">
-            <li aria-selected="true" role="tab">
-              <div class="pill-selected-item" tabindex="0">
-                <div class="big-number-smaller">
-                  <div class="big-number-number">0</div>
-                </div>
-                <div class="pill-label">total</div>
+      <div class="bottom-gutter ajax-block-container">
```
benthorner, line 56: Since we're changing this, can we ditch most of it? I think all we need for this test is just something like the very short examples you have in the commit messages. While the example is more realistic, it just makes the test harder to read and for me doesn't communicate anything valuable.

(We could also simplify the `resourceURL` while we're here.)

package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "description": "Admin front end for GOV.UK Notify",
   "engines": {
-    "node": "10.15.3"
+    "node": ">=10.15.3",
```
benthorner: Optional: the commit message says this is mentioned in the README, but I can't see anything about the NodeJS version in there. Am I missing something?

package.json
```diff
@@ -3,8 +3,7 @@
   "version": "0.0.1",
   "description": "Admin front end for GOV.UK Notify",
   "engines": {
-    "node": ">=10.15.3",
-    "npm": "6.14.15 || >= 7.21.0"
```
benthorner: I don't understand why we're removing this again in the last commit. Could you explain a bit more?

benthorner (commented):
Thanks for the helpfully small commits and detailed commit messages 💯.

I've left one comment where I don't understand why you're going back on a change in a previous commit. Arguably not major as you could have just omitted it entirely and then I'd never have known 🙈. I was also wondering if the deprecation of node-sass and us moving away from it should impact [this line in the Makefile](https://github.com/alphagov/notifications-admin/blob/6ef39e2ebf75404329331175f6b9923e9e6763b0/Makefile#L28).

From a functionality perspective, I've tested this and it's (ahem) no more broken than it was before (referring to the job sending videos)...

## Videos

| Before (diffDOM) | After (domdiff) |
| https://user-images.githubusercontent.com/9029009/133613368-90fb5287-45b9-4783-90b1-72e6a8bd76ea.mov | https://user-images.githubusercontent.com/9029009/133613475-e93cf95c-bb50-4413-aed1-d25e46c46fbe.mov |
| https://user-images.githubusercontent.com/9029009/133613507-bd20f42c-8aab-4599-8de0-2639e49011fc.mov | https://user-images.githubusercontent.com/9029009/133613530-d684b329-f9d5-4f10-870c-691e023400cc.mov |

*I'm _pretty_ sure I rebuilt / reinstalled everything cleanly each time.

package.json
tombye: No, I just haven't kept up to date with changes to the README. It did used to say that (see https://github.com/alphagov/notifications-govuk-alerts/blame/c9bec7254fae39aac86932f0f122698bf5d9b802/README.md) but now doesn't.

I'll update the commit message to reflect this.

tests/javascripts/updateContent.test.js
tombye: I personally find both of those things quite important in giving me confidence that the tests will catch issues from changes, like changing the diffing library. That's why I tend to use HTML 'fixtures' in most of the JS tests copied from pages that use those scripts. I'm not saying they need to be exactly the same, which is a problem in itself as HTML in pages tends to change faster than that in tests, but for the HTML having a roughly similar depth and to be composed of similar node types. Likewise with URLs, for them to have a similar structure and for most of their parts to be match those in the format expected.

tests/javascripts/updateContent.test.js
```diff
@@ -31,9 +31,11 @@ beforeAll(() => {
 
   $.ajax.mockImplementation(() => jqueryAJAXReturnObj);
 
-  // because we're running in node, diffDOM executes as a module
-  // in the normal browser environment it will attach to window so we replicate that here
-  window.diffDOM = require('../../node_modules/diff-dom/diffDOM.js');
+  // because we're running in node, we can't execute the normal domdiff code like a browser because
+  // its variables (actually only one: domdiff) stay local to it
+  // in the normal browser environment it will attach to window so we replicate that here by using
```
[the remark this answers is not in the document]
tombye: Looking at my comments again I think they miss some crucial information:
- it's not just that we're running in Node, it's because we're bringing the `domdiff` code in using `require` instead of just having it execute as lines of code in the same file, previous to our code (as we do in browsers)
- JavaScript brought in using `require` will still execute before our test code but with its own scope so any variables global to it but not exported by it will be hidden from our test code

It feels like those points are missing from both our versions but not totally sure. Does that change how you would explain it?

package.json
tombye: I didn't have confidence that the environments this app would be deployed in would be set up with a version of`npm` that fits in the `"6.14.15 || >= 7.21.0"` range. If they didn't, this would just cause more problems in their logging and may even block deploys. Mainly guesswork here but I just thought that this was a change we could add outside of this pull request when we were sure `npm` had been updated.

tombye:
Thanks for the review @benthorner, particularly in taking the time to document a few content changes in action.

> I was also wondering if the deprecation of node-sass and us moving away from it should impact [this line in the Makefile](https://github.com/alphagov/notifications-admin/blob/6ef39e2ebf75404329331175f6b9923e9e6763b0/Makefile#L28).

This is a great catch. I didn't know that was there but assume the [LibSass](https://github.com/sass/libsass) library `node-sass` used needed to be recompiled for some people's local environments. The new [sass](https://www.npmjs.com/package/sass) module is JavaScript (compiled from [dart-sass](https://github.com/sass/dart-sass)) so shouldn't need it. I've added a commit to change that line here: https://github.com/alphagov/notifications-admin/pull/4023/commits/3fcbfddc24f5ec6ab0db481291f738deb9521649.

tests/javascripts/updateContent.test.js
benthorner: I can see where you're coming from. Just copying code from pages does have a couple of downsides, though:

- It really does make these tests harder to review. That's because it's not clear to me where the HTML came from. Do I need to review all of it, or some of it? What's special about it? You could partially address that by commenting the range of source code it's copied from, which is a nice signal that it doesn't need reviewing.

- It sidesteps the real issue of what we gain from the realism. That suggests there are certain parts of the HTML we're copying that have play a role in making the test more robust. Likewise, there are probably other parts that play no role e.g. we probably don't need three repetitions of those big stat blocks to be robust.

Despite all the writing, I don't feel strongly about this, so I'm comfortable leaving it as-is. I just wanted to share my perspective from the reviewer side of the fence in case it helps next time :-).

package.json
benthorner: Sure, I get it now 👍.

tests/javascripts/updateContent.test.js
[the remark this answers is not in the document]
benthorner: Hmm...sort of. Thanks for explaining, though: I wasn't aware of the extra subtleties! I think you points are really a more detailed version of "we're not a browser so we need to manually attach domdiff to window".

Someone might ask "why can't we just do `require...`, so adding the extra detail could still be valuable in future - there's little harm in being more detailed and although I didn't need to know, someone else might.

benthorner (approved):
Thanks @tombye!

Looks like there's just one outstanding comment (about a comment). That's nothing major so I'll approve this now and leave it up to you what to do with that.

tests/javascripts/updateContent.test.js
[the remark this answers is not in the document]
tombye: I've rewritten the comments based on that in https://github.com/alphagov/notifications-admin/pull/4023/commits/e1c750927d8a21ac8bb9b266f639ceb111b9c638 so gonna mark this as resolved but feel free to reopen if you think it needs any more work.

PR #4027, opened 2021-09-24
Use new Zendesk form everywhere

klssmith:
We started using the new form for go live tickets in alphagov/notifications-admin#4024.

This change updates the two other ticket types that admin creates (branding requests and general enquiries) to also use the new form.

[Pivotal story](https://www.pivotaltracker.com/story/show/179005041)

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -130,7 +130,7 @@ def feedback(ticket_type):
         ticket = NotifySupportTicket(
             subject='Notify feedback',
             message=feedback_msg,
-            ticket_type=ticket_type,
+            ticket_type=get_zendesk_ticket_type(ticket_type),
             p1=out_of_hours_emergency,
             user_name=user_name,
             user_email=user_email,
@@ -231,3 +231,14 @@ def needs_escalation(ticket_type, severe):
         not current_user.is_authenticated,
         not in_business_hours(),
     ))
+
+
+def get_zendesk_ticket_type(ticket_type):
+    # Zendesk has 4 ticket types - "problem", "incident", "task" and "question".
+    # We don't want to use a Zendesk "problem" ticket type when someone reports a
+    # Notify problem because they are designed to group multiple incident tickets together,
+    # allowing them to be solved as a group.
+    if ticket_type == PROBLEM_TICKET_TYPE:
+        return NotifySupportTicket.TYPE_INCIDENT
+
+    return NotifySupportTicket.TYPE_QUESTION
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -4,6 +4,9 @@ import pytz
 from flask import redirect, render_template, request, session, url_for
 from flask_login import current_user
 from govuk_bank_holidays.bank_holidays import BankHolidays
+from notifications_utils.clients.zendesk.zendesk_client import (
+    NotifySupportTicket,
+)
 
 from app import convert_to_boolean, current_service
 from app.extensions import zendesk_client
@@ -124,14 +127,19 @@ def feedback(ticket_type):
             service_string,
         )
 
-        zendesk_client.create_ticket(
+        ticket = NotifySupportTicket(
             subject='Notify feedback',
             message=feedback_msg,
             ticket_type=ticket_type,
             p1=out_of_hours_emergency,
+            user_name=user_name,
             user_email=user_email,
-            user_name=user_name
+            org_id=current_service.organisation_id if current_service else None,
+            org_type=current_service.organisation_type if current_service else None,
+            service_id=current_service.id if current_service else None,
         )
+        zendesk_client.send_ticket_to_zendesk(ticket)
+
         return redirect(url_for(
             '.thanks',
             out_of_hours_emergency=out_of_hours_emergency,
--- a/tests/app/main/views/test_feedback.py
+++ b/tests/app/main/views/test_feedback.py
@@ -5,6 +5,9 @@ import pytest
 from bs4 import BeautifulSoup, element
 from flask import url_for
 from freezegun import freeze_time
+from notifications_utils.clients.zendesk.zendesk_client import (
+    NotifySupportTicket,
+)
 
 from app.main.views.feedback import in_business_hours
 from app.models.feedback import (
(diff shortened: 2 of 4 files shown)


app/main/views/feedback.py
```diff
@@ -231,3 +231,10 @@ def needs_escalation(ticket_type, severe):
         not current_user.is_authenticated,
         not in_business_hours(),
     ))
+
+
+def get_zendesk_ticket_type(ticket_type):
+    if ticket_type == PROBLEM_TICKET_TYPE:
```
idavidmcdonald, line 237: So this says, if it is 'report-problem' then we should use `NotifySupportTicket.TYPE_INCIDENT` ie `incident`.

Should `report-problem` not become `problem` instead of `incident`?

tests/app/main/views/test_feedback.py
```diff
@@ -136,12 +136,19 @@ def test_get_feedback_page(client, ticket_type, expected_status_code):
...
+    (GENERAL_TICKET_TYPE, 'question'),
+])
+def test_passed_non_logged_in_user_details_through_flow(client, mocker, ticket_type, zendesk_ticket_type):
     mock_ticket = mocker.patch(
         'app.main.views.feedback.NotifySupportTicket',
         return_value='feedback_ticket',
     )
+    mock_ticket.TYPE_QUESTION = 'question'
+    mock_ticket.TYPE_INCIDENT = 'incident'
+    (GENERAL_TICKET_TYPE, 'question'),
```
idavidmcdonald, line 151: I'm confused at what this line is doing? It even looks like invalid python, just creating a tuple?

tests/app/main/views/test_feedback.py
```diff
@@ -180,18 +187,26 @@ def test_passed_non_logged_in_user_details_through_flow(client, mocker, ticket_t
...
+    zendesk_ticket_type,
     data
 ):
     mock_ticket = mocker.patch(
         'app.main.views.feedback.NotifySupportTicket',
         return_value='feedback_ticket',
     )
+    mock_ticket.TYPE_QUESTION = 'question'
+    mock_ticket.TYPE_INCIDENT = 'incident'
+    mock_ticket.TYPE_GENERAL = 'question'
```
idavidmcdonald, line 209: This feels a bit like a code smell that we are redefining the values here to match exactly what they do in the original class we are mocking. It feels like maybe

```
mock_ticket = mocker.patch(
        'app.main.views.feedback.NotifySupportTicket',
        return_value='feedback_ticket',
    )
```

is too general and the mock needs to be more targeted.

I'm not immediately sure what I would suggest on this (maybe because it's the end of a Monday) but do give it some thought and we can discuss tomorrow potentially

idavidmcdonald (commented):
Generally looks good. I'm a bit suspicious of the pattern of how we are doing the mocking in this but I don't yet have a suggestion for how to improve it... will see if I can offer any constructive suggestions but thought I'd at least leave some smaller comments to start you off

tests/app/main/views/test_feedback.py
klssmith: That line wasn't meant to be there, it was just a copy and paste error.

app/main/views/feedback.py
klssmith: Zendesk defines four ticket types - `problem`, `incident`, `task` or `question`, though we have only ever used `question` or `incident` in our code. The different ticket types do different things in Zendesk, and our definition of a problem ticket doesn't really match the Zendesk definition of a problem ticket - I think a "Notify problem ticket" is the equivalent of a "Zendesk incident ticket". Zendesk problem tickets are designed to link together multiple Zendesk incident tickets.

Here's the relevant part of the Zendesk docs (I couldn't link directly to the part):

> There are four values for type. Setting the type helps you to categorize your tickets, which you can then use in your workflow. For example, you can create views of tickets by their type. While the field can be blank initially (and through any number of updates), once you change the field to a specified type, you can't change it to blank again.
> * **Question** is used to indicate that the requester's issue is a question rather than a problem that needs to be solved.
> * **Incident** is used to indicate that the requester is experiencing a single occurrence of a larger problem that is affecting multiple users.
> * **Problem** is used to indicate that the requester is having an issue with your product or service that needs to be resolved and that is affecting multiple users. For example, if the wireless network in an office stops working, the problem will probably generate several support requests. Instead of handling each ticket separately, create one ticket describing the problem and set the type to Problem.
> * Link incident tickets to problem tickets. When you solve the problem ticket, all of the linked incident tickets are solved, too. See Working with problem and incident tickets.
> * **Task** is used when you want to assign the ticket as a task to a specific agent. When you select Task, you also set the Task Due Date. The due date is defined as 12pm in the account's local timezone on the date specified.

tests/app/main/views/test_feedback.py
klssmith: I also didn't really like the way things were mocked here, but hadn't really managed to get an alternative version working. Here's a different way I've found of writing the test:
```python
mock_create_ticket = mocker.spy(NotifySupportTicket, '__init__')
mock_send_ticket_to_zendesk = mocker.patch(
    'app.main.views.feedback.zendesk_client.send_ticket_to_zendesk',
    autospec=True
)

client.request.post(....)

mock_create_ticket.assert_called_once_with(...)
mock_send_ticket_to_zendesk.assert_called_once
```
One minor difference here is that we now have to use `mock_send_ticket_to_zendesk.assert_called_once` instead of `mock_send_ticket_to_zendesk.assert_called_once_with(...)` but I'm not sure if that matters.

Let me know if you think the new way is better and should be used for the tests (if so I'll change all the tests to mock in this way) or if you have another suggestion for the mocks.

idavidmcdonald: I think I like your suggestion! Feel free to try changing your tests to use this approach. If it doesn't quite work or you run into further problems let me know though and we can always rethink.

app/main/views/feedback.py
idavidmcdonald: Thanks Katie, well backed up with docs there (although I'm not really a fan of the zendesk types but I don't think we can do much about that).

I think this is OK then. I think a code comment or something in the commit message please to indicate why these two lines of code are indeed intentional and correct would be useful as otherwise looking at it seems strange.

This might be worth flagging to Ale too as an FYI if he isn't already aware of the slightly misleading different definitions of incidents and problems we have.

klssmith: 👍 I've added a comment to the function. I have mentioned the ticket types to Ale too, although we don't check or do anything with the Zendesk ticket types, so this shouldn't have any impact for him.

tests/app/main/views/test_feedback.py
klssmith: I've changed all the mocks added here (and rebased the changes), then added a new commit to update the mocks that were added in a previous PR.

tests/app/main/views/test_feedback.py
```diff
@@ -154,30 +162,44 @@ def test_passed_non_logged_in_user_details_through_flow(client, mocker, ticket_t
...
-        p1=ANY
+        ticket_type=zendesk_ticket_type,
+        p1=False,
+        user_name='Anne Example',
+        user_email='anne@example.com',
+        org_id=None,
+        org_type=None,
+        service_id=None
     )
+    mock_send_ticket_to_zendesk.assert_called_once
```
idavidmcdonald, line 177: I think this is missing brackets on the end as `assert_called_once` is a method that needs calling rather than a property. https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

I think at the moment, it may just be passing silently without actually checking it has been called

idavidmcdonald (approved):
I think I spotted one assertion that needs to change everywhere as it may not be doing what you want it to do.

After that is changed, then this is good to go.

Note, I do prefer the mocking and tests now you've made these changes. I assume you do too?

tests/app/main/views/test_feedback.py
klssmith: Good spot, I've changed this everywhere

PR #4030, opened 2021-09-30
Update Privacy page content

karlchillmaid:
This PR updates the content on the Privacy page. Specifically:

* clarifying the use of third party suppliers in data processing
* explaining that data is stored in the UK and EEA
* updating the ICO email address
* making a few minor edits for style and clarity

These changes have been developed with and agreed by Caley.


app/templates/views/privacy.html
```diff
@@ -63,12 +63,11 @@ <h2 class="heading-medium">How long we keep your data</h2>
...
     <h2 class="heading-medium">Where your data is processed and stored</h2>
 
-    <p class="govuk-body">We design, build and run our systems to make sure that your data is as safe as possible at any stage, both while it’s
+    <p class="govuk-body">We design, build and run our systems to make sure that your data is as safe as possible at every stage, both while it’s
       processed and when it’s stored.</p>
 
-    <p class="govuk-body">Your personal data will not be transferred outside of the European Economic Area (EEA) throughout the course of its
-      processing at GDS or any third party suppliers.</p>
-
+    <p class="govuk-body">Your personal data will be processed both in the United Kingdom and the European Economic Area (EEA). Your data receives the same level of protection in the EEA as it does in the UK through the safeguard of <a class="govuk-link govuk-link--no-visited-state" href="https://ico.org.uk/for-organisations/dp-at-the-end-of-the-transition-period/data-protection-and-the-eu-in-detail/adequacy/">Adequacy Decisions</a>.</p>
```
karlchillmaid, line 69:
```suggestion
    <p class="govuk-body">Your personal data will be processed both in the UK and the European Economic Area (EEA). Your data receives the same level of protection in the EEA as it does in the UK through the safeguard of <a class="govuk-link govuk-link--no-visited-state" href="https://ico.org.uk/for-organisations/dp-at-the-end-of-the-transition-period/data-protection-and-the-eu-in-detail/adequacy/">Adequacy Decisions</a>.</p>
```

I don’t think we need to write ‘United Kingdom’ in full.

klssmith:
Do you need to change "Last update: 5 March 2019" at the top of the page, or are all these change too minor for that to be necessary?

karlchillmaid:
Good point. I think the change in how we describe where the data is stored probably counts as significant enough for a date change too.

app/templates/views/privacy.html
```diff
@@ -63,12 +63,11 @@ <h2 class="heading-medium">How long we keep your data</h2>
 
     <h2 class="heading-medium">Where your data is processed and stored</h2>
 
-    <p class="govuk-body">We design, build and run our systems to make sure that your data is as safe as possible at any stage, both while it’s
+    <p class="govuk-body">We design, build and run our systems to make sure that your data is as safe as possible at every stage, both while it’s
```
karlchillmaid, line 66: Might need to revert this one – I’m waiting for confirmation.

```suggestion
    <p class="govuk-body">We design, build and run our systems to make sure that your data is as safe as possible at any stage, both while it’s
```

klssmith (approved):

app/templates/views/privacy.html
```diff
@@ -28,25 +28,25 @@ <h2 class="heading-medium">Who we are</h2>
...
 
     <h2 class="heading-medium">Why we need your data</h2>
 
-    <p class="govuk-body">We require your personal data to ensure only legitimate users of GOV.UK Notify can use it to send notifications.</p>
+    <p class="govuk-body">We need your personal data to ensure that only legitimate users of GOV.UK Notify can use it to send notifications.</p>
 
     <h2 class="heading-medium">What we do with your data</h2>
 
-    <p class="govuk-body">We use your data to check that you are allowed to access GOV.UK Notify and to record your activity when using it.</p>
+    <p class="govuk-body">We use your data to check that you are allowed to access GOV.UK Notify and to record your activity when using it. We use third party suppliers to do this.</p>
```
idavidmcdonald, line 49: What do we mean by we use third party suppliers to do this?

When I read this, I immediately think it is saying that we are using a third party to do our log in system, or sending off your data to some agency to check who you say you are, or we have added some third party individual user tracking to our site? All of which aren't the case.

I guess I get a very different impression from reading this paragraph to what I think it is trying to communicate.

karlchillmaid: Good shout, David. I’ll investigate.

karlchillmaid: Update: This PR is blocked until we can clarify the language around third parties.

PR #4034, opened 2021-10-04
Refactor services and organisations models for users 

quis:
Best reviewed commit-by-commit, but generally this makes things similar to elsewhere in terms of how we:
- turn a list of JSON objects into an iterable of Python models
- sort iterables of Python models

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -292,10 +292,8 @@ class User(JSONModel, UserMixin):
 
     @property
     def services(self):
-        from app.models.service import Service
-        return [
-            Service(service) for service in self.orgs_and_services['services']
-        ]
+        from app.models.service import Services
+        return Services(self.orgs_and_services['services'])
 
     @property
     def services_with_organisation(self):
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -290,16 +290,12 @@ class User(JSONModel, UserMixin):
     def orgs_and_services(self):
         return user_api_client.get_organisations_and_services_for_user(self.id)
 
-    @staticmethod
-    def sort_services(services):
-        return sorted(services, key=lambda service: service.name.lower())
-
     @property
     def services(self):
         from app.models.service import Service
-        return self.sort_services([
+        return sorted(
             Service(service) for service in self.orgs_and_services['services']
-        ])
+        )
 
     @property
     def services_with_organisation(self):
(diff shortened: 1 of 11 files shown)


app/models/user.py
```diff
@@ -293,9 +293,9 @@ def orgs_and_services(self):
     @property
     def services(self):
```
benthorner, line 294: There are a few other places where this ordering was used:

- [user-information.html](https://github.com/alphagov/notifications-admin/blob/ccebae6c751ac6a0f8638bc0794e25961ecff7fb/app/templates/views/find-users/user-information.html#L20) (similarly for trial mode services)
- [go-live-request.txt](https://github.com/alphagov/notifications-admin/blob/ccebae6c751ac6a0f8638bc0794e25961ecff7fb/app/templates/support-tickets/go-live-request.txt#L28)

app/templates/views/choose-account.html
```diff
@@ -19,7 +19,7 @@ <h2>
   {% else %}
     <ul>
   {% endif %}
-  {% for org in organisations %}
+  {% for org in organisations|sort %}
```
benthorner, line 22: This sorting is new (no problem with that), but it looks like [the model itself](https://github.com/alphagov/notifications-admin/blob/ccebae6c751ac6a0f8638bc0794e25961ecff7fb/app/models/organisation.py) isn't sortable. Did you mean to have that in a previous commit?

(There are also several other places we list orgs. It would be nice to propagate this feature there if we're adding it here.)

benthorner (commented):
This was easy to review, thanks for the small commits 💯.

I noticed one change which may not have the effect you were intending it to...

app/models/user.py
quis: I’ve updated `user-information.html` in 18dfc44

`go-live-request.html` passes the list to `format_yes_no`, which casts the it to a boolean, so the ordering isn’t relevant.

app/templates/views/choose-account.html
quis: I’ve made sure the model is sortable and added some test coverage for the place where we list organisations in 0c2b586e4

app/models/user.py
benthorner: Thanks!

> go-live-request.html passes the list to format_yes_no

Oops :-).

benthorner (approved):

PR #4035, opened 2021-10-05
Ensure only logged in users can see /webauthn/register

klssmith:
There are no links to the `webauthn_begin_register` route - you are only
taken there if you are logged in and have clicked to register a key.
However, we have seen this route being crawled by bots making a GET
request which gives a `500` status code error because there isn't a
logged in `current_user`. For consistency, this also adds the decorator to
the POST route.

### DIFF
--- a/app/main/views/webauthn_credentials.py
+++ b/app/main/views/webauthn_credentials.py
@@ -13,9 +13,11 @@ from app.utils.login import (
     log_in_user,
     redirect_to_sign_in,
 )
+from app.utils.user import user_is_logged_in
 
 
 @main.route('/webauthn/register')
+@user_is_logged_in
 def webauthn_begin_register():
     if not current_user.can_use_webauthn:
         abort(403)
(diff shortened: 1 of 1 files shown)


app/main/views/webauthn_credentials.py
```diff
@@ -13,9 +13,11 @@
     log_in_user,
     redirect_to_sign_in,
 )
+from app.utils.user import user_is_logged_in
 
 
 @main.route('/webauthn/register')
+@user_is_logged_in
```
idavidmcdonald, line 20: Optional: For consistency should we also add this decorator to `webauthn_complete_register` which is supposed to only be accessed by a logged in user too (even though it would be much less likely given it is a POST route and we would 400 them anyway)

idavidmcdonald (approved):
Nice proactive work

app/main/views/webauthn_credentials.py
klssmith: 👍 I've added it to the POST route for consistency.

PR #4036, opened 2021-10-05
Update the How to pay page

karlchillmaid:
As part of the work we’re doing for _Better messages_ and the Notify onboarding journey, this PR updates the ‘How to pay’ page to:

* add an introduction
* front-load the fact that users might need to raise a purchase order
* remove the link to contact support because we’ve added a ‘Billing details’ page to deal with most enquiries


tombye (approved):
All seems sensible to me 👍🏻

PR #4037, opened 2021-10-06
Move end of tour route into `tour.py`

quis:
In https://github.com/alphagov/notifications-admin/pull/3663 we made specific routes and a file for sending the ‘tour’ text message, rather than sharing the ‘one-off’ routes in `send.py`.

This commit moves the final route in the tour journey into `tour.py` as well, which is where I expected to find it when I was looking for it just now.


idavidmcdonald (approved):
Looks good although I'd also suggest moving `test_go_to_dashboard_after_tour_link` from `test_send` into the equivalent for the tour test file.

quis:
@idavidmcdonald good point, I found a couple of tests in `test_send.py` which I’ve now moved

idavidmcdonald (approved):

PR #4039, opened 2021-10-11
Bump python-client to 6.3.0

benthorner:


benthorner:
@quis spotted an interesting phenomenon that I think is worth documenting somewhere (here): the [/platform-admin/live-services page](https://www.notifications.service.gov.uk/platform-admin/live-services) _can_ (from a cold start) take around 90s to return a (CloudFront) error.

It turns out the behaviour is due to the CloudFront configuration for the Admin app, but it's worth exploring all the way back as well to see what's actually going on. First, the origin requests:

screencapture-kibana-logit-io-s-9423a789-282c-4113-908d-0be3b1bc9d1d-app-kibana-2021-10-15-12_40_19

This shows 3 interesting things (ignoring the unreliable timestamps):

- A single request by me is leading to 3 separate origin requests. More on this below.
- The initial Admin requests are 499ing after 30s, which means CloudFront has given up on them.
- The upstream requests to API are 499ing after 60s, which is when [Nginx gives up](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout).*

Admin CloudFront is configured to auto-retry up to 3 times:

Screenshot 2021-10-15 at 12 52 37

This explains why some pages can _appear_ to take over 30s to load. All this just reinforces leaving the default timeout at 30s, since the downstream request will have stopped by this point anyway.

*I'm not sure how Nginx is able to continue with these request for a whole 60s, as I would expect the API to have timed out before then. However, [it does look like PaaS can go up to 90s](https://docs.cloudfoundry.org/adminguide/routing-keepalive.html) and perhaps my above statement of 30s for gunicorn is flawed somehow. @sakisv @idavidmcdonald this may be something to investigate before we try to migrate API to CloudFront again: given the above observation and the current CloudFront config for API - the same as that for Admin - the effect of pushing API traffic through CloudFront would be to limit request durations to a max of 30s (down from at least 60s).

PR #4044, opened 2021-10-13
Require the newest version of itsdangerous

quis:
Flask version 2 requires itsdangerous version 2. However rather than bumping both at once it feels safer to do this incrementally, so we can isolate anything which breaks.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -29,7 +29,7 @@ pyproj==3.2.1
 
 # PaaS
 awscli-cwlogs>=1.4,<1.5
-itsdangerous==1.1.0  # pyup: <2
+itsdangerous==2.0.1
 notifications-utils @ git+https://github.com/alphagov/notifications-utils.git@50.0.0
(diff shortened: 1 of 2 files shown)


quis:
This is broken by https://github.com/pallets/itsdangerous/pull/133

When testing certain routes we:
- use a fixture like `logged_in_user` or `client_request`
  - under the hood this logs in the user with `TestClient.login` 
  - logging in the user signs their session with a secret and the current time
- wrap the test method with a `freeze_time()` decorator to simulate a past date and time
- call the wrapped test method
  - now any application code which tries to get the current time will get the frozen time
  - getting the current user means decoding the session
  - the code which decodes the session will see that the session was created in the future, in other words it has a negative age
  - as of ItsDangerous 2.0.0 signatures with a negative age raise an exception

quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/4255

PR #4047, opened 2021-10-18
Change ‘title’ to ‘reference’ when writing alert

quis:
On the screen where you write an alert without a template Andy pointed out that ‘title’ feels  bit out of place.

We chose this because ‘Template name’ definitely didn’t fit, but I agree that reference is better.

Via https://trello.com/c/EqUln5yD/60-changing-title-to-reference-for-alert-template-new-message

# Before

image

# After

image

### DIFF
--- a/app/templates/views/broadcast/write-new-broadcast.html
+++ b/app/templates/views/broadcast/write-new-broadcast.html
@@ -24,7 +24,7 @@
         {{ form.name(param_extensions={
           "classes": "govuk-!-width-full",
           "hint": {"text": "Your recipients will not see this"},
-          "label": {"text": "Title"}
+          "label": {"text": "Reference"}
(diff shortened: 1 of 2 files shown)


karlchillmaid:
I don’t have a strong objection to `reference` – although I’m not sure it’s entirely clear what users should enter here. Have we heard users using this term before?

I agree that `title` doesn’t work, and that `template name` feels weird when you’re not creating a template.

Did you discuss any other options, like `name of alert`, `alert name`, `description`? I’m not saying those are better (some of them definitely aren’t!), I just want to understand how we landed on `reference`.

Another question: where does the reference get played back to users?

quis:
Going to merge this since we haven’t had any better ideas. Can always change it again.

PR #4053, opened 2021-10-22
Replace 'x' psuedo element with SVG

tombye:
Part of these cards:
- [Fix admin broadcast pages in high contrast mode](https://www.pivotaltracker.com/story/show/179852075)
- [Improve accessibility of admin broadcasts content](https://www.pivotaltracker.com/story/show/179850485)

image

Changes the following things with the 'x' link-button that removes an area:
- makes the 'x' an SVG instead of [a ::before pseudo element](https://developer.mozilla.org/en-US/docs/Web/CSS/::before) so we can make screen readers announce "Remove [area]" instead of "x"
- fixes the visual design in high contrast modes
- makes it into a button, instead of a link

The last change is because, unlike the other link-buttons (ie. 'Send' & 'Edit' on template pages), the text isn't underlined so it has no link styling.

## Notes for reviewers

You don't need to follow all this but here are some tips in case useful.

### Testing the 'x' being changed to an SVG

If you're comfortable using a screen reader, try using the button with one. If you aren't, worth knowing I tested with the NVDA, Voiceover and JAWS screen readers and that you can check the accessibility info the new HTML produces in devtools. That'll be the 'Accessibility' tab in Firefox devtools. In Chrome you need to inspect the element and look in the 'Accessibility' tab in the pane to the right of the code view.

### Testing in high contrast mode

I have access to Window High Contrast mode via our testing solution, assistivelabs, and recorded some screengrabs you can see below. High contrast views can also be checked manually using Firefox:

Settings > General > Colours > Override the colours specified by the page with your selections above > Always.

### Testing it being a button

Try the following:
- check the role it has in the 'Accessibility' section of devtools (see guidance above for how to see that)
- tab to it and try pressing space to activate

## Before/after screenshots

In all of these, I tab through the buttons and hover the first one.

#### Before

https://user-images.githubusercontent.com/87140/138466178-f7495a11-4905-4bdd-bcd0-2ada875ffc18.mov

#### After

https://user-images.githubusercontent.com/87140/138466250-c22aa2e7-5cb9-478a-b4a4-9cfa7789f698.mov

#### Before

https://user-images.githubusercontent.com/87140/138473571-590e675a-27c9-425e-866a-6f4cd4f424c6.mov

#### After

https://user-images.githubusercontent.com/87140/138466548-9db60810-0171-4de1-b9da-6f163d107722.mov

#### Before

https://user-images.githubusercontent.com/87140/138466618-46f68404-0073-42d9-9916-b98d32891cd2.mov

#### After

https://user-images.githubusercontent.com/87140/138466653-5889017f-dc06-4c22-ace3-fe588dafa7c4.mov

#### Before

https://user-images.githubusercontent.com/87140/138466723-2ad877b0-1308-4b84-8250-c28ddc278ef2.mov

#### After

https://user-images.githubusercontent.com/87140/138466754-ee79a37a-1bcb-4616-847b-18e977ca0372.mov

### DIFF
--- a/app/assets/stylesheets/components/area-list.scss
+++ b/app/assets/stylesheets/components/area-list.scss
@@ -40,6 +40,11 @@
         color: $light-blue-25;
       }
 
+      // set styles so they don't get overriden by govuk-button styles
+      &:active {
+        top: -2px;
+      }
+
       // The box-shadow that separates the remove link from the list item is hidden in
       // high contrast mode. Remake using a border instead
       // Also sets text colour to match govuk-button
--- a/app/assets/stylesheets/components/area-list.scss
+++ b/app/assets/stylesheets/components/area-list.scss
@@ -19,51 +19,58 @@
       right: -2px; // cover list-item border-right
       bottom: -2px; // cover list-item border-bottom
       width: 35px;
-      background: $govuk-blue;
-      color: $white;
-      box-shadow: -2px 0 0 0 $black, inset 1px 0 0 0 rgba($white, 0.1);
       padding: (govuk-spacing(1) + 1px) 0 govuk-spacing(1); 
       border: 2px solid $govuk-blue;
       border-left: none;
       text-align: center;
       text-decoration: none;
+      font-size: inherit; // counter govuk-button styles
+      line-height: inherit; // counter govuk-button styles
 
-      // The box-shadow that separates the remove link from the list item is hidden in
-      // high contrast mode. Remake using a border instead
-      @media (-ms-high-contrast: active), (forced-colors: active) {
-        & {
-          border-left: 2px solid $govuk-blue;
-        }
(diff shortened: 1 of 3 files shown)


quis:
Checked this out locally. The **&times;** is looking good now.

I did notice that in the `:active` state (in order words when you hold the mouse button down on the element) it gets squished down from the top:

image

Just the `:focus` state looks correct:

image

tombye:
> I did notice that in the :active state (in order words when you hold the mouse button down on the element) it gets squished down from the top:

Sorry, not sure why I didn't spot that before, looks like it comes from the GOVUK button styles. Added some CSS to override them in https://github.com/alphagov/notifications-admin/pull/4053/commits/7e2bd45c73a3e2d182a6f33464caeacea4b26456

quis (approved):

PR #4055, opened 2021-10-26
Fix headings on choose account page

quis:
You’re supposed to see the two column layout on this page if you have multiple categories of things to show.

We weren’t counting the ‘platform admin’ section as one of these categories so platform admin users with only live services or only trial mode services were inadvertently seeing a mixture of the one column and two column layout.

Also this logic around the headings wasn’t tested before – now it is.

Screenshot of the problem:

Screenshot 2021-10-26 at 14 38 36

How the page looks with the non-broken two-column layout:

Screenshot 2021-10-26 at 14 48 25

### DIFF
--- a/app/templates/views/choose-account.html
+++ b/app/templates/views/choose-account.html
@@ -7,7 +7,7 @@
   organisations=[],
   services=[]
 ) %}
-  {% if show_heading %}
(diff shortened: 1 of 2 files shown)


tombye (approved):
Looks good to me 👍🏻

PR #4059, opened 2021-10-29
Upgrade utils to 48.0.0

idavidmcdonald:
Fixes a bug with non breaking spaces being removed from templates

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -2295,8 +2295,8 @@ def test_send_one_off_letter_address_shows_form(
     }),
     # it ignores empty lines and strips whitespace from each line.
     # It also strips extra whitespace from the middle of lines.
-    ('\n  a\ta  \n\n\n      \n\n\n\nb  b   \r\n sw1a1aa \n\n', {
-        'address_line_1': 'a\ta',
+    ('\n  a\ta  \n\n\n      \n\n\n\nb  b   \r\n sw1a\u00A01aa \n\n', {
+        'address_line_1': 'a a',
         'address_line_2': 'b b',
         'address_line_3': '',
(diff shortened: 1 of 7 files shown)


tests/app/main/views/test_send.py
```diff
@@ -2295,8 +2295,8 @@ def test_send_one_off_letter_address_shows_form(
     }),
     # it ignores empty lines and strips whitespace from each line.
     # It also strips extra whitespace from the middle of lines.
-    ('\n  a\ta  \n\n\n      \n\n\n\nb  b   \r\n sw1a1aa \n\n', {
-        'address_line_1': 'a\ta',
+    ('\n  a\ta  \n\n\n      \n\n\n\nb  b   \r\n sw1a\u00A01aa \n\n', {
+        'address_line_1': 'a a',
```
idavidmcdonald, line 2299: @quis I think it is worth pointing out this change that I had to make to fix the tests as it was something I had forgotten/ not noticed.

`normalise_lines` (now renamed to get_lines_with_normalised_whitespace) used to

- do nothing to single tab characters
- change two tab characters next to each other in to a single space character


Now
- changes a single tab character into a single space character
- change two tab characters next to each other in to a single space character

It is called https://github.com/alphagov/notifications-utils/blob/0060b763aea57581c68ca6813a65d27ade2979c2/notifications_utils/postal_address.py#L40

Given the changes, this means if someone has in their letter address a tab character, then we will now be replacing it with a single space, unlike before.

This is a breaking change. Are we OK with that?

quis: This is an improvement :shipit: 

You’ll need to do the same change bump in template preview for this to also work for precompiled letters.

PR #4062, opened 2021-11-04
Fix edit reply to email bug

idavidmcdonald:
https://www.pivotaltracker.com/story/show/180026726

Can be reviewed commit by commit (commit 1 fixes the bug, commit 2 is a slight refactor to a related test).

There was a bug where if you enter an invalid email address in
to the edit reply to email address form and click save, the
form you get shown with your error message will always contain
the field to set as default the reply to and also delete. This should
not have been the case. If you make an error on the form when
changing a reply to that is already a default, then you should not
be given the chance to change it to not default, nor should you
be able to delete it.

This PR fixes that bug by making sure the additional form fields
are only shown if the reply to being changed is not the default.

### Before
image

### After
image

### Before
image

### After
image

### Before
image


CrystalPea (approved):
Looks good to me 👏🏼

PR #4063, opened 2021-11-04
Fix spacing on send one-off SMS

quis:
Previous the text input on the send one off SMS page was wrapped in an element with a class of `form-group`. We overrode the spacing on this element to make the links under the textbox move closer to the textbox, so that visually they group together.

This was broken when we moved to GOV.UK frontend, which uses a class of `govuk-form-group`. This commit changes the CSS to account for both possibilities, because the send-a-one-off-letter page is still using a textbox, not a text input, and textboxes haven’t moved to GOV.UK Frontend yet.

— | Before | After
---|---|---
**Text message** | image | image
**Letter** | image | image

### DIFF
--- a/app/assets/stylesheets/views/send.scss
+++ b/app/assets/stylesheets/views/send.scss
@@ -1,6 +1,7 @@
 .send-one-off-form {
 
-  .form-group {
+  .form-group,
+  .govuk-form-group {
     margin-bottom: 20px;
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
Wow, we still haven't fully moved to gov-uk-frontend? :o

PR #4064, opened 2021-11-04
Fix autofocus in places where it wasn’t working

quis:
There are quite a few places where autofocusing text entry elements wasn’t working because the HTML is using a mixture of GOV.UK-Elements-style HTML and class names and new GOV.UK Frontend components.

This PR makes the Javascript work with both types, until such a time as we’ve moved entirely to GOV.UK Frontend.


tombye (approved):
Looks good to me 👍🏻

PR #4065, opened 2021-11-04
Make space for link in a sticky page footer

quis:
If there’s a sticky page footer which only has a right aligned link it needs another element to give the container its height. This is because the link itself is absolutely positioned so takes up no height.

On the _notification_ page we do this with a non-breaking space:
https://github.com/alphagov/notifications-admin/blob/1c2b65356f4b048ddea65cb5084417b2d6b68436/app/templates/views/notifications/notification.html#L101

This commit applies that same approach in the other places where we have right-aligned links in a sticky page footer elsewhere.

# Before 

image

# After 

image

### DIFF
--- a/app/templates/views/notifications/check.html
+++ b/app/templates/views/notifications/check.html
@@ -76,6 +76,10 @@
         {{ govukButton({ "text": button_text }) }}
       {% endif %}
       {% if template.template_type == 'letter' %}
+        {% if error %}
+          {# no continue button so need to pad the space #}
+          <div>&nbsp;</div>
+        {% endif %}
         <a href="{{ url_for('no_cookie.check_notification_preview', service_id=current_service.id, template_id=template.id, filetype='pdf') }}" download class="govuk-link govuk-link--no-visited-state page-footer-right-aligned-link{% if error %}-without-button{% endif %}">Download as a PDF</a>
       {% endif %}
     </form>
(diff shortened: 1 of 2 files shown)


app/templates/views/notifications/check.html
```diff
@@ -76,6 +76,10 @@
         {{ govukButton({ "text": button_text }) }}
       {% endif %}
       {% if template.template_type == 'letter' %}
+        {% if error %}
```
CrystalPea, line 79: Why we only need it if error?

app/templates/views/uploads/contact-list/contact-list.html
```diff
@@ -129,7 +129,9 @@ <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
 
   <div class="js-stick-at-bottom-when-scrolling">
     <div class="page-footer">
-      {% if not confirm_delete_banner %}
+      {% if confirm_delete_banner %}
+        <div>&nbsp;</div>
+      {% else %}
```
CrystalPea, line 134: code below that line was previously part of this conditional, now it's not, how come?

CrystalPea (commented):
Hey, I left a few comments, I don't understand some things here 😳

samuelhwilliams:
@quis does this still need to go in or can we close this off?

PR #4067, opened 2021-11-10
High contrast issues alerts pages

tombye:
Fixes for a few issues on this card:

https://www.pivotaltracker.com/story/show/179852075

Attempts to fix the following issues:
- blue buttons on template page don't have an outline so look like floating text
- the mock up of the alert on 'Preview' and other pages doesn't have an outline so looks wrong
- the message textarea has a line below the text you enter that moves when the lines increase

### Before

pill_items_high_contrast_dark_before

### After

pill_items_high_contrast_dark_after

### Before

mock_up_high_contrast_mode_dark_before

### After

mock_up_high_contrast_mode_dark_after

#### Note about the alert icon

The alert icon, to the left of the mock up heading, will need to be made into an inline SVG so it can match the text colour in all modes (including high contrast). This is what it looks like in high contrast light:

image

The HTML for this is in [notifications-utils](https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/jinja_templates/broadcast_preview_template.jinja2) so I'll update it outside of this pull request.

### Before

textbox_high_contrast_before

### After

textbox_high_contrast_after

## Notes for reviewers

The screenshots above are from Windows high contrast mode. You can use [Firefox's color modes](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use) to check this normally but, if you're using v94 and above, it'll look broken. I'm looking into this but it affects design system too worth knowing that the approach used in these changes is at least consistent with them and so would be fixed in tandem with any work they do on this.

### DIFF
--- a/app/assets/stylesheets/components/broadcast-message.scss
+++ b/app/assets/stylesheets/components/broadcast-message.scss
@@ -6,6 +6,7 @@
     max-width: 464px;
     box-sizing: border-box;
     padding: govuk-spacing(9) govuk-spacing(3) govuk-spacing(3) govuk-spacing(3);
+    border: solid 1px transparent; // to show it's area in high contrast mode
     background: $panel-colour;
     box-shadow: inset 0 -1px 0 0 rgba(0, 0, 0, 0.05);
     border-radius: 5px;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/components/broadcast-message.scss
```diff
@@ -18,10 +19,11 @@
   &-heading {
     display: block;
     position: absolute;
-    top: 0;
-    left: 0;
+    top: -1px;
+    left: -1px;
     padding: govuk-spacing(2) + 1px govuk-spacing(3) (govuk-spacing(2) - 1px) 46px;
-    width: 100%;
+    border: solid 1px transparent; // to show it's area in high contrast mode
```
CrystalPea, line 25: nice trick :D

app/assets/stylesheets/components/textbox.scss
```diff
@@ -39,6 +39,11 @@
     padding-bottom: govuk-spacing(3);
     z-index: 10;
 
+    // transparent borders become visible in high contrast modes so set to match background
+    @media (-ms-high-contrast: active), (forced-colors: active) {
```
CrystalPea, line 43: wow, advanced CSS magic 🧙🏼

CrystalPea (approved):
🚢

PR #4071, opened 2021-11-17
Add downloadable report for org usage

CrystalPea:
In this PR we took the info that org level users see for each service on the org dashboard, and put it into a downloadable report.

This is so org level users can use this data easier for things like determining spending per service.

Pivotal ticket: https://www.pivotaltracker.com/story/show/180276865

Next steps: think / research if we could add anything to this report to make it even more useful.

Org dashboard with a link to download the report:
Screenshot 2021-11-17 at 17 56 51

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from datetime import datetime
 from functools import partial
 
 from flask import flash, redirect, render_template, request, session, url_for
@@ -43,6 +44,7 @@ from app.main.views.dashboard import (
 from app.main.views.service_settings import get_branding_as_value_and_label
 from app.models.organisation import Organisation, Organisations
 from app.models.user import InvitedOrgUser, User
+from app.utils.csv import Spreadsheet
 from app.utils.user import user_has_permissions, user_is_platform_admin
 
 
@@ -158,16 +160,46 @@ def organisation_dashboard(org_id):
             for key in ('emails_sent', 'sms_cost', 'letter_cost')
         },
         download_link=url_for(
-            '.download_services_report_for_org',
+            '.download_organisation_usage_report',
             org_id=org_id,
+            selected_year=year
         )
     )
 
 
-@main.route("/organisations/<uuid:org_id>", methods=['GET'])
+@main.route("/organisations/<uuid:org_id>/download-usage-report.csv", methods=['GET'])
 @user_has_permissions()
-def download_services_report_for_org(org_id):
-    pass
+def download_organisation_usage_report(org_id):
+    selected_year = request.args.get('selected_year')
+    services_usage = current_organisation.services_and_usage(
+        financial_year=selected_year
+    )['services']
+
+    column_names = OrderedDict([
+        ('service_id', 'Service ID'),
+        ('service_name', 'Service Name'),
+        ('emails_sent', 'Emails sent'),
+        ('sms_remainder', 'Free text message allowance remaining'),
+        ('sms_cost', 'Spent on text messages (£)'),
+        ('letter_cost', 'Spent on letters (£)')
+    ])
+
+    org_usage_data = [[x for x in column_names.values()]]
+
+    for service in services_usage:
+        org_usage_data.append([service[attribute] for attribute in column_names.keys()])
+
+    return Spreadsheet.from_rows(org_usage_data).as_csv_data, 200, {
+        'Content-Type': 'text/csv; charset=utf-8',
+        'Content-Disposition': (
+            'inline;'
+            'filename="{} organisation usage report for year {}'
+            ' - generated on {}.csv"'.format(
+                current_organisation.name,
+                selected_year,
+                datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+            ))
+    }
 
 
 @main.route("/organisations/<uuid:org_id>/trial-services", methods=['GET'])
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -156,8 +156,18 @@ def organisation_dashboard(org_id):
         **{
             f'total_{key}': sum(service[key] for service in services)
             for key in ('emails_sent', 'sms_cost', 'letter_cost')
-        }
+        },
+        download_link=url_for(
+            '.download_services_report_for_org',
+            org_id=org_id,
+        )
     )
+
+
+@main.route("/organisations/<uuid:org_id>", methods=['GET'])
+@user_has_permissions()
+def download_services_report_for_org(org_id):
+    pass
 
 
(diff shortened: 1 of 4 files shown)


app/templates/views/organisations/organisation/index.html
```diff
@@ -112,4 +112,10 @@ <h3 class="govuk-!-font-weight-bold govuk-!-font-size-24 govuk-!-margin-bottom-3
     <div class="keyline-block govuk-!-margin-top-2"></div>
   {% endif %}
 
+  <div class="js-stick-at-bottom-when-scrolling">
+    <p class="bottom-gutter">
+      <a href="{{ download_link }}" download="download" class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download this report</a>
```
quis, line 117: In other places where we have download links in a sticky footer (for example after sending a letter) the link isn’t bold:
```suggestion
      <a href="{{ download_link }}" download="download" class="govuk-link govuk-link--no-visited-state">Download this report</a>
```

app/templates/views/organisations/organisation/index.html
```diff
@@ -112,4 +112,10 @@ <h3 class="govuk-!-font-weight-bold govuk-!-font-size-24 govuk-!-margin-bottom-3
...
+    </p>
```
quis, line 118: It’s not consistent with other sticky footer links, but I think here the link looks best bold and left-aligned. It just needs the spacing tweaking slightly so there’s equal space above and below the link:
```suggestion
    <p class="govuk-!-margin-bottom-1">
      <a href="{{ download_link }}" download="download" class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download this report</a>
    </p>
```

It’ll look like this:
image

app/main/views/organisations.py
```diff
@@ -156,10 +158,51 @@ def organisation_dashboard(org_id):
...
+        financial_year=selected_year
+    )['services']
+
+    column_names = OrderedDict([
+        ('service_id', 'Service ID'),
+        ('service_name', 'Service Name'),
+        ('emails_sent', 'Emails sent'),
+        ('sms_billable_units', 'SMS fragments sent'),
+        ('sms_remainder', 'Free SMS fragments left'),
+        ('sms_cost', 'Spent on SMS'),
```
quis, line 184: I know the concepts here are tricky but I’d prefer that we avoid ‘SMS’ and ‘fragments’.

‘SMS’ because we try to say ‘text messages’ (except in our technical documentation and client libraries).

‘Fragments’ because it’s a new term and it doesn’t quite mesh with how we talk about long messages being ‘charged as more than one message‘.

My attempt, cherry-picking language from elsewhere would be:
```suggestion
        ('sms_billable_units', 'Text messages charged'),
        ('sms_remainder', 'Free text message allowance remaining'),
        ('sms_cost', 'Spent on text messages'),
```

I’d be keen to know what @karlchillmaid thinks too.

CrystalPea: I am ok to avoid SMS, but `"text messages charged"` sounds really unclear to me. They are not charged for all those messages, as some will be in free allowance, and I think it would be good to indicate about fragments, even if historically we didn't do it much.

quis: Oof yeah. It really is tricky.

Maybe we just lose the `sms_billable_units` column for now until we have a better way of explaining this stuff? We don’t show this information on the page now.

CrystalPea: Could do, although I'd wait for @karlchillmaid to weigh in, I will link him this PR :)

karlchillmaid: I agree with Chris that – in order to get this PR merged – we might be better leaving off the title of the `sms_billable_units` column as ‘Sent’ for now.

We know that this is confusing because:
* the figures on the dashboard and usage page don’t tally
* the labelling on both of those pages does not make the distinction clear
* the Pricing page does not work well enough, or users aren’t finding that information when they need it
* we haven’t done the hard work to establish some clear language and definitions around this stuff

I’d like to see us prioritise this piece of work asap.

servingUpAces: What if the page showed the number of whole text messages sent and text message fragments sent.

karlchillmaid: @servingUpAces yeah, I’d like it to do that – I think it makes a lot of sense. If we do this (and I hope we do) we still need to get the language around those 2 things clear and consistent across the rest of Notify too.

CrystalPea: > we might be better leaving the title of the sms_billable_units column as ‘Sent’ for now

Just `Sent`? But then how do they know if it's letters, or sms fragments, or full sms or what? Or do you mean cutting this column out completely?

> I’d like to see us prioritise this piece of work asap.

I agree, it's a pretty important one.

> What if the page showed the number of whole text messages sent and text message fragments sent.

I'd like that, too, but I feel it's a separate story as we would have to alter the query for that, and if we do that we may want to think what other things to include.

CrystalPea:
Note: I just saw this bug card, which I will start working on tomorrow, and which should be fixed before I merged this PR I think: https://www.pivotaltracker.com/story/show/180154729

app/main/views/organisations.py
karlchillmaid: Sorry, yes ‘Text messages sent’ – to be consistent with the emails and letters columns.
I could have been clearer about that.

CrystalPea: Hey @karlchillmaid after more discussion with Ben I am leaning towards getting rid of this column altogether for now. We can think about reintroducing it later after we have a strategy for presenting this data.

The reason I am leaning this way is that the value in this column is: the number of sms fragments multiplied by international weighing. Meaning if someone sends a 2-fragment message abroad to a country with x2 multiplier, it would add 4 units to this value, if that makes sense. We currently don't have a good way of relaying this to our users.

Thinking about how other companies, like mobile networks, do this, I would say "credits" is the right word, for example "sms credits used". At present we don't use this term anywhere else though.

app/main/views/organisations.py
```diff
@@ -156,10 +158,50 @@ def organisation_dashboard(org_id):
...
+        financial_year=selected_year
+    )['services']
+
+    column_names = OrderedDict([
+        ('service_id', 'Service ID'),
+        ('service_name', 'Service Name'),
+        ('emails_sent', 'Emails sent'),
+        ('sms_remainder', 'Free text message allowance remaining'),
+        ('sms_cost', 'Spent on text messages'),
+        ('letter_cost', 'Spent on letters')
```
quis, line 184: Optional suggestion to make it clear what units we‘re using:
```suggestion
        ('letter_cost', 'Spent on letters (£)')
```

quis (approved):

PR #4075, opened 2021-11-24
Update ‘Send’ button content

karlchillmaid:
This PR updates the ‘Send’ button content to ‘Get ready to send’.

The aim is to make users feel more confident about using this button.

In research and through user support we’ve heard that users:
* worry that clicking the button will send the message immediately
* cannot find how to add or upload recipients


quis:
# Before

image

# After

image

quis (approved):

PR #4077, opened 2021-11-25
Reduce impact of bug with performance page caching stale data

idavidmcdonald:
Review commit by commit.

Note, the final commit is optional. It seemed sensible to do but I'm very happy to drop it.

Note, tests will fail until https://github.com/alphagov/notifications-admin/pull/3994 is merged and then this PR is rebased on top.

### DIFF
--- a/tests/app/notify_client/test_status_api_client.py
+++ b/tests/app/notify_client/test_status_api_client.py
@@ -33,7 +33,7 @@ def test_sets_value_in_cache(mocker):
     mock_redis_set.assert_called_once_with(
         'live-service-and-organisation-counts',
         '{"data_from": "api"}',
-        ex=604800
+        ex=3600
     )
 
 
--- a/tests/app/notify_client/test_status_api_client.py
+++ b/tests/app/notify_client/test_status_api_client.py
@@ -1,0 +1,59 @@
+from app.notify_client.status_api_client import StatusApiClient
+
+
+def test_get_count_of_live_services_and_organisations(mocker):
+    mocker.patch('app.extensions.RedisClient.get', return_value=None)
+    client = StatusApiClient()
+    mock = mocker.patch.object(client, 'get', return_value={})
+
+    client.get_count_of_live_services_and_organisations()
+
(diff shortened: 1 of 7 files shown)


tests/app/notify_client/test_status_api_client.py
```diff
@@ -0,0 +1,70 @@
...
+        'app.notify_client.NotifyAdminAPIClient.get',
+        return_value={'data_from': 'api'},
+    )
+    mock_redis_set = mocker.patch(
+        'app.extensions.RedisClient.set',
+    )
+
+    assert client.get_count_of_live_services_and_organisations() == {'data_from': 'api'}
+
+    assert mock_redis_get.call_args_list == [
```
benthorner, line 35: Optional: this is an unusual style of assertion compared with most of our tests. It proves that only one call was made, but that doesn't seem critical to check.

We could just have (more conventionally):

```
mock_redis.assert_called_with('live-service-and-organisation-counts')
```

(Similarly for all the other assertions.)

benthorner (approved):
Nice! Thanks for addressing my original concern. 

It looks like the org / service counts are kept fresh by [manually managing the cache](https://github.com/alphagov/notifications-admin/blob/5e8d0623de85eb0b10572635cdfc15ff6f35db32/app/notify_client/service_api_client.py#L121). However, manual cache management is inherently error-prone, so I agree with making the org/service counts expire regularly. I also had a look through all our `@cache` decorations and can see this is the only other "global" type stat where we might want to do this.

The value of the new `_cache` tests puzzles me - this isn't a reflection on the new tests but in general the thinking behind testing one declarative, unconditional `@cache` line with ~50 lines of Python. Seems overkill to me. I'm happy to visually check it's there, and the non-cache test proves it doesn't fundamentally break anything. On the other hand, I can see it's a bit of a convention so happy for you to leave them in. I just wouldn't add them myself.

idavidmcdonald:
To your wider point about the use and value of the test assertions, I completely agree, I spent a while wondering if it was even worth adding any tests.

In the end I chose to add them and keep them near essential to what is already in place but I'd also agree they are overkill. I didn't take the time to figure out what the right approach was.

PR #4084, opened 2021-12-01
Don’t force users to re-enter their password when renaming a service

quis:
The original idea behind this was to always ask users to re-enter their password any time:
- we want them to be sure that they want to do what they’re about to do
- we want to be sure it’s really the user trying to do the thing (and not someone malicious)

In reality we:
- removed this from the initial place it was added<sup>1</sup>
- only ever added it to the ‘rename service’ feature

So in reality it’s not a pattern we have persisted with. Arguably there are several things you can now do in the admin app without re-entering your password which are much more high consequence than changing the service name.

Also, with browser autofill there’s a lot less chance that forcing someone to re-enter a password really gives much defence against an unattended laptop, for example.

I also wonder whether we might get people to give better service names if we make the process of renaming the service less intimidating.

So this commit removes the need to re-enter your password when renaming a service.

Note that re-naming an organisation still has the same check, but I haven’t removed that too for the sake of keeping scope of the PR small.

***

1. An ancestor of the ‘suspend service’ feature https://github.com/alphagov/notifications-admin/pull/51

### DIFF
--- a/tests/app/main/views/test_service_settings.py
+++ b/tests/app/main/views/test_service_settings.py
@@ -15,7 +15,6 @@ from notifications_utils.clients.zendesk.zendesk_client import (
 )
 
 import app
-from app.formatters import email_safe
 from tests import (
     find_element_by_tag_and_partial_text,
     invite_json,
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_service_settings.py
```diff
@@ -809,39 +785,18 @@ def test_should_not_allow_duplicate_names(
...
-        session['service_name_change'] = service_new_name
-    page = client_request.get(
-        'main.service_name_change_confirm',
-        service_id=SERVICE_ONE_ID,
-    )
-    assert 'Change your service name' in page.text
-    app.service_api_client.get_service.assert_called_with(SERVICE_ONE_ID)
 
 
 def test_should_redirect_after_service_name_confirmation(
```
idavidmcdonald, line 790: Optional: may want to change the name of this test to remove the word 'confirmation' which feels misleading as there is no confirmation stage anymore

idavidmcdonald (approved):
Very nice

idavidmcdonald:
Oh forgot to say that I tested this locally, for both local and non local organisations. I also tested the error case as well of a duplicate service name

tests/app/main/views/test_service_settings.py
quis, line 790:
```suggestion
def test_should_redirect_after_service_name_change(
```

PR #4088, opened 2021-12-06
Add security policy page

idavidmcdonald:
This follows the guidance in
https://gds-way.cloudapps.digital/standards/vulnerability-disclosure.html#vulnerability-disclosure-and-security-txt

Note, I plan on just adding this for www.notifications.service.gov.uk and documents.service.gov.uk.

I think we can get away with not adding it to the tech-docs or the API but happy to be challenged on that

### DIFF
--- a/tests/app/main/views/test_security_policy.py
+++ b/tests/app/main/views/test_security_policy.py
@@ -1,0 +1,13 @@
+import pytest
+
+
+@pytest.mark.parametrize('url', [
+    '/security.txt',
+    '/.well-known/security.txt',
+])
+def test_security_policy_redirects_to_policy(client_request, url):
+    client_request.get_url(
+        url,
+        _expected_status=302,
+        _expected_redirect="https://vdp.cabinetoffice.gov.uk/.well-known/security.txt",
+    )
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_security_policy.py
```diff
@@ -0,0 +1,8 @@
+def test_security_redirects_to_policy(
+    client_request
+):
+    client_request.get(
+        'main.security_policy',
```
quis, line 5: For endpoints with a ‘standard’ URL I think it’s better to test the path itself which gives us the extra confidence that the `main.route` lines are correct:
```suggestion
    client_request.get_url(
        '/security.txt',
```

You’d probably want to parametrize the text to cover both of the two paths as well.

There’s precedent with `robots.txt` here:
https://github.com/alphagov/notifications-admin/blob/787cb3ef1f525daef3191d48fa87290aa1cd7126/tests/app/main/views/test_index.py#L63-L64

idavidmcdonald: Sensible, have rebased this change in. Should be good for a second review please

PR #4089, opened 2021-12-06
Show flash instead of inline upload errors

benthorner:

This has several advantages:

- It gives us more room to explain the error and actions. This will
be useful for upcoming work we want to do, which will add yet more
validations for CSV uploads.

- We already use a flash to show certain kinds of errors on these
pages (just above). This is more consistent.

- It's potentially more accessible. Previously the error and the
button text used to be read out as a single sentence. Now the page
reloads and reads the flash error alone.

In theory we should show an error in both places, but this can be
confusing on pages where there's only a single form control, and
especially if the error is long.

## Screenshots

| Before  | After |
| Screenshot 2021-12-06 at 17 13 14 | Screenshot 2021-12-06 at 17 13 01 |
| Screenshot 2021-12-06 at 17 13 23 | Screenshot 2021-12-06 at 17 10 17 |

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -175,6 +175,11 @@ def send_messages(service_id, template_id):
             ).format(
                 form.file.data.filename
             ))
+    elif form.errors:
+        # just show the first error, as we don't expect the form to have more
+        # than one, since it only has one field
+        first_field_errors = list(form.errors.values())[0]
+        flash(first_field_errors[0])
 
     column_headings = get_spreadsheet_column_headings_from_template(template)
 
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -417,6 +417,11 @@ def upload_contact_list(service_id):
             ).format(
                 form.file.data.filename
             ))
+    elif form.errors:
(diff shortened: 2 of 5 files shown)


benthorner:
@tombye not something to block this PR, but I observed a strange behaviour when using VoiceOver: the "Skip to main content" banner steals the focus from the flash banner. Is that meant to happen?

Screenshot 2021-12-06 at 17 16 19

tombye:
>@tombye not something to block this PR, but I observed a strange behaviour when using VoiceOver: the "Skip to main content" banner steals the focus from the flash banner. Is that meant to happen?
No. It looks like the action of Voiceover acquiring focus when the page loads, moving to the first interactive element on the page, is running after the script that focuses the error banner. This is a guess, based on anecdotes I've heard of developers adding short delays to JS that changes DOM so Voiceover can catch up.

@benthorner would you mind adding a bug for this as I think it will need some more investigation to figure out if I'm right and to find a solution that works (which will be a process of trial and error I think)?

benthorner:
@tombye sure, as discussed: https://www.pivotaltracker.com/story/show/180556274

quis (approved):

PR #4090, opened 2021-12-07
Reject CSV / Spreadsheet files larger than 10Mb

benthorner:

This is a quick additional check to protect the user:

- From getting a CloudFront 502 error if the file takes too
long to upload. I was surprised to find it takes about 1 minute
to upload a 70Mb file to S3.*

- From getting a CloudFront 502 error when we follow the redirect
and run through the slow processing code in utils that builds a
RecipientCSV [1].

For context, a CSV with 100K rows and a few columns is around 5Mb,
so a 10Mb limit should be enough. Analysis over the past week shows
that the vast majority of CSV uploads are actually < 2.5Mb.

I haven't added any tests for this because:

- The check isn't critical, as the worst case scenario is the user
gets a worse error than this in-app one.

- There's no easy way to mock the validation, and I didn't want to
have a test that depends on a 10Mb+ file.

*We're using "key.put" to upload the file, when we could be doing 
a multipart upload [2]. However, I tried this myself with a chunk 
size of 1000 bytes and found it only led to a marginal improvement.

[1]: https://github.com/alphagov/notifications-utils/pull/930
[2]: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-uploading-files.html

## Screenshot

Screenshot 2021-12-07 at 12 37 14

## Size analysis

Screenshot 2021-12-03 at 10 48 08

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -9,6 +9,7 @@ from flask_login import current_user
 from flask_wtf import FlaskForm as Form
 from flask_wtf.file import FileAllowed
 from flask_wtf.file import FileField as FileField_wtf
+from flask_wtf.file import FileSize
 from notifications_utils.columns import Columns
 from notifications_utils.countries.data import Postage
 from notifications_utils.formatters import strip_all_whitespace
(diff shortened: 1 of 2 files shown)


quis:
Good catch on the S3 upload time. That is surprisingly slow and makes this check well worth it.

I think you can test this at the view layer. Would look something like 

```python
from random import randbytes

page = client_request.post(
    'main.send_messages',
    …
    _data={'file': (BytesIO(randbytes(10_000_001), 'Big file')},
    content_type='multipart/form-data',
)
```

benthorner:
Oh nice, I'd not heard of / didn't think of `randbytes`.

I've amended the commit to add a test. I've only put the test for "send" flow as:

- We only need one test to cover the behaviour (we know the contact flow uses the same form).
- The validation isn't as critical for the contact flow.
- The future of the contact flow is still a bit unclear e.g. we may want to severely limit max rows.

I noticed there are quite a few tests missing for the contact flow, acknowledging that it was done in a hurry. I've got a follow-up task to look at that, but for now I'd like to stick with just the one test.

quis (approved):

PR #4095, opened 2021-12-09
Remove redundant 413 error page

benthorner:

This was used when there was an Nginx instance sitting in front of
Admin [1], but nowadays traffic goes through CloudFront, where we
decided not to implement the same protection:

- The likelihood of large requests being a security threat is small
because it's a difficult attack vector.

- We have put in place specific limits on routes where we the size
of the request is actually important [2].

Note that the other error pages can all still be used based on the
response code we get from API requests [3]. Also worth noting we've
had 0 413 response codes for Admin in the last month.

[1]: https://github.com/alphagov/notifications-aws/blob/master/ansible/roles/nginx/templates/nginx.conf.j2#L29-L30
[2]: https://github.com/alphagov/notifications-admin/pull/4090
[3]: https://github.com/alphagov/notifications-admin/blob/b3c0abc4966c847b9a4ccc33a1cf1a148620261e/app/__init__.py#L407-L416


benthorner:
Note: I will likely close this in favour of https://github.com/alphagov/notifications-admin/pull/4096, depending on how that goes.

CrystalPea (approved):
That looks good to me 👍🏼

PR #4096, opened 2021-12-09
Improve handling of large CSVs / spreadsheets

benthorner:

This pulls in a bunch of performance improvements from utils, which
include raising a CSVTooBigError if a CSV takes too long to process.

I want to get these improvements out quickly so we can monitor their
impact. To that end, I've reused an old error page to handle the new
"too big" error, until we have some more time to iterate the content
and overall error handling for large CSVs, as it's...complicated.

## Screenshot

Screenshot 2021-12-09 at 17 23 25

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -28,7 +28,10 @@ from notifications_utils.formatters import (
     formatted_list,
     get_lines_with_normalised_whitespace,
 )
-from notifications_utils.recipients import format_phone_number_human_readable
+from notifications_utils.recipients import (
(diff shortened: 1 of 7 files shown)


benthorner:
Closing in favour of https://github.com/alphagov/notifications-admin/issues/4099 - the next step for this new effort would be impractical due to the complexity of adding more error handling in multiple places where we already have a lot of complex error handling. See https://github.com/alphagov/notifications-utils/pull/936 for more details.

PR #4097, opened 2021-12-10
Add link to user page to go live tickets

klssmith:
This adds a link to the user profile of the person who requested to go
live for "Request to go live" Zendesk tickets. Viewing a user's profile
page helps us to check for duplicate organisations and services from
that user.

[Pivotal story](https://www.pivotaltracker.com/story/show/180521177)

### DIFF
--- a/app/templates/support-tickets/go-live-request.txt
+++ b/app/templates/support-tickets/go-live-request.txt
@@ -30,4 +30,5 @@ Other live services for that user: {{ user.live_services|format_yes_no }}
 Service reply-to address: {{ service.default_email_reply_to_address or "not set" }}
 
 ---
-Request sent by {{ current_user.email_address }}
+Request sent by {{ user.email_address }}
+Requester’s user page: {{ url_for('main.user_information', user_id=user.id, _external=True) }}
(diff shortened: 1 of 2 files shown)


app/templates/support-tickets/go-live-request.txt
```diff
@@ -31,3 +31,4 @@ Service reply-to address: {{ service.default_email_reply_to_address or "not set"
 
 ---
 Request sent by {{ current_user.email_address }}
+Requester’s user page: {{ url_for('main.user_information', user_id=user.id, _external=True) }}
```
quis, line 34: Think this need to be:
```suggestion
Requester’s user page: {{ url_for('main.user_information', user_id=current_user.id, _external=True) }}
```

klssmith: This is working because we have `{% set user = current_user -%}` at the top of the file. We use `user.` in most places and `current_user.` once. It would be nice to be consistent though - any preference for which form to keep? If not, I might just get rid of the one instance of `current_user`.

quis: I think keep the variables defined at the top of the file. They make the rest of it a bit easier to read. So 👍🏻 to getting rid of the one instance of `current_user`

quis (approved):

PR #4103, opened 2021-12-20
Add hscni.net to list of allowed domains

klssmith:
We can't assign `hscni.net` to an organisation because it is used
by GP surgeries, so we don't always know which org it should be
associated with. This change allows people with the an `hscni.net`
email address to sign up and create a service.


quis (approved):
:thum

PR #4105, opened 2021-12-20
Update `get_notifications_for_service` to make POST requests

klssmith:
This updates the `get_notifications_for_service` of the
`NotificationApiClient` to make POST request to the api when the `to`
field is provided. This is done so that we avoid logging personal
details such as email addreses.

If the `to` field is not present, the method will make a GET request and
there will be no change to how it works.

Merge after https://github.com/alphagov/notifications-api/pull/3403

### DIFF
--- a/app/notify_client/notification_api_client.py
+++ b/app/notify_client/notification_api_client.py
@@ -19,7 +19,6 @@ class NotificationApiClient(NotifyAdminAPIClient):
         to=None,
         include_one_off=None,
     ):
-        # TODO: if "to" is included, this should be a POST
         params = {
             'page': page,
             'page_size': page_size,
@@ -35,17 +34,22 @@ class NotificationApiClient(NotifyAdminAPIClient):
 
         params = {k: v for k, v in params.items() if v is not None}
 
+        # if `to` is set it is likely PII like an email address or mobile which
+        # we do not want in our logs, so we do a POST request instead of a GET
+        method = self.post if to else self.get
+        kwargs = {'data': params} if to else {'params': params}
+
         if job_id:
-            return self.get(
+            return method(
                 url='/service/{}/job/{}/notifications'.format(service_id, job_id),
-                params=params
+                **kwargs
             )
         else:
             if limit_days is not None:
                 params['limit_days'] = limit_days
-            return self.get(
+            return method(
                 url='/service/{}/notifications'.format(service_id),
-                params=params
+                **kwargs
             )
 
     def send_notification(self, service_id, *, template_id, recipient, personalisation, sender_id):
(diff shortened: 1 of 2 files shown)


app/notify_client/notification_api_client.py
```diff
@@ -35,17 +34,22 @@ def get_notifications_for_service(
...
         if job_id:
-            return self.get(
+            return method(
                 url='/service/{}/job/{}/notifications'.format(service_id, job_id),
-                params=params
+                **kwargs
             )
         else:
             if limit_days is not None:
                 params['limit_days'] = limit_days
```
idavidmcdonald, line 49: Nothing needed from you but just to share that I learnt something new here

I'm aware of the behaviour where you might expect the value of `b` at the end to be `3` but it is actually `1`

```
>>> a = 1
>>> b = a
>>> a = 3
>>> b
1
```


However, I assumed that behaviour would be the same for adding keys to a dict but it is not!

```
>>> d = {}
>>> e = d
>>> d['hi'] = 'hello'
>>> d
{'hi': 'hello'}
>>> e
{'hi': 'hello'}
```

Which means that this line is correctly doing what we wanted :)

app/notify_client/notification_api_client.py
```diff
@@ -35,17 +34,22 @@ def get_notifications_for_service(
 
         params = {k: v for k, v in params.items() if v is not None}
 
+        # if `to` is set it is likely PII like an email address or mobile which
+        # we do not want in our logs, so we do a POST request instead of a GET
+        method = self.post if to else self.get
+        kwargs = {'data': params} if to else {'params': params}
```
idavidmcdonald, line 40: Nicely done for this bit of logic

idavidmcdonald (approved):
Nice work :)

app/notify_client/notification_api_client.py
klssmith: I'm pretty sure you wrote that line when we paired on this 😄

app/notify_client/notification_api_client.py
klssmith: Interesting... I think the examples have a difference, which is why they behave differently. In the first one, `a` is reassigned so I'd expect `a` and `b` to be different. In the second example, `d` is not reassigned, so `d` and `e` still point to the same object.

If you change the second example to do the same thing (to reassign the first variable) then they do behave in the same way:
```python
>>> d = {}
>>> e = d
>>> d['hi'] = 'hello'
>>> d = {}                         # the added line
>>> d
{}
>>> e
{'hi': 'hello'}
```

app/notify_client/notification_api_client.py
idavidmcdonald: Ha, I thought I wrote some of this logic but not all of it. I thiiiink you extended on what I wrote....

Regardless, I'll make sure to continue giving myself praise when reviewing others PRs ;)

PR #4106, opened 2021-12-21
Revert "Remove free text allowance remaining column from organisation report"

CrystalPea:
The bug that led to us temporarily removing this column from the report has now been fixed, so we can reinstate the column.

### DIFF
--- a/tests/app/main/views/organisations/test_organisations.py
+++ b/tests/app/main/views/organisations/test_organisations.py
@@ -696,7 +696,7 @@ def test_download_organisation_usage_report(
                 'letter_cost': 30.5,
                 'sms_billable_units': 122,
                 'sms_cost': 1.934,
-                'sms_remainder': None
+                'sms_remainder': 0
             },
             {
                 'service_id': SERVICE_TWO_ID,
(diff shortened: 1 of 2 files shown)


tests/app/main/views/organisations/test_organisations.py
```diff
@@ -720,10 +720,10 @@ def test_download_organisation_usage_report(
     )
 
     assert csv_report.string == (
-        "Service ID,Service Name,Emails sent,"
+        "Service ID,Service Name,Emails sent,Free text message allowance remaining,"
         "Spent on text messages (£),Spent on letters (£)"
-        "\r\n596364a0-858e-42c8-9062-a8fe822260eb,Service 1,13000,1.93,30.50"
-        "\r\n147ad62a-2951-4fa1-9ca0-093cd1a52c52,Service 1,23000,3.94,60.50\r\n"
+        "\r\n596364a0-858e-42c8-9062-a8fe822260eb,Service 1,13000,,1.93,30.50"
+        "\r\n147ad62a-2951-4fa1-9ca0-093cd1a52c52,Service 1,23000,,3.94,60.50\r\n"
```
klssmith, line 726: Really minor thing, but when I tried this out instead of having a missing value there is a `0` in the "Free text message allowance remaining" column. I think this is because API returns sms_remainder as 0 and not as None, so the mocks above needs updating.

klssmith (approved):

PR #4107, opened 2021-12-22
Show validation-failed status for templated letters over 10 pages 

servingUpAces:
If a letters that has been posted via the API has more than 10 pages it would not get a validation-failed status. This also happens for letters in a CSV upload, only the first row has been validated for having too many pages, because you need to created the pdf before getting an accurate page count.

The API has been updated to mark these letters as invalid and move the
letter to the invalid s3 bucket, the meta data is also set with the
error message and page count.

This PR updates the notification page to display the validation error.

related PRs: 
alphagov/notifications-template-preview#619
alphagov/notifications-api#3418

https://www.pivotaltracker.com/story/show/169209742

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -18,6 +18,7 @@ from flask import (
     url_for,
 )
 from notifications_python_client.errors import APIError, HTTPError
+from notifications_utils import LETTER_MAX_PAGE_COUNT
 from notifications_utils.letter_timings import (
     get_letter_timings,
     letter_can_be_cancelled,
@@ -79,6 +80,13 @@ def view_notification(service_id, notification_id):
             )
     else:
         page_count = get_page_count_for_letter(notification['template'], values=personalisation)
+        if page_count and page_count > LETTER_MAX_PAGE_COUNT:
+            # when a templated letter is sent via the api and the personalisation pushes
+            # the page count over 10 we need to validate that. The letter is not in validation failed,
+            # this will enable us to show the error. Perhaps there is a better way to do this.
+            error_message = get_letter_validation_error(
+                "letter-too-long", [1], page_count
+            )
 
     if notification.get('postage'):
         if notification["status"] == "validation-failed":
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -81,9 +81,15 @@ def view_notification(service_id, notification_id):
     else:
         page_count = get_page_count_for_letter(notification['template'], values=personalisation)
         if page_count and page_count > LETTER_MAX_PAGE_COUNT:
-            # when a templated letter is sent via the api and the personalisation pushes
-            # the page count over 10 we need to validate that. The letter is not in validation failed,
-            # this will enable us to show the error. Perhaps there is a better way to do this.
+            # We check page count here to show the right error message for a letter that is too long.
+            # Another way to do this would be to get the status and error message from letter metadata.
+            # This would be a significant amount of work though, out of scope for this bug fix.
+            # This is because currently we do not pull the letter from S3 when showing preview.
+            # Instead, we generate letter preview based on the letter template and personalisation.
+            # Additionally, when a templated letter is sent via the api and the personalisation pushes the
+            # page count over 10 pages, it takes a while for validation status to come through.
+            # Checking page count here will enable us to show the error message even if the letter is not
+            # fully processed yet.
             error_message = get_letter_validation_error(
                 "letter-too-long", [1], page_count
             )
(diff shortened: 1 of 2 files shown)


app/main/views/notifications.py
```diff
@@ -79,6 +80,13 @@ def view_notification(service_id, notification_id):
             )
     else:
         page_count = get_page_count_for_letter(notification['template'], values=personalisation)
+        if page_count and page_count > LETTER_MAX_PAGE_COUNT:
+            # when a templated letter is sent via the api and the personalisation pushes
+            # the page count over 10 we need to validate that. The letter is not in validation failed,
```
CrystalPea, line 85: > The letter is not in validation failed, this will enable us to show the error

why is the letter not in validation failed?

servingUpAces: The letter is in validation-failed. 
The trouble with this page is that it doesn't get the file from S3, which does have the metadata with the error message. Instead it uses a preview of the letter using template-preview. This is why I had to check the page count in this way. 
I am open to other options, do you have any ideas?

leohemsted: hmmmm, why aren't we getting the file from s3? we put it in the `letters-invalid-pdf` bucket with metadata attached in your template preview PR

servingUpAces: Would this code be more clear if I add a `if validation-failed` statement?

servingUpAces: Only trouble with adding that statement is that it could take time to go to validation-failed because the template-preview app has to first create the pdf, then call a task in API to update the status of the notification. Without the `if validation-failed` the user could see the failure as soon as the notification is uploaded.

CrystalPea: I think it's ok (although downloading file and metadata from S3 would be even better = one source of truth), I'd change the comment to be more clear though, maybe something like:

```
when a templated letter is sent via the api and the personalisation pushes the page count over 10 pages, it takes a while for validation status to come through. Checking page count here will enable us to show the error even if the letter is not in validation failed yet.
```

leohemsted: i think this whole function could be really refactored, at least the letter part of it. I found it pretty hard to follow and confusing as to why there's all these separate blocks. (i also _really_ don't like how admin is parsing precompiled PDFs for its page count itself if there isn't info in the metadata, but that's a separate issue). There's other stuff I don't understand like "why are we modifying the notification['template']['postage'] rather than reading from the notification itself". I think we could shake this whole block up by starting again from what we actually need and laying it out sensible.

But that said, i don't think that needs to block this PR going in, and I think pea's comment change is a great suggestion.

app/main/views/notifications.py
```diff
@@ -81,9 +81,11 @@ def view_notification(service_id, notification_id):
     else:
         page_count = get_page_count_for_letter(notification['template'], values=personalisation)
         if page_count and page_count > LETTER_MAX_PAGE_COUNT:
-            # when a templated letter is sent via the api and the personalisation pushes
-            # the page count over 10 we need to validate that. The letter is not in validation failed,
-            # this will enable us to show the error. Perhaps there is a better way to do this.
+            # This will enable us to show the error message for a letter that is too long.
+            # When a templated letter is sent via the api and the personalisation pushes
```
CrystalPea, line 85: This comment still feels a bit chaotic. It mostly explains why we went for this solution - that would really fit better as a part of commit message.

Also the part that explains validation process for templated API letters doesn't state how this is connected with the decision to count pages here.

I spent some time on this, I hope I arrived at something that can satisfy us both. Up to you if you'd rather have it inline with the code, or as a "WHY" part in the commit message ☺️ .


> We check page count here to show the right error message for a letter that is too long. Another way to do this would be to get the status and error message from letter metadata. This would be a significant amount of work though, out of scope for this bug fix. This is because currently we do not pull the letter from S3 when showing preview. Instead, we generate letter preview based on the letter template and personalisation.
> Additionally, when a templated letter is sent via the api and the personalisation pushes the page count over 10 pages, it takes a while for validation status to come through. Checking page count here will enable us to show the error message even if the letter is not fully processed yet.

CrystalPea (changes requested):
Hey, good work fixing this bug. 

I found the comment was unclear, and maybe it would fit in better as a part of commit message?

I rewrote it a bit, let me know what you think.

app/main/views/notifications.py
servingUpAces: Yes I agree that it would be better to refactor this whole page, and I don't like how I had to update the page to get the validation error to show. 
However, I would like to put this in to complete the story and look at the large refactor at another time.

CrystalPea (approved):

PR #4114, opened 2021-12-30
Trial running Sentry in a Flask app

benthorner:
https://www.pivotaltracker.com/story/show/180766893

This will capture and send various events to Sentry:

- Any unhandled exceptions.
- Any logger.error calls.
- Some request traces.

The latter are severely limited to avoid going over the free tier limits
for Sentry, and to avoid excess effort on our end.

## App selection

This app was chosen because:

- It emits a variety of errors (logs and exceptions) on a regular basis.
- It doesn't have any significant performance concerns (relative to API).

## Deployment

This PR is intended to be a trial: after merging, I'll raise a PR to revert, with the intention of deploying in about a week.

In order to deploy this, we'll need to temporarily set the `SENTRY_DSN` environment variable in all environments. One option is to add it to `credentials`, which is annoying as it's another PR to revert. Another option is to manually set it and remove it (we'll need to do this anyway) by doing `cf set-env notify-admin SENTRY_DSN <dsn>`.

I prefer the second option.

### DIFF
--- a/application.py
+++ b/application.py
@@ -1,6 +1,19 @@
+import os
+
+import sentry_sdk
 from flask import Flask
+from sentry_sdk.integrations.flask import FlaskIntegration
 
 from app import create_app
+
+sentry_sdk.init(
+    dsn=os.environ['SENTRY_DSN'],
+    integrations=[FlaskIntegration()],
+    environment=os.environ['NOTIFY_ENVIRONMENT'],
+    attach_stacktrace=True,
+    traces_sample_rate=0.00005  # avoid exceeding rate limits in Production
+)
+sentry_sdk.set_level('error')  # only record error logs or exceptions
 
 application = Flask('app')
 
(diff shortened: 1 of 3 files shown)


benthorner:
Works in Preview, so removing the test route:

```
    @application.route('/test')
    def test():
        application.logger.error('error me')
        application.logger.warning('warn me')
        1/0
```

## Screenshot

Screenshot 2021-12-30 at 14 50 47

idavidmcdonald (approved):
Looks good. How do you intend to give it to the environment variable it needs?

benthorner:
@idavidmcdonald I've amended the PR description to explain that. Open to doing either option. I don't see any risk from doing it manually - we're not going to re/create this app in any environment in the next week.

idavidmcdonald:
I'm happy to go with the second option. One thing to point out (although you may already be aware) is that you will need to do it before the deployment happens to any environment as otherwise we would get a `KeyError` when starting the app.

Note, https://github.com/alphagov/notifications-manuals/wiki/Production-access means we shouldn't do any `cf env` command in production without a two eye (just in case you want to run that command after `set-env` to check you've done it correctly.

PR #4118, opened 2021-12-31
Standardise on use of `client_request` fixture

quis:
My late Christmas present to you all.

# Before

There are 5 different fixtures for making a request in a test:
- `client`
- `client_request`
- `logged_in_client`
- `logged_in_client_with_session`
- `platform_admin_client`

Depending which one you used you:
- would get a different type back
- may or may not be logged in
- may or may not get useful extra checks like asserting for a sensible HTTP status or checking the markup for common accessibility problems

# After

There is one fixture for making a request:
- `client_request`

Want to be logged in with a different user? Call `client_request.login(user)` first.

Don’t want to be logged in? Call `client_request.logout()` first (most of our tests need to be logged in).

Need an instance of `Response` object not an instance of `BeautifulSoup`? Use `client_request.get_response` or `client_request.post_response`.

Need to pass in a URL, not arguments to `url_for`? Use `client_request.get_url(…)` or `client_request.post_url(…)`.

Need to pass in a URL and get a `response` back? Use `client_request.get_response_from_url(…)` or `client_request.post_response_from_url(…)`.


servingUpAces (approved):
Nice work

PR #4122, opened 2022-01-06
Don’t allow <text> elements in letter logos 

quis:
SVG images are XML documents which can contain text, for example `<text>This is an example</text>`.

To render text in an SVG consistently the system rendering the SVG must have the fonts specified by the SVG installed.

If the fonts are not installed then the renderer will fall back to a system font and the text will look different. This is especially bad news for branding where the right font is an integral part of any brand:

How it should look | How it looks with the fonts missing
---|---
image | image

To fix this, the text should instead be converted to `<path>` elements. This process is sometimes called ‘outlining’.

A few of our logos had this problem, and I’ve fixed most of them by hand. Adding this validation will stop the problem, coming up again.


app/main/validators.py
```diff
@@ -77,18 +78,30 @@ def __call__(self, form, field):
...
-        is_image_embedded = '<image' in field.data.stream.read().decode("utf-8")
+        svg_contents = field.data.stream.read().decode("utf-8")
         field.data.stream.seek(0)
-        if is_image_embedded:
+        if f'<{self.element}' in svg_contents.lower():
             raise ValidationError(self.message)
 
 
+class NoEmbeddedImagesInSVG(NoElementInSVG):
+    element = 'image'
```
benthorner, line 101: Optional: this is concise but surprising because the type of the class-level attribute (string) is different that of the base class (property). I can see the temptation to do it this way and don't feel strongly about it.

Did you consider writing a method?

benthorner (approved):
Nice one ⭐.

Can you update [the kickoff doc](https://docs.google.com/document/d/16RL_aBBa4DWUT6vr2GGHH9O-Gaf3NVmQo3KnOpLMzxc/edit#heading=h.cubb7jk12ev9) when this goes in?

I tried this out locally and it works just fine ✅. Will be interesting to see what happens the first time someone on support encounter this!

app/main/validators.py
quis: This is a slightly awkward nuance of how abstract attributes have to be defined in Python.

Both of these are equivalent:
```python
class Example:
    @property 
    def a(self):
        return True
```

```python
class Example:
    a = True
```

In both cases the type returned is `bool`:
```python
Example().a
>>> True
```

However only the first way lets you mark `a` as being abstract:
```python
class AbstractExample(ABC):
    @property 
    @abstractmethod
    def a(self):
        return True
```

So that:
```python
class ConcreteExample(AbstractExample):
    pass

ConcreteExample()
>>> TypeError: Can't instantiate abstract class ConcreteExample with abstract method a
```

quis: Which might be slightly over-engineering things in this case…

benthorner: Yes, I think this might be going that way - just a little bit 😄. 

They aren't totally equivalent, by the way:

```
type(BaseClass.a) => property
type(Subclass.a) => bool
```

There's some irony there about which principles we choose to follow. But I'll stop self-indulging now: they're functionally equivalent, and that's what matters; the abstract method makes the trait clear 👍.

PR #4127, opened 2022-01-11
Allow organisation team members to be removed

klssmith:
Needs to be merged after
- [x] https://github.com/alphagov/notifications-api/pull/3423

We had some endpoints and pages to remove a user from an organisation but these weren't working. This makes it all work (once the api PR is merged) and removes the old code that is either no longer needed or no longer works. We've decided not to have a page for each individual organisation team member, so the "Team members" page is used to show the confirmation banner.

More details in the individual commit messages. 

[Pivotal story](https://www.pivotaltracker.com/story/show/180573449)


servingUpAces (commented):
Looks good. Quick question: can a user remove themselves from the organisation?
I can't remove myself from a service. Maybe I should be able to do that. 
(I've corrected the typo in this comment)

klssmith:
I've written this so that users can remove themselves - I double checked that with Chris after and he thought that was ok. You can't remove yourself from a service, but it would probably be a bit more work to implement that since we have requirements about minimum number of team members for a service and the permissions they need to have.

servingUpAces:
When you remove yourself from an organisation and you are not a platform admin, you get "You're not allowed to see this page" because you are still on the /organisation page. Perhaps we can redirect them to their /accounts page.

klssmith:
@servingUpAces thanks, I've changed where the redirect goes on this line (and have rebased into the original commit) https://github.com/alphagov/notifications-admin/pull/4127/files#diff-11eecc50c648e6b1c5c13b38769c3e8b3135e392651c134664c6de99ee0de65eR287

servingUpAces (approved):
Good work. I've tested removing myself (not as a platform admin), removing a user as someone else, having more than 7 users on the team members page (search form works as expected). And I've removed the last user from organisation (that' okay to do).

PR #4128, opened 2022-01-11
Don’t force users to re-enter their password when renaming an organisation

quis:
This replicates https://github.com/alphagov/notifications-admin/pull/4084 but for renaming organisations rather than services.


app/main/views/organisations.py
```diff
@@ -330,12 +320,16 @@ def edit_organisation_name(org_id):
...
+        except HTTPError as http_error:
+            error_msg = 'Organisation name already exists'
+            if http_error.status_code == 400 and error_msg in http_error.message:
+                form.name.errors.append('This organisation name is already in use')
+            else:
+                raise http_error
+        else:
+            return redirect(url_for('.organisation_settings', org_id=org_id))
 
     return render_template(
```
benthorner, line 334: You can move this up one level as it's now the only template in that directory.

app/notify_client/organisations_api_client.py
```diff
@@ -81,12 +81,6 @@ def remove_user_from_organisation(self, org_id, user_id):
         data = _attach_current_user({})
         return self.delete(endpoint, data)
 
-    def is_organisation_name_unique(self, org_id, name):
-        return self.get(
-            url="/organisations/unique",
```
benthorner, line 87: Are you planning to remove these APIs as well?

app/notify_client/organisations_api_client.py
```diff
@@ -56,11 +56,10 @@ def update_organisation(self, org_id, cached_service_ids=None, **kwargs):
         if cached_service_ids:
             redis_client.delete(*map('service-{}'.format, cached_service_ids))
 
-        return api_response
+        if 'name' in kwargs:
+            redis_client.delete(f'organisation-{org_id}-name')
```
benthorner, line 60: Optional: could we just have `@cache.set` on `get_organisation`? Then we could just have `@cache.delete` on `update_organisation`.

app/main/views/organisations.py
```diff
@@ -329,6 +329,10 @@ def edit_organisation_name(org_id):
         form.name.data = current_organisation.name
 
     if form.validate_on_submit():
+
+        if form.name.data == current_organisation.name:
```
benthorner, line 320: Are you sure this is necessary? I'm unable to reproduce it locally.

benthorner (commented):
Screenshot 2022-01-12 at 15 22 32

I like it when we can delete things for the benefit of the user. Right now the users are just Platform Admins, but I'm assuming this will get exposed as we overhaul organisations.

Before this goes in I want to double check the same-org-name scenario, as it's not behaving as stated. I also found one extra tidy-up change you could make.

app/notify_client/organisations_api_client.py
quis: I think the reason for not doing this previously was because there’s stuff in the organisation response which is hard to cache. Maybe some of this stuff isn’t returned any more, but the [count of live services](https://github.com/alphagov/notifications-api/blob/081e0cab8898df6f268a3da7dd2dae32e6790e54/app/models.py#L449) still strikes me as something which would be difficult to invalidate from all the places which can change it.

app/main/views/organisations.py
quis: Changed in 473bb89cb

app/notify_client/organisations_api_client.py
quis: Have started a draft PR over at https://github.com/alphagov/notifications-api/pull/3427

app/main/views/organisations.py
quis: I can’t reproduce it for renaming a service either. It definitely was a problem at one point though: https://github.com/alphagov/notifications-admin/pull/2137

app/main/views/organisations.py
benthorner: ...but broke the build ;-).

app/notify_client/organisations_api_client.py
benthorner: Fair enough, thanks for giving it some thought 💯.

app/main/views/organisations.py
benthorner: Aha, this was [due to the `/unique` endpoint](https://github.com/alphagov/notifications-admin/pull/2137/files#diff-a87e3297d1ea8866e40da7b2861cd7d5d8a604e7847705ebb8d5d455ff6a6877R126), which [looks like it got subsequently fixed](https://github.com/alphagov/notifications-api/commit/e1cc8175d7278c2f723e8e8efe847d93cd81fe6b). Either way, this means we don't need the check here, as we always do an `update()` on the current organisation.

Can you remove it?

benthorner (approved):
Thanks for addressing my comments 🏅.

Just one or two things to fix (e.g. so the tests pass).

app/main/views/organisations.py
quis: I came to the same conclusion eventually – removed in 3ab30f27b

PR #4133, opened 2022-01-18
Point people to Zendesk if there's a problem

benthorner:
This was a minor issue in a recent incident [1].

We don't want people emailing us directly:

- It's harder to collaborate on replies than in Zendesk, where I
can see who's editing.

- We can't take advantage of macros like we can in Zendesk.

- People emailing us directly creates an unnecessary distraction
and it's easy to lose track. We want to minimise distractions in
an incident situation.

- We use notify-support@ for monitoring, so we don't want noise.

[1]: https://docs.google.com/document/d/1eArpcYoCmhlgB45FZpGj1_HnkAgpgtgwr-KBiUq7h8c/edit#

### DIFF
--- a/app/assets/error_pages/5xx.html
+++ b/app/assets/error_pages/5xx.html
@@ -75,7 +75,7 @@
               Try again later.
             </p>
             <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


idavidmcdonald (approved):
Looks good technically. Probably worth giving Ale a heads up about this change please

PR #4134, opened 2022-01-18
Reintroduce DiffDOM

tombye:
Fixes an issue raised in this story: https://www.pivotaltracker.com/story/show/180011155 by re-introducing the [DiffDOM](https://github.com/fiduswriter/diffDOM/blob/main/browser/diffDOM.js) library to our JavaScript and making it work in Internet Explorer (IE) down to version 9.

## What DiffDOM does in our code

Our `updateContent.js` script keeps parts of our pages in sync with the app's data. For example, it updates the number of emails sent on the dashboard when this changes.

It works by:
1. polling the server
2. getting new HTML for them
3. diff'ing it against the existing HTML
4. applying any changes
 
We used to use the DiffDOM library for this but changed it when its code was updated to a version of JavaScript that didn't work with IE (see https://www.pivotaltracker.com/story/show/165380360). Turns out, the replacement library just swapped all the old HTML for the new every time, irrelevant of differences which caused issues like https://www.pivotaltracker.com/story/show/180011155/comments/227516175.

DiffDOM, by contrast, only applies changes if there were some and only changed the different parts that differed. It also seems to be the only decently supported library that does this. So we have to reintroduce DiffDOM and make it work with IE (down to 9).

## Making it work with Internet Explorer

DiffDOM is now written in ECMAScript (the spec' for JavaScript) version 6 whereas IE9 uses a version closer to 3.

We therefore need to:
1. rewrite the ES6 syntax (ie. arrow functions, for...of loops, etc) to an ES3 equivalent
2. polyfill any new features
    - JS built-in objects (ie. Promises)
    - extensions to existing ones (ie. Object.entries)
    - DOM APIs (ie. Nodelist.prototype.forEach)

Because DiffDOM is 3rd party, we need to work out which polyfills it needs and in a repeatable way so we can add the right polyfills now but also in the future when it changes.

This pull request proposes using [BabelJS](https://babel.dev/) to determine which polyfills are needed, and the changes it makes come from this.

## BabelJS

BabelJS is a JS compiler. It turns JS code into an [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) (AST) which can then be used by other JS (plugins) to transform it into another type of code or to infer things about it. You can use it with a group of plugins (presets) called `preset-env` to transform newer versions of ECMAScript syntax into older. It can also be used to add polyfills based on target browsers and the language features used in the code.

If you use BabelJS to add polyfills, it will use those from [the CoreJS library](https://github.com/zloirock/core-js/), based on the target browser's support for the latest stable ECMAScript. This means any feature used where the target browsers don't support the latest ECMAScript version will be polyfilled. As an example, [here's BabelJS transpiling ES6 and adding polyfills to it, in its REPL (targeting IE9)](https://babeljs.io/repl#?browsers=ie%209&build=&builtIns=usage&corejs=3.6&spec=false&loose=false&code_lz=MYewdgzgLgBAhgJwQRhgXhgbWQGgEw4DMOALALoBQFoksiCe68SyAdBADYCWwApgBR4AlEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Cstage-2&prettier=false&targets=&version=7.16.9&externalPlugins=&assumptions=%7B%7D). Note: it adds imports for the polyfill code as it assumes you will bundle the code up afterwards.

This can mean it adds a lot of code. In my tests with DiffDOM, BabelJS added ~60KB of polyfills vs ~8KB from Polyfill.io.

## Using Polyfill.io to add only the polyfills needed

The [Polyfill.io](https://polyfill.io/v3/) service takes a list of feature names and will respond with a bundle of those needed for the browser requesting them. You can try it out by [building a URL for your bundle](https://polyfill.io/v3/url-builder/) and `curl`ing it with the user agent header, like so:

```bash
curl https://polyfill.io/v3/polyfill.js?features=Object.entries -H 'User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; WOW64; Trident/5.0)'
``` 

It uses Mozilla Developers Network (MDN) pages for browser support data and will assume any features a browser supports don't need polyfilling unless the newer version includes a major change (ie. `Array.prototype.sort`, which introduced [stable sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability) in newer versions). 

I sent the names of the features BabelJS used CoreJS to polyfill to Polyfill.io with an IE9 user agent string. The bundle it responded with is used in these changes to polyfill DiffDOM. I've used the uncompressed, commented version of the bundle because it lists all the polyfills used, those not included and all the internal dependencies.

I've recorded the decision-making process in [this document](https://docs.google.com/drawings/d/1k0z-TQdzuNb3nqgVwMfsobjEHp36kDwvW4zMJnnR3rw/edit).

## Caveats to the approach

There are necessary manual parts to this process:
- you need to be able to infer the features used from the core-js polyfills BabelJS adds
- some features polyfill.io doesn't have are because MDN says all browsers support them, some because it just doesn't have the polyfill yet. You need to figure out why a polyfill isn't included and whether it should be or not

Both require looking at the code you're polyfilling (DiffDOM here) for uses of the feature and making a decision about whether it needs a polyfill or not. For example, BabelJS added a polyfill for `RegExp.prototype.sticky` for DiffDOM. Polyfill.io didn't include a polyfill for this but it is a modern feature (so not omitted because all browsers support it) so the code needed checking. Having found no uses of it, I decided not to include it.

## How these changes are tested

I tested these changes by looking at the main dashboard when sending a notification (to test updates to notification numbers) and the 'current alerts' page when adding and rejecting an alert (to test additions to the DOM). Both tests were done in IE9 and IE11 and worked without errors.

### DIFF
--- a/tests/javascripts/updateContent.test.js
+++ b/tests/javascripts/updateContent.test.js
@@ -31,6 +31,9 @@ beforeAll(() => {
 
   $.ajax.mockImplementation(() => jqueryAJAXReturnObj);
 
+  // RollupJS assigns our bundled module code, including DiffDOM, to window.GOVUK.
+  // DiffDOM is assigned to its vendor property so we need to copy that here for the updateContent
+  // code to pick it up.
   window.GOVUK.vendor = {
     DiffDOM: require('diff-dom').DiffDOM
   };
--- a/tests/javascripts/updateContent.test.js
+++ b/tests/javascripts/updateContent.test.js
@@ -31,13 +31,9 @@ beforeAll(() => {
 
   $.ajax.mockImplementation(() => jqueryAJAXReturnObj);
 
-  // using require to execute the version we use in our our frontend build here can't add
-  // the domdiff variable to this scope like it does when executed in browsers because
-  // that version doesn't export it
-  // we use CommonJS version instead because it does (as the default property)
-  // see https://nodejs.org/en/knowledge/getting-started/what-is-require/ for more info
-  // also, we're not a browser so we need to manually attach domdiff to window
-  window.domdiff = require('domdiff/cjs').default;
+  window.GOVUK.vendor = {
+    DiffDOM: require('diff-dom').DiffDOM
+  };
   require('../../app/assets/javascripts/updateContent.js');
 
 });
--- a/.jshintignore
+++ b/.jshintignore
@@ -1,0 +1,1 @@
+app/assets/javascripts/modules/polyfills-for-diffdom.js
--- a/app/assets/javascripts/modules/all.mjs
+++ b/app/assets/javascripts/modules/all.mjs
@@ -6,6 +6,7 @@
 //
 // Exported items will be added to the window.GOVUK namespace.
 // For example, `export { Frontend }` will assign `Frontend` to `window.Frontend`
+import './polyfills-for-diffdom.js';
 import Header from 'govuk-frontend/components/header/header';
 import Details from 'govuk-frontend/components/details/details';
 import Button from 'govuk-frontend/components/button/button';
--- a/app/assets/javascripts/modules/all.mjs
+++ b/app/assets/javascripts/modules/all.mjs
@@ -10,6 +10,7 @@ import Header from 'govuk-frontend/components/header/header';
 import Details from 'govuk-frontend/components/details/details';
 import Button from 'govuk-frontend/components/button/button';
 import Radios from 'govuk-frontend/components/radios/radios';
+import {DiffDOM} from 'diff-dom';
 
 /**
  * TODO: Ideally this would be a NodeList.prototype.forEach polyfill
@@ -62,6 +63,11 @@ var Frontend = {
   "initAll": initAll
 }
 
+var vendor = {
+  "DiffDOM": DiffDOM
+}
+
 export {
-  Frontend
+  Frontend,
+  vendor
 }
--- a/app/assets/javascripts/modules/polyfills-for-diffdom.js
+++ b/app/assets/javascripts/modules/polyfills-for-diffdom.js
@@ -1,0 +1,1009 @@
+/* Polyfill service v3.109.0
+ * For detailed credits and licence information see https://github.com/financial-times/polyfill-service.
+ * 
+ * Features requested: Array.prototype.concat,Array.prototype.includes,Array.prototype.join,Array.prototype.slice,Array.prototype.sort,Array.prototype.splice,Function.prototype.name,JSON.Stringify,NodeList.prototype.forEach,Object.entries,Object.prototype.toString,RegExp.prototype.constructor,RegExp.prototype.exec,RegExp.prototype.sticky,RegExp.prototype.toString,String.prototype.match,String.prototype.replace,String.prototype.split,String.prototype.startsWith,String.prototype.trim
+ * 
+ * - _ESAbstract.Call, License: CC0 (required by "String.prototype.startsWith", "_ESAbstract.ToString", "_ESAbstract.ToPrimitive", "_ESAbstract.OrdinaryToPrimitive")
+ * - _ESAbstract.CreateMethodProperty, License: CC0 (required by "Object.entries", "_ESAbstract.EnumerableOwnProperties", "Object.keys")
+ * - _ESAbstract.Get, License: CC0 (required by "String.prototype.startsWith", "_ESAbstract.ToString", "_ESAbstract.ToPrimitive", "_ESAbstract.OrdinaryToPrimitive")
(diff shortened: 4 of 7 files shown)


tests/javascripts/updateContent.test.js
```diff
@@ -31,13 +31,9 @@ beforeAll(() => {
...
 
-  // using require to execute the version we use in our our frontend build here can't add
-  // the domdiff variable to this scope like it does when executed in browsers because
-  // that version doesn't export it
-  // we use CommonJS version instead because it does (as the default property)
-  // see https://nodejs.org/en/knowledge/getting-started/what-is-require/ for more info
-  // also, we're not a browser so we need to manually attach domdiff to window
-  window.domdiff = require('domdiff/cjs').default;
+  window.GOVUK.vendor = {
+    DiffDOM: require('diff-dom').DiffDOM
```
benthorner, line 35: As discussed, I found this line confusing because I noticed it's different from [the original](https://github.com/alphagov/notifications-admin/commit/87f54d1e886f5c45ef7f12d88ef988c3f9343641#diff-c256e69af3e4a832646bcc1331d4bb89cebb6259b1ef1955d26d35c408aa3ea6L36). I wasn't sure if this was due to the use of Babel, or an aspect of the newer version of diffDOM, or something else. Thanks for explaining it a bit in person. I think it would help someone else reviewing this to add some written expalantion and why the approach here has changed.

benthorner (commented):
Thanks for the detailed PR description @tombye! I found it very helpful 💯.

I am similarly nervous about the polyfill approach here: it's manual, complex and may potentially cause more problems than it solves because of this. We should consider other options and weigh them up.

Options we discussed, in rough order of preference:

## 1. Regress to an older version of diffDOM

This would fix all the issues we're aware of. But we'd be back in a place where we can't upgrade if we need to e.g. to fix a vulnerability. However, we now have this PR and all of the following to fall back on.

This is the least work and most understood option, hence it being the first choice. We should make sure we comment the pinned version thoroughly so someone can easily pick this up if they need to.

## 2. Apply this PR but with all of the polyfills

Adding ~60KB to our responses isn't great, but:

- In our discussion, you explained to me how this issue - packages randomly requiring modern JS - is a new and growing problem. This ~60KB isn't exclusive to diffDOM: it will help if we have the same issue with other packages as well.

- It's unclear how much we _actually_ care about 60KB. Notify nor GDS doesn't have a general strategy for response sizes that takes the above issue into account.

- Using comprehensive polyfills means we could standardise our code to consistently use  more modern JS. We'd have the opportunity to simplify our code and maybe, possibly offset some of the increased response size in doing so.

## 3. Disable AJAX code for all IE users

It's not critical functionality, but it's also not great to withdraw it. @tombye also said it may be hard to do this in practice, as IE11 doesn't support conditional comments.

IE11 users are still a significant proportion of our user base. We expect that to decline over time. GDS as a whole is still undecided on when we should drop IE / IEx.

However, this is still less work than the following, for a non-critical feature. We could potentially replace the AJAX feature with a visual prompt to refresh the page manually.

## 4. Rewrite our AJAX to avoid diffing

This would mean introducing new endpoints that return actual data / instructions instead of dumb HTML. We would then use this to manually update parts of the page.

It would be a lot of work to migrate to such new functionality - especially if we want to avoid current functionality breaking for users who haven't refreshed a page yet.

@tombye said he wants to talk to @quis to get some more thoughts and ideas about this or other approaches. @quis is one of the main authors of the current AJAX code.

## 5. This PR, as-is

See above re. complexity, hidden problems.

## 6. Stick with domDiff and its problems

So far we're only aware of one (accessibility) issue caused by it. We should exhaust all other options before we accept that.

## 6. Switching to the latest diffDOM

@tombye clarified this would break all our JS code for IE users, as an error in one part would prevent the rest of it from executing. This would be a significant, general degradation.

tests/javascripts/updateContent.test.js
tombye: This PR might not be merged but I added a comment to these changes (in https://github.com/alphagov/notifications-admin/pull/4134/commits/670032e054f04bcfe689fd434f72b19c7e9ae39b) to, hopefully, explain this better.

tombye:
Slightly embarassingly, [I did find an alternative library](https://github.com/alphagov/notifications-admin/pull/4143) so all this work is now not needed.

PR #4135, opened 2022-01-18
Display if broadcast was cancelled via API

CrystalPea:
If broadcast_message has no value under cancelled_by_id, display in the view that it was cancelled by API integration.

### DIFF
--- a/app/models/broadcast_message.py
+++ b/app/models/broadcast_message.py
@@ -181,7 +181,9 @@ class BroadcastMessage(JSONModel):
 
     @cached_property
     def cancelled_by(self):
-        return User.from_id(self.cancelled_by_id)
+        if not self.cancelled_by_id:
+            return "an API call"
+        return User.from_id(self.cancelled_by_id).name
 
     @cached_property
     def count_of_phones(self):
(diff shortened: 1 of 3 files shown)


app/models/broadcast_message.py
```diff
@@ -181,7 +181,9 @@ def approved_by(self):
 
     @cached_property
     def cancelled_by(self):
-        return User.from_id(self.cancelled_by_id)
+        if not self.cancelled_by_id:
+            return "API integration"
```
benthorner, line 185:
```suggestion
            return "an API call"
```

This is the terminology we use [elsewhere](https://github.com/alphagov/notifications-admin/blob/2635c80ca4659483cd01dd72d9da577ab5b653e8/app/templates/views/broadcast/view-message.html#L200). I prefer having the text as part of the method here so it's all in one place 👍.

benthorner (approved):
Looks good - just one comment about wording.

I also checked this works locally in tandem with the API PR:

Screenshot 2022-01-18 at 17 40 08

app/models/broadcast_message.py
CrystalPea: Good spot, I have amended this.

PR #4141, opened 2022-01-26
Fix current alerts focus styles

tombye:
Fixes for the focus style on a few alerts pages. Part of https://www.pivotaltracker.com/story/show/180011155, specifically:
- the focus style of alert list items on the current/past/rejected alerts pages
- the focus style for area list items on the alerts page for selecting sub-areas

## What caused the bugs

These items use a variant of the 'file-list' pattern, so like the uploads page but with slightly different HTML which caused the bugs.

The items on the current/past/rejected alerts pages wrapped their links in a `<h2>` tag (correctly), which broke the CSS we use to show the items underneath the link and give them enough colour contrast when focused.

The items on the select a sub-area page had no hint text underneath the link, which the CSS assumes is always there.

### Before

current_alerts_before

### After

current_alerts_after_with_js

Note: this now uses JS to apply the styles. Full details are in the related commit message.

### Before

large_link_no_hint_before

### After

large_link_no_hint_after

## Notes for reviewers

Details of the decisions behind the changes are in the commit messages so it's worth working commit-by-commit.

This can't be reviewed until issue https://github.com/alphagov/notifications-admin/pull/4134 tries to fix is resolved as it causes the current/past/rejected alerts items to be replaced every few seconds so it's almost impossible to test focus styles.


tombye:
The issue with the focus style on the current/past/rejected alerts page is a bit more complicated so I'm shifting it out from these changes to unblock them.

klssmith (approved):
I've tried this in a few browsers and can see how highlighting the selected the area now works as intended. I couldn't replicate the Safari issue with the template lists though - things look the same for me both before and after. I'm assuming that could be because of a different version we're using or something though, so I'm happy to approve anyway.

PR #4142, opened 2022-01-26
Make logo CDN domain into simple config

quis:
Having this as a function which does string parsing and manipulation surprised me a bit when I was trying to figure out why something wasn’t working.

It’s more in line with the way we do other config like this (for example `ASSET_PATH`) to make it a simple config variable, rather than trying to be clever and guess things based on other config variables.

It’s also less code, and is explicit enough that it doesn’t need tests.


tombye (approved):
I can't really comment on the history of why it was done this way but changing to something more standard seems sensible and the code looks good to me 👍🏻

PR #4144, opened 2022-01-28
Add postage changes

karlchillmaid:
This PR updates the pricing page to warn users about upcoming changes to postage rates.

### DIFF
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -250,7 +250,12 @@
   </div>
 
   <div class="panel panel-border-wide">  
-    <p class="govuk-body">These prices were updated on 1 February 2021.</p>
+    <p class="govuk-body">On 1 March 2022:</p>
+    <ul class="list list-bullet">
+    <li>second class postage will increase by 2 pence, plus VAT</li>
+    <li>international postage will increase by 7 pence, plus VAT</li>
+    <li>first class postage will decrease by 6 pence, plus VAT</li>
+  </ul>
   </div>
 
 {% endblock %}
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -252,9 +252,9 @@
   <div class="panel panel-border-wide">  
     <p class="govuk-body">On 1 March 2022:</p>
     <ul class="list list-bullet">
-    <li>second class postage will increase by 2 pence, plus VAT</li>
-    <li>international postage will increase by 7 pence, plus VAT</li>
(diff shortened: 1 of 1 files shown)


app/templates/views/pricing/index.html
```diff
@@ -250,7 +250,12 @@ <h2 class="heading-medium" id="letters">Letters</h2>
   </div>
 
   <div class="panel panel-border-wide">  
-    <p class="govuk-body">These prices were updated on 1 February 2021.</p>
+    <p class="govuk-body">On 1 March 2022:</p>
+    <ul class="list list-bullet">
+    <li>second class postage will increase by 2 pence, plus VAT</li>
+    <li>international postage will increase by 7 pence, plus VAT</li>
+    <li>first class postage will decrease by 6 pence, plus VAT</li>
```
quis, line 257: Do you think ’go up’ and ’go down’ might work better? I find it quite easy to lose ‘decrease’ after the two instances of ‘increase’. Or ‘2 pence more expensive’ and ‘6 pence cheaper’ maybe?

karlchillmaid: Yeah, we played with that in some of the drafts of the accompanying email.

I think I like go up/down more as it makes the `plus VAT` bit sound less awkward.

karlchillmaid, line 257:
```suggestion
    <li>second class postage will go up by 2 pence, plus VAT</li>
    <li>international postage will go up by 7 pence, plus VAT</li>
    <li>first class postage will go down by 6 pence, plus VAT</li>
```

app/templates/views/pricing/index.html
```diff
@@ -250,7 +250,12 @@ <h2 class="heading-medium" id="letters">Letters</h2>
   </div>
 
   <div class="panel panel-border-wide">  
-    <p class="govuk-body">These prices were updated on 1 February 2021.</p>
+    <p class="govuk-body">On 1 March 2022:</p>
+    <ul class="list list-bullet">
```
idavidmcdonald, line 254: A quick one before I head off but this may need to be `govuk-list govuk-list--bullet` as the classes

Apart from that it looks good I think but I'm not too familiar with the CSS styling.

quis: It’ll look the same either way. This is the old naming, your suggestion is the new GOV.UK Frontend naming. There’s loads of lists we’ll need to convert at some point, I wouldn’t worry about introducing one more instance of the old naming

quis (approved):
image

PR #4151, opened 2022-02-03
Tidy up new email branding routes

klssmith:
**Remove old branding URLs**
We've changed the URLs for these endpoints but needed to keep the old URLs in place in addition to the new URLs in case anyone was in the middle of changing their branding as the change was being deployed. We can now safely remove the old URLs.

**Make separate endpoints for GOV.UK email branding options**
The endpoint to change the email branding to "GOV.UK" branding and "GOV.UK and organisation" branding was the same but with a query string used to determine which of the two options had been selected. This makes them two separate endpoints, which makes the code a bit simpler and hopefully means there is less chance of things not working as expected.

[Pivotal story](https://www.pivotaltracker.com/story/show/180807160)

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1153,7 +1153,6 @@ def create_email_branding_zendesk_ticket(form_option_selected, detail=None):
     zendesk_client.send_ticket_to_zendesk(ticket)
 
 
-@main.route("/services/<uuid:service_id>/branding-request/email", methods=['GET', 'POST'])
 @main.route("/services/<uuid:service_id>/service-settings/email-branding", methods=['GET', 'POST'])
 @user_has_permissions('manage_service')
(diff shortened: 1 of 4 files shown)


app/main/views/service_settings.py
```diff
@@ -1169,19 +1168,10 @@ def email_branding_request(service_id):
...
-                'govuk_and_org': '.email_branding_govuk',
-                'nhs': '.email_branding_nhs',
-                'organisation': '.email_branding_organisation',
-                'something_else': '.email_branding_something_else',
-            }[form.options.data]
-
             return redirect(
                 url_for(
-                    endpoint,
+                    f'.email_branding_{form.options.data}',
```
quis, line 1173: 👨🏻‍🍳 💋

quis (approved):

PR #4154, opened 2022-02-09
Proof of concept for redis producing dashboard stats

idavidmcdonald:
https://www.pivotaltracker.com/story/show/179338377

image

This PR is a proof of concept/discussion starter for how to use redis to give our dashboard page its stats. I've implemented a very basic version of one of our options and have created the equivalent bit needed on the API in https://github.com/alphagov/notifications-api/pull/3450.

Have a read, have a browse of the code and then please let me know your thoughts!

## Option 1 - a redis key that doesn't include a day

We could store redis keys of the following

`service-id-xxxxxxxx-email-sending`
`service-id-xxxxxxxx-sms-failed`

These keys would never expire.

Everytime an email is created, we add one to the count of `service-id-xxxxxxxx-email-sending`
Everytime an email is delivered, we add one to the count of `service-id-xxxxxxxx-email-delivered` and subtract one from `service-id-xxxxxxxx-email-sending`

We would need a nightly job of some sort to remove any notifications at the end of 7 days from the count.

The admin app can go and grab these keys and display them directly on the dashboard page.

## Option 2 - a redis key, per day, that counts the number of messages in the past 7 days

`service-id-xxxxxxxx-date-2022-02-09-email-sending` which would represent the number of messages that are in sending that have been created in the last 7 days

These keys would expire after about 8 days.

Everytime an email is created, we add one to the count of:

`service-id-xxxxxxxx-date-2022-02-09-email-sending`
`service-id-xxxxxxxx-date-2022-02-10-email-sending`
`service-id-xxxxxxxx-date-2022-02-11-email-sending`
...
`service-id-xxxxxxxx-date-2022-02-17-email-sending`

Everytime an email is delivered, we add one to the count of 
`service-id-xxxxxxxx-date-2022-02-09-email-delivered`
`service-id-xxxxxxxx-date-2022-02-10-email-delivered`
`service-id-xxxxxxxx-date-2022-02-11-email-delivered`
...
`service-id-xxxxxxxx-date-2022-02-17-email-delivered`

and subtract one from:

`service-id-xxxxxxxx-date-2022-02-09-email-sending`
`service-id-xxxxxxxx-date-2022-02-10-email-sending`
`service-id-xxxxxxxx-date-2022-02-11-email-sending`
...
`service-id-xxxxxxxx-date-2022-02-17-email-sending`

The admin app can grab the key for todays date and display it directly on the dashboard page. (it would have to do the same for each of the other notification types and for each of the other notification statuses).

## Option 3 - store each days count by itself and get the admin app to sum up the last 7 days worth

We could store redis keys of the following

`service-id-xxxxxxxx-date-2022-02-09-email-sending`
`service-id-xxxxxxxx-date-2022-02-09-sms-failed`

These keys would expire after about 8 days.

Everytime an email is created, we add one to the count of `service-id-xxxxxxxx-date-2022-02-09-email-sending`
Everytime an email is delivered, we add one to the count of `service-id-xxxxxxxx-date-2022-02-09-email-delivered` and subtract one from `service-id-xxxxxxxx-date-2022-02-09-email-sending`

The admin app would then ask for the past 7 days by summing 
`service-id-xxxxxxxx-date-2022-02-02-email-sending`
`service-id-xxxxxxxx-date-2022-02-03-email-sending`
...
`service-id-xxxxxxxx-date-2022-02-09-email-sending`

## Option 4 - use redis to support our current API route and the admin app stays as is

We could store redis keys of the following

`service-id-xxxxxxxx-date-2022-02-09-email-sending`
`service-id-xxxxxxxx-date-2022-02-09-sms-failed`

These keys would expire after 24 hours.

Everytime an email is created, we add one to the count of `service-id-xxxxxxxx-date-2022-02-09-email-sending`
Everytime an email is delivered, we add one to the count of `service-id-xxxxxxxx-date-2022-02-09-email-delivered` and subtract one from `service-id-xxxxxxxx-date-2022-02-09-email-sending`

When the API comes to get the data for how many notifications have been sent in the last 7 days, it gets the previous days data from the `ft_notification_status` table and then gets todays data from todays key `service-id-xxxxxxxx-date-2022-02-09-email-sending`.

## Things I haven't dealt with here

The following will need to be solved but I don't see them as particularly difficult or noteworthy compared to picking the larger strategy from above

BST/UTC - will be some nuance around making sure notifications are marked against the correct days redis key. 

Which statuses to have keys for. We could split our keys by just 'sending', 'delivered', 'failed' which would match up exactly to what we show in the admin app. Or we could split them by every status we have (ie splitting failed in to tech failure, perm failure, temp failure)

Template stats. The API route that is slow also gives us stats on how many times a template has been used in the past 7 days. We would need to take care of this either by making the suggested keys more granular (adding in template ids) or by creating different redis keys just to count template usage. 

Making the code be correct and pretty. I've scraped some stuff together, more to illustrate. It would need to be rewritten but should roughly demonstrate what option 3 would look like. It works locally for me :) Everytime I sent an email, the number on the dashboard goes up by 1!

Storing the data in simple keys like this or storing it in something like a redis hash.

## Thoughts on which option is best

I think option 4 is not the best option. It doesn't also solve the problem of the dashboard not being in realtime (bug card https://www.pivotaltracker.com/story/show/172026169), which I think we should try and solve as part of this. You technically could solve it as notifications enter their final state after 72 hours, so by getting the most recent 72 hours of data from redis and the older days from `ft_notification_status` but at this point, you might as well be doing one of the other options.

I think option 2 is not the best option, it is a more complex and confusing version of option 3, with not that much more benefit.

That leaves option 1 or option 3.

I think I prefer option 3 because 

- option 3 is easier to inspect and debug. You can match daily tallys. If there are any problems, you can always track it down to a single day causing the issue.
- option 1 may be harder to debug. If something goes wrong and our count accidentally goes wrong, it won't get fixed over time. For example, if our nightly job failed and didn't remove 300 delivered emails from the count of delivered emails, then those 300 delivered emails would stay counted and would never be removed. It might be hard to figure out what is going wrong with a number that is constantly moving.
- option 3 will closer match the `ft_notification_status` table which gives some nicer consistency

Therefore I have implemented a rough sketch of what option 3 could look like in this PR and the related API PR.

## One final note

This is very different usage of redis to all our other usage. Our current usage of redis is as a cache and redis should always hold the same value as a value in the database.

The exception to this is the rate limits, which count how many notifications someone has sent in the past minute and in the past day. However, these are not explicitly shown to the user. And if something went wrong with them, it generally wouldn't be the end of the world if we lost the data for a day. 

This new use case is different because if there is a problem with redis, lets say redis went down for 2 hours, then our dashboards would just say 0 or break. Equally, we can't just solve problems with purging redis. If we delete these keys, we lose our lovely tallys of how many notifications have been sent and we don't get those back (unless we add some sort of command/job that would repopulate these keys for every service for every notification type based on the database).

## What I would l like from you please

- Comment and critique on the above options
- Which option/options do you think we should go for?
- Any other solutions I haven't considered

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -69,7 +69,7 @@ class Config(object):
     ANTIVIRUS_ENABLED = True
 
     REDIS_URL = os.environ.get('REDIS_URL')
-    REDIS_ENABLED = os.environ.get('REDIS_ENABLED') == '1'
+    REDIS_ENABLED = True
 
     ASSET_DOMAIN = ''
     ASSET_PATH = '/static/'
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -1,6 +1,6 @@
 import calendar
 from collections import namedtuple
-from datetime import datetime
+from datetime import datetime, timedelta
 from functools import partial
 from itertools import groupby
 from operator import itemgetter
@@ -24,6 +24,7 @@ from app import (
     service_api_client,
     template_statistics_client,
 )
+from app.extensions import redis_client
 from app.formatters import format_date_numeric, format_datetime_numeric
 from app.main import main
 from app.statistics_utils import get_formatted_percentage
@@ -280,12 +281,48 @@ def aggregate_notifications_stats(template_statistics):
     return notifications
 
 
+def notification_count_cache_key(
+    service_id, created_at_utc, notification_type, simplified_status
+):
+    # this is not actually turning it in to bst, will need to fix this
+    bst_date = created_at_utc.strftime("%Y-%m-%d")
+    x = "service-{}-bst-date-{}-{}-{}".format(
+            str(service_id), bst_date, notification_type, simplified_status
+        )
+    return x
+
+def get_last_7_days_total(service_id, notification_type, simplified_status):
+    total = 0
+    for i in range(7):
+        total += int(
+            redis_client.get(
+                notification_count_cache_key(
+                    service_id, datetime.utcnow() - timedelta(days=i) , notification_type, simplified_status
+                )
+            ) or 0)
+
+    return total
+
+
+def get_dashboard_stats_from_redis(service_id):
+    results = {}
+    for notification_type in ['email', 'text', 'letter']:
+        results[notification_type] = {}
+        for status in ['sending', 'delivered', 'failed']:
+            results[notification_type][status] = get_last_7_days_total(
+                service_id, notification_type, status
+            )
+        results[notification_type]['requested'] = sum(results[notification_type].values())
+
+    return results
+
+
 def get_dashboard_partials(service_id):
     all_statistics = template_statistics_client.get_template_statistics_for_service(service_id, limit_days=7)
     template_statistics = aggregate_template_usage(all_statistics)
     stats = aggregate_notifications_stats(all_statistics)
 
-    dashboard_totals = get_dashboard_totals(stats),
+    dashboard_totals = get_dashboard_totals(get_dashboard_stats_from_redis(service_id)),
     free_sms_allowance = billing_api_client.get_free_sms_fragment_limit_for_year(
         current_service.id,
         get_current_financial_year(),
--- a/app/templates/views/dashboard/_totals.html
+++ b/app/templates/views/dashboard/_totals.html
@@ -16,11 +16,11 @@
     </div>
     <div id="total-sms" class="govuk-grid-column-one-third">
       {{ big_number_with_status(
-        statistics['sms']['requested'],
-        statistics['sms']['requested']|message_count_label('sms', suffix='sent'),
-        statistics['sms']['failed'],
-        statistics['sms']['failed_percentage'],
-        statistics['sms']['show_warning'],
+        statistics['text']['requested'],
+        statistics['text']['requested']|message_count_label('sms', suffix='sent'),
+        statistics['text']['failed'],
+        statistics['text']['failed_percentage'],
+        statistics['text']['show_warning'],
         failure_link=url_for(".view_notifications", service_id=service_id, message_type='sms', status='failed'),
         link=url_for(".view_notifications", service_id=service_id, message_type='sms', status='sending,delivered,failed'),
         smaller=True,


rjbaker:
Good writeup! Agree that option three seems the best bet.

I feel this might be a good use of Redis' Hash data type. Hashes support storage key-value data in an object using a single key. Something a bit like:

```
service:some-long-uuid:2022-02-09:delivery-stats
  email-sending: 1
  email-delivered: 10
  sms-sending: 10
  sms-delivered: 20
  ...
``` 
We can use the [HINCRBY](https://redis.io/commands/hincrby) command to increment a value stored at a given key in the hash.

`HINCRBY service:some-long-uuid:2022-02-09:delivery-stats email-sending 1`

This way you can grab all the service's stats for a given day via a single key:

`HGETALL service:some-long-uuid:2022-02-09:delivery-stats`

Wwith [pipelining](https://redis.io/topics/pipelining), we can grab the keys for each of the 7 days in a single operation.

Notify utils Redis client appears to use the redis-py client via FlaskRedis which [supports multi command pipelining](https://github.com/redis/redis-py#pipelines). Using this we can do:

```python
# psuedocode
r.pipe()
r.hgetall('service:some-long-uuid:2022-02-09:delivery-stats')
r.hgetall('service:some-long-uuid:2022-02-08:delivery-stats')
r.hgetall('service:some-long-uuid:2022-02-07:delivery-stats')
...
r.execute()
```

This will return a list of dictionaries which we can reduce to a single dictionary containing the summed count keys.

We can also use pipelines to _atomically_ increment & decrement different keys in the hash in a transactional command using [MULTI](https://redis.io/commands/multi).

```python
# psuedocode
r.pipe()
r.multi()
r.hincrby('service:some-long-uuid:2022-02-09:delivery-stats', 'email-sending', -1)
r.hincrby('service:some-long-uuid:2022-02-09:delivery-stats', 'email-delivered', 1)
r.execute()
```

A minor point on the Redis key format used in your example. Consider using colons to separate the key elements and hyphens for multi-word key elements. This is the [convention generally recommended by Redis](https://redis.io/topics/data-types-intro#redis-keys) and makes is easier to understand keys once there are a few different types and lots of denormalised data floating about. e.g. `service:some-long-uuid:2022-02-09:delivery-stats`

A possible thought about UTC/BST. Consider that we store most dates etc in UTC and convert to the user's local timezone. This might cause issues when BST offsets are applied throughout the year, as well possible users in other timezones. We could store rolling hourly counts of each metric, and when calculating a total in BST (or some other timezone), offset the "hour" by one (or some other offset), so that the count for the correct 24 hour period is displayed.

If you've not read it already, there's a really good chapter on [counters and stats in the Redis In Action site/ebook](https://redis.com/ebook/part-2-core-concepts/chapter-5-using-redis-for-application-support/5-2-counters-and-statistics/5-2-1-storing-counters-in-redis/).

rjbaker:
For template stats we could use a [sorted set](https://redis.io/topics/data-types-intro#redis-sorted-sets), not for the sorting ability but for the ability to increment scores of the members in the set and then use some of the useful sorted set commands.

e.g. given the key
`service:some-long-uuid:2022-02-09:template-stats`

We can increment scores for each template as they're used throughout a given day:
```
ZINCRBY service:some-long-uuid:2022-02-09:template-counts 1 some-template-uuid
ZINCRBY service:some-long-uuid:2022-02-09:template-counts 1 some-template-uuid
ZINCRBY service:some-long-uuid:2022-02-09:template-counts 1 some-other-template-uuid
```

Get a single day's totals for the templates using [ZRANGE](https://redis.io/commands/zrange) to get all the set members with scores:

```
ZRANGE service:some-long-uuid:2022-02-09:template-counts 0 -1 WITHSCORES
[some-template-uuid: 2, some-other-template-uuid: 1]
```

We can use a [ZUNION](https://redis.io/commands/zunion) to sum the scores of elements stored in different sorted sets.

E.g. given the sorted sets:
```
service:some-long-uuid:2022-02-09:template-counts
   some-template-uuid: 2
   some-other-uuid: 4

service:some-long-uuid:2022-02-08:template-counts
  some-template-uuid: 1
  some-other-uuid: 5
  
ZUNION 2 service:some-long-uuid:2022-02-09:template-counts service:some-long-uuid:2022-02-08:template-counts WITHSCORES AGGREGATE SUM

[ some-template-uuid: 3, some-other-uuid: 9]
```

benthorner:
Lots to think about here. I'm pleased we're still pursuing this (rather epic!) bug card.

When you first mentioned Redis my immediate thought was Option 4, which is still my preference. By the way, I noticed we can also benefit from these changes on [the monthly statistics dashboard](https://github.com/alphagov/notifications-api/blob/9ce6d2fe924626c578cdb77467d8df0e842e9821/app/service/rest.py#L590). Lots of potential here!

Let me structure my thoughts a bit:

- Option 4 addresses the main problem: performance of querying the Notifications table. It's also a precursor to Option 3 - we could still do Option 3 after doing Option 4, but we'll understand the decision better by then.

- Fixing the realtime bug is tangential (worth remembering). Another way to fix it would be to update Fact..Status immediately for late receipts. The volume of these is low, so it's a practical option performance-wise. 

- Option 3 also concerns me as it brings the purpose of Fact..Status into question (in my head, at least). What was it meant to be used for? How does Option 3 fit into our broader data architecture?

- I think we should keep the current queries as a fallback, as they work for the majority of services. This means we maintain our current level of service if Redis goes down. What are your thoughts on having a fallback?

- The extra Redis calls are in performance critical parts of the code (sending, receipts). Although I expect them to be quick, we should consider doing some performance testing so we're clear on the impact.

- We should ideally have a way to refill the cache if it's faulty, especially if we go with Option 3. A pragmatic solution would be to not bother, and have a marker that says Redis data is unavailable for that day.

Keen to hear what others think too...

leohemsted:
some thoughts.

* i think code should live in API, not admin. since API will be doing all the updating of these values it feels more self contained and means its easier to see code what is used/not used.
* i also like 4, and bens point about 4 leading in to 3 is great - we can deploy 4 to solve the performance issue while working on a widespread overhaul of 3. the bug is not great but i think the priority should be to fix performance while ensuring we're not closing off solutions to that bug, rather than to fix both at once as the mvp. 
* performance: unlike ben im not worried about redis performance. we investigated it before in summer 2020 and it's really negligible, plus will happen in async tasks.
* question: does "sending" there include created, pending virus scan, etc? (itll also need pending)
* question: does failed there include failed virus scan, failed validation, etc?
* question: should we do one redis entry per db status or just "sending", "delivered" and "failed"
* data loss: im not too worried about redis losing data, in practice that doesnt happen very often (once in 4 years?). that said it would be prudent to make sure we consider it. a fallback to the notification table and/or a task we can manually kick off to rebuild those stats from the db would be good.

quis:
I’d also think about an approach which could support [making the usage information on the dashboard real time](https://www.pivotaltracker.com/story/show/172026169) which at the moment we don’t do because it’s too expensive to compute.

klssmith:
I was trying to work out how this would work for delivery receipts coming back on a different day to the day when the notification was created - whether we'd update Redis or not (for option 4 if we get a receipt for an old notification) and which day we'd update (for option 3). For options 1 and 2, the value in Redis is just for the full 7 days, so this wouldn't complicate anything.

For option 3, would that mean that every time a notification is delivered you'd need to check what day it was created so that you'd know which day to adjust the key for? If we're updating Redis at the same point at which we update the notification, then I guess we wouldn't need to make an extra database request to find the created_at datetime. Unless the intention is that we just that record how many delivery receipts arrived each day, but then we'd need to make sure we don't count duplicate receipts or status changes we don't care about.

I think it would be good to have a way to repopulate Redis, not because of the risk that it might have issues but more because we might bulk change data. Although it's rare, there are cases when we've done things such as cancelled 1000s of letters a service has wrongly sent out, so it would be nice for the dashboard to show that change. At the moment, the dashboard would only show the wrong thing for a day, but if all the data comes from Redis it could show the wrong thing for 7 days unless we have a way of updating the data there.

idavidmcdonald:
Going to close this because I haven't had the time to work on it properly. Can be used as a starter for ideas when we start tackling this again at some point this quarter.

PR #4157, opened 2022-02-16
Show link to service history page

benthorner:
This makes the feature more visible to platform admins, although a
service manager can technically view the page as well.

I toyed with putting the link at the bottom of the settings page,
but this was a bit easier than untangling the "if" blocks there.

## Screenshot

Screenshot 2022-02-16 at 13 05 28

### DIFF
--- a/app/navigation.py
+++ b/app/navigation.py
@@ -282,6 +282,9 @@ class MainNavigation(Navigation):
             'guest_list',
             'old_guest_list',
         },
+        'history': {
+            'history'
+        }
     }
 
 
--- a/app/templates/main_nav.html
+++ b/app/templates/main_nav.html
@@ -32,6 +32,9 @@
     <li><a class="govuk-link govuk-link--no-visited-state{{ main_navigation.is_selected('usage') }}" href="{{ url_for('.usage', service_id=current_service.id) }}">Usage</a></li>
     <li><a class="govuk-link govuk-link--no-visited-state{{ main_navigation.is_selected('team-members') }}" href="{{ url_for('.manage_users', service_id=current_service.id) }}">Team members</a></li>
   {% endif %}
+  {% if current_user.platform_admin %}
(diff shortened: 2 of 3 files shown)


quis:
Can we try untangling the `if` blocks instead? In the past I’ve tried to avoid changes that make the UI that platform admins see different to the UI that service teams see. With this change someone on support might tell a user to click the _Service history_ link in the left nav, assuming the user can also see it.

benthorner:
Does that reduce the risk, though? I can see the problem: there's nothing to indicate who the feature is available to. But how would that change if we put it on the settings page instead? The "delete" link at the bottom is available to users as well, so it's still at least slightly ambiguous who can do what.

This feature is actually available to service admins, so we could go the other way and expose the link to them as well. This may be where I'm missing some context: the feature is available to them, but its current state suggests it was never "shipped". Did we have to abandon it or something?

Possible actions:

- Remove access to service admins to clarify feature status (double check no one's using it first).
- Move link as separate item below the delete / suspend ones on the settings page; or
- Move link as inline item with delete / suspend ones (styles seem to prevent this though); or
- Leave it as-is (risk does seem relatively small in this case).

quis:
There are two places I can think of where we are ambiguous:
- delete link (as you pointed out)
- priority queue setting on templates

These are exceptions to the rule that everything platform admin for a service should be at the bottom of the settings page. I’d prefer us to stick to this rule for new things we add.

I think the reason this was never fully shipped is because we haven’t done the research to understand whether the design works for users and where they would expect to find it.

benthorner:
Fair enough: thanks for explaining the research situation.

I guess it doesn't do any harm to let service admins continue to use it (if they've found it), but otherwise it sounds like we're viewing this as a longterm Platform Admin only feature. With that in mind, I've done a different PR to put the link on the settings page like you suggested: https://github.com/alphagov/notifications-admin/pull/4198.

PR #4158, opened 2022-02-18
Fix 500 error due to inconsistent recipient check

benthorner:
This strengthens the initial check of what's in the session to make
sure it contains some kind of recipient. Without this, we get:

      File "/home/vcap/deps/0/python/lib/python3.9/site-packages/flask/app.py", line 1950, in full_dispatch_request
        rv = self.dispatch_request()
      File "/home/vcap/deps/0/python/lib/python3.9/site-packages/flask/app.py", line 1936, in dispatch_request
        return self.view_functions[rule.endpoint](**req.view_args)
      File "/home/vcap/app/app/utils/user.py", line 26, in wrap_func
        return func(*args, **kwargs)
      File "/home/vcap/app/app/main/views/send.py", line 1041, in send_notification
        recipient=session['recipient'] or InsensitiveDict(session['placeholders'])['address line 1'],
      File "/home/vcap/deps/0/python/lib/python3.9/site-packages/notifications_utils/insensitive_dict.py", line 41, in __getitem__
        return super().__getitem__(self.make_key(key))
    KeyError: 'addressline1'

I'm not sure how to reproduce this, but this should at least give
the user a better experience, instead of a 500 page.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -1025,7 +1025,9 @@ def get_template_error_dict(exception):
 @main.route("/services/<uuid:service_id>/template/<uuid:template_id>/notification/check", methods=['POST'])
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def send_notification(service_id, template_id):
-    if {'recipient', 'placeholders'} - set(session.keys()):
+    recipient = get_recipient()
+
+    if not recipient:
         return redirect(url_for(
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):
Nice one 👏🏼 . Less un-actionable alerts for us, too :D

PR #4159, opened 2022-02-18
Catch error if letter does not exist on send

benthorner:
This repeats the pattern we already have for previewing a letter,
where we assume the error is because the notification has already
been sent and redirect the user to see it.

I've improved the original pattern a bit:

- I've DRYed-up the low-level boto code and moved the error handler
there so it can be reused.

- I've introduced a custom exception, which the calling code can
choose to log.

- I've introduced the moto library, which we use elsewhere, to make
it easier to test S3 code.

I've used an error level log when sending a notification - now that
we have a more descriptive log, we can verify the assumption is true
and then make an informed decision to downgrade the log.

In future we may want to merge this handler with the similar code
in utils [1], but we'll need to be careful as the utils handler is
superficial - it doesn't check the reason for the error.

[1]: https://github.com/alphagov/notifications-utils/blob/bce0f4e596af9451212d0026c8ecc27b636eed38/notifications_utils/s3.py#L52

### DIFF
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -7,7 +7,6 @@ from functools import partial
 from io import BytesIO
 from zipfile import BadZipFile
 
-from botocore.exceptions import ClientError
 from flask import (
     abort,
     current_app,
@@ -39,6 +38,7 @@ from app.main import main
 from app.main.forms import CsvUploadForm, LetterUploadPostageForm, PDFUploadForm
 from app.models.contact_list import ContactList
 from app.s3_client.s3_letter_upload_client import (
+    LetterNotFoundError,
     backup_original_letter_to_s3,
     get_letter_metadata,
     get_letter_pdf_and_metadata,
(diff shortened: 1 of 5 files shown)


app/main/views/uploads.py
```diff
@@ -352,7 +351,18 @@ def send_uploaded_letter(service_id, file_id):
...
+        metadata = get_letter_metadata(service_id, file_id)
+    except LetterNotFoundError as e:
+        current_app.logger.error(e)
+
+        # if the file's not there, it's probably because we've already created the notification and the letter has been
+        # moved to the normal letters-pdf bucket. So lets just bounce out to the notification page
+        return redirect(url_for(
+            '.view_notification',
+            service_id=service_id,
+            notification_id=file_id,
```
CrystalPea, line 364: file_id will match the notification id? :o

benthorner: Good question: [yes it will](https://github.com/alphagov/notifications-api/blob/master/app/service/send_notification.py#L189) (but it's quite hard to figure it out!).

CrystalPea: Ahh, good find :o. Do we want to add a little inline comment so it's less mysterious? Something like `# when we persist uploaded letter, we set notification id to qual file id, so this will match` ?

benthorner: Sure, I've rewritten the comment here: https://github.com/alphagov/notifications-admin/pull/4159/commits/3effb75045b84e940c70b4e0fbfe353fe87263c6.

benthorner:
Just a note about steps to reproduce this:

- Upload a letter.
- Duplicate the tab.
- Try to send in both tabs.

However, it's still unclear if this is what's triggering the errors in Production (see PR description).

## Screenshots (for the second tab)

| Before | After |
| - | - |
| Screenshot 2022-02-24 at 11 04 16 | Screenshot 2022-02-24 at 11 04 24 |

PR #4160, opened 2022-02-21
Make it easy to clear cache for all key formats

benthorner:
https://www.pivotaltracker.com/story/show/180663519

This adds a new option to the form and tweaks the flash confirmation
so it's more closely aligned with the action performed.

Please review commit by commit.

## Screenshots

| Before | After |
| - | - |
| Screenshot 2022-02-21 at 11 46 27 | Screenshot 2022-02-21 at 14 15 45 |


quis:
Is there stuff which ‘All’ doesn’t include, for example rate limits?

It might be better making it the last option and labelling as ‘All of the above’ or similar.

app/main/views/platform_admin.py
```diff
@@ -420,17 +420,26 @@ def clear_cache():
...
             redis_client.delete_cache_keys_by_pattern(pattern)
-            for pattern in CACHE_KEYS[to_delete]
+            for pattern in patterns
         )
-        msg = 'Removed {} {} object{} from redis'
-        flash(msg.format(num_deleted, to_delete, 's' if num_deleted != 1 else ''), category='default')
+
+        msg = f'Removed {num_deleted} objects ' \
+              f'across {len(patterns)} key formats ' \
+              f'for {", ".join(group_keys)}'
```
quis, line 440: I’d prefer we didn’t use backslashes because:
- it makes the amount of indentation arbitrary, which is unusual for Python
- renaming the variable means you need to change the indentation of the subsequent lines, which makes for less-clean diffs

```suggestion
        msg = (
            f'Removed {num_deleted} objects ' 
            f'across {len(patterns)} key formats ' 
            f'for {", ".join(group_keys)}'
        )
```

quis (approved):
Checkboxes are definitely the way to go. One optional comment on code style.

app/main/views/platform_admin.py
benthorner: Nice! I'd seen the first technique elsewhere but also prefer your suggestion.

I've amended the last commit with it 👍.

PR #4162, opened 2022-02-22
Allow template changes if Redis is down

benthorner:
https://www.pivotaltracker.com/story/show/180663519

This updates the version of -utils, which provides this
new behaviour. See the screenshots for the main effect.

I've also included some extra commits to tidy up how we
manage the cache in service_api_client.py. The last commit
is an optional suggestion - let me know what you think.

## Before

Users can't change templates - inconsistent.

Screenshot 2022-02-22 at 12 39 53

## After

This is consistent with the behaviour for all other cached items e.g. services, broadcasts.

| 1. Added "redis offline" to template | 2. Refresh after Redis returns | 3. Refresh after cache clear |
| - | - | - |
| Screenshot 2022-02-22 at 12 35 03 | Screenshot 2022-02-22 at 12 35 13 | Screenshot 2022-02-22 at 12 36 34

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -352,10 +352,10 @@ def platform_admin_returned_letters():
 
         try:
             letter_jobs_client.submit_returned_letters(references)
-            redis_client.delete_cache_keys_by_pattern(
+            redis_client.delete_by_pattern(
                 'service-????????-????-????-????-????????????-returned-letters-statistics'
             )
-            redis_client.delete_cache_keys_by_pattern(
+            redis_client.delete_by_pattern(
                 'service-????????-????-????-????-????????????-returned-letters-summary'
             )
         except HTTPError as error:
@@ -431,7 +431,7 @@ def clear_cache():
         patterns = list(itertools.chain(*groups))
 
         num_deleted = sum(
-            redis_client.delete_cache_keys_by_pattern(pattern)
+            redis_client.delete_by_pattern(pattern)
             for pattern in patterns
         )
 
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -216,8 +216,9 @@ class ServiceAPIClient(NotifyAdminAPIClient):
             })
         data = _attach_current_user(data)
         endpoint = "/service/{0}/template/{1}".format(service_id, id_)
+        ret = self.post(endpoint, data)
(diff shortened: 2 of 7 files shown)


benthorner:
Note that this doesn't actually cause the wrong content to be sent out:

| Email before failure | Email during failure (insert "redis down") | Email after failure (correct) |
| - | - | - |
| Screenshot 2022-02-23 at 16 04 02 | Screenshot 2022-02-23 at 16 04 10 | Screenshot 2022-02-23 at 16 04 15 |

rjbaker (approved):
Nice work. Probably needs a review by someone more familiar with this area, but having a data source in the event of unavailable cache is always a good thing 👍

idavidmcdonald (commented):
Looks good.

I think for your last commit,  you should move this in to utils so it can be reused and it also means we can have consistency with calling `cache.delete` and `cache.delete_by_pattern`.

I am a little bit cautious of this change, in particular around the change of behaviour for utils meaning that if Redis goes down, someone changes a template, redis goes up, and then sends a message using the API, we will send out stale messages. A trade off between consistency and showing users errors. We would also have to get fairly unlucky to fall in to the above example, especially given we generally see little issue with Redis and its availability. I think it is probably OK though.

If you are happy to add the decorator to a new version of utils, pull it in to this PR and let me know and I can approve.

idavidmcdonald:
Oh thanks for writing good commit messages by the way. Stellar work!

benthorner:
Thanks @richardc0 for taking a look and @idavidmcdonald for the praise ☺️.

I've raised a PR to replace the last commit: https://github.com/alphagov/notifications-utils/pull/950

> I am a little bit cautious of this change

Yes, it's a case of the least worst option. Personally I think it would be better to error for everything, but at least we can now make that decision uniformly when we're analysing a new outage - when we'll have clearer idea of what we want. We could even go further and expose [the `raise_exception` parameters](https://github.com/alphagov/notifications-utils/blob/7e2add5fb9b915fffa4a850e83e178713e5fa523/notifications_utils/clients/redis/redis_client.py#L65) if we want to deviate on a per-method basis - then it would still be clear what the behaviour was, without having to go digging into multiple layers of utils.

benthorner:
@idavidmcdonald I'll merge this now as it's been approved and the change you asked for in -utils is now incorporated into the final commit. Let me know if there's anything more you'd like done here though.

PR #4166, opened 2022-02-23
Let users on email auth delete their mobile numbers

CrystalPea:
Sometimes users ask us to delete their mobile numbers for them.

If those users are on email auth, they should be able to delete their number themselves.

This will save them writing a support ticket and save us going into the database.

Here is the flow:

Screenshot 2022-02-24 at 09 25 16

Screenshot 2022-02-24 at 09 25 23

Screenshot 2022-02-24 at 09 28 04

### DIFF
--- a/app/templates/views/user-profile/change.html
+++ b/app/templates/views/user-profile/change.html
@@ -20,7 +20,7 @@
     <div class="govuk-grid-column-three-quarters">
       {% call form_wrapper() %}
         {{ form_field(error_message_with_html=True) }}
-        {% if current_user.auth_type == 'email_auth' %}
+        {% if current_user.auth_type == 'email_auth' and (current_user.mobile_number and thing == "mobile number") %}
           {{ page_footer(
             'Save',
             delete_link=url_for(
--- a/tests/app/main/views/test_user_profile.py
+++ b/tests/app/main/views/test_user_profile.py
@@ -12,6 +12,8 @@ from app.models.webauthn_credential import (
 )
 from tests.conftest import (
     create_api_user_active,
+    create_user,
+    fake_uuid,
     normalize_spaces,
     url_for_endpoint_with_token,
 )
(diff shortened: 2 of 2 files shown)


app/templates/views/user-profile/change.html
```diff
@@ -20,9 +20,20 @@
     <div class="govuk-grid-column-three-quarters">
       {% call form_wrapper() %}
         {{ form_field(error_message_with_html=True) }}
-        {{ page_footer('Save') }}
+        {% if current_user.auth_type == 'email_auth' %}
```
benthorner, line 23:
```suggestion
        {% if current_user.auth_type != 'sms_auth' %}
```

This is inconsistent with the conditional in the endpoint.

CrystalPea: You are right it is inconsistent, I would change it the other way round - as people who use webauthn should still have their phone numbers as a backup for if they lose the key we downgrade them to sms auth if I remember the process around this correctly.

benthorner (commented):
Fair enough 👍.

I tried this out locally and noticed we still show "Delete your number":

- After I've deleted my number. 
- On the form to change my email.

Tweaking the conditional should fix the first one. We can also switch to using a dedicated template for the mobile number form so it won't leak into the other one.

CrystalPea:
@benthorner 

> I tried this out locally and noticed we still show "Delete your number":
>
> After I've deleted my number.
> On the form to change my email.
> Tweaking the conditional should fix the first one. We can also switch to using a dedicated template for the mobile number form so it won't leak into the other one.

Nice find 👏🏼 . I fixed it with a slightly clunky conditional, do you think it is ok?

tests/app/main/views/test_user_profile.py
```diff
@@ -183,6 +183,59 @@ def test_should_show_mobile_number_page(
 ):
     page = client_request.get(('main.user_profile_mobile_number'))
     assert 'Change your mobile number' in page.text
+    assert 'Delete your number' not in page.text
+
+
+def test_change_your_mobile_number_page_shows_delete_link_if_user_on_email_auth(
```
benthorner, line 189: Should we have a test to cover the other side of the conditional (not on email auth or doesn't have a mobile number setup)?

benthorner (approved):
Works fine locally for me now ✅.

I think the clunky conditional is OK - if anything else gets added in future then we should consider having a dedicated partial instead.

tests/app/main/views/test_user_profile.py
CrystalPea: That's a good idea, I will do that

PR #4167, opened 2022-02-24
Allow platform admins to change user auth in the UI

CrystalPea:
So we do not have to go into the db when we need to change user auth

Part of this story: https://www.pivotaltracker.com/story/show/180597596

Screenshot 2022-02-24 at 10 15 51

Screenshot 2022-02-24 at 10 16 02
auth.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1015,6 +1015,16 @@ def filter_by_broadcast_permissions(valuelist):
         return [entry for entry in valuelist if any(entry in option for option in broadcast_permission_options)]
 
 
+class AuthTypeForm(StripWhitespaceForm):
+    auth_type = GovukRadiosField(
+        'Sign in using',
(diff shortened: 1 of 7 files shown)


idavidmcdonald:
This will be helpful!

Just a quick comment from me, I haven't looked at the code properly.

How does this work with a user using webauthn? What do they see on the page? Will it allow you to change it to something that isn't webauthn etc?

CrystalPea:
@idavidmcdonald that's a good question. For now, to be on the safe side, I opted for raising 403 when you try to change auth type if the user uses webauthn, to minimise security risk. I will also hide the button in that case 🤔 .

We can always revisit that later, or if you think it's a bad decision let me know.

leohemsted (approved):
This looks pretty good - just one small suggestion on user model usage

PR #4168, opened 2022-02-25
Allow SMS fragment limit to be 0

klssmith:
We want to be able to set the free allowance for a service to 0, but the form was not allowing this - it gave an error message of `Cannot be empty`. This can be fixed by changing the WTForms validator from `DataRequired` (which coerces 0 to falsey) to the `InputRequired` validator.

Should be merged after https://github.com/alphagov/notifications-api/pull/3469

[Pivotal story](https://www.pivotaltracker.com/story/show/181282523)


benthorner (approved):
Works for me locally ✅ .

Only small issue is that the edit form doesn't replay a value of "0" if I go to change it again - normally it shows the current free SMS allowance in the input field.

klssmith:
I've had a look at why `0` isn't appearing and it's caused by how `govuk-frontend` renders the field, so not something I can change easily here unfortunately.

PR #4169, opened 2022-02-25
Downgrade error log for expected exception

benthorner:
Previously we weren't sure if the cause of this exception was what
the comment below suggested [1]. I've now verified this from:

        Letter not found for service 0bd1d970-f11c-40e1-8319-4baefe6239d7 and file aa07ed06-3161-4795-93b3-b45d7c576af9

I checked that aa07ed06-3161-4795-93b3-b45d7c576af9 exists already
as a notification i.e. the comment is correct and we're not sending
users to a 404 page. It's possible there are other scenarios where
the comment is wrong, but I don't think it's worth keeping the error.

[1]: https://github.com/alphagov/notifications-admin/pull/4159

### DIFF
--- a/app/main/views/uploads.py
+++ b/app/main/views/uploads.py
@@ -270,7 +270,7 @@ def uploaded_letter_preview(service_id, file_id):
     except LetterNotFoundError as e:
         current_app.logger.warning(e)
 
-        # If the file is missing it could be because this is a duplicate
+        # If the file is missing it's likely because this is a duplicate
         # request, the notification already exists and the file has been
         # moved to a different bucket. Note that the ID of a precompiled
         # notification is always set to the file_id.
@@ -356,9 +356,9 @@ def send_uploaded_letter(service_id, file_id):
     try:
         metadata = get_letter_metadata(service_id, file_id)
     except LetterNotFoundError as e:
-        current_app.logger.error(e)
+        current_app.logger.warning(e)
 
-        # If the file is missing it could be because this is a duplicate
+        # If the file is missing it's likely because this is a duplicate
         # request, the notification already exists and the file has been
         # moved to a different bucket. Note that the ID of a precompiled
         # notification is always set to the file_id.


app/s3_client/s3_letter_upload_client.py
```diff
@@ -81,7 +81,9 @@ def get_letter_s3_object(service_id, file_id):
         return s3.Object(current_app.config['TRANSIENT_UPLOADED_LETTERS'], file_location).get()
     except botocore.exceptions.ClientError as e:
         if e.response['Error']['Code'] == 'NoSuchKey':
-            raise LetterNotFoundError(f'Letter not found for service {service_id} and file {file_id}')
+            message = f'Letter not found for service {service_id} and file {file_id}'
+            current_app.logger.warning(message)
```
leohemsted, line 85: Loosely held opinion, but in general i prefer to have exception logs higher up - ie: we should only log the exception when we're actively suppressing the exception's bubble-up. That way, the function calling can decide whether it's important if the exception is thrown or not. It also means that when calling code, we don't need to worry "are any potential exceptions already logged?" - as either we will write the handling code ourselves or let it bubble up further to app level handlers where things will get logged.

For example, hypothetically we might decide it's a warning if `uploaded_letter_preview` fails but an error if `send_uploaded_letter`.

so in summary, i'd rather we kept this file as it is and kept the log lines in uploads.py, but just downgraded them to `warning` instead of `error`.

Thoughts?

benthorner: That's a fair point. I was attempting to DRY-up the log but it doesn't save much and I agree with your logic - logic I normally follow myself but for wanting to de-duplicate this bit of code. On reflection, it's the whole block of code that would benefit from a refactor, but that's out-of-scope (and it's only repeated once so arguably not an issue).

I've amended the commit to simply downgrade the second instance of the log.

leohemsted (approved):

PR #4171, opened 2022-02-27
Remove grid from alerts file-list

tombye:
The alert listings on the current/past/rejected alerts pages need these parts to be clickable:
- the link
- the description text (underneath the link)
- the status text (underneath the link)

alert_grid

The last 2 are positioned in a grid which stops them being clickable. This removes it in favour using flexbox to position the elements in a similar way.

Part of this story: https://www.pivotaltracker.com/story/show/180011155

## Use of flexbox and fallbacks

Flexbox is a fairly complex layout system and Notify doesn't use it much so worth explaining these changes.

Setting `display: flex` on an element makes any direct descendents 'flex items', meaning flexbox styles are added to change their size and position. We're only using those that control position here, specifically the `justify-content: space-between` setting to arrange their children like so:

flexbox-justify-content-space-between-visual

(Image 'borrowed' from https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

In our component, this pushes both bits of text to the outer edges. Their width is set using `calc` to make them take up half the space each but with a 30px gap between. This is intended to mimic the layout they had before by being in a 2-column govuk grid.

Note: these changes are only meant to replicate the current layout using flexbox. We could use flexbox rules to redesign how the elements adapt to different context sizes but that's out of scope for these changes.

### Fallbacks

These changes include fallback styles for browsers that don't support flexbox (IE<11). They work by making both bits of text display as inline content (like parts of text on the same line) and setting the alignment of the line to the 'justify' value. This spreads them out similar to our flexbox styles.

Note: 'display' is overriden on child items under flexbox so setting display: inline-block is ignored.

## Testing these changes

I've tested these changes on IE9-11, as well as Chrome, Edge, Safari and Firefox.

### DIFF
--- a/app/assets/stylesheets/views/dashboard.scss
+++ b/app/assets/stylesheets/views/dashboard.scss
@@ -63,11 +63,13 @@
       // and making children inline mimic `justify-content: space-between`
       display: inline-block;
 
-      // simulate govuk-grid-column-one-half spacing - smaller screens
+      // This simulates a 50% column in a govuk grid on smaller screens
+      // govuk grid columns go to 100% width on smaller screens
       width: 100%;
-      flex-grow: 1;
 
-      // simulate govuk-grid-column-one-half spacing - larger screens
+      // This simulates a 50% column in a govuk grid on larger screens
+      // as with govuk grid, this includes a gap between columns to ensure the contents
+      // are separated by a space
       @include govuk-media-query($from: tablet) {
         max-width: calc(50% - #{$govuk-gutter-half});
       }
@@ -76,10 +78,6 @@
     & > .file-list-status {
       overflow: hidden; // old IE hack to make it vertically line up with the hint
       margin-bottom: 0; // cancel margin-bottom from .govuk-hint class
-
-      @include govuk-media-query($from: tablet) {
-        margin-bottom: 10px; // match margin-bottom on hint
-      }
     }
 
     & > .file-list-hint-large {
--- a/app/assets/stylesheets/views/dashboard.scss
+++ b/app/assets/stylesheets/views/dashboard.scss
@@ -44,6 +44,49 @@
 
 .file-list {
 
+  // for file-lists with section-like content and a single item
+  &--sectioned {
+    display: flex;
+    flex-wrap: wrap;
+    text-align: justify; // fallback for browsers that don't support flexbox
+    justify-content: space-between;
+
+    // note: first-child of a section should be a heading
+    & > :first-child,
+    & > .area-list {
+      width: 100%;
+    }
+
+    & > .file-list-hint-large,
+    & > .file-list-status {
+      // fallback for browsers that don't support flexbox - let `text-align: justify` on parent
+      // and making children inline mimic `justify-content: space-between`
+      display: inline-block;
+
+      // simulate govuk-grid-column-one-half spacing - smaller screens
+      width: 100%;
+      flex-grow: 1;
+
+      // simulate govuk-grid-column-one-half spacing - larger screens
+      @include govuk-media-query($from: tablet) {
+        width: calc(50% - #{$govuk-gutter-half});
+      }
+    }
+
+    & > .file-list-status {
+      overflow: hidden; // old IE hack to make it vertically line up with the hint
+      margin-bottom: 0; // cancel margin-bottom from .govuk-hint class
+
+      @include govuk-media-query($from: tablet) {
+        margin-bottom: 10px; // match margin-bottom on hint
+      }
+    }
+
+    & > .file-list-hint-large {
+      text-align: left; // reset for text-align fallback on parent
+    }
+  }
+
   &-hint,
   &-hint-large,
   &-status {
@@ -136,10 +179,10 @@
 
 // Note: this selector should use the :has() pseudo-class in the future when it is supported
 //       which will remove the need for JS
-.file-list .js-child-has-focus + .govuk-grid-row {
+.file-list .js-child-has-focus {
 
-  & .file-list-hint-large,
-  & .file-list-status > .govuk-hint {
+  & + .file-list-hint-large,
+  & ~ .file-list-status {
     color: $govuk-focus-text-colour;
   }
 
--- a/app/assets/stylesheets/views/dashboard.scss
+++ b/app/assets/stylesheets/views/dashboard.scss
@@ -69,7 +69,7 @@
 
       // simulate govuk-grid-column-one-half spacing - larger screens
       @include govuk-media-query($from: tablet) {
-        width: calc(50% - #{$govuk-gutter-half});
+        max-width: calc(50% - #{$govuk-gutter-half});
       }
     }
 
--- a/app/templates/views/broadcast/partials/dashboard-table.html
+++ b/app/templates/views/broadcast/partials/dashboard-table.html
@@ -3,36 +3,30 @@
 <div class="ajax-block-container js-mark-focus-on-parent" data-classes-to-persist="js-child-has-focus">
 {% for item in broadcasts|sort|reverse|list %}
   <div class="keyline-block">
-    <div class="file-list govuk-!-margin-bottom-2">
+    <div class="file-list file-list--sectioned govuk-!-margin-bottom-2">
       <h2>
         <a class="file-list-filename-large govuk-link govuk-link--no-visited-state" href="{{ url_for(view_broadcast_endpoint, service_id=current_service.id, broadcast_message_id=item.id) }}">{{ item.template.name }}</a>
       </h2>
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-one-half">
-          <span class="file-list-hint-large govuk-!-margin-bottom-2">
-            {{ item.content }}
-          </span>
-        </div>
-        <div class="govuk-grid-column-one-half file-list-status">
-          {% if item.status == 'pending-approval' %}
-            <p class="govuk-body govuk-!-margin-bottom-0 govuk-hint">
-              Waiting for approval
-            </p>
-          {% elif item.status == 'rejected' %}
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/views/dashboard.scss
```diff
@@ -44,6 +44,49 @@
...
+
+    // note: first-child of a section should be a heading
+    & > :first-child,
+    & > .area-list {
+      width: 100%;
+    }
+
+    & > .file-list-hint-large,
+    & > .file-list-status {
+      // fallback for browsers that don't support flexbox - let `text-align: justify` on parent
```
benthorner, line 62: Would be helpful to distinguish which of these attributes are related to the fallback approach. As far as I can see this is visually equivalent to what we had before, but if we remove `width: 100%` in future then we'll have some changes.

| Better wrapping on small screens | `margin-bottom: 0` causes overlap with the focus area |
| - | - |
| Screenshot 2022-03-07 at 15 06 02 | Screenshot 2022-03-07 at 15 06 11 |

For reference, the date not overlapping the focus area is unchanged:

Screenshot 2022-03-07 at 15 06 38

app/assets/stylesheets/views/dashboard.scss
```diff
@@ -44,6 +44,49 @@
...
+      flex-grow: 1;
+
+      // simulate govuk-grid-column-one-half spacing - larger screens
+      @include govuk-media-query($from: tablet) {
+        width: calc(50% - #{$govuk-gutter-half});
+      }
+    }
+
+    & > .file-list-status {
+      overflow: hidden; // old IE hack to make it vertically line up with the hint
```
benthorner, line 77: These bits are puzzling me: I tried them in IE9 and the text areas still aren't clickable; some of the styles also appear to be overridden (crossed out in the inspector).

Just to avoid me going down a rabbit hole: can you confirm the behaviour is meant to be the same in IE9, or are we targeting higher versions?

benthorner (commented):
Nice one getting the old and the new to play nicely 🏅. I'm surprised it works together so well, considering they're two completely different sets of attributes!

I got stuck on the IE aspect of the fix though - left a comment about that and one or two other things to check I'm understanding all this fully.

A note for others reading this: when I first came to this PR I was confused why we would have two approaches (flexbox + IE hack) when the IE hack should work in all cases. @tombye explained that flexbox is the conventional way to achieve this, so we should go with that in the first instance and remove the IE hack whenever we drop IE.

app/assets/stylesheets/views/dashboard.scss
tombye: Only line 64 is directly related to the fallback, to make the child elements render inline and so be positioned according to the parent's `text-align` value.

I'm glad you mention the potential benefits this approach could give smaller screens. One of the reasons I think we should move more UI to flexbox is its better adaptability. The only problem is that the design does need to change from being based on certain ranges (640px or less, 640px and up) to being rules-based (how much space each child is allowed to occupy + min-max dimensions + what content it contains) so these changes are intended to just mimic the current layout but allow for that change in design in future by moving to flexbox.

app/assets/stylesheets/views/dashboard.scss
tombye: I couldn't get the text to be clickable in IE9. We do support IE9 but as 'functional', meaning it should work without errors. I think that means the text doesn't need to be clickable in IE9 but do say if you think otherwise. IE9 doesn't really appear in the last year's analytics and I do plan to dump it from our support matrix once I've actually done the research for this.

Also: I'd guess that the styles being overridden might be due to `overflow: hidden` giving the text hasLayout, which is an old IE-specific state that gets triggered by certain styles and changes how things are laid out. This isn't well documented anymore and I'm not 100% sure tbh,

app/assets/stylesheets/views/dashboard.scss
benthorner: Hmm...from my own experiments I found a couple of other styles that are at least have some IE-specific behaviour, which I've left some comments about.

This one is clear though 👍.

app/assets/stylesheets/views/dashboard.scss
benthorner: Ahh, I understand now: the goal for IE9 is just to preserve the layout, given that we're changing the elements used and their structure.

I've had another look at these attributes:

- `overflow` does behave as expected after all, it's just IE9 dev tools are unintuitive. I confirmed it fixes the right-hand text being aligned vertically.

- `text-align: left` and `margin-bottom: 10px` have no effect for me in Chrome or IE9, irrespective of the window size. Are they necessary?

benthorner (commented):
Phew, thanks for chatting about this offline. I've made some suggested comments to embed what I've learned - trying to think what would have helped me when I first looked at this.

[There are two styles that I still don't understand](https://github.com/alphagov/notifications-admin/pull/4171/files#r820809472), so I'd like to defer approving in case there are more changes to come - nearly there though!

app/assets/stylesheets/views/dashboard.scss
tombye: Digging into the `text-align: left` and `margin-bottom: 10px`:

Both the description and status need to be 10px above the area list on larger screens. We only need `margin-bottom: 10px` on the description if the status text never goes to 2 lines (because it wouldn't have a margin so wouldn't push the content below down). I added it as a precaution, do you think we can just assume this won't happen?

The `text-align: left` is more complex. Because the description is forced onto 1 line, by having `whitespace: nowrap`, it isn't affected by `text-align` because its text doesn't have to be arranged to fit inside the containing space. So we could do without it but that might raise the question of why it should inherit `text-align: justify` from its parent to anyone reading the code. I'm just guessing though, and that was my intention by including it, what do you think?

app/assets/stylesheets/views/dashboard.scss
```diff
@@ -44,6 +44,49 @@
...
+    }
+
+    & > .file-list-hint-large,
+    & > .file-list-status {
+      // fallback for browsers that don't support flexbox - let `text-align: justify` on parent
+      // and making children inline mimic `justify-content: space-between`
+      display: inline-block;
+
+      // simulate govuk-grid-column-one-half spacing - smaller screens
+      width: 100%;
```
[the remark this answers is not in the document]
tombye: It does have a bit of an effect on the flexbox approach. This was unclear because I explained the flexbox approach incorrectly.

I said that an element becoming a flex item meant its set dimensions (via height/width) were overridden by those assigned by flexbox. This isn't strictly true. If the set dimensions of all flex items fill the container, they won't grow.

Our non-flexbox styles:
- tell these flex items grow to fill the container (via `width: 100%`)
- prevent them from growing past `max-width: calc(50% - $govuk-gutter-half)`

So, in other words, `width: 100%` does the same as `flex-grow: 1`, in this case at least, so:
- it does have some effect on the way we're using flexbox
- we don't actually need `flex-grow: 1`

Sorry about all this. I've added the changes suggested here to a new commit https://github.com/alphagov/notifications-admin/pull/4171/commits/a82a67aba361e003a2f3efc3cbcd42445b9185ae which also includes revised comments based on your advice here and in other comments.

app/assets/stylesheets/views/dashboard.scss
```diff
@@ -44,6 +44,49 @@
...
+      flex-grow: 1;
+
+      // simulate govuk-grid-column-one-half spacing - larger screens
+      @include govuk-media-query($from: tablet) {
+        max-width: calc(50% - #{$govuk-gutter-half});
```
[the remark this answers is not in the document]
tombye: I've changed the comments in https://github.com/alphagov/notifications-admin/pull/4171/commits/a82a67aba361e003a2f3efc3cbcd42445b9185ae, along with other changes. Hope it does roughly what you describe here but say if not.

app/assets/stylesheets/views/dashboard.scss
benthorner: > I added it as a precaution but it wasn't there before

Ah, I understand now. I'd be open to keeping it with a comment to explain that, but removing it is reasonable so let's leave it at that.

> So we could do without it but that might raise the question of why it should inherit

I'd totally forgotten about CSS inheritance. What you say makes sense - worth a comment to clarify perhaps, but I'll leave that up to you.

app/assets/stylesheets/views/dashboard.scss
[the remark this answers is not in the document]
benthorner: Yup, that looks great 👍.

app/assets/stylesheets/views/dashboard.scss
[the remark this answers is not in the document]
benthorner: No problem - it's a toughie!

I wonder if it's worth adding a comment somewhere explaining why we're using flexbox the way we are - it's quite subtle that we're only using it for alignment because:

- Make it easier to support the sizing aspect in IE.
- To emulate the previous grid behaviour (arguably arbitrary so subtle).
  - This includes an (arguably unnecessary as we control both) gap in case of text filling both columns.

All the existing comments allude to these reasons but don't explain the _why_ explicitly.  Another option might be to revise the PR description slightly.

This is an optional suggestion, though - the important thing is I now get it 🎉.

benthorner (approved):
Yay, we're done, finally 😆.

While my in-depth review did uncover a bug, it wasn't a big one, so thanks for being patient and helping me (and also you 😉) understand the goals and techniques at play here.

I've left two parting comments where you may want to add a bit more documentation, just considering how surprisingly hard this stuff was to understand.

app/assets/stylesheets/views/dashboard.scss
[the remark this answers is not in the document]
tombye: I see you point and I've rewritten the description to make it clearer that the changes simulate the grid and that we're only using flexbox for position.

PR #4173, opened 2022-02-28
Refactor sort mixin to work on attributes other than `name`

quis:
`SortByNameMixin` is working well for classes where we want to sort objects by name. But it’s limited to the `name` attribute.

This commit makes it more generic, so that subclass-ees can choose which attribute to sort on. For example this means that users can be sorted by email address without having to do a custom call to `sorted` with an icky `lambda`.

This could also let us implement enhancements ascending/descending sorting in the future, for example.

### DIFF
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -8,22 +8,10 @@ from rtreelib import Rect
 from werkzeug.utils import cached_property
 
 from app.formatters import square_metres_to_square_miles
-from app.models import SortByStringAttributeMixin
+from app.models import IdEqualityMixin, SortByStringAttributeMixin
 
 from .populations import CITY_OF_LONDON
 from .repo import BroadcastAreasRepository, rtree_index
-
-
-class IdEqualityMixin:
-
-    def __repr__(self):
-        return f'{self.__class__.__name__}(<{self.id}>)'
-
-    def __eq__(self, other):
-        return self.id == other.id
-
-    def __hash__(self):
-        return hash(self.id)
 
 
 class GetItemByIdMixin:
--- a/app/broadcast_areas/models.py
+++ b/app/broadcast_areas/models.py
@@ -8,7 +8,7 @@ from rtreelib import Rect
 from werkzeug.utils import cached_property
 
 from app.formatters import square_metres_to_square_miles
-from app.models import SortByNameMixin
+from app.models import SortByStringAttributeMixin
 
 from .populations import CITY_OF_LONDON
 from .repo import BroadcastAreasRepository, rtree_index
@@ -75,7 +75,9 @@ class BaseBroadcastArea(ABC):
         return max(500, min(estimated_bleed, 5000))
 
 
-class BroadcastArea(BaseBroadcastArea, IdEqualityMixin, SortByNameMixin):
+class BroadcastArea(BaseBroadcastArea, IdEqualityMixin, SortByStringAttributeMixin):
+
+    __sort_attribute__ = 'name'
 
     def __init__(self, row):
         self.id, self.name, self._count_of_phones, self.library_id = row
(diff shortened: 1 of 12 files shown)


app/models/service.py
```diff
@@ -181,10 +183,7 @@ def active_users(self):
 
     @cached_property
     def team_members(self):
-        return sorted(
```
benthorner, line 184: You could also make a similar change in [organisation.py](https://github.com/alphagov/notifications-admin/blob/3033c29c93ff99d4a62616ad5acff9254691f988/app/models/organisation.py#L159) / [the view](https://github.com/alphagov/notifications-admin/blob/3033c29c93ff99d4a62616ad5acff9254691f988/app/templates/views/organisations/organisation/users/index.html#L33).

benthorner (commented):
Overall approach seems good to me 👍.

Just a couple of small things to tie up before approving.

samuelhwilliams:
@quis does this still need to go in or can we close this off?

app/models/service.py
quis: Fixed and squashed into b5db8f0bc16b92f78ba4f92412c07b8ef483499a

app/models/user.py
```diff
@@ -32,7 +32,11 @@ def _get_org_id_from_view_args():
     return str(request.view_args.get('org_id', '')) or None
 
 
-class User(JSONModel, UserMixin, SortByStringAttributeMixin):
+class BaseUser(JSONModel, SortByStringAttributeMixin):
```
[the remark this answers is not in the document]
quis: Having separate `Mixin`s for different attributes is the thing I’m trying to get away from

app/broadcast_areas/models.py
```diff
@@ -75,9 +77,7 @@ def estimated_bleed_in_m(self):
         return max(500, min(estimated_bleed, 5000))
 
 
-class BroadcastArea(BaseBroadcastArea, IdEqualityMixin, SortByStringAttributeMixin):
-
-    __sort_attribute__ = 'name'
```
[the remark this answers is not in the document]
quis: Think this was referring to a refactoring which didn’t actually work, have dropped the commit now

quis:
@samuelhwilliams I‘ve rebased and addressed Ben’s comments – mind taking a look?

app/models/__init__.py
```diff
@@ -62,7 +62,16 @@ def __init__(self, *args, page=None, **kwargs):
...
 
-class SortByNameMixin():
+class SortByStringAttributeMixin(ABC):
+
+    @property
+    @abstractmethod
+    def __sort_attribute__(self):
+        pass
 
     def __lt__(self, other):
```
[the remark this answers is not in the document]
quis: I’ve done some more refactoring (b810d40, 75258b8, f69d05c) to get to a point where we can use total ordering in 5cc323ea83ce9e5986fd12073f60348eda15d876

app/models/__init__.py
```diff
@@ -7,7 +8,32 @@
...
+            getattr(self, self.__sort_attribute__).lower()
+        ) < (
+            getattr(other, self.__sort_attribute__).lower()
+        )
+
+    def __eq__(self, other):
+        return self.id == other.id
+
+    def __hash__(self):
+        return hash(self.id)
```
samuelhwilliams, line 33: Do we need to update these to use `self.__sort_attribute__`?

app/models/__init__.py
```diff
@@ -7,7 +8,32 @@
...
+    @abstractmethod
+    def __sort_attribute__(self):
+        pass
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}(<{self.id}>)'
+
+    def __lt__(self, other):
+        return (
+            getattr(self, self.__sort_attribute__).lower()
```
samuelhwilliams, line 24: This makes it so that the sort attribute has to be a string. We can probably accept that, but it's not super clear that this is the case.

app/models/__init__.py
```diff
@@ -7,7 +8,32 @@
 )
 
 
-class JSONModel(SerialisedModel):
+@total_ordering
+class SortingAndEqualityMixin(ABC):
+
+    @property
+    @abstractmethod
+    def __sort_attribute__(self):
```
quis, line 16:
```suggestion
    def __sort_attribute__(self):
        """
        Subclasses that want sorting to work must set this property to the
        string name of the attribute on which the instances should be
        sorted. For example 'email_address' or 'created_at' to sort on
        instance.email_address or instance.created_at respectively.
        """
```

PR #4175, opened 2022-03-01
Prevent scrolling of branding previews

quis:
In user research we quite often saw people accidentally scroll the branding preview `<iframe>` when trying to scroll the page.

This is suboptimal because they:
- were confused why the page wasn’t scrolling
- lost visibility of the branding they were trying to preview because it scrolled outside the bounds of the `<iframe>`

### DIFF
--- a/app/assets/javascripts/previewPane.js
+++ b/app/assets/javascripts/previewPane.js
@@ -13,7 +13,7 @@
   const $paneWrapper = $('<div class="govuk-grid-column-full"></div>');
   const $form = $('form');
   const previewType = $form.data('previewType');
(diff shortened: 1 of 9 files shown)


quis:
@klssmith good spot – I’ve added the same change to `previewPane.js` and squashed it into b32cd7c

klssmith (approved):
Looks good to me. We might be able to delete some of the [preview email content](https://github.com/alphagov/notifications-admin/blob/cc876194d05c19a68a87bdbed818a954da3bfb91/app/main/views/index.py#L100) now if we won't ever display all of it, but it's also not doing any harm to keep it

PR #4176, opened 2022-03-01
Replay falsey values in input fields

benthorner:
Currently an integer of 0 doesn't get shown because it fails the
truthiness check in the govuk-frontend template [^1]. Note that
we can't just do str(value) as for None this would be "None".

I struggled to find a place to add a test for this, so one of the 
commits does some refactoring of existing tests for code in 
forms.py to make that easier here and in future.

## Screenshots

| Before | After |
| - | - |
| Screenshot 2022-03-01 at 13 59 16 | Screenshot 2022-03-01 at 13 59 06 |

[^1]: https://github.com/alphagov/govuk-frontend/blob/fd4952f1c02c79056e09fc9c9ebae5c078140e49/src/govuk/components/input/template.njk#L51


benthorner:
TODO: add a test for this somewhere.

benthorner:
@tombye do you have any thoughts on the approach here?

tombye (approved):
Seems like a very sensible approach to me 👍🏻

tombye:
Ah, sorry, I approved this before you finished it. I agree a test would be good too.

benthorner:
I've added a new commit to fix a gnarly behaviour this change has churned up: https://github.com/alphagov/notifications-admin/pull/4176/commits/21452649fdace2ba16778e3213ff59b09e8e4527.

That took a while so I'll take a break from this - just needs a test and then it's ready to 🚢.

benthorner:
Added the extra test - this is now ready for review.

tombye:
I think the new test structure is nice and matches the conventions elsewhere in `tests`. I was a little thrown by the moving of some tests in `test_validators.py` but after thinking it over, they stand out due to using a form in their tests so there seems a good argument for moving them to the test file for that form, like you did. If we need to test more form-agnostic scenarios, they can always be moved back as function tests, like the rest.

The only thing that I'd question is keeping `test_validators.py` outside of `tests/app/main/forms`. The validators follow the WTForms convention so should go with the forms, no?

benthorner:
Let me clarify: we want to have a single pattern for how we organise our tests. Having multiple or unclear patterns just means we have to make more and pointless decisions when we add new tests.

The dominant (and definitely my preferred) pattern around our repos is that the file a test goes in should have the same name as the file of the top-level code its testing. The tests should also start with the name of the function (or class if there are multiple classes) under test. This makes it clear what's being tested and where it is.

Example: [test_generate_previous_next_dict_adds_other_url_args](https://github.com/alphagov/notifications-admin/blob/a555ac8d076c7e3f6d95597ff3ac5225e6ea5dda/tests/app/utils/test_pagination.py#L18) in `test_pagination.py` is indeed testing `generate_previous_next_dict`, which is defined in `pagination.py` (with the same directory structure).

Where we should have some flexibility is if there are _lots_ of tests for lots of code, as is the case here. What we should really do is split up `forms.py` into a package, but that would be a big change for this PR. Equally, I don't want to go backwards by cramming all the existing tests into a single `test_forms.py` file, as that'll make it harder for someone to split up `forms.py` in future (and keep the tests consistent with the split).

> The only thing that I'd question is keeping test_validators.py outside of tests/app/main/forms

Hopefully this is now clearer: `test_validators.py` contains tests for functions defined in `validators.py`, which is located in `app/main` (not `app/main/forms(.py)`).

Does that make more sense?

tombye:
> Hopefully this is now clearer: test_validators.py contains tests for functions defined in validators.py, which is located in app/main (not app/main/forms(.py)).

> Does that make more sense?

Yes, now you put it like that, it makes total sense. No blockers now, will approve.

tombye (approved):
All comments addressed so looks good to me 👍🏻

PR #4174, opened 2022-03-01
Update letter prices

karlchillmaid:
This PR updates the:
* postage pricing table
* last updated information

To make the pricing page consistent with the roadmap, privacy notice and accessibility statement, we’re moving the ‘last updated’ information to the top of the page.

PR #4177, opened 2022-03-02
Update price of letters on home page

quis:
Follows https://github.com/alphagov/notifications-admin/pull/4174


klssmith:
We have the old prices on a couple of other pages too - https://www.notifications.service.gov.uk/features/letters and when you toggle letter sending on and off on the settings page

idavidmcdonald:
Katie you are so thorough, thanks for doing a better job than me!

quis:
@klssmith thanks – I’ve added commits for those two other instances.

I did a `cmd` + `f` and couldn’t find any other mentions of 39 or 1.21

PR #4180, opened 2022-03-02
Link directly to email branding "something else"

benthorner:
https://www.pivotaltracker.com/story/show/181415991

Previously we duplicated the "something else" email branding form
on its own page and embedded in the choices form (if it was the
only option). See [^1] for how this looks - it's inconsistent.

This DRYs-up the "something else" form by bypassing the choices
form when "something else" is the only option. I've also tweaked
the "Back" button to be consistent with this behaviour.

Making this change also simplifies the choices form, which we'll
be adding pool options to shortly. I'd like to make the letters
form consistent, but let's see how emails pan out first.

Note that the choices form will now show a single radio button if
"something else" is the only option. I think that's OK as nothing
will link to the page, and the form still works.

## Screenshots

Note that the point of this PR is to make the forms consistent, favouring the old / basic version without a preview. We're not sure if the version with a preview makes sense to users and we expect to make further changes to it anyway as part of the next stage of the branding work, so for now the priority is just to simplify.

### Multiple options

After clicking "Change" in service settings, then "Something else".

| Before | After (unchanged) |
| - | - |
| Screenshot 2022-03-02 at 15 45 28 | Screenshot 2022-03-02 at 15 45 28 |

### Just "something else"

| Before | After |
| - | - |
| Screenshot 2022-03-02 at 15 47 40 | Screenshot 2022-03-02 at 15 45 28 |

[^1]: https://github.com/alphagov/notifications-admin/pull/4163#issuecomment-1050088088

### DIFF
--- a/tests/app/main/views/test_service_settings.py
+++ b/tests/app/main/views/test_service_settings.py
@@ -5424,7 +5424,7 @@ def test_get_email_branding_govuk_and_nhs_pages(
     assert normalize_spaces(page.select_one('.page-footer button').text) == 'Use this branding'
 
 
-def test_get_email_branding_something_else_page(client_request, service_one, mocker):
+def test_get_email_branding_something_else_page(client_request, service_one):
     # expect to have a "NHS" option as well as the
     # fallback, so back button goes to choices page
     service_one['organisation_type'] = 'nhs_central'
@@ -5441,7 +5441,7 @@ def test_get_email_branding_something_else_page(client_request, service_one, moc
     )
 
 
-def test_get_email_branding_something_else_page_is_only_option(client_request, mocker, service_one):
+def test_get_email_branding_something_else_page_is_only_option(client_request, service_one):
     # should only have a "something else" option
     # so back button goes back to settings page
     service_one['organisation_type'] = 'other'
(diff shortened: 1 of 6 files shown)


tests/app/main/views/test_service_settings.py
```diff
@@ -5509,14 +5424,35 @@ def test_get_email_branding_govuk_and_nhs_pages(
     assert normalize_spaces(page.select_one('.page-footer button').text) == 'Use this branding'
 
 
-def test_get_email_branding_something_else_page(client_request):
+def test_get_email_branding_something_else_page(client_request, service_one, mocker):
```
klssmith, line 5427: Really minor, but the `mocker` fixture isn't need here or in other two new tests that were added

tests/app/main/views/test_service_settings.py
```diff
@@ -5802,6 +5736,37 @@ def test_service_settings_links_to_branding_request_page_for_letters(
...
+    mocker,
+    service_one,
+    client_request,
+    no_reply_to_email_addresses,
+    no_letter_contact_blocks,
+    single_sms_sender,
+    single_branding_option,
+    expected_href,
+):
+    service_one['permissions'].append('letter')
```
klssmith, line 5753: This doesn't seem necessary

tests/app/main/views/test_service_settings.py
benthorner: Ah thank you: minimal fixtures is definitely better.

Fixed in https://github.com/alphagov/notifications-admin/pull/4180/commits/c22d3412fad4b74490f85365317dbe0777480bd6.

klssmith (approved):

PR #4183, opened 2022-03-03
Report for total notifications sent per day for each channel.

servingUpAces:
Daily volumes report: total volumes across the platform aggregated by whole business day (bst_date)
Volumes by service report: total volumes per service aggregated by the date range given.

NB: start and end dates are inclusive

- [ ] https://github.com/alphagov/notifications-api/pull/3477

### DIFF
--- a/app/templates/views/platform-admin/daily-volumes-report.html
+++ b/app/templates/views/platform-admin/daily-volumes-report.html
@@ -31,11 +31,11 @@
       {% for message_length, charge in [
         ('day', 'The whole business day in BST.'),
         ('sms totals', 'The number of text messages sent'),
+        ('sms fragments', 'The number of text message fragments sent times the rate multiplier'),
         ('sms chargeable units', 'The number of text message fragments sent'),
-        ('sms_fragments_times_multiplier', 'The number of text message fragments sent times the rate multiplier'),
         ('email totals', 'The number of emails sent'),
         ('letter totals', 'The number of letters sent'),
-        ('letter sheet totals', The number of sheets sent)
+        ('letter sheet totals', 'The number of sheets sent')
       ] %}
         {% call row() %}
           {{ text_field(message_length) }}
--- a/app/templates/views/platform-admin/daily-volumes-report.html
+++ b/app/templates/views/platform-admin/daily-volumes-report.html
@@ -1,0 +1,48 @@
+{% extends "views/platform-admin/_base_template.html" %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/table.html" import mapping_table, row, text_field %}
+
+{% block per_page_title %}
+  Daily volumes report
+{% endblock %}
+
+{% block platform_admin_content %}
+
+  <h1 class="heading-large">
+    Daily volumes Report
+  </h1>
+
+  {% call form_wrapper() %}
+    {{ form.start_date(param_extensions={"hint": {"text": "Use the format YYYY-MM-DD"}}) }}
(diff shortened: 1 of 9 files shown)


app/templates/views/platform-admin/daily-volumes-report.html
```diff
@@ -0,0 +1,48 @@
...
+      caption_visible=False
+    ) %}
+      {% for message_length, charge in [
+        ('day', 'The whole business day in BST.'),
+        ('sms totals', 'The number of text messages sent'),
+        ('sms chargeable units', 'The number of text message fragments sent'),
+        ('sms_fragments_times_multiplier', 'The number of text message fragments sent times the rate multiplier'),
+        ('email totals', 'The number of emails sent'),
+        ('letter totals', 'The number of letters sent'),
+        ('letter sheet totals', The number of sheets sent)
```
klssmith, line 38: I had to add quotes around `The number of sheets sent` when running this locally in order to avoid a Jinja error

app/templates/views/platform-admin/daily-volumes-report.html
```diff
@@ -0,0 +1,48 @@
...
+      caption='Descriptions of daily volumes data',
+      field_headings=['Name', 'Description'],
+      field_headings_visible=True,
+      caption_visible=False
+    ) %}
+      {% for message_length, charge in [
+        ('day', 'The whole business day in BST.'),
+        ('sms totals', 'The number of text messages sent'),
+        ('sms chargeable units', 'The number of text message fragments sent'),
+        ('sms_fragments_times_multiplier', 'The number of text message fragments sent times the rate multiplier'),
```
klssmith, line 35: The 'sms_fragments_times_multiplier' is named slightly differently in the report - it's called 'sms_fragment_totals' there

klssmith (commented):
A basic test for each of the new report endpoints seems like it would be worthwhile having (we have a test for the other report endpoints that could just be copied). I know you said that the aim was to test this with production data though, so that could be done later once it's been confirmed that this report doesn't need any changes.

app/templates/views/platform-admin/volumes-by-service-report.html
```diff
@@ -0,0 +1,46 @@
...
+    {% call mapping_table(
+      caption='Descriptions of volumes by service data',
+      field_headings=['Name', 'Description'],
+      field_headings_visible=True,
+      caption_visible=False
+    ) %}
+      {% for message_length, charge in [
+        ('free allowance', 'Free allowance set for the service. This is the latest free allowance for the date range given'),
+        ('sms notifications', 'The number of text messages sent by the service.'),
+        ('sms volume', 'The number of text message fragments times the rate multiplier sent by the service.'),
```
klssmith, line 34: The header in the CSV file is "sms chargeable units" instead of "sms volume".

There are also a few columns in the CSV file which aren't described here ("letter_cost", "letter_sheet_totals"), but that should be fine if they're self-explanatory.

app/templates/views/platform-admin/daily-volumes-report.html
servingUpAces: I'll add that thanks

app/templates/views/platform-admin/daily-volumes-report.html
servingUpAces: Thanks - we changed the name of the column, but I missed this one.

app/templates/views/platform-admin/volumes-by-service-report.html
servingUpAces: Fixed - thanks

klssmith (approved):

PR #4184, opened 2022-03-07
Simplify "activate_user" to clear session earlier

benthorner:
A snippet of old code [^1] in "activate_user" was forcing us to
keep "user_details" in the session until the very last moment with
a "try / finally" combo. But "activate_user" already knows the ID
of the user: it's the argument we pass to the function.

None of the functions called by "activate_user" require the session
to have that key either i.e. it's definitely redundant.

It's unclear if we need to pop the key from the session in both
"verify" methods - there are no tests covering this behaviour. For
now, we can at least make the flow clearer by adjusting where we do
the "pop" and the assignment.

[^1]: https://github.com/alphagov/notifications-admin/commit/bbc7b173f05305fca3386138479ccdba257be302#diff-d12384ece5ad90e9b66063fd3ab170453788d36b7e0babf49ee016f0a880f251L71

### DIFF
--- a/app/main/views/verify.py
+++ b/app/main/views/verify.py
@@ -31,10 +31,8 @@ def verify():
     form = TwoFactorForm(_check_code)
 
     if form.validate_on_submit():
-        try:
-            return activate_user(user_id)
-        finally:
-            session.pop('user_details', None)
+        session.pop('user_details', None)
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
Nice one 👍🏼

PR #4190, opened 2022-03-14
Format `auth_type` in a consistent way in the UI

quis:
On the ‘find user’ page it says ‘sms_auth’ instead of ‘Text message code’.

This commit fixes that, and adds a handy formatter so it’s easier to do the right thing in the future.

### DIFF
--- a/app/formatters.py
+++ b/app/formatters.py
@@ -528,3 +528,16 @@ def format_yes_no(value, yes='Yes', no='No', none='No'):
 
 def square_metres_to_square_miles(area):
     return area * 3.86e-7
+
+
+def format_auth_type(auth_type, with_indefinite_article=False):
+    indefinite_article, auth_type = {
+        'email_auth': ('an', 'Email link'),
+        'sms_auth': ('a', 'Text message code'),
+        'webauthn_auth': ('a', 'Security key'),
+    }.get(auth_type, ('a', auth_type))
+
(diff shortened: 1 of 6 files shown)


app/formatters.py
```diff
@@ -528,3 +528,16 @@ def format_yes_no(value, yes='Yes', no='No', none='No'):
...
 def square_metres_to_square_miles(area):
     return area * 3.86e-7
+
+
+def format_auth_type(auth_type, with_indefinite_article=False):
+    indefinite_article, auth_type = {
+        'email_auth': ('an', 'Email link'),
+        'sms_auth': ('a', 'Text message code'),
+        'webauthn_auth': ('a', 'Security key'),
+    }.get(auth_type, ('a', auth_type))
```
benthorner, line 538: I don't think we should have a default here: "a sms_auth" doesn't make sense and it would be better for us to error. Are we expecting to fall back here?

benthorner (approved):
This caught my eye as well - thanks for making it more readable 💯.

app/formatters.py
quis: Good point, I think it’s better to raise an exception and force us to fix it if we ever add a new `auth_type`, or use this formatter on something that isn’t a valid `auth_type`.
```suggestion
    }[auth_type]
```

PR #4191, opened 2022-03-15
Remove Sizewell from the demo areas library

quis:
We added this for the 25 May demo. We’re unlikely to need it again.

We’re also unlikely to need this library again, so this commit removes the source and the code that creates it.

***

This is a re-attempt of https://github.com/alphagov/notifications-admin/pull/3956

I can confirm that the admin app pages now render fine:

image

Instead of an error we just get a log message when viewing this page:
```
WARNING None "BroadcastMessage has 1 area IDs but 0 found in the library. Treating 02de5a38-e12c-480a-b3d2-2d3b28811a22 as a custom broadcast."
```


benthorner (approved):
Tested and confirms this works locally with an old Sizewell broadcast ✅.

PR #4196, opened 2022-03-23
Final set of refactorings for email branding

benthorner:
https://www.pivotaltracker.com/story/show/181415991

This continues where I left off in [^1]. Most of the
commits are about splitting out letters and emails in
the branding forms and the new branding module, which
makes it easier to fill all the holes in the tests.

Best to review this one commit by commit.

[^1]: https://github.com/alphagov/notifications-admin/pull/4182


tests/app/main/views/test_service_settings.py
```diff
@@ -4752,6 +4752,33 @@ def test_update_service_organisation_does_not_update_if_same_value(
     assert mock_update_service_organisation.called is False
 
 
+@pytest.mark.parametrize('single_branding_option, expected_href', [
+    (True, f'/services/{SERVICE_ONE_ID}/service-settings/email-branding/something-else'),
+    (False, f'/services/{SERVICE_ONE_ID}/service-settings/email-branding'),
+])
+def test_service_settings_links_to_branding_request_page_for_emails(
```
leohemsted, line 4759: i like this commit, but i'm not sure it goes far enough

But this is a 6200 line file (!) with 179 separate test definitions in it - can we go one further and move these (and any other) email/letter branding tests out to a new `test_service_settings_branding.py` file (or two)? We've already done that for a couple of other cases - inbound sms and service permissions.

leohemsted (commented):
Generally really sensible stuff, the commits were really cleanly broken up and straightforward. But a couple of things bugged me enough that I think we can go a bit further.

app/models/organisation.py
```diff
@@ -23,12 +23,16 @@ class Organisation(JSONModel, SortByNameMixin):
...
+    NHS_TYPES = (
         (TYPE_NHS_CENTRAL, 'NHS – central government agency or public body'),
         (TYPE_NHS_LOCAL, 'NHS Trust or Clinical Commissioning Group'),
         (TYPE_NHS_GP, 'GP practice'),
+    )
+
+    TYPES = (
+        (TYPE_CENTRAL, 'Central government'),
+        (TYPE_LOCAL, 'Local government'),
+    ) + NHS_TYPES + (
```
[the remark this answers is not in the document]
benthorner: You're right it's icky. At the time it seemed the lesser evil to keep it as-is.

I've added a commit to decouple the two concepts as per your suggestion: https://github.com/alphagov/notifications-admin/pull/4196/commits/0b8016bd1fab7ae10d3d978de038564057c066a0 - nice suggestion with `OrderedDict` while we're here 👍. Also, it turns out nothing needs `Organisation.TYPES` so 🔥.

tests/app/main/views/test_service_settings.py
leohemsted: very fair point about not knowing there was a separate test directory - i'd also be up for moving `tests/app/main/views/test_service_settings.py` to `tests/app/main/views/service_settings/test_service_settings.py` if that helps make it easier to discover the other files and go "aha, there's a better place for [whatever test i am writing]").

yeah i think changing the service_settings file is deffo out of scope for this PR, though would be great to do. 
i don't think we need to wait to do that for us to start moving tests though. but i'll leave it up to you

leohemsted (approved):

tests/app/main/views/test_service_settings.py
benthorner: Oh go on then - `/service_settings/test_service_settings.py` never occurred to me 😆.

It's definitely easier to navigate the email branding tests now: https://github.com/alphagov/notifications-admin/pull/4196/commits/c7951f8667f725c4419a9a6f0cdea687bc927c78 - a mere ~500 lines.

PR #4199, opened 2022-03-25
Support changing priorities with inactive SMS provider

benthorner:
https://www.pivotaltracker.com/story/show/181574489

This fixes a bug where a third inactive provider would (potentially) appear
in place of one of the two active ones, depending on the order of its
identifier compared to the other two.

I found the tests for the provider pages were quite hairy - lots of array
indices! - so the first few commits are some refactoring to make the actual
bug fix easier to follow.


tests/app/main/views/test_providers.py
```diff
@@ -10,103 +10,103 @@
 from app.main.views.providers import add_monthly_traffic
 from tests.conftest import normalize_spaces
 
+sms_provider_1 = {
```
idavidmcdonald, line 12: This commit is a real positive refactoring. Much more readable

tests/app/main/views/test_providers.py
```diff
@@ -443,16 +443,17 @@ def test_should_show_version_history_for_first_two_sms_providers(
         'app.provider_client.get_all_providers',
         return_value=stub_providers
     )
-    # second_sms_international will be the primary provider because it’s
-    # the first in the list when reverse sorting the SMS providers
+
+    # we get only get version info for the "primary" (first) provider
```
idavidmcdonald, line 447: I find this comment a little bit confusing as I can't quite work out why we are only getting version info for the first provider. Is the comment telling me that we are choosing to only get the version info for the first provider of is it telling me that we get a result of the only version info is for the first provider?

Also it might be helpful to capitalise and use full stops as then it is clear this is a single sentence (I think) rather than each line being two disjoint points?

app/main/views/providers.py
```diff
@@ -73,19 +73,19 @@ def edit_sms_provider_ratio():
     if len(providers) < 2:
         abort(400)
 
-    primary_provider, secondary_provider = providers[0:2]
+    first_provider, second_provider = providers[0:2]
```
idavidmcdonald, line 76: Good commit again, this is a helpful language change

tests/app/main/views/test_providers.py
idavidmcdonald: Do we need to remove the word 'primary' from this comment now we have removed that term from the rest of the code?

idavidmcdonald (approved):
Nice, some good refactorings in here.

I'll approve this now. I think there is one comment that ideally be made a bit clearer but I trust that if you do decide to change it, it will be appropriate and an improvement.

All the rest is pretty optional and doesn't require a change unless you feel strongly.

app/templates/views/providers/providers.html
```diff
@@ -23,7 +23,7 @@ <h2 class="heading-medium">SMS</h2>
       field_headings_visible=True
   ) %}
 
-      {{ text_field(item.display_name) }}
+      {{ link_field(item.display_name, url_for('main.view_provider', provider_id=item.id)) }}
```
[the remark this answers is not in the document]
benthorner: Good questions! Although I haven't used these pages either, they will shortly become more useful when we remove the history element of the edit page for provider ratios - it won't work with 3 providers.

I haven't checked why the links weren't there before, but I suspect it's related to the above: if we only have two providers, then the same information is (currently) available on the ratio edit page.

Yes, it's not ideal these are slow #pagination. At least the ratio edit page will become faster when we remove the history element from there, so the end result should be even more tolerable.

tests/app/main/views/test_providers.py
```diff
@@ -10,103 +10,103 @@
...
+    'id': '0bd529cd-a0fd-43e5-80ee-b95ef6b0d51b',
+    'notification_type': 'email',
+    'updated_at': None,
+    'version': 1,
+    'created_by': None,
+    'supports_international': False,
+    'current_month_billable_sms': 0,
+}
+
+sms_provider_int_1 = {
```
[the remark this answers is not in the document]
benthorner: Fair point, I've stretched it to `intl` in https://github.com/alphagov/notifications-admin/pull/4199/commits/a2fb386957ad0b76c495c6b5d76d936630a7bb51 😄.

PR #4203, opened 2022-04-06
Remove importlib_metadata as a dependency and bump dependencies that look relatively safe

klssmith:
**Remove importlib-metadata from requirements file**

This was pinned in https://github.com/alphagov/notifications-admin/commit/ee3fb2f175724d304fba9b90aba24890c4de1147 to prevent version conflicts. It only seems to be required by packages when the version of Python is 3.8 or lower (e.g. https://github.com/PyCQA/flake8/blob/21d3c70d676007470908d39b73f0521d39b3b997/docs/source/release-notes/3.8.0.rst#new-dependency-information), so is now not required anyway.

**Bump various dependencies, including all test dependencies**
This bumps the dependencies which don't appear to have anything in the changelog that could cause issues as well as all the test dependencies.


CrystalPea (approved):
Looks good. So some of our code is now bad style according to latest flake8 criteria? 😳

klssmith:
It was flake8-bugbear, not "main" flake8 that didn't like loop variables the way we had them in a few places. I thought how we had them was ok though, so decided to ignore it

CrystalPea:
Makes sense @klssmith 👍🏼

PR #4206, opened 2022-04-08
add new daily sms provider volumes report

leohemsted:
nearly identical to the daily-volumes-report but sms only, and split up by provider

see https://github.com/alphagov/notifications-api/pull/3502


app/templates/views/platform-admin/daily-sms-provider-volumes-report.html
```diff
@@ -0,0 +1,47 @@
...
+<div class="bottom-gutter-3-2">
+    {% call mapping_table(
+      caption='Descriptions of daily SMS provider volumes data',
+      field_headings=['Name', 'Description'],
+      field_headings_visible=True,
+      caption_visible=False
+    ) %}
+      {% for message_length, description in [
+        ('day', 'The whole business day in BST.'),
+        ('provider', 'The SMS provider')
```
benthorner, line 33:
```suggestion
        ('provider', 'The SMS provider'),
```

app/templates/views/platform-admin/daily-sms-provider-volumes-report.html
```diff
@@ -0,0 +1,47 @@
...
+        ('sms totals', 'The number of text messages sent'),
+        ('sms fragments', 'The number of text message fragments sent'),
+        ('sms chargeable units', 'The number of text message fragments sent times the rate multiplier'),
+        ('sms cost', 'The cost of text messages sent')
+      ] %}
+        {% call row() %}
+          {{ text_field(message_length) }}
+          {{ text_field(description | safe) }}
```
benthorner, line 41: Why are we using this filter?

tests/app/main/views/test_platform_admin.py
```diff
@@ -1113,3 +1113,46 @@ def test_get_daily_volumes_report_when_calls_api_and_download_data(
 
         '\r\n'
     )
+
+
+def test_get_daily_sms_provider_volumes_report_when_calls_api_and_download_data(
```
benthorner, line 1118:
```suggestion
def test_get_daily_sms_provider_volumes_report_calls_api_and_downloads_data(
```

tests/app/main/views/test_platform_admin.py
```diff
@@ -1113,3 +1113,46 @@ def test_get_daily_volumes_report_when_calls_api_and_download_data(
...
+        "day,provider,sms totals,sms fragment totals,sms chargeable units,sms cost\r\n" +
+
+        '2019-01-01,' +
+        'foo,' +
+        '20,' +
+        '40,' +
+        '60,' +
+        '80'
+
+        '\r\n'
```
benthorner, line 1157: This is using implicit concatenation whereas all the other lines are explicit. I think we prefer implicit, so you could remove the "+" on all the other lines.

Appreciate this is just following the tests above, which you could make consistent as well. There's quite a bit of variety going on in this file for some reason:

- `+ 'abc'`
- `'abc' +`
- `'abc' 'def'`

app/main/views/platform_admin.py
```diff
@@ -381,6 +381,46 @@ def get_daily_volumes():
...
 
+@main.route("/platform-admin/reports/daily-sms-provider-volumes-report", methods=['GET', 'POST'])
+@user_is_platform_admin
+def get_daily_sms_provider_volumes():
+    form = BillingReportDateFilterForm()
+
+    if form.validate_on_submit():
+        start_date = form.start_date.data
+        end_date = form.end_date.data
+        headers = [
```
benthorner, line 392: Out of scope: we repeat this pattern quite a lot in here; at some point we should DRY it up into a model or maybe just `Spreadsheet.from_json(column_name_dict)`.

benthorner (commented):
Works locally with the suggested edits 👇.

Could do with an extra test [over hear](https://github.com/alphagov/notifications-admin/blob/26e6e3e40cef1bdfc4e340bde330b7971744f5c5/tests/app/notify_client/test_billing_client.py#L55) so we've got coverage end-to-end. I did ponder if we should add an extra test for the "No results" case but it's just a cosmetic thing so I don't think it's worth it.

tests/app/main/views/test_platform_admin.py
leohemsted: whoops this is deffo a typo. i'll clean up the file

app/notify_client/billing_api_client.py
```diff
@@ -55,5 +55,14 @@ def get_data_for_daily_volumes_report(self, start_date, end_date):
                             'end_date': str(end_date),
                         })
 
+    def get_data_for_daily_sms_provider_volumes_report(self, start_date, end_date):
+        return self.get(
+            url='/platform-stats/daily-sms-provider-volumes-report',
```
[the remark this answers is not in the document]
leohemsted: i'll change API as "daily sms provider volumes report" reads better i think

app/templates/views/platform-admin/daily-sms-provider-volumes-report.html
leohemsted: I wasn't sure so I followed the code lineage - it's because on one line here, there's some HTML to add a link.

https://github.com/alphagov/notifications-admin/blob/master/app/templates/views/platform-admin/get-billing-report.html#L33

I'll move the `| safe` to just that single line, rather than applying it to every description, since it's quite a dangerous filter

benthorner: Great digging - much simpler!

tests/app/main/views/test_platform_admin.py
leohemsted: fixed in https://github.com/alphagov/notifications-admin/pull/4206/commits/971101d07c142cd770d01e75fb59500f3cdb2a31

tests/app/main/views/test_platform_admin.py
leohemsted: fixed in https://github.com/alphagov/notifications-admin/pull/4206/commits/971101d07c142cd770d01e75fb59500f3cdb2a31 (and elsewhere in this file where it was copy pasted from)

benthorner: Thanks you ⭐ .

(I see you found [some extra anarchy I didn't spot](https://github.com/alphagov/notifications-admin/commit/971101d07c142cd770d01e75fb59500f3cdb2a31#diff-aecc5f300f15c7061868939e3ede4cb98a154ef674901f94845d2f499b5e900eL961).)

app/main/views/platform_admin.py
leohemsted: yeah agreed. i did pause to look when writing this but there's a few gotchas in this file (eg, some where we're calling `.get` to access data or `.strip` on data, and also the header names for the output spreadsheet are not the same as the fields they're accessing from the API) so i wasn't sure if i could nicely make stuff consistent in a way that really improves rather than just moves stuff around.

benthorner (approved):
Looks good ✅

PR #4221, opened 2022-04-25
Show monthly usage rate correctly if the sms rate has changed during the financial year

idavidmcdonald:
https://www.pivotaltracker.com/story/show/181935935

The first two commits are little variable name refactoring.

The last commit is the proper one that makes the sms rate come from monthly rates rather than assuming a single rate for the whole year. I've written this a bit quickly before my training course so would suggest thinking about

- where would more test coverage be beneficial
- have I made correct assumptions about data I will be getting from the API
- might I be missing anything else?

I haven't spun this up locally yet so suggest you do as part of a review.

Note, this screenshot is the part of the page that this will affect however the screenshot does not show the new code and is just the current production. I've included the screenshot so you know which exact part of the app I am expecting this to make changes to.
image


app/main/views/dashboard.py
```diff
@@ -476,23 +477,31 @@ def get_free_paid_breakdown_for_month(
 ):
     allowance = free_sms_fragment_limit
 
+    # makes the assumption that there is either no item in `monthly_usage` because they have not sent any SMS
+    # or that they have sent SMS and that there is only a single item in `monthly_usage` because they have only
+    # been sent at a single rate during the month
+    sms_rate = monthly_usage[0]['rate'] if len(monthly_usage) else 0
```
quis, line 483: `if len(monthly_usage)` is equivalent to `if monthly_usage`, unless you want an exception raised should there be a non-iterable value for `monthly_usage`

benthorner (approved):
Works locally for me (doing some manual DB insertions):

Screenshot 2022-04-26 at 13 41 59

The second row shows how this won't cope if a change happens part way during a month, but that's still an improvement over assuming it's the same for the whole year.

As far as testing goes, I found [one other relevant test](https://github.com/alphagov/notifications-admin/blob/2706ec4c73ec22701b4f9f1dd4aa44cc020ba64a/tests/app/main/views/test_dashboard.py#L990), where you could adjust the monthly rate so that it's distinct from the yearly one - to prove it's used in preference. I've added a quick commit to tweak the tests to do that (don't think it needs another review).

Much of what's here is about to change anyway as we integrate https://github.com/alphagov/notifications-api/pull/3520, which will replace most of the calculations, so I don't think it's worth doing anything more here - going to merge to avoid conflicts.

PR #4222, opened 2022-04-26
Update text message rate on 1 May 2022

karlchillmaid:
This PR updates the pricing and product pages so they display the current text message rate automatically.

This means we no longer need to manually update this content when we change our prices.

We’re making this change now so we can automate the text message price increase that comes into effect over the bank holiday weekend.

### DIFF
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -42,7 +42,7 @@
 
   <h2 class="heading-medium" id="text-messages">Text messages</h2>
   <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
-  <p class="govuk-body">When a service has used its annual allowance, it costs 1.72 pence (plus VAT) for each text message you send.</p>
+  <p class="govuk-body">When a service has used its annual allowance, it costs {{ sms_rate }} pence (plus VAT) for each text message you send.</p>
   <p class="govuk-body">You’ll use more free messages, or pay more for each message, if you send <a class="govuk-link govuk-link--no-visited-state" href="#long-text-messages">text messages longer than 160 characters</a>.</p>
   <p class="govuk-body">You may also use more free messages, or pay more for each message, if you:</p>
   <ul class="list list-bullet">
--- a/app/templates/views/pricing/index.html
+++ b/app/templates/views/pricing/index.html
@@ -15,7 +15,7 @@
 
     {{ content_metadata(
       data={
-        "Last updated": "22 April 2022"
+        "Last updated": "1 May 2022"
       }
     ) }}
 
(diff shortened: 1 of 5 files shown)


app/templates/views/pricing/index.html
```diff
@@ -42,10 +42,7 @@ <h2 class="heading-medium" id="emails">Emails</h2>
 
   <h2 class="heading-medium" id="text-messages">Text messages</h2>
   <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
-  <p class="govuk-body">When a service has used its annual allowance, it costs 1.61 pence (plus VAT) for each text message you send.</p>
-  <div class="panel panel-border-wide">  
-    <p class="govuk-body">On 1 May 2022 the cost of sending a text message will go up to 1.72 pence (plus VAT).</p>
-  </div>
+  <p class="govuk-body">When a service has used its annual allowance, it costs 1.72 pence (plus VAT) for each text message you send.</p>
```
klssmith, line 45: Should the "Last updated" date at the top of the page change too?

karlchillmaid: It absolutely should, though @klssmith

app/templates/views/pricing/index.html
```diff
@@ -42,10 +42,7 @@ <h2 class="heading-medium" id="emails">Emails</h2>
 
   <h2 class="heading-medium" id="text-messages">Text messages</h2>
   <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
-  <p class="govuk-body">When a service has used its annual allowance, it costs 1.61 pence (plus VAT) for each text message you send.</p>
-  <div class="panel panel-border-wide">  
-    <p class="govuk-body">On 1 May 2022 the cost of sending a text message will go up to 1.72 pence (plus VAT).</p>
-  </div>
```
karlchillmaid, line 48: @quis if we deploy this change on 29 April it probably won’t matter too much - but in future it would be nice to have a way to keep this line until the price change comes into effect.

tests/app/main/views/test_index.py
```diff
@@ -356,3 +356,36 @@ def test_font_preload(
     for element in preload_tags:
         assert element['href'].startswith('https://static.example.com/fonts/')
         assert element['href'].endswith('.woff2')
+
+
+@pytest.mark.parametrize('current_date, expected_rate', (
+    ('2022-04-30', '1.61'),
+    ('2022-05-01', '1.72'),
+))
+def test_sms_price(
```
benthorner, line 365: While this is clear in the context of this PR, longterm I think we should have one test for each endpoint, or perhaps just one test for the function itself.

But I think we'll replace this code soon anyway so happy for it to stay as-is for the time being.

benthorner (approved):
Works locally for me (tweaking the date to verify) ✅.

Currently the only way to get rate data is in the context of billing for a service, so we'd need to expose a new API if we wanted to make this properly automatic.

PR #4225, opened 2022-04-26
Change annual usage to work with multiple SMS rates

benthorner:
https://www.pivotaltracker.com/story/show/181934027

This starts shifting SMS usage calculations to new
fields in the API [^1]. Key changes include:

- Costs are now calculated correctly in the API -
Admin just sums them together to get a total.

- We now show a breakdown by rate for each distinct
rate that SMS was charged at.

The first few commits are some general changes and
refactoring to make the main changes clearer.

## Screenshots

| Before | After |
| - | - |
| Screenshot 2022-04-26 at 17 52 50 | Screenshot 2022-04-26 at 17 30 50 |

[^1]: https://github.com/alphagov/notifications-api/pull/3520

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -328,8 +328,6 @@ def get_dashboard_totals(statistics):
 
 def get_annual_usage_breakdown(usage, free_sms_fragment_limit):
     sms = get_usage_breakdown_by_type(usage, 'sms')
-    # this relies on the assumption: only one SMS rate per financial year.
-    sms_rate = 0 if len(sms) == 0 else sms[0].get("rate", 0)
     sms_chargeable_units = sum(row['chargeable_units'] for row in sms)
     sms_free_allowance = free_sms_fragment_limit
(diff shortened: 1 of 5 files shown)


leohemsted (approved):
i think this looks really good! loved the commits, really helped chunk it up into manageable reviews. 

didn't have a chance to test this locally though - it might be worth thinking about if there's a way we can test this on preview with some DB jiggery pokery

benthorner:
Thanks for reviewing @leohemsted !

> didn't have a chance to test this locally though - it might be worth thinking about if there's a way we can test this on preview with some DB jiggery pokery

I'm not sure what more there is to test beyond what I've already done locally, populated `ft_billing` manually. I'm confident this works as expected, but to be on the safe side I'll:

- Have some before / after tabs open to check [the three services I looked at in the API PR](https://github.com/alphagov/notifications-api/pull/3520).
- I'll make a note to check these services again on Tuesday to see if the new rate shows up.

PR #4228, opened 2022-05-03
Clean up old text message price

quis:
Now that the new price is in effect, we don’t need to check the date to see which price should be displayed to the user.

### DIFF
--- a/app/main/views/pricing.py
+++ b/app/main/views/pricing.py
@@ -8,15 +8,14 @@ from app.main import main
 from app.main.forms import SearchByNameForm
 from app.main.views.sub_navigation_dictionaries import pricing_nav
 
-
-SMS_RATE = '1.72'
+CURRENT_SMS_RATE = '1.72'
 
 
 @main.route('/pricing')
 def pricing():
     return render_template(
         'views/pricing/index.html',
-        sms_rate=SMS_RATE,
+        sms_rate=CURRENT_SMS_RATE,
         international_sms_rates=sorted([
             (cc, country['names'], country['billable_units'])
             for cc, country in INTERNATIONAL_BILLING_RATES.items()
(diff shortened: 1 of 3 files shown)


app/main/views/pricing.py
```diff
@@ -13,8 +13,6 @@
 
 
 def get_current_sms_rate():
```
idavidmcdonald, line 15: Is this function still worth having rather than hardcoding the price in the view like we did before https://github.com/alphagov/notifications-admin/pull/4222/files was introduced?

tests/app/main/views/test_index.py
```diff
@@ -359,7 +359,6 @@ def test_font_preload(
 
 
 @pytest.mark.parametrize('current_date, expected_rate', (
```
idavidmcdonald, line 361: Optional: You can remove the parametrisation from this test if you want as there is only one test case

app/main/views/pricing.py
```diff
@@ -1,28 +1,22 @@
...
 from app.main import main
 from app.main.forms import SearchByNameForm
 from app.main.views.sub_navigation_dictionaries import pricing_nav
 
 
-def get_current_sms_rate():
-    if convert_utc_to_bst(datetime.utcnow()) < convert_utc_to_bst(datetime(2022, 5, 1)):
-        return '1.61'
-    return '1.72'
+SMS_RATE = '1.72'
```
idavidmcdonald, line 12: Optional: I slightly prefer `CURRENT_SMS_RATE` over `SMS_RATE` but don't feel strongly

idavidmcdonald (approved):
Looks good, thanks

tests/app/main/views/test_index.py
quis: I’ll keep it there since it makes for a cleaner diff if we need to make another change in the future.

PR #4234, opened 2022-05-05
Fix alignment regression error with area lists

tombye:
When we changed the layout for each alert on the
current/past/rejected alerts pages to use flexbox,
we added a fallback for older browsers that set
text-align: justify on the container: https://github.com/alphagov/notifications-admin/pull/4171.

This has led to items in the list of areas an
alert will be sent to being laid out as justified
content when they were left-aligned.

These changes set the correct alignment.

## Before

list_of_areas_before

## After

list_of_areas_after


benthorner (approved):
Nice find 👍.

PR #4235, opened 2022-05-05
Fix incorrect group for NPM dependencies

benthorner:
https://www.pivotaltracker.com/story/show/181889734

Most of these are only used for development / build. This means we can use
use tools like "npm audit" to look for vulnerabilities we definitely need
to fix as they could pose a direct risk to users. I've also removed one of
the packages where @tombye noticed it was unused.


tombye (approved):
Looks good to me.

By the way, I checked and the 'del' package should have been removed in https://github.com/alphagov/notifications-admin/commit/aff73293e4604be8b745a4adeabd426727b8b481.

PR #4238, opened 2022-05-09
Remove whitespace around API key in JS templates

quis:
We removed whitespace in the HTML of the copy to clipboard component in https://github.com/alphagov/notifications-admin/pull/4236

When the Javascript on the page loads it re-renders the component, using HTML which is embedded in the .js file.

This means we also need to apply the same change to the .js file to remove any extraneous whitespace.

### DIFF
--- a/tests/javascripts/copyToClipboard.test.js
+++ b/tests/javascripts/copyToClipboard.test.js
@@ -25,9 +25,7 @@ describe('copy to clipboard', () => {
         ${options.thing}
       </h2>
       <div data-module="copy-to-clipboard" data-value="${apiKey}" data-thing="${options.thing}" data-name="${options.name}">
-        <span class="copy-to-clipboard__value" aria-live="assertive">
-          ${(options.name === options.thing) ? '<span class="govuk-visually-hidden">' + options.thing + ': </span>' : ''}${apiKey}
-        </span>
+        <span class="copy-to-clipboard__value" aria-live="assertive">${(options.name === options.thing) ? '<span class="govuk-visually-hidden">' + options.thing + ': </span>' : ''}${apiKey}</span>
         <span class="copy-to-clipboard__notice" aria-live="assertive"></span>
       </div>`;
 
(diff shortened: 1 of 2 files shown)


benthorner (commented):
The main code change makes sense, but I'm confused by [the "prefix" test change](https://github.com/alphagov/notifications-admin/pull/4238/files#diff-00acf83f8d250a0e3a86e93c6cf7a03de2913fd3715f5a80c8cbceb6f1894e19R357). It's unclear what is meant by a prefix and why the main code change affects it by a single decrement. Could you help me understand?

tests/javascripts/copyToClipboard.test.js
```diff
@@ -346,11 +350,11 @@ describe('copy to clipboard', () => {
 
         test("the copied selection (range) should start after the prefix of the id", () => {
 
-          // that selection (a range) should have a startOffset past the first two nodes:
-          // index 0: text node containing the whitespace before the prefix
-          // index 1: the prefix node
+          // that selection (a range) should have a startOffset of 1:
+          // index 0: the prefix node
```
benthorner, line 354:
```suggestion
          // index 0: the hidden prefix node e.g. "ID: "
```

tests/javascripts/copyToClipboard.test.js
```diff
@@ -346,11 +350,11 @@ describe('copy to clipboard', () => {
 
         test("the copied selection (range) should start after the prefix of the id", () => {
```
benthorner, line 351:
```suggestion
        test("the copied selection (range) should start after the hidden prefix", () => {
```

benthorner: (And similarly in the previous test - avoid use of "id" terminology)

benthorner (approved):
Thanks for thorough explanation. I get it now.

My confusion was a combo of:

1. I forgot that a `range` is actually specified in terms of elements; I was thinking it was some kind of character offset.

2. The "prefix of the id" contained two unfamiliar terms. It looks like this component was written for "ids" but is now being used for many other things.

I've suggested a couple of small textual tweaks to try and hint at the behaviour. It was mainly (1) that got me confused, though, so I'll leave it up to you.

tests/javascripts/copyToClipboard.test.js
quis: Easier to understand with some real examples I think:
```suggestion
          // index 0: the visually hidden prefix node, for example "Template ID: " or "API key: "
```

benthorner: Yes, that's much more useful.

PR #4244, opened 2022-05-12
Upgrade xlrd to version 2 and drop `.xlsm` support

quis:
xlrd is a library for reading data and formatting information from Excel files in the historical `.xls` format.

Version 2 of xlrd no longer supports anything other than `.xls` files.

We were using it to also support reading .xlsm files (old Excel files with macro support).

We could keep using the old version of this dependency, but hopefully this niche version of an ancient file format is obscure enough that no-one is using it, and we can drop support, keeping our dependencies up to date.


quis:
Average number of uploads by day, by file extension (log scale):

image

As a table:

File extension | Mean number of uploads per day
-- | --
`.xlsx` | 376
`.csv` | 444
`.xls` | 37
`.ods` | 3
`.xlsm` | 5
`.tsv` | 4


***

This means that we still have a handful of `.xlsm` files being uploaded almost every day.

quis:
# Force xlsm files to open with pyexcel-xlsx

`.xlsm` files are like `.xlxs` files but with macros enabled. They store data in the same XML-based format as `.xlsx` files.

Pyexcel will try to use the xlrd package to parse `.xlsm` files. This used to work because xlrd used to support reading `.xlsx` files. xlrd has dropped support for `.xlsx` files in version 2 because of security concerns. This means that when pyexcel asks xlrd to parse a `.xlsm` file it causes an error.

This commit adds some branching to force `.xlsm` files to be opened with pyexcel-xlsx instead, which does support `.xlsx` files.

klssmith (approved):
I can't try this out since I don't have any xlsm files, but it looks good

quis:
@klssmith the tests check against [`tests/spreadsheet_files/excel 2007 with macro support.xlsm`](https://github.com/alphagov/notifications-admin/blob/master/tests/spreadsheet_files/excel%202007%20with%20macro%20support.xlsm) which gives me confidence that this will work ok

PR #4245, opened 2022-05-12
Run `pip-compile --upgrade`

quis:
This upgrades all the sub-dependencies which we don’t pin in `requirements.txt`

I’ve already upgraded any dependencies which needed code changes in separate PRs.


rjbaker (approved):
Looks good. Tests and Pyup are both happy.

PR #4248, opened 2022-05-19
Stop pyup trying to upgrade blocked dependencies

quis:
We can’t upgrade Jinja or Werkzeug until we’re on Flask 2.x.x. We can’t upgrade Flask to 1.1.3 because it pins older versions of Jinja and Werkzeug than the ones we’re using. We can’t upgrade Flask to 2.x.x until we upgrade itsdangerous to 2.x.x, which is blocked by https://github.com/alphagov/notifications-admin/pull/4044

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -4,12 +4,12 @@
 ago==0.0.93
 govuk-bank-holidays==0.11
 humanize==4.1.0
-Flask==1.1.2  # pyup: <2
+Flask==1.1.2  # pyup: <1.1.3
 Flask-WTF==1.0.1
 wtforms==3.0.1
 Flask-Login==0.6.1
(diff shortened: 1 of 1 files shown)


CrystalPea (commented):
Is this permanently blocked by `itsdangerous`, or do we have a ticket to try and unblock it at some point?

CrystalPea:
Hey @quis I think a card would be helpful, because then we can prioritise that work and make sure it is not forgotten about, and also link this PR here in the card, so we don't forget to turn PyUp back on for these deps when the card is done.

quis:
Makes sense @CrystalPea – I have added a card here: https://www.pivotaltracker.com/story/show/182244431

CrystalPea (approved):

PR #4256, opened 2022-05-27
Bump Flask, Jinja and Werkzeug to their latest versions

quis:
Now that we’ve upgraded itsdangerous to the latest version (https://github.com/alphagov/notifications-admin/pull/4255) we are:
- unblocked from upgrading to Flask 2, which requires a recent version of itsdangerous
- unblocked from upgrading Jinja and Werkzeug to the latest versions, which require Flask 2

There are a few breaking changes to account for, details of which are in the individual commits.

### DIFF
--- a/tests/app/utils/test_user.py
+++ b/tests/app/utils/test_user.py
@@ -24,9 +24,10 @@ def _test_permissions(
         decorated_index()
     else:
         try:
-            if (
-                decorated_index().location != '/sign-in?next=%2F' or
-                decorated_index().status_code != 302
+            response = decorated_index()
+            if not (
+                response.location.startswith('/sign-in?next=') and
+                response.status_code == 302
             ):
                 pytest.fail("Failed to throw a forbidden or unauthorised exception")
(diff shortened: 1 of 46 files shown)


tests/app/utils/test_user.py
```diff
@@ -24,9 +24,10 @@ def _test_permissions(
         decorated_index()
     else:
         try:
-            if (
-                decorated_index().location != '/sign-in?next=%2F' or
-                decorated_index().status_code != 302
+            response = decorated_index()
```
benthorner, line 27: 😷

Nice one figuring this out 🏅. I agree this is a reasonable compromise for now. Given that we're just hiding a bug though, I should at least ask: what would it take to make this test realistic? I'm wondering if we can separate testing the decorator can decorate (one test) vs. the internals of how the decorators decides who can access what (many tests).

quis: I’ll address this separately in https://github.com/alphagov/notifications-admin/pull/4261

PR #4261, opened 2022-06-06
Refactor permissions tests

quis:
The tests in this file are very old, and don’t make use of the fixtures and techniques we’ve since developed.

By refactoring them a bit we:
- delete a bunch of lines of code
- make it easier to see what cases the tests are covering

I don’t think it’s worth trying to cover more or fewer cases at this point, but this should make it easier for anyone who comes along and wants to add to or fix one of these tests.

### DIFF
--- a/tests/app/utils/test_user.py
+++ b/tests/app/utils/test_user.py
@@ -1,6 +1,6 @@
 import pytest
 from flask import request
-from werkzeug.exceptions import Forbidden, Unauthorized
+from werkzeug.exceptions import Forbidden
 
 from app.main.views.index import index
 from app.utils.user import user_has_permissions
@@ -30,7 +30,7 @@ def _test_permissions(
                 response.status_code == 302
             ):
                 pytest.fail("Failed to throw a forbidden or unauthorised exception")
-        except (Forbidden, Unauthorized):
+        except Forbidden:
             pass
 
 
--- a/tests/app/utils/test_user.py
+++ b/tests/app/utils/test_user.py
@@ -10,7 +10,7 @@ def _test_permissions(
     client_request,
     usr,
     permissions,
-    will_succeed,
+    will_succeed=True,
     kwargs=None,
 ):
     request.view_args.update({'service_id': 'foo'})
@@ -19,19 +19,7 @@ def _test_permissions(
 
     decorator = user_has_permissions(*permissions, **(kwargs or {}))
     decorated_index = decorator(index)
-
-    if will_succeed:
-        decorated_index()
-    else:
-        try:
-            response = decorated_index()
-            if not (
-                response.location.startswith('/sign-in?next=') and
-                response.status_code == 302
-            ):
-                pytest.fail("Failed to throw a forbidden or unauthorised exception")
-        except Forbidden:
-            pass
+    return decorated_index()
 
 
 def test_user_has_permissions_on_endpoint_fail(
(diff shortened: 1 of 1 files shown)


tests/app/utils/test_user.py
```diff
@@ -122,22 +102,18 @@ def test_no_user_returns_redirect_to_sign_in(
     client_request
 ):
     client_request.logout()
-    response = _test_permissions(
-        client_request,
-        None,
-        [],
-    )
+    decorator = user_has_permissions()
+    decorated_index = decorator(index)
```
benthorner: Note: this is sufficient as we check a user is logged in before trying to access [the current service](https://github.com/alphagov/notifications-admin/pull/4261/commits/15131d003b10633910e251a22144705a79286ae4#diff-4d7a26a1ab505f7460b3087e21a674d8b62662f1857dd4650e66a2a80f3b9a2cR15).

tests/app/utils/test_user.py
```diff
@@ -4,98 +4,77 @@
...
-def test_exact_permissions(
-    client_request,
-):
-    _test_permissions(
-        client_request,
+        {},
+    ),
+    (
         _user_with_permissions(),
         ['manage_service', 'manage_templates'],
```
benthorner: This test is the same as the one above. I appreciate it's testing from a different perspective, but now we've lost the test name to explain it I think it's just confusing.

Can we remove it?

tests/app/utils/test_user.py
```diff
@@ -4,98 +4,77 @@
...
+    return user_data
 
 
-def test_user_has_permissions_success(
-    client_request,
-    mocker,
-):
-    _test_permissions(
-        client_request,
+@pytest.mark.parametrize('user, permissions, kwargs', (
```
benthorner: Stretch goal: sometimes I find it hard to understand what a particular set of parameters is trying to test and find it helpful when people leave a comment to explain. In addition, there's some subtlety here of ["UI permissions" vs "DB permissions"](https://github.com/alphagov/notifications-admin/blob/3c25de3f8572442d216243e822588edea9510bfe/app/utils/user_permissions.py#L35), which would be helpful to mention in comments and/or the param name.

What do you think?

tests/app/utils/test_user.py
```diff
@@ -1,151 +1,77 @@
...
-    mocker,
-):
-    user = _user_with_permissions()
-    mocker.patch('app.user_api_client.get_user', return_value=user)
-    _test_permissions(
-        client_request,
-        user,
-        ['manage_templates', 'manage_service'],
-        will_succeed=True)
+    user = create_user(
```
benthorner, line 25: Optional: could we use one of the fixtures instead? Like this:

```
api_user_active['permissions'] = ...
api_user_active['services'] = ...
```

This is a bit more simpler; we also get the ID generation for free. Similarly for the other tests where we're using `create_user`.

tests/app/utils/test_user.py
```diff
@@ -1,151 +1,77 @@
...
+        id=sample_uuid(),
+        permissions={'foo': ['manage_users', 'manage_templates', 'manage_settings']},
+        services=['foo', 'bar'],
+    )
+    client_request.login(user)
 
+    @user_has_permissions(*permissions)
```
benthorner, line 32: Nice consistency improvement (noticing we did this further down already) 💯

benthorner (approved):
A clear net improvement 👏.

I've got a couple of optional suggestions where you could go a bit further, but this is also good to go as-is if you prefer.

tests/app/utils/test_user.py
quis: Changed in 41c9136

PR #4265, opened 2022-06-07
Fix relative URLs appearing in support tickets

quis:
When we get a support ticket we put a link to the service at the end.

As part of https://github.com/alphagov/notifications-admin/commit/8b7f2fbf041c182156cbdb4d3fbadc1c851e6653 we accidentally made these URLs relative, rather than absolute. This means they aren’t made into links by email clients or Zendesk.

For example:

image

This commit fixes the links to include the domain again.


benthorner (approved):
Nice refactoring 💯.

Also my bad for not seeing this - it's one of just two `app/` changes in that commit.

PR #4269, opened 2022-06-09
Allow organisations to be archived

klssmith:
This adds a "Delete this organisation" link to the organisation settings
page. The link shows a confirmation banner and makes a request to API to
archive the organisation if the user clicks confirm.

This is similar to what we have for services, but a little simpler. Only
platform admin users can see the settings page, so no one else can
archive an organisation. We also don't create an event.

The API endpoint expects the org to already have no team members or
services, which means that we don't need to clear the cache for those
things. It removes branding and domains for an org when archiving it, so
we do need to clear the domains cache.

[Pivotal story](https://www.pivotaltracker.com/story/show/178519743)

_Needs to be merged after https://github.com/alphagov/notifications-api/pull/3554_

Screenshot 2022-06-09 at 16 55 33

Screenshot 2022-06-09 at 16 55 41


app/main/views/organisations.py
```diff
@@ -596,6 +596,32 @@ def edit_organisation_billing_details(org_id):
...
+@user_is_platform_admin
+def archive_organisation(org_id):
+    if not current_organisation.active:
+        abort(403)
+
+    if request.method == 'POST':
+        try:
+            organisations_client.archive_organisation(org_id)
+        except HTTPError as e:
+            if 'team members' in e.message or 'services' in e.message:
```
idavidmcdonald, line 609: It looks like some of our other code also checks that `e.status_code == 400`, just to cover the case where there may be some other type of HTTP error which contains say the word 'services'. I suggest we follow this pattern for consistency please

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -953,6 +953,142 @@ def test_organisation_settings_for_platform_admin(
...
+    organisation_one['active'] = False
+    mocker.patch('app.organisations_client.get_organisation', return_value=organisation_one)
+
+    client_request.login(platform_admin_user)
+    page = client_request.get('.organisation_settings', org_id=organisation_one['id'])
+
+    assert not page.select('.page-footer-link a')
+
+
+def test_archive_organisation_is_platform_admin_only(
```
idavidmcdonald, line 988: Optional: Might also be good to add a test to checks we don't show the link to archive the organisation (on the off chance that we make a mistake and show the link, even if the link is blocked from working)

app/notify_client/organisations_api_client.py
```diff
@@ -86,5 +86,13 @@ def get_services_and_usage(self, org_id, year):
             params={"year": str(year)}
         )
 
+    @cache.delete('organisations')
+    @cache.delete('domains')
+    def archive_organisation(self, org_id):
```
idavidmcdonald, line 91: Realised that my comment about changing the organisation name as part of the API pull request, may mean that you would need to add `@cache.delete('organisation-{org_id}-name`) to this. Does that sound right?

app/notify_client/organisations_api_client.py
```diff
@@ -86,5 +86,13 @@ def get_services_and_usage(self, org_id, year):
             params={"year": str(year)}
         )
 
+    @cache.delete('organisations')
+    @cache.delete('domains')
```
idavidmcdonald, line 90: Very optional: Do we also need to add `live-service-and-organisation-counts` to the keys deleted from the cache?

Looking at https://www.notifications.service.gov.uk/performance, it looks like we don't include organisations with no services on anyway on the page so maybe the change wouldn't actually have an affect anyway?

Looking at the `archive_service` method in the `service_api_client.py` file, it looks like we don't bother there so maybe it isn't worth bothering with anyway...

I'll leave you to decide what you think is appropriate

idavidmcdonald (commented):
Looking good!

It all looks fairly reasonable. I've given a bit of extra investigation into redis and deleting cache keys just to make sure we aren't missing anything there.

I support your choice for no audit event for deleting an organisation.

A few minor comments but nothing substantial to change in here. The overall approach is all good!

tests/app/main/views/organisations/test_organisations.py
klssmith: I have this test `test_organisation_settings_does_not_show_delete_link_for_archived_organisations` already, did you mean something else apart from that one?

app/main/views/organisations.py
klssmith: Now done 👍

app/notify_client/organisations_api_client.py
klssmith: Yes, I've updated this to remove that key from the cache too

app/notify_client/organisations_api_client.py
klssmith: I don't think we need to remove that - it gets its data from the [dao_count_organisations_with_live_services](https://github.com/alphagov/notifications-api/blob/5cd6fcbb4feec983ae39a7ae8b8ad3132f534e5d/app/dao/organisation_dao.py#L14) DAO function, which only returns organisations with services. We don't allow you to archive organisations with services, so the value won't change.

idavidmcdonald: You are correct, no action needed on this

idavidmcdonald (approved):
Fantastic, all looks great :shipit:

PR #4276, opened 2022-06-20
Try simplifying the way we filter templates

benthorner:
https://www.pivotaltracker.com/story/show/179736794

This is one of the smaller follow-up actions to [^1], where we 
moved a bunch of template rendering code into a model file.

In the second commit I'm arguing for a really small change in 
how we show empty folders to users, so that we can simplify 
some of the rendering code.

## Screenshots (folder with other templates)

| Before | After |
| - | - |
| Screenshot 2022-06-20 at 16 50 31 | Screenshot 2022-06-20 at 16 50 17 |

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -485,7 +485,7 @@ def delete_template_folder(service_id, template_folder_id):
     template_folder = current_service.get_template_folder_with_user_permission_or_403(template_folder_id, current_user)
     template_list = TemplateList(service=current_service, template_folder_id=template_folder_id)
 
-    if not template_list.folder_is_empty:
+    if any(template_list):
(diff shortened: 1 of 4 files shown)


quis (commented):
Not a fan of this change sorry. Karl and I worked very carefully to make sure those bits of microcopy are as helpful and precise as possible. This change implies that – while there are no email templates in the folder – there might still be other kinds of template in the folder

benthorner:
@quis thanks for taking a look. I could see a lot of effort had gone into this part of the UI, but I wasn't sure why. The exact user need here is still mysterious to me - as in, what practical impact does this knowledge have? - but I'm happy to close as it was just a stretch goal to see where we could simplify some more.

PR #4279, opened 2022-06-21
Remove `<nav>` around template list items

klssmith:
The `<nav>` tag we have around the template list items, on the templates
page, should be removed.

This issue was raised by the Digital Accessibility Centre (DAC), in their
audit on May 18th 2022. Their tester said it creates cognitive load and
doesn't really add much for it.

[Pivotal story](https://www.pivotaltracker.com/story/show/182500752)


tombye (approved):
Looks good to me 👍🏻

PR #4280, opened 2022-06-22
Fix back link on final page to copy a template

benthorner:
https://www.pivotaltracker.com/story/show/179736794

This was broken in two ways:

- It would take the user to a 404 page if the template they copy is from a
different service. To reproduce:

- In all other cases it would exit the copy workflow and go to the preview
page for the original template.

The new tests aren't comprehensive but I think they're sufficient to cover 
the majority of cases for a fairly low use feature.

## Screenshots (before, after clicking "Back")

| Copy from another service (bug) | Copy from the same service (bug) |
| - | - |
| Screenshot 2022-06-22 at 11 09 14 | Screenshot 2022-06-22 at 11 09 43 |

## Screenshots (after, after clicking "Back")

| Copy from another service (fixed) | Copy from the same service (fixed) |
| - | - |
| Screenshot 2022-06-22 at 11 10 25 | Screenshot 2022-06-22 at 11 09 56 |

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -391,13 +391,27 @@ def copy_template(service_id, template_id):
     template['name'] = _get_template_copy_name(template, current_service.all_templates)
     form = form_objects[template['template_type']](**template)
 
+    if template['folder']:
+        back_link = url_for(
+            '.choose_template_to_copy',
+            service_id=current_service.id,
+            from_service=from_service,
+            from_folder=template['folder'],
+        )
+    else:
+        back_link = url_for(
+            '.choose_template_to_copy',
+            service_id=current_service.id,
+            from_service=from_service,
+        )
+
     return render_template(
         'views/edit-{}-template.html'.format(template['template_type']),
         form=form,
         template=template,
         heading_action='Add',
         services=current_user.service_ids,
-        back_link='#',  # TODO: broken
+        back_link=back_link,
     )
 
 
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -397,6 +397,7 @@ def copy_template(service_id, template_id):
         template=template,
         heading_action='Add',
         services=current_user.service_ids,
+        back_link='#',  # TODO: broken
     )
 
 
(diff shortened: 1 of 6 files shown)


app/main/views/templates.py
```diff
@@ -391,13 +391,27 @@ def copy_template(service_id, template_id):
     template['name'] = _get_template_copy_name(template, current_service.all_templates)
     form = form_objects[template['template_type']](**template)
 
+    if template['folder']:
```
idavidmcdonald, line 394: I'm guessing this means we can't just always set the back link as 

```
back_link = url_for(
            '.choose_template_to_copy',
            from_service=from_service,
            from_folder=template['folder'],
        )
```

and that something breaks if template['folder'] is None if that were the case?

idavidmcdonald (approved):
Thanks Ben.

I've replicated the bug in production and looked at the code but I haven't spun this up locally as I'm rushing through but wanted to get this approved before the end of the day.

Am I right in thinking there are there 4 cases here for which we want to make sure the back link works:

- copying a template from your current service that is not in a folder
- copying a template from your current service that is in a folder
- copying a template from another current service that is not in a folder
- copying a template from another current service that is in a folder

If so, might just be worth stepping those through on preview or local environment if you haven't already as I note your test coverage doesn't cover all of those (although I see your point about not aiming for full test coverage so a quick manual test may be good enough).


On a more general note, I found this PR a bit harder to review than some of your others. The first commit was a little difficult to review on its own because it also changed some behaviour (changing the backlink to `#`) where I might have instead to preferred to have seen some tests added that either were telling us what we wanted to happen and marked as expected to fail or adding some tests that checked the existing buggy behaviour (so then your next commit can make it explicit what they are changing from and to). Not a big deal though, it wasn't like it made it actually difficult.

Nice work on spotting this and tidying it up :)

app/main/views/templates.py
benthorner: Good question. The distinction is annoying but important:

- With `from_folder=None` we get `/services/<uuid:service_id>/templates/copy/from-service/<uuid:from_service>/from-folder/<**missing-id**>`

- Without `from_folder` we get `/services/<uuid:service_id>/templates/copy/from-service/<uuid:from_service>`

The new tests cover both cases and the other two tests cover the scenarios of adding or editing a template. What's specifically missing is we're only checking a specific type of template in each test.

benthorner:
Thanks for the review David!

The "copy from my one service" vs. "copy from some other service" are actually the same in practice, but I have manually double checked that all 4 of the scenarios you mention do work.

I also see what you mean about the first commit - it goes too far in the main code and not far enough with the tests. Thanks for highlighting it - I'll leave it as-is though as, like you say, it's not critical.

PR #4281, opened 2022-06-23
Use summary-list component on `/user-profile` page

klssmith:
The `/user-profile` page was using a table to display the user data, and this commit changes it to use the GOV.UK Frontend [summary-list component](https://design-system.service.gov.uk/components/summary-list/) instead. The component takes rows as an input and it's not possible to use conditional logic inside the component. This means that we need to build all the data for the rows in variables first, then pass it to the component.

It's not possible to pass ids to the component, but the existing ids on the page were only being used in tests, so I've changed how we find rows in the tests.

[Pivotal story](https://www.pivotaltracker.com/story/show/182498266)

### Desktop
Desktop screenshot before change

### Mobile
Mobile screenshot before change

### Desktop
Desktop screenshot after change

### Desktop with some fields missing and email address not able to be changed
Desktop screenshot after change with some fields missing and email address not able to be changed

### Mobile
Mobile screenshot after change


tombye:
I'm happy to approve the code, which looks good to me. The experience for most screen reader users is also improved. JAWS and NVDA, the most used screen readers, let you cycle through the rows by field name ("Name, Email address..."), which is great for quickly moving to a field, while also letting you move within a row to see the current value and change it.

The visual experience looks similar though I'm holding off giving this a 👍🏻 as would like @quis 's opinion on the move to bold field names first.

quis:
New mobile layout is a good improvement.

I’d like to match the non-bold labels on desktop and keep the same grid as before. Happy to pick this up or pair – whatever you think is easier.

klssmith:
Thanks @quis I'm not too sure how this should look, so if you can pick this up it would be great

tests/app/main/views/test_user_profile.py
```diff
@@ -37,7 +37,7 @@ def test_overview_page_shows_disable_for_platform_admin(
     client_request.login(platform_admin_user)
     page = client_request.get('main.user_profile')
     assert page.select_one('h1').text.strip() == 'Your profile'
-    disable_platform_admin_row = page.select_one('#disable-platform-admin')
+    disable_platform_admin_row = page.find_all(class_='govuk-summary-list__row')[-1]
```
quis, line 40: I prefer CSS-style selectors:
```suggestion
    disable_platform_admin_row = page.select('.govuk-summary-list__row')[-1]
```

tests/app/main/views/test_user_profile.py
```diff
@@ -62,7 +62,7 @@ def test_overview_page_shows_security_keys_if_user_they_can_use_webauthn(
         return_value=credentials,
     )
     page = client_request.get('main.user_profile')
-    security_keys_row = page.select_one('#security-keys')
+    security_keys_row = page.find_all(class_='govuk-summary-list__row')[-2]
```
quis, line 65: Same
```suggestion
    security_keys_row = page.select('.govuk-summary-list__row')[-2]
```

quis:
I’ve added a commit to tweak the spacing and grid alignment to how it should be (turns out it was broken before anyway).

# Before 

Screenshot 2022-08-03 at 13 23 47

# With just your changes

Screenshot 2022-08-03 at 13 22 42


# With both of our changes

Screenshot 2022-08-03 at 13 19 40

tests/app/main/views/test_user_profile.py
klssmith: I've updated this (and the other similar test line) and rebased the change in

app/templates/views/user-profile.html
```diff
@@ -149,8 +149,10 @@ <h1 class="heading-large">Your profile</h1>
   ) %}
 {% endif %}
 
+<div class="settings-summary-list settings-summary-list--full-width">
```
tombye, line 152: We can add classes to the summary list container so we shouldn't need this wrapper div, I think.

tombye:
Most of these changes make sense to me (only one I have a comment on is [here](https://github.com/alphagov/notifications-admin/pull/4281#discussion_r937620500), but I would like to question how we apply the CSS they have in `app/assets/stylesheets/components/settings-summary-list.scss`.

It's not helpful that I didn't document the decisions made when first adding GOVUK Frontend (sorry, I will try to write this up). The intention though, was that our use of GOVUK Frontend components focused on extending or overriding their default styles/API rather than wrapping them in our own components[^1]. I also intended us to follow GOVUK Frontend coding conventions when we do so. That's the basis for how the `app/assets/stylesheets/govuk-frontend` folder is laid out.

That being the case, I'd like to make an argument for the following changes to the Sass in `app/assets/stylesheets/components/settings-summary-list.scss`:
1. apply the border-top through a single class on the container element, like `govuk-summary-list--border-top` and put those styles in `app/assets/stylesheets/govuk-frontend/extensions.scss`
2. make the row headers non-bold through a `govuk-!-font-weight-regular` class on each `.govuk-summary-list__key` element
3. apply the extra padding on row items through a single class on the `.govuk-summary-list__row` element, like `govuk-summary-list__row--padding-settings` and put those styles in `app/assets/stylesheets/govuk-frontend/extensions.scss`
4. give the `.govuk-summary-list__key` elements 25% width by adding a `govuk-summary-list__key--one-quarter` class and put those styles in `app/assets/stylesheets/govuk-frontend/extensions.scss`

That's a quick guess at the changes you could make to achieve the changes as extensions on the summary-list component but worth noting I'd be happy for any to be changed, I just want the approach to match what we've done so far.

[^1]: To reduce the amount of components we are responsible for and allow the possibility of our submitting any extensions/changes upstream to design system.

tombye:
@klssmith @quis and me had a chat about the comments I made in https://github.com/alphagov/notifications-admin/pull/4281#issuecomment-1205094436 and decided:

## Changes that should apply to all summary lists on Notify

We think all these things should apply to all summary lists on Notify:
- adding a border to the top
- making the `<dt>`s non-bold
- increasing the row height (so height of `<dt>` and `<dd>`)

GOVUK Frontend have a [single responsibility principle](https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md#single-responsibility-principle) convention saying you should have one-class-per-element (sorry @klssmith @quis, I didn't know this when we chatted). Because of that, we could add classes to all elements involved, namespaced to `notify` instead of `govuk`, like:

```html
<dl class="govuk-summary-list notiy-summary-list">
  <div class="govuk-summary-list__row">
    <dt class="govuk-summary-list__key notify-summary-list__key">...</dt>
    <dd class="govuk-summary-list__value notify-summary-list__value">...</dd>
    <dd class="govuk-summary-list__value notify-summary-list__action">...</dd>
  </div>
</dl>
```

## Assigning 'column width'

We can use the proposed approach of giving the `<dt>` elements 25% width by using a `notify`-namespaced class that uses the units of width used by [the design system grid](https://design-system.service.gov.uk/styles/layout/#grid-system). So, to give the `<dt>`s 25% width:

```html
<dt class="govuk-summary-list__key notify-summary-list__key notify-summary-list__key--one-quarter">...</dt>
```

Worth noting that all the `<dt>`s and `<dd>`s have `display: table-cell` so any widths assigned don't include the padding. We are able to use [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) so it could be solved by something like `calc(25% - 20px)`.

tombye:
Update: I noticed design system have guidance for extending and modifying their code: https://design-system.service.gov.uk/get-started/extending-and-modifying-components/

We don't actually follow this in the code that's been written so far when extending/overriding GOVUK Frontend, because I didn't know it existed, but should do. The changes proposed here seem to align with it here so no **I think** don't need alterations but I thought it was worth adding here.

PR #4284, opened 2022-06-29
Add message count to link

karlchillmaid:
We got some feedback about the Received messages page from GOVwifi: “I didn’t realise you get a few days worth of responses when you download the CSV, I thought it was only the first 50 responses”

The current link `Download these messages` really doesn’t tell you that you’re going to get *everything*. There isn’t a count on the page, and you can’t see that it’s paginated until you scroll down. Even then you can’t see how many pages there are.

We should help.

At some point we might want to update our pagination to tell users how many pages there are, but for now we can probably just make the download link text better.

## Current page
Screenshot 2022-08-03 at 10 25 12

## New content

`Download all messages`

### DIFF
--- a/app/templates/views/dashboard/_inbox_messages.html
+++ b/app/templates/views/dashboard/_inbox_messages.html
@@ -4,7 +4,7 @@
 <div class="ajax-block-container">
   {% if messages %}
     <p class="bottom-gutter-2-3 top-gutter-1-2">
-      <a href="{{ url_for('.inbox_download', service_id=current_service.id) }}" download class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download {{ current_service.inbound_sms_summary.count|format_thousands }} messages</a>
+      <a href="{{ url_for('.inbox_download', service_id=current_service.id) }}" download class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download all messages</a>
     </p>
   {% endif %}
   {% call(item, row_number) list_table(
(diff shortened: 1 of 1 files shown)


app/templates/views/dashboard/_inbox_messages.html
```diff
@@ -4,7 +4,7 @@
 <div class="ajax-block-container">
   {% if messages %}
     <p class="bottom-gutter-2-3 top-gutter-1-2">
-      <a href="{{ url_for('.inbox_download', service_id=current_service.id) }}" download class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download these messages</a>
+      <a href="{{ url_for('.inbox_download', service_id=current_service.id) }}" download class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download {{ current_service.inbound_sms_summary.count|format_thousands }} messages</a>
```
karlchillmaid, line 7: Options:

- Download all messages
- Download 10,000 messages
- Download all 10,000 messages
- Download all messages from the last 7 days
- Download 10,000 messages from the last 7 days
- Download all 10,000 messages from the last 7 days

idavidmcdonald: I think this is a hard choice with no clear winner for me.

I think my top two preferences are  `Download all messages` and `Download all messages from the last 7 days`.

I think they may clear up the confusion with a minimum amount of words. I think adding the 10,000 part may not be needed (or at least we could avoid adding it until we have evidence that it isn't working).

Also you may be aware but 'the last 7 days' will need to be flexible per service due to their data retention schedule.

quis: _Download all messages_ would be my preference

karlchillmaid, line 7:
```suggestion
      <a href="{{ url_for('.inbox_download', service_id=current_service.id) }}" download class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">Download all messages</a>
```

samuelhwilliams (approved):
👍

PR #4285, opened 2022-07-01
Bug fix pivotal ticket 179643190

BlessedDev:
https://www.pivotaltracker.com/n/projects/1443052/stories/179643190

Currently a user without platform admin privileges cannot create a new service after their current service is inactive(has been deleted or suspended).

This can be recreated by creating a service and deleting it and immediately attempting to create a new service. Ensure that this is not done as a platform admin.
Screenshot 2022-06-30 at 16 18 52
Screenshot 2022-06-30 at 16 19 24
Screenshot 2022-06-30 at 16 19 47
Screenshot 2022-06-30 at 16 20 15
Screenshot 2022-06-30 at 16 20 30

This is due to check_inactive_service() https://github.com/alphagov/notifications-admin/blob/a92376ae8469d758da0062901e9a62df8fa193b7/app/notify_client/__init__.py#L47-L55

I have added a new method to unset a service that is inactive if it is the current service so that a non platform admin can now create a new service after the current service has been rendered inactive:
Screenshot 2022-06-30 at 16 23 11
Screenshot 2022-06-30 at 16 23 27
Screenshot 2022-06-30 at 16 24 00
Screenshot 2022-06-30 at 16 24 44
Screenshot 2022-06-30 at 16 25 07

The present course of action has rendered the check_inactive_service() redundant. Further discussion is needed as per wether to get rid of it and the test written for it which currently fails as expected.


app/notify_client/__init__.py
```diff
@@ -55,6 +68,7 @@ def check_inactive_service(self):
             abort(403)
 
     def post(self, *args, **kwargs):
+        self.unset_inactive_current_service()
```
CrystalPea, line 71: Doesn't executing this line make the line below it actually not check anything? Like, if you deleted `self.check_inactive_service()` instead of adding a new method, wouldn't it achieve the same result?

CrystalPea (commented):
Well done looking at this. This part of the code is a bit unusual, working with Flask "innards". I left a comment in the code, seems like what you did there could be possibly achieved in a simpler manner? But I may be wrong, I haven't touched this part of the code much - that is also why I hesitated before reviewing this PR earlier in the week.

Also `test_inactive_service_cannot_be_modified_by_normal_user` is now failing. I tried to look at what that test is actually testing, but it seems a bit arcane. The test name says it is testing that normal user cannot edit an archived service, which sounds like what we want, but I feel the test doesn't do exactly what it says it does - it seems to test performing operations on any current service, including any new service user wants to create.

If that's the case, maybe deleting `self.check_inactive_service()` check from that `post` method would be the right thing, but probably also we would still like to make sure that users cannot modify/use inactive services, maybe by stripping all user permissions when archiving a service in `dao_archive_service()` in the API repo?

I feel it would be useful if @leohemsted or @idavidmcdonald could weigh in on that one, since it's an area of code we don't work with much.

app/notify_client/__init__.py
BlessedDev: I agree It is a bit of an odd one at the moment. I personally favour getting rid of the check_inactive_service() method and the test associated with it, as my work renders it obsolete and I see no justification for it.

There is some value to unsetting unsetting the current service in a session if the service has been deactivated so it hasn't been a wasted effort.

It would be great to get @quis to weigh in as well.

app/notify_client/__init__.py
```diff
@@ -44,6 +45,18 @@ def _add_request_id_header(headers):
...
+        """
+        This method unsets the current service to None if the current service is inactive.
+        This is to fix the current issue where a new service is prevented from being created if the current service
+        is inactive
+        """
+        # this file is imported in app/__init__.py before current_service is initialised, so need to import later
+        # to prevent cyclical imports
+        from app import current_service
+        if current_service and not current_service.active:
+            _request_ctx_stack.top.service = None
```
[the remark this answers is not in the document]
leohemsted: This is called mid-way through many many many different functions. What are the implications of current_service being unset half-way through all those different code paths?

leohemsted (commented):
i'm a bit unsure about this.

currently, any time someone makes a stateful change to notify, we check that if `current_service` is set, then make sure it's active. If it's not (and the user isn't a platform admin) then we return a 403.

Instead, the proposal is to clear the current_service instead. This means the 403 check is reundant (as pea pointed out). But this also means the POST will always be attempted, regardless of whether the service is inactive. It used to always fail, now it will always succeed, and that's no good! That means someone could send a POST request (to, eg, send a notification, unsuspend the service, etc) and we would happily process it.

So we'll need to have another look at this.

Also, there's another indicator about this proposed flow being a bit odd - for an inactive service, before the POST request current_service would be set, we'd then clear it down during the POST, and then current_service would be unset. This feels like a big gotcha that I really don't feel comfortable with.

i'd much rather not change the current_service state mid-way through the endpoint at all.

------- 

At first I thought we could clear down the `current_service` at the beginning of the request if that request is not tied to a service (see `load_service_before_request`). However, that might have some implications for other places in the code (I think off the top of my head, we might lose the ability to show the "back to <service name>" button) so I'm not sure if that's viable.

This current check (`check_inactive_service`) is happening on every api client POST. 
Does it make more sense to just not call `check_inactive_service` when we're doing something like adding a service, which is a function that does not operate on a service at all?

There's also a real gotcha here which is that it's checking `current_service.active`, but there's no formal guarantee that the endpoint is operating on that particular service. Indeed, the bug was found because it was triggering on an endpoint that doesn't operate on any existing service. (this bug also manifests if for example you view a suspended service and then try and update your username).

Unfortunately, I can't think off the top of my head a nice way to separate all the various API endpoints in to things that affect a service (not just service endpoints, but also things like templates, api keys, inviting a user, etc) . Versus endpoints that explicitly aren't tied to a service (creating a new service, updating your user profile, organisation endpoints, etc).

But I think if we can crack that, that's the real distinction as to whether we should be calling `check_inactive_service` at all. Rather than fixing a faulty check, we shouldn't be performing that check at all if it's not necessary.

Sorry this has been super rambly as I've learnt more about what checks we have and why they're there.

Thoughts? Happy for us to chat about this in tech catchup tomorrow

app/notify_client/__init__.py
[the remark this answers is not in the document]
BlessedDev: I take your point regarding platform admin check and will add a check to the new function. That should address the issue raised as per platform admins.

I did consider moving the old function from the NotifyAdminApiClient post method as it called by several functions and the bug that was raised by this ticket was an unintended consequence of where it resides. I shall have another go at finding a more suitable place for the new code to reside.

tests/app/notify_client/test_notify_admin_api_client.py
```diff
@@ -42,6 +42,22 @@ def test_active_service_can_be_modified(notify_admin, method, user, service):
...
+        api_user_active
+):
+    api_client = NotifyAdminAPIClient()
+    method = 'post'
+    with notify_admin.test_request_context() as request_context, notify_admin.test_client() as client:
+        client.login(api_user_active)
+        request_context.service = Service(service_json(active=False))
+
+        with patch.object(api_client, 'request') as request:
+            getattr(api_client, method)('url', 'data')
```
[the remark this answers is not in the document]
BlessedDev: That is a good point. I will adjust it accordingly.

BlessedDev:
I have removed the "suspend a service and resume a service features.
There was a bit of clean up to do.
I have also done the same in [notifications-api](https://github.com/alphagov/notifications-api/pull/3579)

tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -149,14 +149,7 @@ def test_service_setting_toggles_show(
 
 @pytest.mark.parametrize('service_fields, endpoint, index, text', [
     ({'active': True}, '.archive_service', 0, 'Delete this service'),
-    ({'active': True}, '.suspend_service', 1, 'Suspend service'),
-    ({'active': True}, '.history', 2, 'Service history'),
-    ({'active': False}, '.resume_service', 0, 'Resume service'),
-    ({'active': False}, '.history', 1, 'Service history'),
```
leohemsted, line 155: i think test should still be here - but with index 0 instead of 1

tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -194,10 +187,7 @@ def test_service_settings_doesnt_show_option_if_parent_permission_disabled(
 
 
 @pytest.mark.parametrize('service_fields, link_text', [
-    # can't archive or suspend inactive service. Can't resume active service.
     ({'active': False}, 'Archive service'),
```
leohemsted, line 190: we can remove the parametrize entirely now if there's only one option. 

the test might need renaming too at that point

BlessedDev: That is a good point and I will look into it.

BlessedDev: It turns out that the test was a bad one as it should have been testing for the presence of a "Delete this service" link not "Archive service".
I have adjusted and renamed the test and also written a test for the reverse, that a "Delete this service" link toggle is displayed when a service is active.

tests/app/main/views/service_settings/test_service_setting_permissions.py
BlessedDev: @leohemsted 
I have renamed this test to compliment the other test you had suggested was in need of a new name.
Service History is not displayed for an archived service.

tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -167,6 +167,22 @@ def test_service_setting_link_toggles(
     assert link['href'] == link_url
 
 
+@pytest.mark.parametrize('service_fields, link_text', [
+    ({'active': False}, 'Delete this service'),
+    ({'active': False}, 'Service history')
```
leohemsted, line 172: the opposite to above - since we're always setting this to be False, we don't need this in the parametrize - we can just hardcode it on the `service_one.update` line

tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -149,16 +149,9 @@ def test_service_setting_toggles_show(
...
     ({'active': True}, '.archive_service', 0, 'Delete this service'),
-    ({'active': True}, '.suspend_service', 1, 'Suspend service'),
-    ({'active': True}, '.history', 2, 'Service history'),
-    ({'active': False}, '.resume_service', 0, 'Resume service'),
-    ({'active': False}, '.history', 1, 'Service history'),
-    pytest.param(
-        {'active': False}, '.archive_service', 2, 'Resume service',
-        marks=pytest.mark.xfail(raises=IndexError)
-    )
+    ({'active': True}, '.history', 1, 'Service history')
```
leohemsted, line 152: since we're always setting this to be True for this test now, we don't need this in the parametrize

also because `service_one` returns an active service, we don't need to call service_one.update at all

leohemsted (commented):
Just a couple of quick comments but think the tests look a lot cleaner now

tests/app/main/views/service_settings/test_service_setting_permissions.py
BlessedDev: In addition to the changes requested I have made a few more changes so that the test is more robust

klssmith:
I'm coming at this without having been at the previous discussions, so might be missing a lot of context or background information here. I like removing the suspend and resume service code, which seems like a good change since we don't really use it.

The change that I'm not certain about (even though it fixes the bug) is the removal of the `check_inactive_service` function that is called before all `POST`, `PUT` or `DELETE` requests. There are two ways to make a service inactive - either by suspending it or by archiving it. We can still make a service inactive by archiving it but admin no longer stops you making a POST request from an archived service. For example, when I tried to do that locally I could send an email from an archived service in admin, and API then errored, put the email in technical-failure, and put it on the retry queue.

BlessedDev:
@klssmith I cannot send an email from an archived service in admin locally
archived service

The check_inactive_service method sought to prevent non platform admins from altering the state of the suspended service. It had the unintended effect of preventing them from making any POST requests, and consequently preventing them from creating a new service which was the initial bug report.

As things stand a non platform admin cannot alter an archived service. I suppose the method could be altered to check the state of the service and if it is not active, then prevent it from making a POST, PUT or DELETE request as a defensive measure. In that case check_inactive_service would be true to it's name.

klssmith:
It's not normally possible to make a POST request from an inactive service since the "Send" button disappears. I should have explained that the way I managed it was having an email ready to send in tab 1 (email address and any placeholders filled in, and just needing the "Send" button to be pressed). Then in tab 2 I archived the service. I then went back to tab 1 and hit "Send". Before this change, admin would have prevented that, but now it lets the request reach api.

I think the chance of anyone doing that by mistake is really slim. What I was less sure about is whether it's ok security-wise that admin no longer checks if a service is active before doing certain actions or if there was a way that someone could deliberately exploit that. Let's see what @idavidmcdonald thinks next week

BlessedDev:
It has been decided that due to the complexity of this issue that for now we will make do with removing the ability to suspend and resume a service. The check_inactive_service will be left intact for now.

tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -147,33 +147,40 @@ def test_service_setting_toggles_show(
...
 
 
+def test_service_setting_links_not_displayed_for_archived_services(
+    get_service_settings_page,
+    service_one,
+):
+    # These are the precise representations of the text for "Delete this service"
+    # and "Service History" and this is essential for the test to be valid
+    link_texts = [
+        '\n            Delete this service\n          ',
```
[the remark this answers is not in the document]
BlessedDev: That was an error, they should be two different items. That has been fixed.
I have also applied .strip() to make it neater.

samuelhwilliams (approved):
👍

PR #4297, opened 2022-07-22
Only show items not in branding pool on add branding pool page

klssmith:
The page to add to the email branding pool for an organisation was showing all email branding, with the items already in the pool checked. We now only want to show items that are not in the pool, so this changes the page to only show branding that doesn't already exist in the pool and can be added.

The page title has also been updated.

[Pivotal story](https://www.pivotaltracker.com/story/show/182771038)

---
Screenshot of updated page

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -455,14 +455,11 @@ def organisation_email_branding_options(org_id):
     form.branding_field.choices = [
         (branding['id'], branding['name'])
         for branding in email_branding_client.get_all_email_branding(sort_key='name')
-    ]
-    form.branding_field.data = [
-        branding['id']
-        for branding in current_organisation.email_branding_pool
+        if branding not in current_organisation.email_branding_pool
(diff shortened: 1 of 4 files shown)


BlessedDev (approved):
Looks good to me.
One issue to be considered for future work in regards to email branding options, is that because default branding are nominally not part of the email branding pool, they are essentially now available to be added to the pool even though they are already available to the organisation as the default branding.

klssmith:
Thanks @BlessedDev. I think it's fine that default branding can be added to the pool - the later work will add the default branding to the pool for each organisation in a migration anyway. I've double checked the code just to see whether adding it now would cause any issues, but I think it's fine. If a service did already have the default branding for its organisation, then we wouldn't show that option to them again: https://github.com/alphagov/notifications-admin/blob/06d13b5c1a08c7dc4bcfc28fd94d5839da81f714/app/utils/branding.py#L23-L25

PR #4300, opened 2022-07-27
Make it more obvious that ‘send files’ is API-only

karlchillmaid:
This PR adds a line to try and make it clearer that ‘send a file by email’ is an API-only feature.

We get questions about this through support, so we’re changing the content here to match what we say in the Zendesk macro.


BlessedDev (approved):
Looks good to me.

PR #4301, opened 2022-07-28
Remove GOVUK Elements [Part 1]

tombye:
This is part 1 of a series of work to remove the GOVUK Elements frontend library from our codebase:

These changes:
- replace [the GOVUK Elements inset text styles](https://govuk-elements.herokuapp.com/typography/#typography-inset-text) with [the GOVUK Frontend inset text component](https://design-system.service.gov.uk/components/inset-text/) or, when they are in the conditional sections of radios replace those radios with [the GOVUK Frontend radios component](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question)
- move [the GOVUK Elements table styles](https://govuk-elements.herokuapp.com/data/) into our codebase

### Moving GOVUK Elements tables CSS into admin

We have stories to convert all our tables to use GOVUK Frontend components (see[^1] and[^2]) so it isn't worth rewriting the GOVUK Elements styles we currently use until that work happens. Because of that, this just moves those styles into our codebase. We added comments to both stories including that work in each story.

### Making some conditional radios use the GOVUK Frontend radios component

Some uses of the GOVUK Elements inset text style were in groups of radios where selecting an option reveals some content styled like inset text. [The GOVUK Frontend radios component has this style built in](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-a-related-question) so we have *just* swapped out the existing radios code for the GOVUK Frontend version.

#### Design changes from move to GOVUK Frontend radios

Reviewers (probably @quis) might be interested to know that the GOVUK Frontend radios styling changes how the conditional sections look:

##### Before

conditional_radios_current

##### After

conditional_radios_govuk_frontend

##### Before

conditional_radios_current_with_errors

##### After

conditional_radios_govuk_frontend_with_errors

[^1]: https://www.pivotaltracker.com/story/show/170030357
[^2]: https://www.pivotaltracker.com/story/show/182595331

### DIFF
--- a/app/assets/stylesheets/govuk_elements/tables.scss
+++ b/app/assets/stylesheets/govuk_elements/tables.scss
@@ -1,0 +1,57 @@
+// Tables
+// ==========================================================================
+
+table {
+    border-collapse: collapse;
+    border-spacing: 0;
+    width: 100%;
+
+    th,
+    td {
+      @include core-19;
+      padding: em(12, 19) em(20, 19) em(9, 19) 0;
+
+      text-align: left;
+      border-bottom: 1px solid $border-colour;
+    }
+
+    thead th {
+      font-weight: 700;
+    }
+
+    td:last-child,
+    th:last-child {
+      padding-right: 0;
+    }
+
+    // Right align table header cells and table cells with a numeric class
+    .numeric {
+      text-align: right;
+    }
+
+    // Allow a qualifying element, only table data cells should use tabular numbers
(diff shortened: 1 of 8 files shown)


tombye:
Chatted with @quis about these changes out of Github. He's happy with the new styles and pointed out they match those we have in our 'tree' of radios elsewhere:

Tree of radio buttons, from the Notify interface

BlessedDev (approved):
Looks good to me.

PR #4302, opened 2022-07-28
Remove GOVUK Elements [Part 2]

klssmith:
This is part 2 of a series of work to remove the GOVUK Elements frontend library from our codebase:

These changes
* Delete a few styles that weren't being used
* Replace the [GOV.UK Elements list styles](https://github.com/alphagov/govuk_elements/blob/v3.1.2/packages/govuk-elements-sass/public/sass/elements/_lists.scss) with the [GOV.UK Frontend list styles](https://design-system.service.gov.uk/styles/typography/#lists)

--- 
Merge after
- [x] https://github.com/alphagov/notifications-admin/pull/4301


tombye (approved):
I didn't work directly on this so can review. Looks good to me 👍🏻 

Also, this will mean less code sent to users so 🚀.

PR #4304, opened 2022-07-30
Remove GOVUK Elements [Part 3]

tombye:
This is part 3 of a series of work to remove the GOVUK Elements frontend library from our codebase:

These changes remove all the forms styles that weren't being used and copy all those that were into our codebase - this will need to be removed separately.

Paired with @klssmith 

---

Merge after

- [x] https://github.com/alphagov/notifications-admin/pull/4301
- [x] https://github.com/alphagov/notifications-admin/pull/4302

### DIFF
--- a/app/assets/stylesheets/govuk_elements/_forms.scss
+++ b/app/assets/stylesheets/govuk_elements/_forms.scss
@@ -1,0 +1,239 @@
+// Forms
+// ==========================================================================
+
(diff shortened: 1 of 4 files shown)


samuelhwilliams (approved):
Looks reasonable to me, and seems good locally from some semi-random looking around... 👍

PR #4306, opened 2022-08-01
Make form to set email branding pool for an org work

klssmith:
We had a form to let you pick email branding to add to the branding
pool for an org, but the functionality had not been implemented. This
adds the button and the POST endpoint for the form. At the moment, we
redirect to the organisation settings page when the form is successfully
submitted. Once the page to preview all email brandings in the pool for
an org exists, we can update the code to redirect there instead.

[Pivotal story](https://www.pivotaltracker.com/story/show/182771318)

**Merge after**
- [x] https://github.com/alphagov/notifications-api/pull/3583

### DIFF
--- a/app/notify_client/organisations_api_client.py
+++ b/app/notify_client/organisations_api_client.py
@@ -61,6 +61,13 @@ class OrganisationsClient(NotifyAdminAPIClient):
 
         return api_response
 
+    @cache.delete('organisation-{org_id}-email-branding-pool')
+    def add_brandings_to_email_branding_pool(self, org_id, branding_ids):
+        return self.post(
+            url=f"/organisations/{org_id}/email-branding-pool",
+            data={'branding_ids': branding_ids}
+        )
+
     @cache.delete('service-{service_id}')
     @cache.delete('live-service-and-organisation-counts')
     @cache.delete('organisations')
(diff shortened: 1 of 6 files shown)


app/notify_client/organisations_api_client.py
```diff
@@ -61,6 +61,13 @@ def update_organisation(self, org_id, cached_service_ids=None, **kwargs):
 
         return api_response
 
+    @cache.delete('organisation-{org_id}-email-branding-pool')
+    def add_branding_to_email_branding_pool(self, org_id, branding_ids):
```
idavidmcdonald, line 65: Optional: Wondering as this takes a list we should name this more explicitely, somethign like `add_brandings_to_email_branding_pool` or `add_list_of_email_branding_to_pool`

I'll leave you to ponder and decide if you think something like that is an improvement (or any other name you can think of that may be better).

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -1747,6 +1747,59 @@ def test_organisation_email_branding_options_shows_branding_not_in_branding_pool
...
+    )
+    assert normalize_spaces(
+        page.select_one('.govuk-error-message').text
+    ) == 'Error: Select at least 1 email branding option'
+
+
+@pytest.mark.parametrize('branding_ids_added, flash_message', [
+    (['1', '2'], '2 email branding options added'),
+    (['1'], '1 email branding option added'),
+])
```
idavidmcdonald, line 1776: Good test cases

idavidmcdonald (approved):
Great work! Looks good!

app/notify_client/organisations_api_client.py
klssmith: This is now updated to `add_brandings_to_email_branding_pool` 👍

PR #4307, opened 2022-08-01
Support API errors for `email_from`

samuelhwilliams:
We are updating the notifications-api to reject `email_from` values
which contain 'uncleanable' characters (basically anything that can't be
reduced to alphanumeric characters or full-stops).

This is because AWS SES doesn't allow sending rich unicode characters in
the local portion of an email address. We could punycode the values, but
this looks very odd in emails and so for now we're leaving this alone -
to be addressed potentially in the future. We're also leaving it as a
future endeavour to allow separating service name and email_from, due to
the need for research+design.

**NOTE**: ~The error message content is a placeholder and waiting for input from @karlchillmaid.~ Error message content has now come from Karl.

This PR can go out any time ahead of the corresponding enforcement PR for notifications-api

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -611,5 +611,16 @@ class ServiceAPIClient(NotifyAdminAPIClient):
         count = redis_client.get(daily_limit_cache_key(service_id)) or 0
         return int(count)
 
+    @classmethod
+    def parse_edit_service_http_error(cls, http_error):
+        """Inspect the HTTPError from a create_service/update_service call and return a human-friendly error message"""
+        if http_error.message.get('email_from'):
+            return "Service name must not include characters from a non-Latin alphabet"
+
+        elif http_error.message.get('name'):
+            return "This service name is already in use"
+
+        return None
+
 
 service_api_client = ServiceAPIClient()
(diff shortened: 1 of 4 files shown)


app/notify_client/service_api_client.py
```diff
@@ -611,5 +611,16 @@ def get_notification_count(self, service_id):
         count = redis_client.get(daily_limit_cache_key(service_id)) or 0
         return int(count)
 
+    @classmethod
+    def parse_edit_service_http_error(cls, http_error):
+        """Inspect the HTTPError from a create_service/update_service call and return a human-friendly error message"""
+        if http_error.message.get('email_from'):
```
klssmith, line 617: This will work for any errors that come from the Marshmallow schemas in api since `http_error.message` will be a dict, but if there are any other errors raised where `http_error.message` is a string then this line raises `AttributeError: 'str' object has no attribute 'get'`. We don't actually expect other error types from API though, so might be ok?

samuelhwilliams: I believe this is OK. Currently we are performing hard item retrieval (eg via `['name']`) on any 400s, and that doesn't appear to be an issue.

klssmith (approved):

PR #4308, opened 2022-08-03
Prefer `page.select` to `page.find` in tests

quis:


idavidmcdonald (approved):
My review has not been very thorough but I'm happy with the intention, this is only changing the tests, the tests still pass and from a good skim it looks correct.

PR #4310, opened 2022-08-04
File upload button changes

tombye:
Fix for accessibility issues with our file upload component, spotted in a previous accessibility audit. Also rolls in changes to the cancel button, to fix another accessibility issue found along the way.

https://www.pivotaltracker.com/story/show/177000534

This work was originally attempted in https://github.com/alphagov/notifications-admin/pull/4268. This takes the resulting changes but merges everything, including suggested corrections, into 2 commits:
- fixing the issue raised in our audit
- fixing the cancel button

## The current approach

It's worth explaining how the existing approach works.

upload_current_state_diagram

It involves the input[type=file] element being visually hidden and it having 3 labels:
1. the real label for the input[type=file] element. It's only visually hidden so it still applies
2. a label styled as a button which:
    - when clicked, activates the input[type=file] element
    - is styled to look focused when the input[type=file] element is focused so it looks like you can tab to it
3. a label that used to hold the name of the file selected. Now redundant since [we made the form submit when the file has been chosen](https://github.com/alphagov/notifications-admin/commit/97a3bf922580f61c81f386f9665a3f05724f4ec4)

This works for visual users with mouse/pointers and keyboard users because clicking on the button activates the file chooser:

show_click_on_current_upload_button

...and it 'looks like' you can tab to it:

show_tab_to_current_upload_button

The problems with it are:
- assistive technologies have problems with multiple labels
- the label styled as a button is announced as a label by screen readers, not a button
- saying 'Click upload file' to speech recognition software, to click a button with 'upload file' text, won't work because it isn't a button
- moving to the input[type=file], which makes the label button look focused, announces the input[type=file] element, which doesn't match the visuals

## Proposed solution

We could keep patching these issues individually: giving the label a 'button' role, hiding the input[type=file] element, etc... but I think it's simpler to just hide all the initial elements and add a real `<button>`. In full, this proposal is to:
1. hide all the existing elements in the upload form, visually and semantically
5. add a button
6. use JS to make clicks on the button click the input[type=file]

Effectively, all that's left in the page is a button and because it's a real `<button>` element it works as expected with all assistive technologies.

## Fixing the cancel button

The cancel button is actually an `<a>` tag, styled to look like a button. This gives it button semantics and uses JS to give it the behaviours expected of a button.

## How do I check this?

I recommend reviewing commit-by-commit. I've checked it in all [the browsers required by the service manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2021) and [all the assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with).

The upload field used when adding a new email brand is a good page to test on.

It's also worth checking without JS and, if you feel comfortable doing it, in a screen reader. I'd also be happy to demo it with some assistive technologies if that would help.

Finally, it's also worth considering that @benthorner did OK the changes here in https://github.com/alphagov/notifications-admin/pull/4268, albeit in a slightly different form.

### DIFF
--- a/app/assets/javascripts/fileUpload.js
+++ b/app/assets/javascripts/fileUpload.js
@@ -5,15 +5,30 @@
 
     this.submit = () => this.$form.trigger('submit');
 
-    this.showCancelButton = () => $('.file-upload-button', this.$form).replaceWith(`
-      <a href="" class='govuk-button govuk-button--warning govuk-!-margin-right-1'>Cancel upload</a>
-    `);
+    this.addCancelButton = () => {
+
+      var $cancelButton = $(`
+        <a href="" role="button" class="file-upload-button govuk-button govuk-button--warning">
+          Cancel upload
+        </a>
+      `);
+
+      $('button[type=button]', this.$form).replaceWith($cancelButton);
+
+      // add GOVUK Frontend behaviours
+      new window.GOVUK.Frontend.Button(this.$form[0]).init();
+
+      // move focus to the cancel button, it is lost when the upload button is removed
+      $cancelButton.focus();
+
+    };
 
     // Add a button that passes a click to the input[type=file]
     this.addFakeButton = function () {
+
       var buttonText = this.$field.data('buttonText');
       var buttonHTMLStr = `
-        <button type="button" class="govuk-button govuk-!-margin-right-1" id="file-upload-button">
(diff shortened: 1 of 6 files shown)


samuelhwilliams (approved):
Thanks for the run-through and demo 👍  This all looks good to me, nice.

PR #4313, opened 2022-08-09
Add sticky footer for the form to add email branding to pool

klssmith:
There are a lot of email brandings in production, so this changes the
form to add branding for an org so that the "Add" button is in a
sticky-footer. Without this change, you have to scroll a long way down
each time you need to add a brand.


samuelhwilliams (approved):
I don't know how to test this locally :( 

But it does look like it should do the job regardless 🤞

PR #4314, opened 2022-08-09
Add police force areas in England and Wales to the library of areas you can send an emergency alert to

quis:
In some areas it's more intuitive to select the area covered by the police force, rather than a local authority boundary.

This is often the case with cities, for example:
- Greater Manchester can’t be directly selected from our library, instead you need to add City of Manchester, Salford, etc.
- London can’t be directly selected from our library, instead you need to add Hackney, Tower Hamlets, etc.

The ONS publishes the [boundaries for police force areas](https://geoportal.statistics.gov.uk/datasets/ons::police-force-areas-december-2020-ew-bgc). They don’t have multiple levels or hierarchies, so we can import them in much the same way as we do for countries. Here’s the new library in the list:

image

And here’s what it looks like selecting a couple of areas in Wales

image

This also has the benefit of having a library of areas which aligns with the areas covered by each Local Resilience Forum – they are the same. There is one exception, London, where the London Resilience Forum covers both the Metropolitan Police area and the City of London Police area.

As requested, we are merging these two areas into one. This should stop someone picking ‘Metropolitan’ and accidentally leaving a hole in the middle for the City of London (it’s easy to miss this on the map). 

Here’s what it looks like picking the single, combined, area:

image


samuelhwilliams (approved):
Tested, looks good 👍

PR #4315, opened 2022-08-09
Create Dockerfile for admin app

rjbaker:
Adds a Dockerfile and various minor app changes to support running Admin in Docker.

We use a [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) to build frontend/Node dependencies first before copying these into Python stage.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -11,7 +11,7 @@ if os.environ.get("VCAP_APPLICATION"):
 
 class Config(object):
     ADMIN_CLIENT_SECRET = os.environ.get("ADMIN_CLIENT_SECRET")
-    API_HOST_NAME = os.environ.get("API_HOST_NAME", "http://localhost:6011")
+    API_HOST_NAME = os.environ.get("API_HOST_NAME")
     SECRET_KEY = os.environ.get("SECRET_KEY")
     DANGEROUS_SALT = os.environ.get("DANGEROUS_SALT")
     ZENDESK_API_KEY = os.environ.get("ZENDESK_API_KEY")
@@ -110,6 +110,7 @@ class Development(Config):
     ADMIN_CLIENT_SECRET = "dev-notify-secret-key"
     DANGEROUS_SALT = "dev-notify-salt"
     SECRET_KEY = "dev-notify-secret-key"
+    API_HOST_NAME = os.environ.get("API_HOST_NAME", "http://localhost:6011")
     ANTIVIRUS_API_HOST = "http://localhost:6016"
     ANTIVIRUS_API_KEY = "test-key"
     ANTIVIRUS_ENABLED = os.getenv("ANTIVIRUS_ENABLED") == "1"
--- a/app/config.py
+++ b/app/config.py
@@ -70,7 +70,7 @@ class Config(object):
     ANTIVIRUS_ENABLED = True
 
     REDIS_URL = os.environ.get("REDIS_URL")
-    REDIS_ENABLED = os.environ.get("REDIS_ENABLED") == "1"
+    REDIS_ENABLED = False if os.environ.get("REDIS_ENABLED") == "0" else True
 
     ASSET_DOMAIN = ""
     ASSET_PATH = "/static/"
--- a/app/config.py
+++ b/app/config.py
@@ -70,7 +70,7 @@ class Config(object):
     ANTIVIRUS_ENABLED = True
 
     REDIS_URL = os.environ.get("REDIS_URL")
-    REDIS_ENABLED = True
+    REDIS_ENABLED = os.environ.get("REDIS_ENABLED") == "1"
 
     ASSET_DOMAIN = ""
     ASSET_PATH = "/static/"
@@ -90,6 +90,8 @@ class Config(object):
             "first.last@digital.cabinet-office.gov.uk",
         ],
     }
+
+    NOTIFY_RUNTIME_PLATFORM = os.environ.get("NOTIFY_RUNTIME_PLATFORM", "paas")
 
 
 class Development(Config):
@@ -116,6 +118,7 @@ class Development(Config):
 
     REDIS_URL = "redis://localhost:6379/0"
     REDIS_ENABLED = os.environ.get("REDIS_ENABLED") == "1"
+    NOTIFY_RUNTIME_PLATFORM = "local"
 
 
 class Test(Development):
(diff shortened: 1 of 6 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM node:16-alpine as frontend-build
+
+WORKDIR /usr/frontend
+COPY app app
+COPY package-lock.json package.json gulpfile.js ./
+
+RUN npm install \
```
idavidmcdonald, line 10: Assuming this dockerfile is to build an image to be deployed into production (I think it should be) then we should use `npm ci` here instead of `npm install` please (https://betterprogramming.pub/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26)

Also worth looking at the audit flag for `npm ci` as somewhere in one of our docker files somewhere in the Notify ecosystem we specify that it should be set to False. Maybe we should do the same here because it will get picked up somewhere else (if we build it that way)?

https://docs.npmjs.com/cli/v8/commands/npm-ci#audit

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    && npm run build
+
+FROM python:3.9-slim-bullseye as python-build
+
+ENV DEBIAN_FRONTEND=noninteractive
```
idavidmcdonald, line 15: I also came across `PYTHONUNBUFFERED` in the template preview dockerfile.

Is it relevant for us?
https://stackoverflow.com/questions/59812009/what-is-the-use-of-pythonunbuffered-in-docker-file

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    build-essential \
```
idavidmcdonald, line 19: I got this image to build without the `build-essential` package. 

Supposedly if we don't plan to build debian packages, we don't need it (I'm guessing this might be the case for us)
https://packages.debian.org/sid/build-essential

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    git \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/* /tmp/*
+
+RUN pip install --upgrade pip
```
idavidmcdonald, line 24: Optional as I don't understand this stuff well enough myself.

Is there a particular reason we want to upgrade pip here?

If we don't, then we will get the deterministic version from the base docker image (currently pip version 21.2.4) or we can let it upgrade to the very latest by calling `pip install --upgrade pip`. Which do we prefer?

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN groupadd -r notify && useradd -r -g notify notify
+
+WORKDIR /home/vcap/app
```
idavidmcdonald, line 27: I'm fine with this but technically we no longer need to place our app in this directory (I think). Don't know if it is actually worth breaking from convention though.

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+COPY requirements.txt ./
+
+RUN echo "Installing python dependencies" \
+    && pip install -r requirements.txt
+
+COPY app app
+COPY application.py gunicorn_config.py ./
+COPY --from=frontend-build /usr/frontend/app/static app/static
```
idavidmcdonald, line 36: In production we will be serving the admin apps static assets directly from S3 instead of from the flask app.
(https://github.com/alphagov/notifications-aws/blob/8be0bd1aa5ed0d0668c84947b1193160e8ccffa1/concourse/templates/admin.yml.j2#L48)

For the moment though this is OK and something we can change later on when we look at a pipeline to deploy this image.

Maybe it is worth adding to TODO comment here for this particular bit (and also for the other commit that sets the `NOTIFY_ENVIRONMENT` as `development` so that we look for static files locally)

idavidmcdonald (commented):
Generally looks good I think. 

I'd like another person to take a look (maybe we should even mob on a review of this for learning purposes) because I'd like our first proper docker app for ECS to start with a great dockerfile as most of the others will probably copy most of its precedence!

docker/Dockerfile
sakisv: It looks like it is, yes

docker/Dockerfile
sakisv: Newer versions may come with different dependency resolution algorithms. This may be ok, but I'd rather aim for more reproducibility.

docker/Dockerfile
sakisv: The context and part of the reasoning here is a discussion we had at some point, about slowly getting containers running on PaaS instead of buildpacks. This would give us a package that we run on either platform. 

Not sure if we're actually going to do it, but this will allows us to do so if we want.

docker/Dockerfile
sakisv: > For the moment though this is OK and something we can change later on when we look at a pipeline to deploy this image.

I think I'd prefered if we kept it like this, i.e. a docker image which contains _everything_ that make up this particular version of our app. In the pipeline we could have a step that builds the image using this entire Dockerfile, and the next step we'd run `make upload-static` within that image. 

As for the `NOTIFY_ENVIRONMENT` yes, we'd need to change it passing in the appropriate build arg[^1]. Keeping a default value of `development` should make this dockerfile reusable for local dev too, or at least help in that direction.

[^1]: https://docs.docker.com/engine/reference/builder/#arg

docker/Dockerfile
idavidmcdonald: Good idea

sakisv:
To add on the commit message on https://github.com/alphagov/notifications-admin/pull/4315/commits/00ca5ad04de2c505eb87522547499954ac79beaa, the left side of the image is the resulting css with `NOTIFY_ENVIRONMENT==preview` and right side with `NOTIFY_ENVIRONMENT==development`

Screenshot 2022-10-24 at 17 16 42

idavidmcdonald (approved):

app/config.py
```diff
@@ -70,7 +70,7 @@ class Config(object):
     ANTIVIRUS_ENABLED = True
 
     REDIS_URL = os.environ.get("REDIS_URL")
-    REDIS_ENABLED = True
+    REDIS_ENABLED = os.environ.get("REDIS_ENABLED") == "1"
```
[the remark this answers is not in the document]
sakisv: Yeah it's safer to default to an enabled redis 👍

PR #4318, opened 2022-08-16
Update geodata for local authorities and electoral wards to the latest versions

quis:
In https://github.com/alphagov/notifications-admin/commit/76244d8c07c3b9d2c7b5bcf6de2e19df767e1fa2 we said:

> At the moment there are some areas which have:
> - a `count_of_phones` value of `None`
> - no sub-areas
>
> This is wrong, but until we fix the data the phone counting code needs to handle this.

We added some code to make these areas return a count of `0` (https://github.com/alphagov/notifications-admin/pull/3617#issuecomment-694135645):

> image

The problem was that the data we were using for the electoral wards was from 2020 while the data we were using for local authorities was from 2019.

This means that local authorities created in 2020 would be missing from our dataset (or we would would have their predecessors) so it was not possible to map their electoral wards to the parent authority. This meant that the local authority appeared not to have any electoral wards.

***

The fix for this is to update our datasets the 2021 releases, which are the latest. Importantly this means that all the local authority datasets are now from the same year, thereby fixing the mismatch:

Before | After
---|---
image | image

***

Using outdated datasets meant that local authorities created in 2021 didn’t exist at all in our database. Now they do:

Before | After
---|---
image | image
image | image

***

For areas that were previously missing electoral wards, the count of phones now works properly: 

image

***

In bringing in the new data there were 3 places which we had to tweak it manually to avoid invalid geometries. In each case one point was manually deleted from the source data. These areas are:

### Bathavon South

This 🍩 problem was also present in the previous dataset, and we fixed it in 779ac74fc726aa1f08be12589555129a5d03dbae

Source | Fixed
---|---
image | image

### Tolleshunt D’Arcy

This is a new polygon which has the same 🍩 problem as Bathavon South

Source | Fixed
---|---
image | image

### Oakham South

This one had two points from separate polygons very close together (not visible in this screenshot), which caused an artefact when merging into a single polygon during our simplification process.

Source | Fixed
---|---
image | image

***

I’ve broken this PR down into several commits. They are not atomic, but hopefully help explain the steps I took to fix this issue.

***

https://www.pivotaltracker.com/story/show/174837293


rjbaker (approved):

rjbaker (approved):
Looks good. Well done sorting the donut-shaped and other weird polygons.

PR #4328, opened 2022-08-22
Platform admin page for email branding pool options

BlessedDev:
This adds an admin page to display email branding pool options as specified in https://www.pivotaltracker.com/n/projects/1443052/stories/182596379

We start with a scenario where no email branding options have been added.
Screenshot 2022-08-22 at 15 56 16
 In that case only the default email branding is displayed
Screenshot 2022-08-22 at 15 13 33
Clicking the "Add other options" button takes the user to the "Add email branding options" page
Screenshot 2022-08-22 at 15 13 55
Clicking "Add", adds the selected option "een" to the email branding pool and the user is returned to the email branding admin page where confirmation of the added option is displayed and the user can now see the preview of the option "een"
Screenshot 2022-08-22 at 15 14 21
Screenshot 2022-08-22 at 15 56 16

The make default and remove buttons for the pool options are now to be added as part of another story.


app/templates/components/page-footer.html
```diff
@@ -19,6 +21,9 @@
       {% set button_data = {"text": button_text, "classes": "page-footer__button"} %}
       {% if destructive %}{% set _ = button_data.update({"classes": "govuk-button--warning page-footer__button"}) %}{% endif %}
       {% if centered_button %}{% set _ = button_data.update({"classes": "page-footer__button--centred"}) %}{% endif %}
+      {% if secondary_button and secondary_button_url%}{% set _ = button_data.update({"classes": "govuk-button--secondary", "type": "button",
```
samuelhwilliams, line 24: Think there was a question around whether we need the `secondary_button_url` part - so maybe we remove that for now? Don't mind too much either way.

samuelhwilliams (commented):
Looks good @BlessedDev - just a few comments :)

app/templates/components/page-footer.html
CrystalPea: yeah, if it's not being used I will remove it 👍🏼

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -0,0 +1,48 @@
...
+{% block backLink %}
+    {{ govukBackLink({ "href": url_for('.organisation_settings', org_id=current_org.id) }) }}
+{%  endblock %}
+
+{% block maincolumn_content %}
+    {{ page_header(page_title) }}
+
+    <!-- display default branding and it's preview -->
+    <p class="govuk-body">
+      <h2 class="heading-small">{{ default_email_branding_name }} Default</h2>
```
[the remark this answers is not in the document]
CrystalPea: I think we have similar formatting for reply-to email addresses so I will take the pattern from there 👍🏼

app/templates/components/page-footer.html
CrystalPea: It is actually being used for `Add other options` button, it's a gov-uk class that makes a button grey instead of green. I misunderstood Gozie during the tech catch-up the other week, I thought he is making it for some later use, but actually it is needed in this PR, so we should keep it. 

Good work staying vigilant though ☺️

samuelhwilliams (approved):
Good work on the updates 👍  Think this is good to go out now.

PR #3164, opened 2019-10-28
Update static error pages

tombye:
Updates the HTML pages we use for errors on the infrastructure that runs this app, outside of the app.

https://www.pivotaltracker.com/story/show/169344247

This updates the following pages:
- `paas-failwhale/*`, used by the PaaS when maintenance means the app needs to be taken offline
- `app/assets/error_pages/5xx.html`, [used by the server for its 500 errors](https://github.com/alphagov/notifications-aws/blob/f63d61ad03586b2a6a15dcb0d8d4d174d2d8520c/ansible/roles/paas-proxy/templates/admin.conf.j2#L25)

It updates them to use the new GOV.UK Frontend template and all associated assets.

## PaaS Failwhale maintenance page

failwhale

This page keeps its own assets in a folder called `static_503` so the whole lot needs to be present for it to work.

## Reverse-proxy 5xx page

5xx html

This page references all its assets from the same `/static` route the admin app uses so doesn't need any assets, just the HTML.

## Deployment

I'm not sure how these get deployed but I've put together a branch with the proposed changes to our AWS configuration:

https://github.com/alphagov/notifications-aws/compare/update-static-error-pages

## How do you test this?

I've written up some tips on how to test these pages locally:

https://gist.github.com/tombye/0ee813aaec5a2788a312ec0b96ac08b0

## Content

~~Would be worth @karlchillmaid checking the content while they're being changed.~~

PR #4330, opened 2022-08-23
Remove GOVUK Elements [Part 8]

tombye:
This is part 8 of a series of work to remove the GOVUK Elements frontend library from our codebase:

The intention of these changes was originally to remove all GOVUK Elements CSS from the page we would use if the app ever needed to be taken offline for maintenance (see https://github.com/alphagov/notifications-admin/pull/3164 for more details). Because this page only contains a few bits of basic content I decided to just redo its CSS to only include styles from GOVUK Frontend, the library for the GOV.UK Design System. Full details are in the commits.


paas-failwhale/index.html
```diff
@@ -16,8 +16,7 @@
     <link rel="apple-touch-icon" sizes="152x152" href="/static_503/images/govuk-apple-touch-icon-152x152.png?40846d46aa37232e2b35065769ce567c">
     <link rel="apple-touch-icon" href="/static_503/images/govuk-apple-touch-icon.png?3c45daa1f1e716458e4da954b973002e">
 
-    <link rel="stylesheet" media="screen" href="/static_503/stylesheets/main.css?d077f86473501ab244de0b30600536ee" />
-    <link rel="stylesheet" media="print" href="/static_503/stylesheets/print.css?28010888ca5719dc83d7c2c80f6ed2b2" />
+    <link rel="stylesheet" media="all" href="/static_503/stylesheets/main.css?d077f86473501ab244de0b30600536ee" />
```
klssmith, line 19: If we're using this query string for cache clearing purposes (I'm guessing?) do we need to change it now that the file has changed?

klssmith (commented):
Looks like we do use print.css [here](https://github.com/alphagov/notifications-admin/blob/d25cf3030296a93569d77c0e8b3403988683098c/app/assets/error_pages/5xx.html#L20) as well as for the failwhale - not sure if that means we do still need it after all or if it's still fine to get rid of and to remove all references to. Although [this is our main 500 page](https://github.com/alphagov/notifications-admin/blob/d25cf3030296a93569d77c0e8b3403988683098c/app/templates/error/500.html) 🤔

tombye:
> Looks like we do use print.css [here](https://github.com/alphagov/notifications-admin/blob/d25cf3030296a93569d77c0e8b3403988683098c/app/assets/error_pages/5xx.html#L20) as well as for the failwhale - not sure if that means we do still need it after all or if it's still fine to get rid of and to remove all references to. Although [this is our main 500 page](https://github.com/alphagov/notifications-admin/blob/d25cf3030296a93569d77c0e8b3403988683098c/app/templates/error/500.html) 🤔

I'm pretty sure those error pages are served by the same server which serves our static assets and use the same stylesheets as our app. Our app still uses separate stylesheets for screen and print styles so we still need `print.css`. The failwhale is a static app, served by PaaS when our apps and servers aren't online. Because of that, it needs all its CSS packaged up with it, separately to our app pages.

paas-failwhale/index.html
tombye: We do! Should now be added in https://github.com/alphagov/notifications-admin/pull/4330/commits/d6467daa1d1249cfccf3af8805c73bddb31bb81f, with details of how I got the hash.

klssmith (approved):

PR #4331, opened 2022-08-24
Bump dependencies

samuelhwilliams:
Bump dependencies

Most of these bumped dependencies come from
https://github.com/alphagov/notifications-admin/pull/4327 - but I've
removed the bump on moto for test dependencies as this was causing a
conflict preventing us from rolling out any of the changes.

Since moto is only installed for tests, and we don't need any of the
changes from the new version, it's feel fine to leave it out for now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,44 +4,44 @@
 #
 #    pip-compile requirements.in
 #
-ago==0.0.93
+ago==0.0.95
     # via -r requirements.in
 async-timeout==4.0.2
     # via redis
-awscli==1.25.2
+awscli==1.25.57
     # via awscli-cwlogs
 awscli-cwlogs==1.4.6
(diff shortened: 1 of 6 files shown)


rjbaker (approved):
Well done. The transitive dependency changes look reasonably safe 🚢

requirements.txt
```diff
@@ -4,44 +4,44 @@
...
-certifi==2022.5.18.1
+certifi==2022.6.15
     # via
     #   pyproj
     #   requests
-cffi==1.15.0
+cffi==1.15.1
     # via cryptography
-chardet==4.0.0
+chardet==5.0.0
```
rjbaker, line 36: This is a major bump. Doesn't look too bad: https://github.com/chardet/chardet/releases/tag/5.0.0

PR #4334, opened 2022-08-31
Add page for removing email branding from pool

samuelhwilliams:
Adds a new page for platform admins to remove email branding options
from an organisation's email branding pool.

Ticket: https://www.pivotaltracker.com/n/projects/1443052/stories/182880012


app/assets/stylesheets/govuk-frontend/overrides.scss
```diff
@@ -58,3 +58,11 @@
...
   display: none !important;
 }
+
+// Override borders
+.govuk-\!-no-top-border {
+  border-top: none !important;
+}
+.govuk-\!-no-bottom-border {
+  border-bottom: none !important;
+}
```
samuelhwilliams, line 68: Adding these because we're now reusing the `user-list` component for the email brand list to get the styling for the action link (remove). If we don't want to use this component, we might be able to remove these but we'll need to fiddle with the styles a lot more.

Happy to do that but direction would be appreciated. This felt like the quickest way to get the style we want without being totally outrageous (user-list component is used in a lot of places already).

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -17,27 +17,68 @@
...
+                </div>
+            </div>
+
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-full">
+                    {{ email_branding_preview(current_org.email_branding_id) }}
+                </div>
+            </div>
+
+                <div class="govuk-grid-row">
```
tombye, line 39: Do we need this row? If it's just for the vertical spacing we should create a variant with it made out of margin.

app/assets/stylesheets/govuk-frontend/overrides.scss
tombye: Design system have some guidance on adding your own override classes: https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#custom-override-classes

I didn't know that so set a bad precedent in `app/assets/stylesheets/govuk-frontend/overrides.scss` which I will have to unpick. I'd suggest you instead create a `app/assets/stylesheets/overrides` folder and put those in a `app/assets/stylesheets/overrides/_border.scss` file. I'd also recommend following the order of existing override classes and doing something like `.govuk-\!-border-bottom-none` instead of `.govuk-\!-no-bottom-border`.

app/templates/views/organisations/organisation/settings/email-branding-options.html
samuelhwilliams: Ahh ok - yes it was just for the spacing. Could you show me how we should do it instead? :)

app/assets/stylesheets/govuk-frontend/overrides.scss
samuelhwilliams: Thank you!

app/templates/views/organisations/organisation/settings/email-branding-options.html
samuelhwilliams: Should we make a variant or just use a govuk margin override?

app/main/views/organisations.py
```diff
@@ -456,6 +456,29 @@ def organisation_email_branding(org_id):
         if option["name"] != current_organisation.email_branding_name
     ]
 
+    remove_branding_id = request.args.get('remove_branding_id')
+    if remove_branding_id:
+        try:
+            remove_branding = next(
+                filter(lambda x: x['id'] == remove_branding_id, current_organisation.email_branding_pool)
```
CrystalPea, line 463: What is the advantage of using this syntax as opposed to a list comprehension, like here: https://github.com/alphagov/notifications-admin/blob/master/app/main/views/service_settings.py#L507 ?

I'm asking as with list comprehension it's easier to see what is happening at a glance. Happy to be convinced if there is a benefit though ☺️

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -17,27 +17,59 @@
...
+                        {{ email_branding_preview(option.id) }}
+                    </div>
+                </div>
+
+                <div class="govuk-grid-row">
+                    <div class="govuk-grid-column-two-thirds">
+                        <p>&nbsp;</p>
+                    </div>
+                    <div class="govuk-grid-column-one-third">
+                        <a href="{{ url_for('main.organisation_email_branding', org_id=current_org.id) }}?remove_branding_id={{ option.id }}"
```
CrystalPea, line 63: Maybe not for this PR, but @tombye would there be a case into making a destructive link into a component? So we could do:

```
{{ delete_link(
    url_for('main.organisation_email_branding', org_id=current_org.id), remove_branding_id=option.id),
    text='Remove',
    visually_hidden='this branding option'
}}
```

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -1754,6 +1754,128 @@ def test_organisation_email_branding_page_shows_all_branding_pool_options(
     )
 
 
+def test_organisation_email_branding_page_shows_remove_brand_links(
```
CrystalPea, line 1757: Nice one, should we also test that `remove` link does not show for the default email branding?

CrystalPea (commented):
Good work 👏🏼 . I left a couple small comments.

app/main/views/organisations.py
samuelhwilliams: There's not a huge difference in practice. This implementation doesn't assign any new memory for the list comprehension, but that's effectively not worth thinking about because it's trivial. Mostly I've just got used to next(filter()) as the way of doing this 🤷  Happy to change to a list comp if you prefer it for readability.

tests/app/main/views/organisations/test_organisations.py
samuelhwilliams: This test does effectively do that already, although perhaps using a non-GOV.UK brand would be a stronger test?

app/main/views/organisations.py
CrystalPea: Ahh, cheers! I do find list comprehension easier to read, but very possibly it's just personal preference based on past exposure, and I just need to get used to the less familiar syntax, so in the end either way is fine :).

tests/app/main/views/organisations/test_organisations.py
CrystalPea: Ahh you are right that it does, I see now, sorry, nothing needs changing after all 😅

CrystalPea (approved):

app/templates/views/organisations/organisation/settings/email-branding-options.html
tombye: This conversation is redundant now https://github.com/alphagov/notifications-admin/pull/4334/commits/8af08ebcd22f4161fd6148e160a4f010fceb766c has gone in so marking as resolved.

app/assets/stylesheets/overrides/_border.scss
```diff
@@ -0,0 +1,7 @@
+.notify-\!-border-top-none {
```
tombye, line 1: Don't think we need these styles any more.

app/templates/components/page-header.html
```diff
@@ -1,8 +1,9 @@
 {% macro page_header(
   h1,
-  size='large'
```
tombye, line 3: We don't need to change this macro's interface. My bad, I thought we would but we don't.

tombye (approved):
The frontend code here now looks good to me so I think it's good to go 👍🏻

PR #4335, opened 2022-08-31
Add IDs to headings on the performance page

idavidmcdonald:
I often try to link to these in support tickets but am not able to
link directly to the section I'd like to.

I think these would be useful.

The ids have not been written or reviewed by a content designer. I
am not sure whether they need to or not.

### DIFF
--- a/app/templates/views/performance.html
+++ b/app/templates/views/performance.html
@@ -15,7 +15,7 @@
     </div>
   </div>
 
-  <h2 class="govuk-heading-m" id='messages-since-may-2016'>
+  <h2 class="govuk-heading-m" id='messages-sent-since-may-2016'>
     Messages sent since May 2016
   </h2>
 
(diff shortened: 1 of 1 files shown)


app/templates/views/performance.html
```diff
@@ -15,7 +15,7 @@
     </div>
   </div>
 
-  <h2 class="govuk-heading-m">
+  <h2 class="govuk-heading-m" id='messages-since-may-2016'>
```
tombye, line 18:
```suggestion
  <h2 class="govuk-heading-m" id='messages-sent-since-may-2016'>
```
This should match the heading text so it's guessable, like the others.

tombye (changes requested):
Only one thing I think needs changing. When that's done I'll be happy to 👍🏻

app/templates/views/performance.html
idavidmcdonald: Thanks Tom, have updated!

PR #4337, opened 2022-09-01
Bump Jest to latest

tombye:
Bumps Jest, our JS test framework, to v19 which should also fix the following vulnerabilities from dependabot alerts:
- [OS Command Injection in node-notifier](https://github.com/alphagov/notifications-admin/security/dependabot/5)
- [Insufficient Granularity of Access Control in JSDom](https://github.com/alphagov/notifications-admin/security/dependabot/13)

## Notes for reviewers

This pull request is written to be reviewed commit-by-commit.

We use [JSDOM](https://github.com/jsdom/jsdom) to simulate the DOM APIs our JS interacts with. Almost all of the fixes that have their own commit involve changes to how we work with JSDOM and so can require knowledge of the DOM, JSDOM or the differences they have. That being the case, it should be sufficient to review those commits on the basis of whether it makes logical sense rather than whether it is correct or not.


samuelhwilliams (approved):
This all _looks_ good to me ... but just on the surface logic, as you say.

PR #4344, opened 2022-09-05
Add `govuk-list` class where it was missing

klssmith:
This adds the `govuk-list` to two `<ul>` where it was missing and where we don't want the bullet points to be visible.

Hopefully I've now caught all the places that were missed in https://github.com/alphagov/notifications-admin/pull/4302

### DIFF
--- a/app/templates/views/templates/choose-reply.html
+++ b/app/templates/views/templates/choose-reply.html
@@ -38,7 +38,7 @@
     {{ live_search(target_selector='#template-list .template-list-item', show=True, form=search_form) }}
 
     <nav id="template-list">
-      <ul>
+      <ul class="govuk-list">
       {% for item in templates_and_folders %}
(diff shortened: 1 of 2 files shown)


quis:
You could add an extra check here to make sure we never have a `<ul>` without a `class` attribute:
https://github.com/alphagov/notifications-admin/blob/3a8c99774a12754548fa64549749e395a1c3f03c/tests/conftest.py#L2905-L2920

app/templates/views/templates/choose-reply.html
```diff
@@ -38,7 +38,7 @@ <h1 class="heading-large">Choose a template</h1>
     {{ live_search(target_selector='#template-list .template-list-item', show=True, form=search_form) }}
 
     <nav id="template-list">
-      <ul>
+      <ul govuk-list>
```
quis, line 41:
```suggestion
      <ul class="govuk-list">
```

klssmith: Whoops, now fixed 😅

quis (approved):

PR #4346, opened 2022-09-06
Update navigation landmarks give them all unique accessible names

tombye:
Makes all `<nav>` tags on our pages uniquely labelled by giving all except one a label.

https://www.pivotaltracker.com/story/show/176180604

Note: the accessible name for a `<nav>` tag is computed as its role (ie. "navigation") prefixed by its label, if it has one (ie. "Service"). So a `<nav>` with `aria-label="Service"` would have an accessible name of "Service navigation"

### Information pages

Each page's side navigation gets its label from the `navigation_label_prefix` variable, like this:

```jinja
{% set navigation_label_prefix = 'Using Notify' %}
```

#### 'Using Notify'

- app/templates/views/get-started.html
- app/templates/views/message-status.html
- app/templates/views/trial-mode.html
- app/templates/views/guidance/index.html
app/templates/views/guidance/create-and-send-messages.html
- app/templates/views/guidance/edit-and-format-messages.html
- app/templates/views/guidance/branding-and-customisation.html
- app/templates/views/guidance/send-files-by-email.html
- app/templates/views/guidance/upload-a-letter.html
- app/templates/views/guidance/letter-specification.html
- app/templates/views/documentation.html

Example on live: https://www.notifications.service.gov.uk/using-notify/get-started

#### 'About Notify'

- app/templates/views/features.html
- app/templates/views/features/emails.html
- app/templates/views/features/text-messages.html
- app/templates/views/features/letters.html
- app/templates/views/roadmap.html
- app/templates/views/guidance/who-can-use-notify.html
- app/templates/views/security.html
- app/templates/views/terms-of-use.html

Example on live: https://www.notifications.service.gov.uk/features

#### Pricing

- app/templates/views/pricing/index.html
- app/templates/views/pricing/billing-details-signed-out.html
- app/templates/views/pricing/billing-details.html
- app/templates/views/pricing/how-to-pay.html

Example on live: https://www.notifications.service.gov.uk/pricing

### Service pages

All the service pages use `app/templates/main_nav.html` for their side navigation so adding a label to this changes all of them.

Example on live: https://www.notifications.service.gov.uk/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553

### Organisation pages

All the organisation pages use `app/templates/org_nav.html` for their side navigation so adding a label to this changes all of them.

Example on live: https://www.notifications.service.gov.uk/organisations/ee0e5030-3a7b-4da8-8685-b6efb770231f

### Platform admin pages

The platform admin pages use `app/templates/views/platform-admin/_base_template.html` for the bits they all share. This includes the side navigation so adding a label here changes all of them.

Example on live: https://www.notifications.service.gov.uk/platform-admin

## Notes for reviewers

The content should be reviewed as well as the code changes, so I've assigned @saimaghafoor for that. For the content review, I'm happy to either demo it or you can check the changes against how the accessible name is computed (see explanation above).

### DIFF
--- a/app/templates/components/sub-navigation.html
+++ b/app/templates/components/sub-navigation.html
@@ -22,9 +22,10 @@
 {% endmacro %}
 
 {% macro sub_navigation(
-  item_set
+  item_set,
+  nav_label_prefix
 ) %}
-  <nav class="sub-navigation">
+  <nav class="sub-navigation" aria-label="{{ nav_label_prefix }}">
     <ol itemscope itemtype="http://schema.org/ItemList">
       {% for item in item_set %}
         {% if item.sub_navigation_items %}
--- a/app/templates/content_template.html
+++ b/app/templates/content_template.html
@@ -1,4 +1,5 @@
 {% extends "withoutnav_template.html" %}
+
 {% from "components/sub-navigation.html" import sub_navigation %}
 {% from "components/page-header.html" import page_header %}
 
@@ -10,9 +11,9 @@
 
   <div class="govuk-grid-row">
 
-    {% if navigation_links %}
+    {% if navigation_links and navigation_label_prefix %}
       <div class="govuk-grid-column-one-quarter">
-        {{ sub_navigation(navigation_links) }}
+        {{ sub_navigation(navigation_links, navigation_label_prefix) }}
       </div>
       <div class="govuk-grid-column-five-eighths">
     {% else %}
--- a/app/templates/main_nav.html
+++ b/app/templates/main_nav.html
@@ -1,7 +1,7 @@
 {% if help %}
   {% include 'partials/tour.html' %}
 {% else %}
-<nav class="navigation">
+<nav class="navigation" aria-label="Service">
   <ul>
   {% if current_user.has_permissions() %}
     {% if current_service.has_permission('broadcast') %}
--- a/app/templates/org_nav.html
+++ b/app/templates/org_nav.html
@@ -1,4 +1,4 @@
-<nav class="navigation">
+<nav class="navigation" aria-label="Organisation">
   <ul>
     <li><a class="govuk-link govuk-link--no-visited-state{{ org_navigation.is_selected('dashboard') }}" href="{{ url_for('.organisation_dashboard', org_id=current_org.id) }}">Usage</a></li>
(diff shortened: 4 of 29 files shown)


saimaghafoor (approved):
Happy with the changes :)

samuelhwilliams:
Couple of comments/questions.

Should we also add a nav aria-label on the (seemingly semi-retired) page at /features/using-notify? It calls the `sub_navigation(navigation_links)` and doesn't provide a `nav_label_prefix`.

Also the platform admin `/accounts` page has a nav for choosing a service that doesn't have an aria-label. Looks like `/find-services-by-name`, `/find-users-by-name`, `/users/<user_id>` are the same as well. Maybe lots of other platform admin pages? They're not the side nav, just other pages with `<nav>`'s on them.

tombye:
@samuelhwilliams good catches!

> Should we also add a nav aria-label on the (seemingly semi-retired) page at /features/using-notify? It calls the sub_navigation(navigation_links) and doesn't provide a nav_label_prefix.

Absolutely. Its still publicly accessible so should be updated. Done in https://github.com/alphagov/notifications-admin/pull/4346/commits/836891a4a875cc438e8a6f90985360acd118aad5.

> Also the platform admin /accounts page has a nav for choosing a service that doesn't have an aria-label. Looks like /find-services-by-name, /find-users-by-name, /users/<user_id> are the same as well. Maybe lots of other platform admin pages? They're not the side nav, just other pages with <nav>'s on them.

This is an interesting case. I was going to say we should add one but remembered that it's an example of a page where `<nav>` is used to wrap lists of links in the main content area. The Digital Accessibility Centre advised us not to do that on our `<service_id>/templates` page in a recent audit as it means more 'navigation' landmarks to wade through without much added value. I think the same logic can be applied here so we should actually remove all these `<nav>`s. It shouldn't be part of this PR though, I've added [this ticket](https://www.pivotaltracker.com/story/show/183221139) to clear all those `<nav>` tags up.

For reference, I added a comment about this last point to the ticket for this story: https://www.pivotaltracker.com/story/show/176180604/comments/233207001

samuelhwilliams (approved):
Looks good to me now. Nice work! 🎉

PR #4347, opened 2022-09-07
Don’t allow @ symbol in organisation domains

quis:
The domain will only work if it’s just the domain. If it’s an email address or part thereof it won’t be accepted.

We have a few of these in our production database, let’s not have any more.

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -6,6 +6,7 @@ from notifications_utils.formatters import formatted_list
 from notifications_utils.recipients import InvalidEmailError, validate_email_address
 from notifications_utils.sanitise_text import SanitiseSMS
 from notifications_utils.template import BroadcastMessageTemplate
+from orderedset import OrderedSet
 from wtforms import ValidationError
 
 from app.main._commonly_used_passwords import commonly_used_passwords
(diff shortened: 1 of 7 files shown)


app/main/validators.py
```diff
@@ -197,3 +197,13 @@ def __init__(
     def __call__(self, form, field):
         if field.data and not re.match(self.regex, field.data):
             raise ValidationError(self.message)
+
+
+class NoAtSymbols:
+
+    def __init__(self, message='Enter a domain name without a leading ‘@’'):
```
idavidmcdonald, line 204: Optional: This error message is specific about domains but this validator doesn't indicate that (we could go and use this validator in a non domain specific context)

You could either change the error message to remove the domain specific part or you could rename the class to something like `DomainWithNoAtSymbols`

I don't feel too strongly though as the chance of reuse of this class is low, this is more to do with sticking with potential convention

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -1478,6 +1478,36 @@ def test_update_organisation_domains_when_domain_already_exists(
...
+):
+    user = create_platform_admin_user()
+    client_request.login(user)
+
+    page = client_request.post(
+        'main.edit_organisation_domains',
+        org_id=ORGANISATION_ID,
+        _data={
+            'domains-0': 'test@example.gov.uk',
+            'domains-1': '@example.gov.uk',
```
idavidmcdonald, line 1496: Optional: I was maybe expecting these two to be different parametrised test cases so we are sure that the error is picked up both cases (whereas at the moment I think it will just if any of these cases break the rule).

Note, this comment may be incorrect as I can't replicate the error messages appearing and so can't tell if we show two errors on the page, one for each failure, or if the fact you are checking the error message twice below is to do with the GOV.UK Design System way of showing the error at the top of the page and inline.

idavidmcdonald (commented):
Have followed up on Slack with a different comment but this will likely be good to go after that.

quis:
I have:
- rewritten the error as a banner, to account for the fact that `list-entry.js` re-renders the individual fields without the error messages
- rewritten the validator class to be generic, such that it can exclude any set of characters and can doesn’t return an error specific to one instantiation

Here’s what it looks like running locally:

image

idavidmcdonald (approved):

PR #4350, opened 2022-09-12
Adjustment to set email branding page

BlessedDev:
As per https://www.pivotaltracker.com/n/projects/1443052/stories/183002322, After a platform admin sets email branding for a service, the platform admin will now be taken to a page and given the option of adding it to the email branding pool of the services' organisation so that it can be accessed by other services in that organisation.
Screenshot 2022-09-09 at 14 12 23
Screenshot 2022-09-09 at 14 12 35
Screenshot 2022-09-09 at 14 12 46

In a scenario where the email branding is not gong to be added to the organisation's email branding pool, the user journey will be as follows:
Screenshot 2022-09-09 at 14 13 08
Screenshot 2022-09-09 at 14 13 16
Screenshot 2022-09-09 at 14 13 30

In each case a flash message is displayed to the platform admin, this is to keep consistency with the work done in the organisations settings page.

The content on the page is a work in progress and will be adjusted in the near future


app/main/views/service_settings.py
```diff
@@ -1016,18 +1018,81 @@ def service_set_email_branding(service_id):
...
+
+            }
+    }
+    form.choice_option.param_extensions['items'][1] = {
+        'hint': {
+            'html': Markup(
+                f"Only apply this branding to {service_name}")
+
+        }
+    }
```
samuelhwilliams, line 1054: It doesn't feel right to have so much content/configuration here in the view - I think we should move more of this to the form.

What do you think of this patch?


```diff
diff --git a/app/main/forms.py b/app/main/forms.py
index ebf399c50..dbf2ab596 100644
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1975,8 +1975,28 @@ class AdminChangeEmailBrandingPoolForm(StripWhitespaceForm):
 
 
 class AdminSetEmailBrandingAddToBrandingPoolStepForm(StripWhitespaceForm):
+    def __init__(self, service_name, org_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.choice_option.param_extensions['items'][0]['hint']['html'] = Markup(
+            f"<p>Apply this branding to {service_name}<p>"
+            f"<p>Let other {org_name} teams apply this branding themselves<p>"
+        )
+        self.choice_option.param_extensions['items'][1]['hint']['html'] = Markup(
+            f"Only apply this branding to {service_name}"
+        )
+
     choice_option = GovukRadiosField(
-        validators=[DataRequired(message="Please select an option")]
+        choices=[
+            ('yes', 'Yes'),
+            ('no', 'No')
+        ],
+        validators=[DataRequired(message="Please select an option")],
+        param_extensions={
+            'items': [
+                {'hint': {'html': ''}},
+                {'hint': {'html': ''}},
+            ]
+        }
     )
 
 
diff --git a/app/main/views/service_settings.py b/app/main/views/service_settings.py
index 99c65001e..cc220d711 100644
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1029,29 +1029,7 @@ def service_set_email_branding_add_to_branding_pool_step(service_id):
     org_id = current_service.organisation.id
     service_name = current_service.name
 
-    form = AdminSetEmailBrandingAddToBrandingPoolStepForm()
-    form.choice_option.choices = [
-        ('yes', 'Yes'),
-        ('no', 'No')
-    ]
-
-    # For clarity we are going to add some hints to the choices "Yes or No"
-    form.choice_option.param_extensions = {'items': [{}, {}]}
-    form.choice_option.param_extensions['items'][0] = {
-        'hint': {
-                'html': Markup(
-                    f"<p>Apply this branding to {service_name}<p>"
-                    f"<p>Let other {org_name} teams apply this branding themselves<p>")
-
-            }
-    }
-    form.choice_option.param_extensions['items'][1] = {
-        'hint': {
-            'html': Markup(
-                f"Only apply this branding to {service_name}")
-
-        }
-    }
+    form = AdminSetEmailBrandingAddToBrandingPoolStepForm(service_name=service_name, org_name=org_name)
 
     if form.validate_on_submit():
         choice = form.choice_option.data
```

app/main/views/service_settings.py
```diff
@@ -1016,18 +1018,81 @@ def service_set_email_branding(service_id):
...
+    email_branding = email_branding_client.get_email_branding(email_branding_id)['email_branding']
+    email_branding_name = email_branding['name']
+    org_name = current_service.organisation.name
+    org_id = current_service.organisation.id
+    service_name = current_service.name
+
+    form = AdminSetEmailBrandingAddToBrandingPoolStepForm()
+    form.choice_option.choices = [
+        ('yes', 'Yes'),
+        ('no', 'No')
```
samuelhwilliams, line 1035: Do you think it'd be cleaner to use True/False as the 'clean' data values here rather than string yes/no?

app/main/views/service_settings.py
```diff
@@ -1016,18 +1018,81 @@ def service_set_email_branding(service_id):
...
+            'html': Markup(
+                f"Only apply this branding to {service_name}")
+
+        }
+    }
+
+    if form.validate_on_submit():
+        choice = form.choice_option.data
+        # If the platform admin chose "yes" the branding is added to the organisation's
+        # branding pool
```
samuelhwilliams, line 1059: I'm not sure this comment should be necessary. We should probably be able to write the code in a more self-explanatory way so that, eg if we changed the name of the `choice_option` field to `add_to_pool`?

app/templates/views/service-settings/set-email-branding-add-to-branding-pool-step.html
```diff
@@ -0,0 +1,31 @@
+{% extends "withnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import page_footer, sticky_page_footer %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/back-link/macro.njk" import govukBackLink %}
+
+{% set page_title = "Apply {} branding".format(email_branding_name) %}
+{% set  page_blurb = "Should other teams in {} have the option of this branding?".format(org_name) %}
```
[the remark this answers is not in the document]
samuelhwilliams: Also, I think that in the screenshot above, there is too big a gap between the two lines of hint text for the 'yes' option.

At the moment we have those two lines in separate `<p>` tags. We probably just want one `<p>` tag, and if we need them to be on separate lines, have a `<br>` between them? That would end up looking like this:

image

But again a call for @quis to make.

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -3745,6 +3745,57 @@ def test_should_set_branding_and_organisations(
...
+    ),
+))
+def test_service_set_email_branding_add_to_branding_pool_step_choices_yes_or_no(
+    mocker,
+    client_request,
+    platform_admin_user,
+    service_one,
+    organisation_one,
+    choice_option,
+    flash_message
```
samuelhwilliams, line 3765: We don't ever use this `flash_message` parameter here. I think it'd be good to check that the flash message is shown in each case - but hopefully we don't need to override it and can just test the legit content maybe?

samuelhwilliams (changes requested):
Have tested this PR locally and the functionality looks good! Nicely done.

I can see you've added a test for the POST endpoint - can we add one for GET as well and cover of the basic rendering of the page if appropriate.

I think there are a few bits of the code that we might want to clean up and I've flagged a couple of potential design quirks that may or may not need revisiting.

app/templates/views/service-settings/set-email-branding-add-to-branding-pool-step.html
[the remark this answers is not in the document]
quis: Will pick these up with @saimaghafoor later today, but no need for them to block this PR being merged.

app/main/views/service_settings.py
BlessedDev: @samuelhwilliams It would indeed be cleaner and I have adjusted it accordingly.

tests/app/main/views/service_settings/test_service_settings.py
BlessedDev: I have fixed the tests to test for the flash messages and other content.

app/main/views/service_settings.py
BlessedDev: add_to_pool is a good name and I have incorporated it.

app/main/views/service_settings.py
BlessedDev: This is a good point. I have streamlined the code for the view and moved the relevant content into the form.

app/main/views/service_settings.py
BlessedDev: I have had to roll back on this. Due to the form validation code that converts form choices to a dict by using the dict constructor.
https://github.com/alphagov/notifications-admin/blob/286073c55e94f7b3453ad0a12ad7628ddcceb54e/app/main/forms.py#L151-L154

So the data values remain 'yes' and 'no'.

app/templates/views/service-settings/set-email-branding-add-to-branding-pool-step.html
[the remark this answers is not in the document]
BlessedDev: I have updated the content to this:
Screenshot 2022-09-16 at 09 11 27

app/main/forms.py
```diff
@@ -1974,6 +1963,34 @@ class AdminChangeEmailBrandingPoolForm(StripWhitespaceForm):
...
+        )
+        self.add_to_pool.param_extensions['items'][1]['hint']['html'] = Markup(
+            f"Only apply branding to {service_name}")
+
+    add_to_pool = GovukRadiosField(
+        choices=[
+            ('yes', 'Yes'),
+            ('no', 'No')
+        ],
+        validators=[DataRequired(message="Please select an option")]
```
samuelhwilliams, line 1990: Is this the content we normally use for a required selection or otherwise explicitly chosen for this error message?

app/main/views/service_settings.py
```diff
@@ -1016,18 +1017,67 @@ def service_set_email_branding(service_id):
...
+        'branding_name': email_branding_name
+    }
+
+    form = AdminSetEmailBrandingAddToBrandingPoolStepForm(**data)
+
+    if form.validate_on_submit():
+        add_to_pool = form.add_to_pool.data
+        # If the platform admin chose "yes" the branding is added to the organisation's
+        # branding pool
+        if add_to_pool == "yes":
```
samuelhwilliams, line 1042: I thought you changed the cleaned data to True/False? 🤔  If so, this should probably be `if add_to_pool is True` and we can drop the comment.

app/main/views/service_settings.py
```diff
@@ -1016,18 +1017,67 @@ def service_set_email_branding(service_id):
...
+            email_branding_ids = [email_branding_id]
```
samuelhwilliams, line 1043: Optional nit: I don't think this variable really needs to exist. It doesn't ever get reused and it doesn't really make anything clearer or easier to read when compared to just having `[email_branding_id]` inline in the function call

samuelhwilliams (commented):
Looks much cleaner now, nice 👍  Couple of small comments and then once the test is fixed I'll be happy to ✅

app/main/forms.py
BlessedDev: This has been explicitly chosen for this error message. I am agnostic about the content and as @quis noted the content will be looked at and updated later if necessary.

app/main/views/service_settings.py
BlessedDev: I rolled it back due to the following:
https://github.com/alphagov/notifications-admin/blob/286073c55e94f7b3453ad0a12ad7628ddcceb54e/app/main/forms.py#L151-L154

The validation favours a string as a dict constructor is being used

app/main/views/service_settings.py
BlessedDev: I tried to make this a bit consistent with other code so that any one looking at this will immediately get that the api is a expecting a list of email branding ids.

app/main/views/service_settings.py
samuelhwilliams: I don't understand what you mean by that. I think this code snippet is equivalent and it works fine. Am I misunderstanding? :)

```
In [1]: choices=((True, 'yes'), (False, 'no'))

In [2]: True not in dict(choices).keys()
Out[2]: False

In [3]: 'blah' not in dict(choices).keys()
Out[3]: True
```

app/main/forms.py
quis: You don’t need to explicitly add a validator here because you are using `RadioField`.

When submitting the form `RadioField` will run `pre_validate` which checks that one of the options has been chosen:
https://github.com/alphagov/notifications-admin/blob/601dd09f34e7a4ac54fae89855f2ee17168b1a58/app/main/forms.py#L151-L154

If one hasn’t then it will return the default error message, which is ‘Select an option’

In this case it might be a bit nicer to say ‘Select yes or no’ which you could achieve with the `thing` argument:
 ```suggestion
        thing='yes or no',
```

Here’s one of many examples where we have that content already:
https://github.com/alphagov/notifications-admin/blob/601dd09f34e7a4ac54fae89855f2ee17168b1a58/app/main/forms.py#L1693-L1700

Comment (quis) [reply app/main/forms.py:1990]: Note that ‘Select [thing]’ is the same format used by the GOV.UK Design System: https://design-system.service.gov.uk/components/radios/#error-messages

image

File: app/main/views/service_settings.py (hunk 7 above)
Comment (BlessedDev) [reply app/main/views/service_settings.py:1042]: Self.data presents True and False as strings hence a ValidationError is thrown fails as it is actually `'True' in dict(choices).keys()` and not `True in dict(self.choices).keys()`

Comment (samuelhwilliams) [reply app/main/views/service_settings.py:1042]: Yeah gross and interesting.

It looks like we've come across this before and ended up with an `OnOffField` that inherits from `GovukRadiosField` to semi work around this issue. Could use that, but I agree at this point it's not overly important either way.

File: app/main/forms.py (hunk 6 above)
Comment (BlessedDev) [reply app/main/forms.py:1990]: Thanks, I will adjust the code accordingly

File: app/main/views/service_settings.py (hunk 7 above)
Comment (samuelhwilliams) [reply app/main/views/service_settings.py:1042]: For future reference, after digging a bit more, WTForms provide an argument to Fields called `coerce` that can be used to cast from strings to an alternative data type. So you could use something like this (probably not worth adding now to be honest - just posting to share):

```
    add_to_pool = GovukRadiosField(
        choices=[
            ('yes', 'Yes'),
            ('no', 'No')
        ],
        validators=[DataRequired(message="Please select an option")],
        coerce=lambda val: val == 'True' if val in {'True', 'False'} else val,
    )
```

samuelhwilliams (approved):
Nice work @BlessedDev 👍

PR #4353, opened 2022-09-15
Remove GOVUK Frontend Toolkit [part 4]

tombye:
This is part 4 of a series of work to remove the GOVUK Frontend Toolkit library from our codebase:

https://www.pivotaltracker.com/story/show/182596710

This part covers removal of the Sass typography code and so replaces all uses of its font mixins with equivalents from GOVUK Frontend, following design system guidelines:

https://design-system.service.gov.uk/get-started/updating-your-code/#typography

## Notes for reviewers

Our version of GOVUK Frontend still has 'compatibility mode' on, as per [design system guidelines for migration](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/#add-gov-uk-frontend-to-your-project), which means we still use the old font and colours. Because of that, these commits don't result in much difference to the resulting CSS. We still checked all the uses and didn't see any differences.

There's also an extra commit in there removing some redundant CSS.

### DIFF
--- a/app/assets/stylesheets/app.scss
+++ b/app/assets/stylesheets/app.scss
@@ -130,7 +130,7 @@ details summary {
 
     th,
     td {
-      @include core-19;
+      @include govuk-font(19);
       word-wrap: break-word;
     }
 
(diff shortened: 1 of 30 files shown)


klssmith (approved):
Approving this since we paired on it and have both checked it

PR #4354, opened 2022-09-16
Change default email branding for organisations

samuelhwilliams:
Make it possible for platform admins to set the default email branding
for an organisation from a single consolidated page, rather than having
one page to set the default and another to add/remove email brands.

Only central organisations are allowed to switch back to the platform
default GOV.UK branding if they are on another brand.

If an organisation is being switched away from GOV.UK, we want to warn
them that any services in that organisation currently also on the
implicit GOV.UK brand will have their brands updated to match. This only
happens when moving from GOV.UK (effectively 'no brand') to something
else. Switching between two explicit brands will not affect services at
all.

Another ticket covers removing the now-redundant default email branding
page.

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -1,10 +1,12 @@
 from collections import OrderedDict
 from datetime import datetime
 from functools import partial
+from typing import Optional
 
 from flask import flash, redirect, render_template, request, send_file, url_for
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
+from werkzeug import Response
 from werkzeug.exceptions import abort
 
 from app import (
@@ -447,6 +449,29 @@ def organisation_preview_email_branding(org_id):
     )
 
 
+def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
+    try:
+        remove_branding = next(
+            filter(lambda x: x['id'] == remove_branding_id, current_organisation.email_branding_pool)
+        )
+    except StopIteration:
+        abort(400, f'Invalid email branding ID {remove_branding_id} for {current_organisation}')
+
+    if request.method == 'POST':
+        organisations_client.remove_email_branding_from_pool(
+            current_organisation.id,
+            remove_branding_id
+        )
+        flash(f'Email branding ‘{remove_branding["name"]}’ removed.', 'default_with_tick')
+        return redirect(url_for('main.organisation_email_branding', org_id=current_organisation.id))
+
+    else:
+        flash(
+            f'Are you sure you want to remove the email brand ‘{remove_branding["name"]}’?',
+            'delete',
+        )
+
+
 @main.route("/organisations/<uuid:org_id>/settings/email-branding", methods=['GET', 'POST'])
 @user_is_platform_admin
 def organisation_email_branding(org_id):
@@ -458,26 +483,8 @@ def organisation_email_branding(org_id):
 
     remove_branding_id = request.args.get('remove_branding_id')
     if remove_branding_id:
-        try:
-            remove_branding = next(
-                filter(lambda x: x['id'] == remove_branding_id, current_organisation.email_branding_pool)
-            )
-        except StopIteration:
-            abort(400, f'Invalid email branding ID {remove_branding_id} for {current_organisation}')
-
-        if request.method == 'POST':
-            organisations_client.remove_email_branding_from_pool(
-                current_organisation.id,
-                remove_branding_id
-            )
-            flash(f'Email branding ‘{remove_branding["name"]}’ removed.', 'default_with_tick')
-            return redirect(url_for('main.organisation_email_branding', org_id=current_organisation.id))
-
-        else:
-            flash(
-                f'Are you sure you want to remove the email brand ‘{remove_branding["name"]}’?',
-                'delete',
-            )
+        if response := __handle_remove_branding(remove_branding_id):
+            return response
 
     return render_template(
         'views/organisations/organisation/settings/email-branding-options.html',
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -3,7 +3,16 @@ from datetime import datetime
 from functools import partial
 from typing import Optional
 
-from flask import flash, redirect, render_template, request, send_file, url_for
+from flask import (
+    Markup,
+    current_app,
+    flash,
+    redirect,
+    render_template,
+    request,
+    send_file,
+    url_for,
+)
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
 from werkzeug import Response
@@ -23,6 +32,7 @@ from app.main.forms import (
     AddGPOrganisationForm,
     AddNHSLocalOrganisationForm,
     AdminBillingDetailsForm,
+    AdminChangeOrganisationDefaultEmailBrandingForm,
     AdminNewOrganisationForm,
     AdminNotesForm,
     AdminOrganisationDomainsForm,
@@ -449,7 +459,11 @@ def organisation_preview_email_branding(org_id):
     )
 
 
-def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
+def _handle_remove_branding(remove_branding_id) -> Optional[Response]:
+    """
+    The user has clicked 'remove' on a brand and is either going to see a confirmation flash message
+    or has clicked to confirm that flash message.
+    """
     try:
         remove_branding = next(
             filter(lambda x: x['id'] == remove_branding_id, current_organisation.email_branding_pool)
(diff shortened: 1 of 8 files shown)


app/main/views/organisations.py
```diff
@@ -471,6 +485,109 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+
+
+def __handle_reset_to_default_branding(is_central_government) -> Optional[Response]:
+    """
+    This handles changing a central government organisation from a custom brand back to the default GOV.UK brand.
+    If we're in here, then the user has either clicked the 'Reset to GOV.UK' link or they're clicking the
+    button in the confirmation dialog.
+    """
+    # Only central government departments can have their brand reset to default GOV.UK
+    if is_central_government:
```
CrystalPea, line 498: shouldn't it be `if not central_government` here?

app/main/views/organisations.py
```diff
@@ -471,6 +485,109 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
             'delete',
         )
 
+    return None
+
+
+def __handle_reset_to_default_branding(is_central_government) -> Optional[Response]:
```
CrystalPea, line 491: this function name is a bit confusing, since "default" means "gov.uk" specifically here. Should we maybe rename it `_change_default_org_branding_to_gov_uk`

app/main/views/organisations.py
```diff
@@ -471,6 +485,109 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+            current_organisation.id,
+            email_branding_id=None,
+        )
+        flash(f'Email branding for {current_organisation.name} reset to GOV.UK.', 'default_with_tick')
+        return redirect(url_for('main.organisation_email_branding', org_id=current_organisation.id))
+
+    else:
+        current_brand = current_organisation.email_branding_name
+        message = Markup(
+            "Are you sure you want to make GOV.UK the default email branding?<br><br>"
```
CrystalPea, line 512: Nice one 👍🏼

app/main/views/organisations.py
```diff
@@ -471,6 +485,109 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+    # Only central government departments can have their brand reset to default GOV.UK
+    if is_central_government:
+        return None
+
+    if request.method == 'POST':
+        organisations_client.update_organisation(
+            current_organisation.id,
+            email_branding_id=None,
+        )
+        flash(f'Email branding for {current_organisation.name} reset to GOV.UK.', 'default_with_tick')
```
CrystalPea, line 506: It would be good if this message was consistent with the one for other brandings, like `The default email branding for {current_organisation.name} services is now GOV.UK.'` . 

But also not sure if we need those flash messages at all, since we don't redirect to another page, so the user sees the effect of their action straight away at the top of the page. For example when we change default reply-to email address we do not have a banner confirming that action is done. I am fine either way though.

app/main/views/organisations.py
```diff
@@ -471,6 +485,109 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+        )
+        flash(
+            message,
+            "yes, update default branding"
+        )
+
+    return None
+
+
+def __handle_change_default_branding(form, confirm_change_branding_id) -> Optional[Response]:
```
CrystalPea, line 524: maybe we could rename `confirm_change_branding_id` to ` new_default_branding_id` as it reflects closer what it is?

app/main/views/organisations.py
```diff
@@ -481,14 +598,31 @@ def organisation_email_branding(org_id):
         if option["name"] != current_organisation.email_branding_name
     ]
 
+    is_central_government = current_organisation.organisation_type == Organisation.TYPE_CENTRAL
     remove_branding_id = request.args.get('remove_branding_id')
+    reset_to_default_branding = 'reset_to_default_branding' in request.args
```
CrystalPea, line 603: could we rename this to `change_default_branding_to_govuk` or something, again because of ambiguity of word  "default" in this context.

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -25,19 +25,43 @@ <h2 class="govuk-heading-s govuk-!-margin-bottom-0">
 
     <!--display the email-branding-pool options and the preview for each one -->
 
+    {% if show_govuk_default_option %}
+      <p class="govuk-body govuk-!-margin-bottom-7">
+        <a
+          class="govuk-link govuk-link--no-visited-state"
+          href="{{ url_for('main.organisation_email_branding', org_id=current_org.id) }}?reset_to_default_branding"
```
CrystalPea, line 32: ohh cool I didn't know we don't have to assign true to the `reset_to_default_branding` for it to work

CrystalPea (commented):
Nice one, good split to avoid making the view too large 👏🏼 

 I know this is only a draft, but since I'm not working Mondays I thought I'd leave some comment, mainly naming suggestions. Sorry if you were going to change those anyway 😅

app/main/views/organisations.py
samuelhwilliams: Yes 😱   Hopefully would've caught that later in a test but great spot 👀

app/main/views/organisations.py
samuelhwilliams: Yeah good call 🙌

app/main/views/organisations.py
samuelhwilliams: Content is all due for review later when the actual PR goes up, so not worried about it now, it's all filler.

But yeah it's possible we don't need this flash at all.

app/main/views/organisations.py
samuelhwilliams: yep

app/templates/views/organisations/organisation/settings/email-branding-options.html
samuelhwilliams: Yeah it will still end up in the `request.args` dict, albeit with an empty string value. We could provide a value but if we have nothing meaningful to convey then it's sometimes just as easy to not have one.

app/main/views/organisations.py
samuelhwilliams: This bit doesn't actually work yet and is going to be the tricky part. Not sure we support rendering lists inside confirmation dialogs yet, and it looks a bit weird when I try. Will have to dig into it a bit more.

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -1747,9 +1748,9 @@ def test_organisation_email_branding_page_shows_all_branding_pool_options(
         normalize_spaces(heading.text) for heading in page.select('.govuk-heading-s')
             ) == expected_branding_options
 
-    add_options_button = page.select_one('.govuk-button--secondary')
-    assert normalize_spaces(add_options_button.text) == 'Add branding options'
-    assert add_options_button.attrs["href"] == url_for(
+    add_options_buttons = page.select('.govuk-button--secondary')
```
CrystalPea, line 1751: There is only one "Add options" button, so maybe we would do it like:

`add_options_button = page.select('.govuk-button--secondary')[-1]`

?

CrystalPea (commented):
Good work, a pretty complex story this one. I left some comments:

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -16,7 +16,8 @@
 {%  endblock %}
 
 {% block maincolumn_content %}
-    {{ page_header(page_title) }}
+    <h1 class="govuk-heading-l">{{ page_title }}</h1>
```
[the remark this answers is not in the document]
samuelhwilliams: The macro is so trivial as to be broadly a waste and it possibly shouldn't exist (as discussed and agreed with Tom).

app/main/views/organisations.py
```diff
@@ -447,6 +449,29 @@ def organisation_preview_email_branding(org_id):
     )
 
 
+def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
```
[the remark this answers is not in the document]
samuelhwilliams: Good catch - should just be one underscore for non-public module functions :)

Double underscores are more reserved for class methods/attributes and special name mangling happens to them in that case to prevent them from being (easily) accessed.

app/main/views/organisations.py
```diff
@@ -462,15 +476,119 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
             remove_branding_id
         )
-        flash(f'Email branding ‘{remove_branding["name"]}’ removed.', 'default_with_tick')
+        confirmation_message = f'Email branding ‘{remove_branding["name"]}’ removed.'
+        flash(confirmation_message, 'default_with_tick')
         return redirect(url_for('main.organisation_email_branding', org_id=current_organisation.id))
 
     else:
+        confirmation_question = Markup(
+            f'<h2 class="banner-title">'
```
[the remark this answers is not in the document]
samuelhwilliams: Hum hummmm... The default banner doesn't actually making the error message a heading, which it should be according to the design system.

I'm not _too_ fussed about getting it 100% correct in this PR, as @tombye has already told me there's a backlog ticket to go over all validation messages and make them consistent and in line with the Design System.

I have had a chat with Tom in the process of building out this work. From my quick look it seems like most confirmation style messages use the flash message system, which is built in to all templates out of the box so it's nice to just be able to re-use that. It does get a little trickier for the cases where we're flashing a confirmation style thing, so might be a good reason to deviate for those ones. But it makes the private helper methods for this view more complex as they need to return flags which can be passed to the template to render the right question, as opposed to just flashing the message and letting the template handle it innately.

I might extract these to template partials which can be rendered into the flash messages, which removes the HTML from the python code and keeps it nicely separate, while still allowing us to use the flash message system to keep the functions/return values cleaner. Let's see what you think of the next patch.

app/main/views/organisations.py
```diff
@@ -462,15 +476,119 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+        organisations_client.update_organisation(
+            current_organisation.id,
+            email_branding_id=None,
+        )
+        return redirect(url_for('main.organisation_email_branding', org_id=current_organisation.id))
+
+    else:
+        current_brand = current_organisation.email_branding_name
+        confirmation_question = Markup(
+            f'<h2 class="banner-title">Are you sure you want to make GOV.UK the default email branding?</h2>'
```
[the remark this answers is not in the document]
samuelhwilliams: As above. Let's keep convo on the above thread.

app/main/views/organisations.py
```diff
@@ -462,15 +476,119 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+            f'<li>will no longer be able to use the GOV.UK branding</li>'
+            f'</ul>'
+        )
+        flash(
+            confirmation_question,
+            'yes, make this email branding the default',
+        )
+
+    if form.validate_on_submit():
+        if current_organisation.email_branding_id is None:
```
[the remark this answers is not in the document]
samuelhwilliams: Will add a comment 👍

app/main/views/organisations.py
```diff
@@ -462,15 +476,119 @@ def __handle_remove_branding(remove_branding_id) -> Optional[Response]:
...
+            return redirect(
+                url_for(
+                    'main.organisation_email_branding',
+                    org_id=current_organisation.id,
+                    new_default_branding_id=form.email_branding_id.data,
+                )
+            )
+
+        organisations_client.update_organisation(
```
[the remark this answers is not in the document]
samuelhwilliams: We do need both as they cover different cases. I'll add a comment to explain, but in summary:

Block 549-572 covers the case where an org is changing their brand from GOV.UK -> something else
Line 
Block 575-582 covers the case where an org is trying to change from GOV.UK -> something else and we need to redirect them to block 549-572 to show the confirmation message
Lines 584-588 (fallthrough from above) covers the case of non-GOV.UK brand -> another brand.

app/templates/views/organisations/organisation/settings/email-branding-options.html
```diff
@@ -25,19 +26,43 @@ <h2 class="govuk-heading-s govuk-!-margin-bottom-0">
...
-         class="govuk-link govuk-link govuk-link--destructive govuk-link--no-visited-state">
-          Remove <span class="govuk-visually-hidden">this branding option</span>
-        </a>
-      </p>
+      <div class="govuk-grid-row govuk-!-margin-top-2 govuk-!-margin-bottom-6">
+        <div class="govuk-grid-column-one-half">
+          {% call form_wrapper(action=url_for('main.organisation_email_branding', org_id=current_org.id)) %}
+            <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+            {{ form.email_branding_id(value=option.id) }}
+            <input type="submit" class="govuk-button govuk-button--secondary" value="Make default" />
```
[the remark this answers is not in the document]
samuelhwilliams: Sure, but I generally think we should be fairly free to just use design system HTML/classes directly and not have macros for trivial things like this.

tests/app/main/views/organisations/test_organisations.py
samuelhwilliams: Good call 👍

samuelhwilliams:
@CrystalPea have pushed updates from the review in a separate commit to be a bit easier to see what's changed. Will merge them all together after we're happy :)

app/main/views/organisations.py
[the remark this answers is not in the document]
CrystalPea: > I might extract these to template partials which can be rendered into the flash messages, which removes the HTML from the python code and keeps it nicely separate, while still allowing us to use the flash message system to keep the functions/return values cleaner. Let's see what you think of the next patch.

That sounds like a good solution.

As to what we usually use, I think it's a mixture. For example in [here](https://github.com/alphagov/notifications-admin/blob/08084dfbd2127df911c227997eb905f9909abf5f/app/templates/views/edit-user-permissions.html) and [here](https://github.com/alphagov/notifications-admin/blob/9d6f8db421b4bfdea0cf4dd591b238d5ed8e0c6d/app/templates/views/templates/template.html) we use  the banner component directly in the template.

app/main/views/organisations.py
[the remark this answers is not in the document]
CrystalPea: Ahhh, ok, thanks! Yeah, comments would be handy with so many cases :)

app/templates/views/organisations/organisation/settings/email-branding-options.html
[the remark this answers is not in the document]
CrystalPea: Ahh ok, personally I quite like that it reads nicely, but not a hill I'm willing to die on :)

CrystalPea (approved):

CrystalPea:
All good to go :D

app/templates/views/organisations/organisation/settings/email-branding-options.html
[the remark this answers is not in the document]
CrystalPea: Possibly I just got very used to the macros :)

PR #4356, opened 2022-09-20
Tweak layout and wording of ‘Apply branding’ page

quis:

Lots of small changes to make the page a bit clearer, and in line with the sizes, spacing, and font weights we use elsewhere.

Based on the content in this document: https://docs.google.com/document/d/1LLkTz5aBCsggERJZPZvZ1E5HSRrKfKVm3Yxcym9z-CY/edit?pli=1

Before | After 
---|---
image | image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1977,9 +1977,12 @@ class AdminSetEmailBrandingAddToBrandingPoolStepForm(StripWhitespaceForm):
             f'this branding?'
         )
 
-        self.add_to_pool.param_extensions = {'items': [{'hint': {}},
-                                                       {'hint': {}}]
-                                             }
+        self.add_to_pool.param_extensions = {
+            'items': [
+                {'hint': {}},
+                {'hint': {}},
+            ]
+        }
         self.add_to_pool.param_extensions['items'][0]['hint']['html'] = Markup(f'''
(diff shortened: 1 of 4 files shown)


idavidmcdonald (approved):
Some lovely refactors here, very very very easy to review. Thanks for breaking down into such small and well written commits Chris.

Two things that aren't my area of expertise and I was just pondering (so this isn't an ask to change them, more just to share my thoughts and you are still good to merge this)

1. I'd maybe expect a bit more of a margin underneath the 'Should other teams...' text, it looks a bit cosy to me and this is the key sentence for the user to read I think

2. It feels like the 'Should other teams...' sentence is the key question being asked to the user here. I'd maybe expect it to be more prominent whereas instead I'm drawn to reading the H1 for whether I should apply the branding and then being presented with a 'yes' or 'no'. Again, I'm not picking up from the layout and content the key question I'm being asked... but I'm not actually a user using this so maybe it is all fine anyway!

quis:
@idavidmcdonald with the spacing it’s a bit of a balance between consistency with other pages and the Design System, and something which might work better in this instance. I’ve gone for the former.

I get your point about making the question the H1/most important thing on the page. But I’m not sure where the line with the name of the branding would go in that case – it doesn’t really work as a question `label`.

PR #4360, opened 2022-09-22
Make further small refinements to the ‘Apply branding’ journey

quis:
The functional changes are:

# Only prompt to add brandings not in the pool

The user should only see this page if the ‘ACAS’ branding is not already in the organisations pool:
image

At the moment the page is showing even though the branding is in the organisation’s pool:
image

This change adds an extra check to look at the contents of the pool before prompting.

# Update branding once the user has confirmed

The form on the _Apply branding_ page says that choosing ‘yes‘ or ‘no’ will apply the new branding to the service:

image

So we shouldn’t do it until the user has clicked the green button on this page.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1072,9 +1072,9 @@ def service_preview_email_branding(service_id):
             return redirect(url_for('main.service_set_email_branding_add_to_branding_pool_step',
                                     service_id=service_id,
                                     email_branding_id=email_branding_id))
-        else:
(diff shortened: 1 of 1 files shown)


samuelhwilliams (approved):
This makes more sense 👍

CrystalPea:
Would a platform admin ever be setting a branding for a service when that branding is already part of the pool? Since that service could set that branding themselves?

quis:
@CrystalPea I guess it is unlikely, but it was enough to confuse me when I was playing around with the feature

PR #4363, opened 2022-09-23
Add tick image as SVG with new `$green` colour

quis:
This PR just adds the image, without using it anywhere.

Once we switch to using the new colours from the Design System, we can replace uses of `tick.png` with this image.

This will give us:
- the correct green
- a resolution-independent image
- a smaller file size (55% of the original)

***

tick


tombye (approved):
😍 👍🏻

PR #4364, opened 2022-09-23
Remove old default email branding page

samuelhwilliams:
We have now integrated the features that this page provided into a single page under `/settings/email-branding`, so this page is redundant.

Let's keep things simple and cleaner by only having one way to do things and remove this duplicated functionality.


samuelhwilliams:
@klssmith Good spot :)

klssmith:
`views/organisations/organisation/settings/preview-email-branding.html` can be deleted too now that the preview route has gone

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -3650,10 +3641,6 @@ def test_should_send_branding_and_organisations_to_preview(
         'main.service_preview_email_branding',
         {'service_id': SERVICE_ONE_ID},
     ),
-    (
-        'main.organisation_preview_email_branding',
-        {'org_id': ORGANISATION_ID},
-    ),
```
klssmith, line 3656: It would make things a bit clearer now if we get rid of the parameterization in this test (and in a couple of others in this PR) where we're now only testing the `service_preview_email_branding` endpoint and not the `organisation_preview_email_branding` endpoint too

klssmith (approved):

PR #4366, opened 2022-09-27
Remove Google Analytics

quis:
This is a spike to show how easy it would be to remove Google Analytics.

Some reasons we might want to remove it:
- we collect lots of data that we very rarely look at or use to make decisions, which goes against the principle of data minimisation 
- it requires setting a non-optional cookie, so we have to show an intrusive cookie banner every time someone visits our service
- because cookies require opt-in, we are only tracking a subset of users (40–60% based on anecdotes from other services)
- there is bias in this subset of users – it might skew to those on mobile, or those who are more likely to click accept without reading
- if we want to continue using Google Analytics we need to do work to upgrade to version 4, which is [a 12 step process](https://support.google.com/analytics/answer/10759417)
- even if we do upgrade to version 4 we still lose all historical data, so won’t be able to do comparisons over longer time periods
- the Google Analytics code itself, plus all the Javascript and CSS for the cookie banner and storing consent is extra bytes we are serving to every user (extra 10kb of Javascript and 6kb of CSS)


kr8n3r:
you might want to create a function to remove the cookie from user on (first) future visit

quis:
Closing this because it will be easier to start again than resolve merge conflicts

PR #4369, opened 2022-09-29
Re-add page to confirm NHS branding

quis:
In 92f76638c891487889b7debcdc8c7914da5a804a, when we let people choose NHS branding for themselves, we added a bespoke confirmation page with guidance about whether using the NHS branding was appropriate.

This was lost once we put NHS branding in the pool, and choosing it was just picking another `id`, rather than `nhs` as a special case.

This commit restores the behaviour by adding an extra check against the `id` of the NHS branding, and treating it the same as if they had previously chosen `nhs` as a special case.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -20,6 +20,7 @@ from notifications_utils.recipients import (
     normalise_phone_number,
     validate_phone_number,
 )
+from orderedset import OrderedSet
 from werkzeug.utils import cached_property
 from wtforms import (
     BooleanField,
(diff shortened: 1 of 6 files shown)


klssmith:
I've noticed that we were sometimes showing the special NHS page, but only in cases where NHS was in the branding pool _and_ the service belonged to an NHS org. In this case, the NHS branding would appear on the form twice by default - once with the real ID and once with the fake ID of "nhs". We would [remove the real ID](https://github.com/alphagov/notifications-admin/blob/master/app/main/forms.py#L2043-L2044) to make the form redirect to the right place.

We could now deduplicate the NHS branding on the form by removing the fake ID instead and passing the real ID to the endpoint (just for consistency), but I'm not too sure it matters since the page now works.

quis:
@klssmith I think we can do one better and remove the fake `nhs` ID entirely now. Have a look at ae0e32e8e737590145dae55e580f3683b47eb3d8 and see what you think

klssmith:
@quis I really like how that last change simplifies things and removes the fake ID, which was a bit of a special case. Could we just remove [these two lines](https://github.com/alphagov/notifications-admin/blob/d4b2ae9abe4197cce9ceaf6cc5edce1e62136e99/app/utils/branding.py#L16-L17) completely though? They're adding the NHS branding for any services that belong to an NHS org, but those organisations should have the NHS branding in their branding pool already, so the services will already see NHS branding on the form

quis:
@klssmith it’s a bit of an edge case but those two lines also work for services where `service.organisation_type` is one of the NHS variants, but where `service.organisation_id` is `None`.

What would need to happen to trigger this:
- someone signs up with an `@nhs.net` email address
- selects something like ‘NHS central’ when asked ‘Who runs this service’
- makes an email branding request for something custom, request is approved
- decides they want to change back to NHS

klssmith:
Oh yep, I forgot you could have an organisation type set on a service but without the service belonging to an org

PR #4371, opened 2022-09-29
Ensure that changing service branding to GOV.UK does not 404

CrystalPea:
When changing service branding to GOV.UK, we are actually changing it to no branding, as GOV.UK is what we show as a default when a service or organisation has no branding set.

When we introduced a page confirming if new service branding should be added to the pool, we introduced a bug where the view was trying to find a branding with the id of __NONE__.

This commit fixes that bug.


samuelhwilliams (approved):
😱 I had no idea about the `__NONE__` special value. Great find @klssmith / @CrystalPea 

Tested and confirm this fixes the bug

PR #4377, opened 2022-10-05
Bump utils to 57.0.1

quis:
Fixes an issue when special characters are used in placeholders:

https://github.com/alphagov/notifications-utils/pull/980


idavidmcdonald (approved):
Can these go out in either order or is there a specific order needed for the admin and api deploys?

PR #4381, opened 2022-10-06
Store emailed address of archived user when archiving a user in the events table

klssmith:
We now store the email address of the user being archived in the events
table as part of the `archive_user` event.

This will enable us, in a separate story, to delete the user's name and
email from their row in the user table when we archive them and not
have to introduce a new job to remove user data that is older than 1
year for archived users.

### DIFF
--- a/app/main/views/find_users.py
+++ b/app/main/views/find_users.py
@@ -33,6 +33,9 @@ def user_information(user_id):
 @user_is_platform_admin
 def archive_user(user_id):
     if request.method == "POST":
+        user = User.from_id(user_id)
+        original_email_address = user.email_address
+
         try:
             user_api_client.archive_user(user_id)
         except HTTPError as e:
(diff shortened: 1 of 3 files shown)


app/main/views/find_users.py
```diff
@@ -42,7 +45,10 @@ def archive_user(user_id):
                     "check all services have another team member with manage_settings"
                 )
                 return redirect(url_for("main.user_information", user_id=user_id))
-        create_archive_user_event(user_id=str(user_id), archived_by_id=current_user.id)
+
+        create_archive_user_event(
```
samuelhwilliams, line 49: This is obviously not related to your PR at all - but it feels less than ideal that we are creating audit events like this rather than 1) in the api as part of the action itself, or 2) encapsulated in some kind of single archive_user function in admin. Do you know why we do this?

samuelhwilliams (approved):
Change looks good 👍 

One question about historical decisions.

app/main/views/find_users.py
klssmith: I'm not sure why we have events done like this, but it appears that we set this up right at the start of Notify - https://github.com/alphagov/notifications-admin/pull/516

PR #4385, opened 2022-10-07
Remove GOV.UK Frontend Toolkit [part 10]

klssmith:
This is part 10 of a series of work to remove the GOV.UK Frontend Toolkit library from our codebase:

https://www.pivotaltracker.com/story/show/182596710

We want to remove GOV.UK Frontend Toolkit from our code to replace it with GOV.UK Frontend.

This is the final part, where we remove the package and all references to it. The last two commits also tidy up a couple of other things that were missed in previous work.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -228,7 +228,6 @@ const sass = () => {
 const images = () => {
   return src([
       paths.src + 'images/**/*',
-      paths.template + 'assets/images/**/*',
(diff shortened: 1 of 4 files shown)


tombye (approved):
The changes look good to me, the frontend build works, the tests run green and I'm seeing no errors in the console for the compiled script so 👍🏻

PR #4390, opened 2022-10-17
Change namespace for Notify's JS modules pattern

tombye:
This changes the name of [the JS modules pattern](https://github.com/alphagov/notifications-manuals/wiki/GOVUK-Modules-pattern) we use to be specific to Notify so it doesn't clash with the one GOVUK Frontend uses.

https://trello.com/c/CWddqvWb/33-rewrite-our-js-modules-to-work-with-govuk-frontend

We identify the elements in our HTML we want to have JS behaviours by giving them a `data-module` attribute, set to the name of the JS we want to run for it. From version 3, GOVUK Frontend does the same so we won't be able to have components associated with both GOVUK Frontend and Notify JS.

GOVUK Frontend 3 also [hard-codes it into the HTML in their templates](https://github.com/alphagov/govuk-frontend/blob/v3.0.0/src/govuk/components/button/template.njk#L36) so any `data-module` attributes passed in are ignored. We already have JS, like `GOVUK.Modules.RegisterSecurityKey.js` which is added to a GOVUK Frontend button component, that would be broken by this.

This proposes changing the names used by our JS modules pattern to avoid the clash.

To that end, this pull request changes the following name changes:

|Current name|New name|Description|
|---|---|---|
|`GOVUK.Modules`|`GOVUK.NotifyModules`|The namespace[^1] used to hold the different module scripts|
|`GOVUK.modules`|`GOVUK.notifyModules`|The code that matches HTML to module script and provides methods to intialise them|
|`data-module`|`data-notify-module`|Attribute used to bind HTML to Notify's JS modules|

[^1]: Note that in client-side JS, the `window` object **is** the global scope. Because of this JavaScripts tend to expose their public interfaces through objects attached to it, like `window.GOVUK`, to avoid naming clashes. This is usually referred to as 'namespacing'.

### DIFF
--- a/app/assets/javascripts/addEmailBrandingOptionsForm.js
+++ b/app/assets/javascripts/addEmailBrandingOptionsForm.js
@@ -177,4 +177,4 @@
     `).get(0);
   };
 
-})(window.GOVUK.Modules);
+})(window.GOVUK.NotifyModules);
--- a/app/assets/javascripts/authenticateSecurityKey.js
+++ b/app/assets/javascripts/authenticateSecurityKey.js
@@ -1,7 +1,7 @@
 (function (window) {
   "use strict";
 
-  window.GOVUK.Modules.AuthenticateSecurityKey = function () {
+  window.GOVUK.NotifyModules.AuthenticateSecurityKey = function () {
     this.start = function (component) {
       $(component)
         .on('click', function (event) {
--- a/app/assets/javascripts/autofocus.js
+++ b/app/assets/javascripts/autofocus.js
@@ -24,4 +24,4 @@
     };
   };
 
-})(window.GOVUK.Modules);
+})(window.GOVUK.NotifyModules);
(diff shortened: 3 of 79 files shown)


klssmith:
I think there are a few places where `data-module` still needs to be replaced with `data-notify-module`, for example in forms.py in the `AdminEditEmailBrandingForm` - the banner colour preview isn't working since we're still using `data-module` there.

Doing a search for `data-module` in the code brings back 44 results. I think some are fine since they are referring to code from GOV.UK Frontend, but others such as [this place](https://github.com/alphagov/notifications-admin/blob/a54f1842aac3efd90f424e6dd9b1e1c197604d4f/tests/app/main/views/organisations/test_organisations.py#L284) (which hasn't been changed) is our JavaScript, so presumably needs changing

tombye:
Great catches! I updated the tests and 'forms.py', as well as the textbox component which sets `data-module` using the WTForms pattern of converting params in snake_case to kebab-case so was ignored in searches. I think that should mean all remaining `data-module`s are either GOVUK Frontend components or our version of one (`app/templates/forms/fields/checkboxes/template.njk`).

klssmith (approved):
Looks good to me and worked when I tried this locally

PR #4392, opened 2022-10-19
Refactor email and letter branding into their own models to make `utils/branding.py` simpler

quis:
I noticed in https://github.com/alphagov/notifications-admin/pull/4282#discussion_r913232533 that @leohemsted thought `get_email_choices` could do with being made less complex.

Feels like now is a good time to do some refactoring, before we get into building a letter branding pool as well.

This pull request broadly moves logic from the view layer and the `utils/branding.py` file into model classes for email and letter branding. 

This generally means we get nicely-named properties, instead of manipulating or introspecting raw data structures.

### DIFF
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -4,6 +4,7 @@ from notifications_python_client.errors import HTTPError
 from app import email_branding_client
 from app.main import main
 from app.main.forms import AdminEditEmailBrandingForm, SearchByNameForm
+from app.models.branding import AllEmailBranding, EmailBranding
 from app.s3_client.s3_logo_client import (
     TEMP_TAG,
     delete_email_temp_file,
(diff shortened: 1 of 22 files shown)


klssmith (approved):
This makes things much clearer to read, particularly the logic in the `get_email_choices` function.

We've added a few more properties to email branding since this was created, so you could potentially add those to the email branding model too - we're not using them yet though, so they could also be added when we need them.

quis:
@klssmith I’ll leave those out for now – not clear if we’re going to use the data from [those new columns](https://github.com/alphagov/notifications-api/commit/e329c0b308683950919406f505adb666534edbf4#diff-90c680eba43456b516da8b4c2573d467ae17d1b0ed4373549f2a593ced3616d5R243-R246) in the UI anywhere. Thanks for the review!

PR #4393, opened 2022-10-21
Record creator/updater for email branding

samuelhwilliams:
When a user creates or updates an email brand, let's send across the user id to the API so it can be recorded in the database.

On updates, let's also record an event including the old email branding, in case we ever need to look back in time.

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -18,6 +18,7 @@ EVENT_SCHEMAS = {
         "provider_restriction",
     },  # noqa: E501 (length)
     "archive_service": {"service_id", "archived_by_id"},
+    "update_email_branding": {"email_branding_id", "updated_by_id", "old_email_branding"},
 }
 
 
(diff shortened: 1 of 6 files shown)


app/event_handlers.py
```diff
@@ -57,6 +58,10 @@ def create_archive_service_event(**kwargs):
     _send_event("archive_service", **kwargs)
 
 
+def create_email_branding_event(**kwargs):
```
klssmith, line 61: I found this name slightly confusing because it's the event that happens when we update email branding, but at first glance I expected this to be called when we _create_ email branding. It could be named `create_update_email_branding_event` instead, but I'm not sure that's much clearer, so feel free to ignore and go with what you think is best 😅

klssmith (approved):

app/event_handlers.py
samuelhwilliams: I think yours is slightly clearer, good call :) But yes it is confusing both ways >_<

PR #4394, opened 2022-10-21
Add pages to generate a government identity system logo

quis:
Test locally at http://localhost:6012/email-branding/create-government-identity?text=Government%20Digital%20Service

image

image

image


tests/app/main/views/test_email_branding.py
```diff
@@ -478,3 +479,88 @@ def test_colour_regex_validation(
...
+        ".create_email_branding_government_identity",
+    )
+    assert [
+        (
+            input["name"],
+            normalize_spaces(page.select_one("label[for=" + input["id"] + "]").text),
+            input["value"],
+        )
+        for input in page.select("input[type=radio]")
+    ] == [
```
CrystalPea, line 569: This is a very busy assertion, a bit hard to follow. It would help if we defined sides of the assertion as variables and then compared those. Like:

```
expected_stripes_options = [all the stripes options]

radios_on_the_page = [ the comprehension list for getting radio labels ]

assert radios_on_the_page ==  expected_stripes_options
```

We could do the same for crests, too.

app/main/forms.py
```diff
@@ -2474,47 +2474,57 @@ class ChangeSecurityKeyNameForm(StripWhitespaceForm):
     )
 
 
-class GovernmentIdentityOptions(StripWhitespaceForm):
+def markup_for_crest_or_insignia(filename):
```
CrystalPea, line 2477: Is there advantage to defining this markup here as opposed to the template? 

Asking since it feels more naturally at home in the template.

app/templates/views/email-branding/government-identity-options-colour.html
```diff
@@ -0,0 +1,26 @@
+{% extends "views/platform-admin/_base_template.html" %}
+{% from "components/file-upload.html" import file_upload %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import sticky_page_footer %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/back-link/macro.njk" import govukBackLink %}
+
+{% block per_page_title %}
+  {{ '{} email branding'.format('Update' if email_branding else 'Add')}}
```
CrystalPea, line 9: same here, don't we create a new branding each time as it is?

app/templates/views/email-branding/government-identity-options-colour.html
```diff
@@ -0,0 +1,26 @@
...
+{% endblock %}
+
+
+{% block backLink %}
+  {{ govukBackLink({ "href": url_for('.email_branding') }) }}
+{% endblock %}
+
+{% block platform_admin_content %}
+
+  {{ page_header('{} email branding'.format('Update' if email_branding else 'Add')) }}
```
CrystalPea, line 19: It would be good to define content for page title and header once only, and then pass that variable. @samuelhwilliams has shown me recently that this also can be done using form's field label (like [here](https://github.com/alphagov/notifications-admin/blob/master/app/templates/views/service-settings/branding/add-new-branding/government-branding-or-own-logo.html#L10))

tests/app/main/views/test_email_branding.py
```diff
@@ -486,70 +486,157 @@ def test_create_email_branding_government_identity_form(client_request, platform
...
+
+@pytest.mark.parametrize(
+    "extra_args",
+    (
+        {},
+        {"filename": "foo"},
+        {"filename": "foo.png"},
+    ),
+)
+def test_create_email_branding_government_identity_colour_400_if_no_filename(
```
CrystalPea, line 617: Could we rename this to `test_create_email_branding_government_identity_colour_400_if_no_filename_or_file_doesnt_exist` ?

CrystalPea (changes requested):
Good work, this will make setting the government logo much easier for the admins 👏🏼 

I left some comments.

Also, I gather integrating this picker with the Zendesk ticket should be a separate PR and story? I will create a story for this in Trello 👍🏼

app/main/forms.py
```diff
@@ -2467,3 +2472,59 @@ class ChangeSecurityKeyNameForm(StripWhitespaceForm):
...
+
+class GovernmentIdentityCoatOfArmsOrInsignia(StripWhitespaceForm):
+
+    coat_of_arms_or_insignia = GovukRadiosField(
+        "Coat of arms or insignia",
+        choices=[
+            (name, markup_for_crest_or_insignia(f"{name}.png") + name)
+            for name in sorted(GOVERNMENT_IDENTITY_SYSTEM_CRESTS_OR_INSIGNIA)
+        ],
+        validators=[DataRequired()],
```
samuelhwilliams, line 2508: I think we actually don't want this validator when using GovukRadiosField. I discovered recently that this changes the error message shown when no option is selected from 'Select an option' to 'Cannot be empty' - I think we prefer the former?

app/models/branding.py
```diff
@@ -106,3 +108,125 @@ def __init__(self, id):
         self.items = tuple()
         if id:
             self.items = self.client_method(id)
+
+
+GOVERNMENT_IDENTITY_SYSTEM_COLOURS = [
```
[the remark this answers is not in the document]
samuelhwilliams: I also think the list-of-dicts structure is slightly verbose here. Could we not just have it be a dict of `{"org_name": "#aabbcc", ...}`?

CrystalPea (commented):
I [added a story](https://trello.com/c/kneFiwYk/98-integrate-crest-picker-with-the-zendesk-ticket) for the integration with Zendesk now.

app/main/views/email_branding.py
```diff
@@ -95,6 +100,36 @@ def update_email_branding(branding_id, logo=None):
...
+@main.route("/email-branding/create-government-identity", methods=["GET", "POST"])
+@user_is_platform_admin
+def create_email_branding_government_identity():
+    form = GovernmentIdentityOptions()
+
+    if form.validate_on_submit():
+        filename = f"{form.coat_of_arms_or_insignia.data}.png"
+        image_file = INSIGNIA_ASSETS_PATH / filename
+        upload_filename = upload_email_logo(
+            email_safe(filename),
```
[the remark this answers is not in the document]
quis: Yes, for example https://static-logos.notifications.service.gov.uk/e688b983-a2ed-4985-a1b6-9e770de1d175-blaby.png

app/models/branding.py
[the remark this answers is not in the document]
quis: Changed to a `dict` in 6e984bbf32004ab3bd0adbe7403d243640f017df

Personally I feel like models are the place to get data about a specific domain. Sometimes that’s from the API, sometimes from the filesystem, sometimes hardcoded.

For me a utils are more like for standalone functions which take primitive data types as arguments. But that distinction might just be in my head 🤷🏻

app/main/forms.py
quis: I don’t think there’s an easier way to inject HTML into the field labels now that we’re using the componentised form fields.

app/templates/views/email-branding/government-identity-options-colour.html
quis: I’ve fixed the page titles by using the form label instead: e35272e64cf33bdc0c527f9c959362310cf0d494

app/templates/views/email-branding/government-identity-options-colour.html
quis: I’ve fixed the page titles by using the form label instead: e35272e64cf33bdc0c527f9c959362310cf0d494

app/main/forms.py
quis: Made the validation better in defd96540c01742a99b277506200a45010184394

tests/app/main/views/test_email_branding.py
quis:
```suggestion
def test_create_email_branding_government_identity_colour_400_if_no_filename_or_file_doesnt_exist(
```

app/templates/views/email-branding/government-identity-options.html
```diff
@@ -0,0 +1,27 @@
+{% extends "views/platform-admin/_base_template.html" %}
+{% from "components/file-upload.html" import file_upload %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import sticky_page_footer %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/back-link/macro.njk" import govukBackLink %}
+
+{% block per_page_title %}
+  {{ '{} email branding'.format('Update' if email_branding else 'Add')}}
```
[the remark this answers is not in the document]
quis: I’ve fixed the page titles by using the form label instead: e35272e64cf33

app/main/views/email_branding.py
```diff
@@ -95,6 +100,36 @@ def update_email_branding(branding_id, logo=None):
...
 
+@main.route("/email-branding/create-government-identity", methods=["GET", "POST"])
+@user_is_platform_admin
+def create_email_branding_government_identity():
+    form = GovernmentIdentityOptions()
+
+    if form.validate_on_submit():
+        filename = f"{form.coat_of_arms_or_insignia.data}.png"
+        image_file = INSIGNIA_ASSETS_PATH / filename
+        upload_filename = upload_email_logo(
```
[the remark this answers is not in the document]
quis: Yes, but this seems to be the way the existing page works. It goes through these steps:
- upload the logo
- page submits the form automatically and logo is stored to S3 with a ‘temporary’ file name
- form is submitted again with text and colour fields filled in
- logo is re-uploaded with a new ‘temporary’ file name
- old temporary logo is deleted
- new temporary logo is copied to the same bucket but with a permanent file name
- new temporary logo is deleted

It’s a mess but:
- at least we’re not storing duplicates permanently
- fixing it is definitely outside the scope of this story

tests/app/main/views/test_email_branding.py
quis: Refactored in aa084100dae14e379a3252cd634eb83fd29097ac

app/main/views/email_branding.py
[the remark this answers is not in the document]
CrystalPea: Ahhh, interesting, thanks!

app/main/views/email_branding.py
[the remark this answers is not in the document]
CrystalPea: I had no idea this is how it's currently working, wow 🤯. But I guess if they are temporary maybe it's not that bad. We can look at it when doing clean up at the finish line for this epic.

app/templates/views/email-branding/government-identity-options-colour.html
CrystalPea: Nice one 👏🏼

app/models/branding.py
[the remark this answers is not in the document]
CrystalPea: I see what you mean. I think it would also go well in app/utils/branding.py, but also I don't feel strongly about this, and I think both perspectives have merit, so I resolved this convo 👍🏼

CrystalPea (approved):
Looks good to me

PR #4395, opened 2022-10-21
Remove JavaScript and styles for RadioSlider

klssmith:
The radio slider was used to change the SMS provider priority, but we stopped using it in 7f333ba5fe0e570284d47c55aa0b7b0c082558d1.


tombye (approved):
Looks good to me 👍🏻

PR #4398, opened 2022-10-26
Refactor email and letter branding into their own models to make `utils/branding.py` simpler

klssmith:
This is re-opening https://github.com/alphagov/notifications-admin/pull/4392 by @quis, which was reverted. The original commits are the same, with two new ones added to the end.

**Original PR description**

I noticed in https://github.com/alphagov/notifications-admin/pull/4282#discussion_r913232533 that @leohemsted thought get_email_choices could do with being made less complex.

Feels like now is a good time to do some refactoring, before we get into building a letter branding pool as well.

This pull request broadly moves logic from the view layer and the utils/branding.py file into model classes for email and letter branding.

This generally means we get nicely-named properties, instead of manipulating or introspecting raw data structures.

**New commits**

1. Since this was merged we've added a new event, so this updates the code for that which would no longer work as expected.
2. We saw an issue when this was deployed, which the second new commit fixes - there was one place where we weren't expecting `LetterBranding` to be a class so there was an error when trying to serialize it

### DIFF
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -65,7 +65,9 @@ def update_email_branding(branding_id, logo=None):
                 updated_by_id=current_user.id,
             )
             create_update_email_branding_event(
-                email_branding_id=branding_id, updated_by_id=str(current_user.id), old_email_branding=email_branding
+                email_branding_id=branding_id,
+                updated_by_id=str(current_user.id),
+                old_email_branding=email_branding.serialize(),
             )
         except HTTPError as e:
             if e.status_code == 400 and "name" in e.response.json().get("message", {}):
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -6,6 +6,7 @@ from app import email_branding_client
 from app.event_handlers import create_update_email_branding_event
 from app.main import main
 from app.main.forms import AdminEditEmailBrandingForm, SearchByNameForm
+from app.models.branding import AllEmailBranding, EmailBranding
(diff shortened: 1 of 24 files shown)


app/main/views/email_branding.py
```diff
@@ -65,7 +65,9 @@ def update_email_branding(branding_id, logo=None):
                 updated_by_id=current_user.id,
             )
             create_update_email_branding_event(
-                email_branding_id=branding_id, updated_by_id=str(current_user.id), old_email_branding=email_branding
+                email_branding_id=branding_id,
+                updated_by_id=str(current_user.id),
+                old_email_branding=email_branding._dict,
```
samuelhwilliams, line 70: Is this the 'official' way to get the data out? Feels a bit dirty. Should we add a `.serialize()` method, or just make it a public attribute?

klssmith: No, we shouldn't really be calling `_dict` outside the class. I've made a `serialize` method instead.

app/models/branding.py
```diff
@@ -40,6 +40,9 @@ def is_nhs(self):
     def is_govuk(self):
         return self.id is None
 
+    def serialize(self):
+        return self._dict
```
samuelhwilliams, line 44: Small thing, but this should probably return a _copy_ rather than the actual internal dict. Otherwise someone calling this and mutating the return value would break the original object.

klssmith: 👍 this is updated now to return a copy

quis:
Thanks @klssmith 🙇🏻

PR #4399, opened 2022-10-26
Always cf-target before cf-deploy

samuelhwilliams:
I accidentally did `make preview cf-deploy` while targetting `production` (from a previous dig into the DB), expecting it to target the correct paas space automatically.

It did not and we got an alert in prod. Fortunately the deploy was rejected because of a config mismatch so nothing went wrong, but let's protect ourselves from my stupidity anyway...


sakisv (approved):
Yikes, I thought it did it

PR #4400, opened 2022-10-26
Forward limited headers from template-preview

samuelhwilliams:
We have experienced issues locally with template-preview responses to the admin app because they include a Transfer-Encoding:chunked header that indicates the response is a streaming response. By including these in the response from admin to the browser, the browser gets confused, because it receives a standard non-streamed HTTP response but has a header telling it to expect chunks. This leads to issues and doesn't render the template preview correctly in the iframe.

Let's only forward specific headers (eg the Content-Type) and allow flask on the admin app to set the rest as it sees fit.

### DIFF
--- a/app/template_previews.py
+++ b/app/template_previews.py
@@ -9,6 +9,12 @@ from app import current_service
 
 
 class TemplatePreview:
+    @staticmethod
+    def get_allowed_headers(headers):
+        header_allowlist = {"content-type", "cache-control"}
+        allowed_headers = {header: value for header, value in headers.items() if header.lower() in header_allowlist}
+        return allowed_headers.items()
+
     @classmethod
     def from_database_object(cls, template, filetype, values=None, page=None):
         data = {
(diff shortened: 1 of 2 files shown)


app/template_previews.py
```diff
@@ -9,6 +9,12 @@
 
 
 class TemplatePreview:
+    @staticmethod
+    def get_allowed_headers(headers):
+        header_allowlist = {"content-type", "cache-control"}
```
samuelhwilliams, line 14: Maybe we don't want `cache-control` here?

CrystalPea (approved):

app/template_previews.py
samuelhwilliams: @klssmith - I mentioned briefly yesterday that removing cache-control here might fix the slow preview loading you'd seen recently, but looks like no - admin sets no-cache on all responses anyway, and that hasn't been added particularly recently, so think the problem may well still be something else :( 🤔

klssmith: Oh well - at least this will make things work locally, even if it doesn't solve the issue of slow loading

PR #4405, opened 2022-10-27
Bump GOVUK Frontend to version 3

tombye:
Brings in version 3 of GOVUK Frontend, the code library for the GOV.UK Design System.

https://trello.com/c/IDaUxo8b/31-bump-govuk-frontend-to-version-3

Note: this pull request was originally raised in draft form, to enable comments from @quis on the changes GOVUK Frontend v3 makes to our visual design.

The list of things we requested comments on specifically were:
- placeholders in textboxes now sit higher on the line, we think because the new font sits differently[^1]
- should we update email styles too (we haven't yet)?
- are there any parts of utils that generate HTML we should update?
- footer links in v3.0.0 have no underline so we hacked one in (this was raised as an accessibility issue before and they have an underline again from v3.12.0
- hints, and anything else using 'dark-grey' or $govuk-secondary-text-colour are still grey-1 in v3. They changed it to be darker in v3.8.0. We haven't hacked this in but could

[^1]: See blog post about this: https://designnotes.blog.gov.uk/2019/07/29/weve-updated-the-gov-uk-colours-and-font/


tombye (commented):
Noticed a few things that don't look right but overall the changes read as sensible and don't break things 👍🏻

quis:
> * placeholders in textboxes now sit higher on the line, we think because the new font sits differently[1](#user-content-fn-1-931a4e870123a8bcce7ec512d9914584)

I think we can adjust for this by changing the `box-shadow` and `border-radius` so the top and bottom edges of the yellow background meet the tips of the inner bracket again. It’s not perfect, but this is closer:

image

```css
border-radius: 11px;
box-shadow: inset 0.47em 0 0 0 #fff, inset -0.47em 0 0 0 #fff, inset 0 -0.05em 0 0 #fff, inset 0 0.25em 0 0 #fff;
```

> * should we update email styles too (we haven't yet)?
> * are there any parts of utils that generate HTML we should update?

HTML in utils (including email templates) is mostly in:
- https://github.com/alphagov/notifications-utils/tree/master/notifications_utils/jinja_templates
- https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/markdown.py

I would make updating the email template a separate step.

> * footer links in v3.0.0 have no underline so we hacked one in (this was raised as an accessibility issue before and they have an underline again from v3.12.0

👍🏻 

> * hints, and anything else using 'dark-grey' or $govuk-secondary-text-colour are still grey-1 in v3. They changed it to be darker in v3.8.0. We haven't hacked this in but could

Should be OK to wait until v4 since we plan to upgrade again imminently.

***

# Other stuff I noticed

## Greys 

The grey buttons here look very washed out. Can we override with our own grey?

image

Where we are still using the old grey, the hue is a bit off and doesn’t match the footer. The old grey is a cool grey, the new grey (footer) is a warm grey:

image

Changing the old grey from `#e0e1e2` to `#e2e1e0` keeps the same luminance but warms the colour up so it matches better:

image

Would suggest also using `#e2e1e0` for secondary buttons.


## Hover states of folders

Blue in the hover state of folders doesn’t match the colour of the text:

image

tombye:
Thanks for all the excellent feedback @quis. I've added the following commits to address the issues you spotted:
- [fix for the placeholder style](https://github.com/alphagov/notifications-admin/pull/4405/commits/b550d9c071024f94d9ad0943d7f179ed1db677bf), only tweaked to stop an odd rendering artefact in Safari
- [change secondary button style to #e2e1e0](https://github.com/alphagov/notifications-admin/pull/4405/commits/aec4eb56778a84e885c4f9ad8deda08b07422944), made from `govuk-shade(govuk-colour("light-grey"), 7%)`
- [change replacement for 'grey-3' to #e2e1e0](https://github.com/alphagov/notifications-admin/pull/4405/commits/d553798c70cf2713cd708fe0fcc4ab3b2b4b4eb9), made from `govuk-shade(govuk-colour("light-grey"), 7%)`
- [fix hover colour for folder icon](https://github.com/alphagov/notifications-admin/pull/4405/commits/96a5d439e551d77c2fa625809eff36685edcf0d8)

Note: I talked with Chris outside of this pull request about the hover colour for the updated button colour. He was happy with `#c7c6c6`, made with `govuk-shade(govuk-colour("light-grey"), 18%);`, which is what went in. It's based on the footer colour and matches the warmth while lowering the lightness.

quis:
Thanks Tom. I’m happy with how everything looks now. This is good to ship as far as I’m concerned.

PR #4413, opened 2022-11-02
Add urls and view names for upload your logo journey

CrystalPea:
So that it is easier to work on those views in parallel and link them to one another.


app/navigation.py
```diff
@@ -223,15 +223,19 @@ class MainNavigation(Navigation):
         "settings": {
             "add_organisation_from_gp_service",
             "add_organisation_from_nhs_local_service",
+            "email_branding_choose_logo",
```
leohemsted, line 226: how come this has been added to nav now? this isn't a new route. if so, should it be in test_nav too?

CrystalPea: It already is in test_nav, I think it got eaten during a tricky rebase yesterday evening, so I re-added it.

leohemsted (approved):

PR #4417, opened 2022-11-03
Enforce `page.select` over `page.find` in tests

quis:
In https://github.com/alphagov/notifications-admin/pull/4308 we replaced uses of `page.find` with `page.select`, to standardise on one way of doing things.

However I’ve seen uses of `page.find` creeping back in. This commit prevents `page.find` by raising an exception when someone tries to use it.

This also has the (welcome) side effect of removing the third way of selecting things with BeautifulSoup, the magic (in a bad way) attribute access, for example `page.form` or `page.h1`.

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,6 +16,7 @@ from notifications_utils.url_safe_token import generate_token
 from app import create_app, webauthn_server
 
 from . import (
+    NotifyBeautifulSoup,
     TestClient,
     api_key_json,
     assert_url_expected,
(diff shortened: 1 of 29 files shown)


tests/conftest.py
```diff
@@ -2742,6 +2751,8 @@ def get_url(
                 assert resp.location == _expected_redirect
 
             page = BeautifulSoup(resp.data.decode("utf-8"), "html.parser")
+            block_method(page, "find", preferred_method_name="select_one")
+            block_method(page, "find_all", preferred_method_name="select")
```
samuelhwilliams, line 2755: Looks like doing this here for now is OK and covers all of the find/find_all uses, but should we just be monkeypatching the BeautifulSoup class directly rather than a single page instance?

samuelhwilliams (approved):
👍

One comment on a possible alternative patch, but good otherwise

tests/conftest.py
quis: Good point, I’ve rewritten as a class in 1632ddf. This has also made it reusable in other places, which caught a few more `find` and `find_all`s.

samuelhwilliams: Mm, that works, but I guess it suffers the same problem that you're trying to fix here: people can forget to use x and use y instead, and that needs to get caught by manual review.

If people use BeautifulSoup directly rather than our custom class, the methods won't be blocked.

If we want to fix this we probably need to be slightly magical and directly monkeypatch the methods on BeautifulSoup.

quis: There are only [two tests](2449828dd3ec32ba74bef86555936f96c57dd69b) which use `BeautifulSoup` directly. All the others get it automatically as the return value of methods on the `client_request` fixture. So while I agree it is possible to to accidentally use the wrong thing it does feel like very diminishing returns trying to stop it automatically.

PR #4418, opened 2022-11-03
Convert text-input fields to GOVUK Jinja macros

leohemsted:
Change how we render text-input fields to use the [text-input component from GOVUK Jinja macros](https://github.com/LandRegistry/govuk-frontend-jinja/blob/1.5.1/govuk_frontend_jinja/templates/components/input/macro.html).

Part of https://trello.com/c/0BFf4jLR/38-move-templating-to-use-govuk-frontend-jinja-macros.

### Addition of render_govuk_frontend_macro helper function

The new jinja macros don't have template files we can call directly using `render_template` so we need to create our own mini template just to call the macro. We'll need to do this for each forms component we use from GOVUK Frontend so we decided to generate the template dynamically.

The dict struct that the template is constructed through has a key which maps to a design system component (in this case, [text-input](https://design-system.service.gov.uk/components/text-input/), and then contains a path to that file (in this case https://github.com/LandRegistry/govuk-frontend-jinja/blob/1.5.1/govuk_frontend_jinja/templates/components/input/macro.html - note the 1.5.1 tag since that's the version of govuk_frontend_jinja we install), and the name of the macro within that file (`govukInput` in this case).

## How to review

To review this PR (and other similar future PRs) you'll want to check out the code and test the component in its various guises around the site. For text-input, this includes text input, email input, password input, search input, uk and international mobile phone numbers.

The GOVUK Frontend jinja macros have a testsuite that uses GOVUK Frontend fixtures (like https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/input/input.yaml) so we can be confident most use cases will be fine. They aren't exhaustive though and the WTForms code and our CSS both have the potential to change how things render so we do need to check them in pages. When you do that, the important things to be aware of are "do the various parameters we pass in still map to values in the jinja template". So in the case of textInput, this is stuff like the `type` field (which sets the html input type), and any params/param_extensions we pass in, such as extra classes or input hints.

I don't think we need to exhaustively check every single usage of every text field, but we should at least cover all the major sub types.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -5,7 +5,7 @@ from itertools import chain
 from numbers import Number
 
 import pytz
-from flask import Markup, render_template, request
+from flask import Markup, render_template, render_template_string, request
 from flask_login import current_user
 from flask_wtf import FlaskForm as Form
 from flask_wtf.file import FileAllowed
@@ -86,6 +86,36 @@ from app.utils.user_permissions import (
     broadcast_permission_options,
     permission_options,
 )
+
+
+def render_govuk_frontend_macro(component, params):
+    """
+    jinja needs a template to render
+
+    This function creates a template string that just calls that macro on its own.
+
+    ```
+    {%- from <path> import <macro> -%}
+
+    {{ macro(params) }}
+    ```
+
+    This function dynamically fills in the path and macro based on the GOVUK_FRONTEND_MACROS dictionary.
+    Then we render that template with any params to produce just the output of that macro.
+    """
+    govuk_frontend_components = {
+        "text-input": {"path": "govuk_frontend_jinja/components/input/macro.html", "macro": "govukInput"},
+    }
+
+    # we need to duplicate all curly braces to escape them from the f string so jinja still sees them
+    template_string = f"""
+        {{%- from '{govuk_frontend_components[component]['path']}'
+        import {govuk_frontend_components[component]['macro']} -%}}
+
+        {{{{ {govuk_frontend_components[component]['macro']}(params) }}}}
+    """
+
+    return Markup(render_template_string(template_string, params=params))
 
 
 def get_time_value_and_label(future_time):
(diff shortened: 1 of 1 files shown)


app/main/forms.py
```diff
@@ -83,6 +83,36 @@
...
+        "text-input": {"path": "govuk_frontend_jinja/components/input/macro.html", "macro": "govukInput"},
+    }
+
+    # we need to duplicate all curly braces to escape them from the f string so jinja still sees them
+    template_string = f"""
+        {{%- from '{govuk_frontend_components[component]['path']}'
+        import {govuk_frontend_components[component]['macro']} -%}}
+
+        {{{{ {govuk_frontend_components[component]['macro']}(params) }}}}
+    """
```
samuelhwilliams, line 111: Gross :D

samuelhwilliams (approved):

app/main/forms.py
leohemsted: it really is! 

for reference here are the options in the three different string formatting options at our disposal - f strings, string concatenation and old percent interpolation. (ignoring `str.format` as it's basically the same as f strings)

```python
f"""
{{%- from '{govuk_frontend_components[component]['path']}'
import {govuk_frontend_components[component]['macro']} -%}}
{{{{ {govuk_frontend_components[component]['macro']}(params) }}}}
"""
```

```python
"{%- from '" + govuk_frontend_components[component]['path'] + "' \n" +
"import " + govuk_frontend_components[component]['macro'] +" -%} \n" + 
"{{ " + govuk_frontend_components[component]['macro'] + "(params) }}"
```

```python
"""
    {{%%- from '%s'
    import %s -%%}}
    {{ %s(params) }}
""" % (
    govuk_frontend_components[component]["path"],
    govuk_frontend_components[component]["macro"],
    govuk_frontend_components[component]["macro"],
)
```

PR #4420, opened 2022-11-03
lookup insignias in compiled assets path

leohemsted:
the normal app/assets path is only for dev (ie: unminified js, uncompiled scss, etc). When we build artifacts to deploy, we move the files in to app/static (and minify, compile, etc) using our gulpfile.

Make sure the new government insignia lookup uses this path rather than the assets path

### DIFF
--- a/app/models/branding.py
+++ b/app/models/branding.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+from app import asset_fingerprinter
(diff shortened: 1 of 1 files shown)


samuelhwilliams (approved):

CrystalPea:
ohh :o That's what it was!

PR #4422, opened 2022-11-03
Allow letter brandings to be removed from pool

klssmith:
This adds a "Remove" link beneath each letter branding in the pool apart from the default brand. The default is shown separately at the top of the page.

#### Depends on
- [x] https://github.com/alphagov/notifications-api/pull/3630

### With no branding as the default
Screenshot 2022-11-03 at 16 17 45

### With a specific branding as the default
Screenshot 2022-11-03 at 16 17 14

### Confirmation banner
Screenshot 2022-11-03 at 16 38 10

### Flash message after the brand has been removed
Screenshot 2022-11-03 at 16 38 15

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -625,9 +625,22 @@ def organisation_preview_letter_branding(org_id):
     )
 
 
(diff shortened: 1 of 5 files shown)


samuelhwilliams (approved):
👍

PR #4423, opened 2022-11-04
Link to government identity builder from ticket

quis:
So that when someone asks for a government identity logo it’s easy for the person on support to create the branding and apply it.

image


leohemsted (approved):

CrystalPea (approved):
Nice work 👏🏼

PR #4424, opened 2022-11-04
Convert radios to govuk jinja macros

leohemsted:
Part of https://trello.com/c/0BFf4jLR/38-move-templating-to-use-govuk-frontend-jinja-macros.

To review, you'll want to check out the branch and ensure that the various radio button fields still function, handle error states, any custom classes etc.

Note that many radio fields (such as the insane nested folder radios) still extend from WtForms radio rather than Govuk ones, so will not be affected by this PR. (We should update them in the future)

The different types of fields we've identified and tested:

GovukRadiosField
- OptionalGovukRadiosField
  - /services/<uuid:service_id>/service-settings/broadcasts/<broadcast_channel>
- OnOffField - service setting toggles
- OrganisationTypeField - edit org settings
- GovukRadiosFieldWithNoneOption - platform admin add email branding style 

The send files by email radio has some nested text-inputs too - we've checked that the validation and display of those still looks good

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -104,6 +104,7 @@ def render_govuk_frontend_macro(component, params):
     Then we render that template with any params to produce just the output of that macro.
     """
     govuk_frontend_components = {
+        "radios": {"path": "govuk_frontend_jinja/components/radios/macro.html", "macro": "govukRadios"},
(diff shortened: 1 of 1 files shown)


quis (approved):
I’ve tried a couple of the more esoteric ones I can find and can’t see any problems

image

image

PR #4426, opened 2022-11-04
Patches to choose logo page

CrystalPea:
Remove validator from form, so default radio button validation message is not overriden.

Include content for when the service doesn't have a logo yet.

Fix the header - only include text and not html for the label we are using as a header.

TODO: change radio class to radio with button, currently being developed by Sam and Tom.


app/main/forms.py
```diff
@@ -2152,7 +2152,7 @@ class GovBrandingOrOwnLogoForm(StripWhitespaceForm):
     options = GovukRadiosField(
         "Choose a logo for your emails",
         choices=branding_options,
-        validators=[DataRequired()],
+        validators=[],
```
quis, line 2155: You can remove the argument entirely:
```suggestion
```

quis (approved):

PR #4427, opened 2022-11-04
Combine letter branding rows in org settings table 

klssmith:
We had two different rows for letter branding in the organisation settings table - "Default letter branding" and "Letter branding options". This change combines them into one row, matching how we display email branding options.

We can now set the letter branding for an organisation from the "main.organisation_letter_branding" endpoint, so can remove the old pages we had.

### With a real brand as the default
Screenshot 2022-11-04 at 14 25 34

### With no brand as the default
Screenshot 2022-11-04 at 14 26 44

### With no brand as the default and no other brands in the pool
Screenshot 2022-11-04 at 14 27 07


klssmith:
This can be reviewed, but can't be merged until the work to make a branding the default from the "Letter branding options" page is merged.

app/templates/views/organisations/organisation/settings/index.html
```diff
@@ -127,22 +127,28 @@ <h1 class="heading-medium">Settings</h1>
...
-            url_for('.edit_organisation_letter_branding', org_id=current_org.id),
-            suffix='default letter branding for the organisation'
-          )
-        }}
-      {% endcall %}
       {% call row() %}
         {{ text_field('Letter branding options') }}
-        {{ optional_text_field(letter_branding_pool_names, default="None") }}
+        {% set letter_branding_html %}
+          <div {% if current_org.email_branding_pool_excluding_default %}class="govuk-!-margin-bottom-3"{% endif %}>
```
leohemsted, line 133:
```suggestion
          <div {% if current_org.letter_branding_pool_excluding_default %}class="govuk-!-margin-bottom-3"{% endif %}>
```

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -972,17 +971,53 @@ def test_organisation_settings_table_shows_letter_branding_pool(
...
+            "id": "5678",
+            "name": "Department for Education",
+            "filename": "dfe",
+        },
+    )
+
+    client_request.login(platform_admin_user)
+    page = client_request.get(".organisation_settings", org_id=organisation_one["id"])
+
+    email_branding_options_row = page.select("tr")[9]
```
leohemsted, line 1010: be good to call this variable `letter_branding_options_row` just to avoid confusion

leohemsted: also i don't have strong opinions on this, but i'm also wondering about just selecting the 9th row. It feels a bit potentially brittle if we add/remove rows, and doesn't make it clear what data this will return.

`page.select('td:contains("Letter branding options") + td')` (find a td containing the header, then get its next td sibling) - this feels a little bit less brittle, but it's not perfect (using the contains selector has similar issues if we want to tweak the wording etc.

ideally we'd just give each `tr` an `id="letter_branding_options"` or something i guess if we want to refer to them? but then adding ids to our HTML just for the benefit of our unit tests also feels a little smelly.

i wonder if @tombye has any opinions.

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -972,17 +971,53 @@ def test_organisation_settings_table_shows_letter_branding_pool(
...
+
+    assert normalize_spaces(email_branding_options_row.text) == (
+        "Letter branding options "
+        "Department for Education Default "
```
leohemsted, line 1014: it might be worthwhile to check that this particular ~row~ div has the `govuk-!-margin-bottom-3 ` class too (if only because I think that would have caught using the email_branding_excluding_default instead of the letter_branding_excluding_default variable above)

tests/app/main/views/organisations/test_organisations.py
klssmith: I've changed the variable name - that was another copy and paste error.

From looking at other tests where we're checking rows in a table, we do access rows by index quite a lot. I also don't like adding an id just for a test. I've updated the test to use [this function](https://github.com/alphagov/notifications-admin/blob/1d90151ba0b16536dc3be2330d34f6edd00841e9/tests/__init__.py#L675) which I discovered we had, which means we can now select the row by partial content. It's maybe a bit less brittle now, but still not perfect so I'm happy to change it back or to something different.

tests/app/main/views/organisations/test_organisations.py
```diff
@@ -972,17 +971,55 @@ def test_organisation_settings_table_shows_letter_branding_pool(
     client_request.login(platform_admin_user)
     page = client_request.get(".organisation_settings", org_id=organisation_one["id"])
 
-    email_branding_options_row = page.select("tr")[10]
+    email_branding_options_row = page.select("tr")[9]
```
leohemsted, line 974: oops, one other copy paste (even if this one is from a previous PR). be good to make this one use that find element function
```suggestion
    letter_branding_options_row = find_element_by_tag_and_partial_text(page, "tr", "Letter branding options")
```

leohemsted (approved):
one additional minor thing, then this is good to go

tests/app/main/views/organisations/test_organisations.py
klssmith: Thanks, this is now fixed too

PR #4429, opened 2022-11-04
Convert checkboxes to govuk frontend jinja macros

leohemsted:
The old checkboxes were vendored in https://github.com/alphagov/notifications-admin/pull/3535

Some of the changes we made aren't present in govuk frontend 3 so we still have to vendor the new template.

In the process of migrating, we were able to remove one override (item.hint.classes now exists as of govuk-frontend v3.5.0), and one override we were able to refactor to remove too (formGroup=False has been replaced by just defining a different class).

Other than that, copy across the overrides as is. Have tested with the templates page (which has some wild custom checkbox html) and the invite page (which has both normal checkboxes for permissions, and also collapsible nested checkboxes for folder permissions).


samuelhwilliams (approved):
Nicely chunked-up commits, made it a lot easier to review, thanks 👍

klssmith:
Are you bumping GOVUK Frontend in another PR? A commit message mentions not needing to add item.hint.classes since this was introduced in GOVUK Frontend 3.5.0, but we only upgraded to version 3.0.0.

leohemsted:
@klssmith not sure how much of this you're already familiar with and im just parroting so forgive me

While the old CCS library took the nunjuck templates that we copy across from our npm govuk-frontend (that we currently pin to 3.0.0) and hacked them into jinja using regexes, the land registry library simply replicates those nunjucks templates in jinja by hand.

We're using govuk-frontend-jinja v1.5.1 (it has some flask fixes), but as the compat chart shows (https://github.com/landregistry/govuk-frontend-jinja#compatibility), that replicates templates from 3.15.0

So when this (and the other related PRs are merged) we'll be on govuk-frontend 3.0.0 for js, css, fonts, etc. And govuk-frontend 3.15.0 for templates.

bit icky but hope that helps clear up

PR #4431, opened 2022-11-04
Direct central org service users towards government identity journey

CrystalPea:
In this PR, we are direct central org service users towards government identity journey. We also make sure that back link on the Something Else page is correct.

Needs to go in after: https://github.com/alphagov/notifications-admin/pull/4426

### To consider:
- do we still need `.email_branding_organisation` and `.email_branding_govuk_and_org` views?
- if users select `organisation` or `govuk_and_org` as their branding choice, should we somehow take this into consideration? If not, should we even still have those options? They seem potentially confusing in the new journey.

### DIFF
--- a/app/templates/support-tickets/government-logo-branding-request.txt
+++ b/app/templates/support-tickets/government-logo-branding-request.txt
@@ -8,7 +8,15 @@ Service: {{ current_service.name }}
 
 ---
 Government logo text requested: {{ logo_text }}
+{% set create_logo_url = url_for(
+  'main.create_email_branding_government_identity_logo', text=logo_text, _external=True
+) %}
+{% if branding_choice == "govuk_and_org" %}This service requested for both GOV.UK and organisation logo to be visible.
+{% set create_logo_url = url_for(
+  'main.create_email_branding_government_identity_logo', text=logo_text, brand_type="both", _external=True
+) %}
+{% elif branding_choice == "organisation" %}This service requested organisation branding.
 
-Create this logo: {{ url_for('main.create_email_branding_government_identity_logo', text=logo_text, _external=True) }}
+{% endif %}Create this logo: {{ create_logo_url }}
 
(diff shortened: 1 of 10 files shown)


quis:
Ideally the choice of `organisation` or `govuk_and_org` would come through in the ticket. In the first case we probably want to add it to the pool once we’ve created the branding, in the second case we need to make sure we pick ‘GOV.UK and logo’ when creating the branding

CrystalPea:
@quis 

> Ideally the choice of organisation or govuk_and_org would come through in the ticket.
Sure, I can implement this.

 In the first case we probably want to add it to the pool once we’ve created the branding
 > ok, that sounds straight forward

in the second case we need to make sure we pick ‘GOV.UK and logo’ when creating the branding
> gotcha, so best to include it as text in the ticket and maybe also as a request arg so this option gets pre-selected?

app/templates/support-tickets/government-logo-branding-request.txt
```diff
@@ -8,7 +8,15 @@ Service: {{ current_service.name }}
 
 ---
 Government logo text requested: {{ logo_text }}
+{% set create_logo_url = url_for(
+  'main.create_email_branding_government_identity_logo', text=logo_text, _external=True
+) %}
+{% if branding_choice == "govuk-and-org" %}This service requested for both GOV.UK and organisation logo to be visible.
```
quis, line 14: This text isn’t showing up in the ticket when I test locally, picking the _Organisation name and GOV.UK_ radio button
```suggestion
{% if branding_choice == "govuk-and-org" %}This service requested for both GOV.UK and organisation logo to be visible.
```

quis (changes requested):

app/templates/support-tickets/government-logo-branding-request.txt
CrystalPea: Ahh, I put in dashes instead of underscores 🤦🏼‍♀️ . Try now ☺️

PR #4434, opened 2022-11-07
Convert the rest of the elements to govuk frontend jinja macros

leohemsted:
these ones are all pretty simple and didn't need any code changes apart from where they're referenced in imports

* back-link - used in loads of places but we don't do anything exciting with it - we only ever pass in an href, the generated html is identical to the old one.
* inset text - used in guidance, very simple
* govukTextarea - we only use this in one place
* govukErrorMessage - only used in webauthn sign in/adding keys (though used in two different ways)
* govukSummaryList - only used on the user profile page
* govukSkiplink - we only invoke directly on the csv check page (it's also used in the main template on every page though)
* govukDetails - used in a few places for detail/summary components. 
* govukErrorSummary - only used in the new email branding flow
* govukFieldset - used in the list view for org email domains and api key guestlists
* [template.html](https://github.com/LandRegistry/govuk-frontend-jinja/blob/main/govuk_frontend_jinja/templates/template.html) - used on every single page as the base that we build from. Take a look at the source code linked to see what it does

as ever, take a look at each commit for a simpler diff

requires the checkboxes PR:

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -1,4 +1,4 @@
-{% extends "template.njk" %}
+{% extends "govuk_frontend_jinja/template.html"%}
 {% from "components/banner.html" import banner %}
 {% from "components/cookie-banner.html" import cookie_banner %}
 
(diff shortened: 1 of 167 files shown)


tombye (approved):
Went through all the changes and this looks good to go 👍🏻

PR #4435, opened 2022-11-07
Convert govuk button to jinja macro

leohemsted:
the last and most-used macro. A few changes to how we query for it in tests (and once in code) but otherwise relatively straightforward.

- [x] https://github.com/alphagov/notifications-admin/pull/4434

### DIFF
--- a/app/templates/components/cookie-banner.html
+++ b/app/templates/components/cookie-banner.html
@@ -5,10 +5,10 @@
     <h2 class="notify-cookie-banner__heading govuk-heading-m" id="notify-cookie-banner__heading">Can we store analytics cookies on your device?</h2>
     <p class="govuk-body">Analytics cookies help us understand how our website is being used.</p>
     <div class="notify-cookie-banner__buttons">
-      <button class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-accept" type="submit" data-accept-cookies="true" aria-describedby="notify-cookie-banner__heading">
+      <button type="button" class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-accept" type="button" data-accept-cookies="true" aria-describedby="notify-cookie-banner__heading">
         Yes<span class="govuk-visually-hidden">, Notify can store analytics cookies on your device</span>
       </button>
-      <button class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-reject" type="submit" data-accept-cookies="false" aria-describedby="notify-cookie-banner__heading">
+      <button type="button" class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-reject" type="button" data-accept-cookies="false" aria-describedby="notify-cookie-banner__heading">
         No<span class="govuk-visually-hidden">, Notify cannot store analytics cookies on your device</span>
       </button>
       <a class="govuk-link notify-cookie-banner__link" href="/cookies">How Notify uses cookies</a>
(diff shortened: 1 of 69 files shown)


tests/javascripts/preventDuplicateFormSubmissions.test.js
```diff
@@ -63,7 +63,7 @@ describe('Prevent duplicate form submissions', () => {
     // JSDOM doesn't implement form submissions
     // see https://github.com/jsdom/jsdom/issues/1937#issuecomment-321575590
     //
-    // If a form submission is caused by a input/button[type=submit], this instead outputs a
+    // If a form submission is caused by a input/button, this instead outputs a
```
tombye, line 66: This, and the other replacement in this file, would read better as:
```suggestion
    // If a form submission is caused by a submit input/button, this instead outputs a
```

Also: could we update the HTML fixture under `// set up DOM` too?

tests/javascripts/previewPane.test.js
```diff
@@ -137,7 +137,7 @@ describe('Preview pane', () => {
         // run preview pane script
         require('../../app/assets/javascripts/previewPane.js');
 
-        expect(document.querySelector('button[type=submit]').textContent).toEqual('Save');
+        expect(document.querySelector('form button').textContent).toEqual('Save');
```
tombye, line 140: Can we update the HTML fixture under `// set up DOM` so the button doesn't have the `type="submit"` any more?

tombye (changes requested):
Only a few, fairly petty requests for change. Overall looks good and I've tested across the browsers with these variants:
- buttons to submit forms sending multiple values (sign in page)
- buttons to submit forms sending single value (settings page)
- buttons to submit in-page search term (notifications page)
- buttons to upload file (add email branding page)

All looked good and worked well so, if the changes are done 👍🏻

leohemsted (commented):
These changes seem pretty reasonable.

what i wouldn't mind, however, is [after this story] us having some thought to figure out how we can avoid swathes of HTML in our javascript and also in our test suites. Be good if we can call govuk-frontend stuff from there and not have to worry when we bump govuk that we might need to update all these things again

app/templates/components/cookie-banner.html
```diff
@@ -5,10 +5,10 @@
     <h2 class="notify-cookie-banner__heading govuk-heading-m" id="notify-cookie-banner__heading">Can we store analytics cookies on your device?</h2>
     <p class="govuk-body">Analytics cookies help us understand how our website is being used.</p>
     <div class="notify-cookie-banner__buttons">
-      <button class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-accept" type="submit" data-accept-cookies="true" aria-describedby="notify-cookie-banner__heading">
+      <button type="button" class="govuk-button notify-cookie-banner__button notify-cookie-banner__button-accept" type="button" data-accept-cookies="true" aria-describedby="notify-cookie-banner__heading">
```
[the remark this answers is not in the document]
tombye: Should be resolved by https://github.com/alphagov/notifications-admin/pull/4435/commits/a3f6a0f280a6ec76b8568be39c2e5ef1de323c4c

tombye (approved):
Since my last review, we found a few things that needed updating. I'm now confident all the buttons are up-to-date and using the right macro so 👍🏻

samuelhwilliams (approved):
Looks good to me 👀 👍  Nice work. All the things I spotted going commit-by-commit got resolved in later commits.

PR #4436, opened 2022-11-07
Remove old ccs jinja library and remove ChainedUndefined too

leohemsted:
- [x] https://github.com/alphagov/notifications-admin/pull/4433
- [x] https://github.com/alphagov/notifications-admin/pull/4434
- [x] https://github.com/alphagov/notifications-admin/pull/4435
- [x] sort out a nice way round the undefined problem

best removed commit by commit

All of the tidy up is basically to get prepared for the biggest change here - we no longer override jinja2's `Undefined`.

Time for a quick segue about jinja's undefined behaviours. Imagine a jinja template where `foo` is not defined or passed in to the environment via render_template or flask context processors. What jinja2 returns when you access that variables is an `Undefined` class - that can be overriden by setting in the jinja_options `jinja_options['undefined'] = SomeOtherClass`

-----------------

#### jinja2.StrictUndefined
this is similar to how python works. i think every single page would probably break in some way if we enabled this, lets not bother
```jinja
{{ foo }} # throws an error
```
#### jinja2.Undefined
jinja2's default
```jinja
{{ foo }} # evaluates as empty string
{{ foo.bar }} # throws an error
```
#### jinja2.ChainedUndefined
The old CCS govuk_frontend_jinja library overrode `jinja2.undefined` with its own [custom class](https://github.com/Crown-Commercial-Service/govuk-frontend-jinja/blob/ddbe208a976ffa4ca330881c506c5200dfa69851/govuk_frontend_jinja/templates.py#L163) that is very similar to this one
```jinja
{{ foo }} # evaluates as empty string
{{ foo.bar }} # evaluates as empty string
{{ foo.bar['baz'] }} # evaluates as empty string
```
--------------

The old library did this to closer replicate how nunjucks handles things. The new jinja library is hand-rolled to specifically work with jinja2's default settings, so already gracefully checks variables being truthy before trying to access their nested children.

What i've done is a bunch of commits that firstly tidy up all these uses. A lot were dead code that was never used - if checks that used to evaluate to false but now started throwing not defined errors. 

The service_navigation commit is because that file was both used for `include` and `import` directives - have a read of the commit message for useful context around when jinja2 context is passed through to another file.

Then finally remove the references to the old CCS library and stop installing side by side! You'll need to remove your `vendor` folder after checking this out.


samuelhwilliams:
> jinja2.StrictUndefined
this is similar to how python works. i think every single page would probably break in some way if we enabled this, lets not bother

Do we/you have an idea how much work this would be? I do tend to err on the side of it would be nice to have confidence that we're not displaying blank strings where we would hope there's actual content.

Makefile
```diff
@@ -26,10 +26,6 @@ VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$V
 bootstrap: generate-version-file ## Set up everything to run the app
 	pip3 install -r requirements_for_test.txt
 
-	# TODO: once CCS govuk_frontend_jinja is removed, replace this with normal reqs.txt (and update PackageLoader)
-	mkdir -p vendor/govuk_frontend_jinja_macros
-	pip3 install govuk-frontend-jinja==1.5.1 --target ./vendor/govuk_frontend_jinja_macros
```
samuelhwilliams, line 31: 🎉

samuelhwilliams (approved):
This also looks good - and everything I spotted going commit-by-commit was resolved in later ones.

Nice clean work 👏

leohemsted:
@samuelhwilliams hah i briefly thought about `StrictUndefined`, but it's a nonstarter. Not only are our own templates not written that way, none of the govuk-frontend-jinja library is either.

StrictUndefined means you can't even do a truthy check, you need to say `{% if foo is not undefined %} {{ foo }} {% endif %}`. we'd need to pass in a huge swathe of variables. Some of these we could definitely make more consistent (`hide_from_search_engines` seen below is only set if it's `True`, and we could default it to `False` for example), but I'm not going through every page to see what's up.

This is all of the undefined variable access on the signed out index page (a helpful snippet for generating this list can be found here https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.make_logging_undefined)

```
'htmlClasses' is undefined
'pageTitleLang' is undefined
'flask.ctx._AppCtxGlobals object' has no attribute 'hide_from_search_engines'
'bodyClasses' is undefined
'bodyAttributes' is undefined
'dict object' has no attribute 'classes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'classes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'serviceName'
'dict object' has no attribute 'serviceName'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'html'
'dict object' has no attribute 'disabled'
'dict object' has no attribute 'isStartButton'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'name'
'dict object' has no attribute 'disabled'
'dict object' has no attribute 'preventDoubleClick'
'dict object' has no attribute 'html'
'iconHtml' is undefined
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'containerClasses'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'attributes'
'dict object' has no attribute 'text'
```

PR #4437, opened 2022-11-08
Tweak display of illustrations in branding radio labels

quis:
Before | After
---|---
image | image


quis:
I’ve updated this again to keep the two column grid (so the label text doesn’t wrap) but without the illustrations getting overly-large:

image

app/assets/stylesheets/components/radios-with-images.scss
```diff
@@ -7,19 +7,25 @@
   @include govuk-responsive-margin(4, "right");
 
   padding-left: 0;
+  cursor: pointer;
```
samuelhwilliams, line 10: I don't think we want this here, do we?

This style gets applied when the JS is loaded, and that's what actually makes it clickable. So with this if the CSS loaded and the JS failed, we'd show a pointer but clicking the image wouldn't do anything.

samuelhwilliams, line 10: Do we definitely want this?

The image is only clickable via JS support. If the CSS loads and the JS fails then people might click on the image but it won't select the input.

quis: Good point, removed in 700539e759db44c2f22105a5e4b2f90bd8627eec

app/assets/javascripts/radios-with-images.js
```diff
@@ -14,7 +14,6 @@
     this.start = function ($radioImage) {
 
       var radioImageNode = $radioImage.get(0);
-      console.log(radioImageNode);
```
samuelhwilliams, line 17: 👀 💥

samuelhwilliams (approved):
👍

PR #4438, opened 2022-11-09
Increase height of placeholder redaction mask

tombye:
The GDS Transport font, which we moved to recently, sits slightly lower on the line which seems to have resulted in the text we display for redacted placeholders poking out from below the bottom of the mask we cover it with.

This increases the height of the mask so it covers its text again.

Note: We also replace the text of any placeholders with 'hidden' so this issue will not have
disclosed any information set as redacted.

## Before these changes

redacted_content_fix_before

## With these changes

redacted_content_fix_after

### DIFF
--- a/app/assets/stylesheets/components/placeholder.scss
+++ b/app/assets/stylesheets/components/placeholder.scss
@@ -43,7 +43,7 @@
   background: currentColor;
   padding: 0 0.5em;
   opacity: 0.8;
-  box-shadow: inset 0 -0.35em 0 0 govuk-colour("white");
+  box-shadow: inset 0 -0.25em 0 0 govuk-colour("white");
   position: relative;
   top: 0.1em;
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/components/placeholder.scss
```diff
@@ -43,7 +43,7 @@
   background: currentColor;
   padding: 0 0.5em;
   opacity: 0.8;
-  box-shadow: inset 0 -0.35em 0 0 govuk-colour("white");
+  box-shadow: inset 0 -0.25em 0 0 govuk-colour("white");
```
quis, line 46: Adding a bit of inset shadow to the top edge helps keep the alignment with the text look balanced:
```suggestion
  box-shadow: inset 0 0.05em 0 0 govuk-colour("white"), inset 0 -0.25em 0 0 govuk-colour("white");
```

Before | After
---|---
image | image
image | image

quis (approved):

PR #4440, opened 2022-11-11
Minor content updates for platform admin branding content

klssmith:
This makes a few small changes to the platform admin content for both letter branding and email branding that were discussed with @karlchillmaid 

Also fixes two really minor HTML / stying things

---

## Use the word "Manage" on the organisation settings page instead of "Change"
We want to start using the term "Manage" when you have the option to change or manage more than one thing, for example email reply-to addresses. This changes the branding rows in the org settings table to fit the new pattern. Other parts of the settings table and app will be looked at later.

Settings page before             |  Settings page after
:-------------------------:|:-------------------------:
Screenshot 2022-11-11 at 08 52 20  |  Screenshot 2022-11-11 at 08 52 35

## Change button text when adding to branding pools
This updates the button content on the pages where a platform admin user can new email or letter brands to a branding pool from "Add options" to "Add selected options".

Button before             |  Button after
:-------------------------:|:-------------------------:
Screenshot 2022-11-11 at 08 55 38  |  Screenshot 2022-11-11 at 08 55 47

## Update the flash confirmation when removing a brand from the pool

Flash message before             |  Flash message after
:-------------------------:|:-------------------------:
Screenshot 2022-11-11 at 08 56 39  |  Screenshot 2022-11-11 at 08 57 17

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -410,14 +410,14 @@ def _handle_remove_branding(remove_branding_id) -> Optional[Response]:
     else:
         confirmation_question = Markup(
             render_template(
-                "partials/flash_messages/email_branding_confirm_remove_brand.html",
+                "partials/flash_messages/branding_confirm_remove_brand.html",
                 branding_name=remove_branding.name,
             )
         )
 
         flash(
             confirmation_question,
-            "delete",
+            "remove",
         )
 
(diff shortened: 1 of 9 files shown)


samuelhwilliams (approved):
👍

PR #4444, opened 2022-11-11
Refactor form widgets

leohemsted:
Inspired by @CrystalPea asking why we had the same comment 10 times throughout forms.py on the widget function, I took the opportunity to move the govuk frontend component logic out to its own base class, and "simply" use that base class from all the various field options.

The new base class has a `govuk_frontend_component_name` variable which needs to be defined. 

It then has a `prepare_params` function which should be overriden to return a params dictionary. This params dict should contain the `errorMessage` key if relevant, however you can use `GovukFrontendField.get_error_message` to help you build that errorMessage up. Those params are then combined with `param_extensions`. Then the component is rendered as normal.

No actual functionality or implementation has changed here. (Perhaps some fields are marginally more flexible now, if they now handle param_extensions where they didn't previously etc).

This is really really definitely best reviewed commit by commit.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -81,7 +81,10 @@ from app.models.branding import (
 from app.models.feedback import PROBLEM_TICKET_TYPE, QUESTION_TICKET_TYPE
 from app.models.organisation import Organisation
 from app.utils import branding, merge_jsonlike
-from app.utils.govuk_frontend_field import render_govuk_frontend_macro
+from app.utils.govuk_frontend_field import (
+    GovukFrontendWidgetMixin,
+    render_govuk_frontend_macro,
+)
 from app.utils.user import distinct_email_addresses
 from app.utils.user_permissions import (
     all_ui_permissions,
@@ -522,7 +525,7 @@ class NestedFieldMixin:
 
             params["items"].append(item)
 
-        return render_govuk_frontend_macro("checkbox", params=params)
+        return render_govuk_frontend_macro(self.govuk_frontend_component_name, params=params)
 
 
 class NestedRadioField(RadioFieldWithNoneOption, NestedFieldMixin):
@@ -789,26 +792,24 @@ class GovukCheckboxField(BooleanField):
         return govuk_checkbox_field_widget(self, field, param_extensions=param_extensions, **kwargs)
 
 
-class GovukTextareaField(TextAreaField):
+class GovukTextareaField(GovukFrontendWidgetMixin, TextAreaField):
+    govuk_frontend_component_name = "textarea"
+
     def __init__(self, label="", validators=None, param_extensions=None, **kwargs):
         super(TextAreaField, self).__init__(label, validators, **kwargs)
         self.param_extensions = param_extensions
 
-    # self.__call__ renders the HTML for the field by:
-    # 1. delegating to self.meta.render_field which
-    # 2. calls field.widget
-    # this bypasses that by making self.widget a method with the same interface as widget.__call__
-    def widget(self, field, param_extensions=None, **kwargs):
+    def prepare_params(self, param_extensions=None, **kwargs):
         # error messages
         error_message = None
-        if field.errors:
-            error_message = {"text": field.errors[0]}
+        if self.errors:
+            error_message = {"text": self.errors[0]}
 
         params = {
-            "name": field.name,
-            "id": field.id,
+            "name": self.name,
+            "id": self.id,
             "rows": 8,
-            "label": {"text": field.label.text, "classes": None, "isPageHeading": False},
+            "label": {"text": self.label.text, "classes": None, "isPageHeading": False},
             "hint": {"text": None},
             "errorMessage": error_message,
         }
@@ -821,7 +822,7 @@ class GovukTextareaField(TextAreaField):
         if param_extensions:
             merge_jsonlike(params, param_extensions)
 
-        return render_govuk_frontend_macro("textarea", params=params)
+        return params
 
 
 # based on work done by @richardjpope: https://github.com/richardjpope/recourse/blob/master/recourse/forms.py#L6
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -683,62 +683,35 @@ def govuk_checkbox_field_widget(self, field, param_extensions=None, **kwargs):
     return render_govuk_frontend_macro("checkbox", params=params)
 
 
-def govuk_checkboxes_field_widget(self, field, wrap_in_collapsible=False, param_extensions=None, **kwargs):
-    def _wrap_in_collapsible(field_label, checkboxes_string):
-        # wrap the checkboxes HTML in the HTML needed by the collapisble JS
-        result = Markup(
-            f'<div class="selection-wrapper"'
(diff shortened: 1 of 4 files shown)


app/utils/govuk_frontend_field.py
```diff
@@ -1,6 +1,36 @@
+from abc import ABC, abstractmethod
+
 from flask import Markup, render_template_string
 
 
+class GovukFrontendField(ABC):
```
samuelhwilliams, line 9: Maybe we should call this something other than field - it's just about rendering the widget isn't it? `AbstractGovukFrontendWidgetMixin` might be clearer. Bit of a mouthful though... This got confusing/looked a bit odd later in the process when we ended up with multiple inheritence for class GovukCheckboxesField(GovukFrontendField, SelectMultipleField).

app/main/forms.py
```diff
@@ -223,131 +242,52 @@ def password(label="Password"):
...
-
-    # self.__call__ renders the HTML for the field by:
-    # 1. delegating to self.meta.render_field which
-    # 2. calls field.widget
-    # this bypasses that by making self.widget a method with the same interface as widget.__call__
-    def widget(self, field, param_extensions=None, **kwargs):
-
-        params = {"attributes": {"spellcheck": "false"}}  # email addresses don't need to be spellchecked
-        merge_jsonlike(params, param_extensions)
+        merge_jsonlike(self.param_extensions, param_extensions)
```
samuelhwilliams: why do we do this here but not in the other inits above?

Also it's kinda gross that this function mutates the first parameter rather than returning a new thing and assigning explicitly, doesn't feel nice (but know it's not new)

samuelhwilliams: oh duh we do it because we have some class-level param_extensions

samuelhwilliams: it's maybe a bit weird that the param_extensions passed into the `__init__` call can't override the class-level param_extensions. Sort of fine for this class as doubt we'd want to enable spellcheck, but on the one below where we're setting width we can't override it.

Although we could override it when calling the field in the template (and that place makes more sense to override it anyway) - it's a weird inconsistency.

app/main/forms.py
```diff
@@ -223,131 +242,52 @@ def password(label="Password"):
...
-    # self.__call__ renders the HTML for the field by:
-    # 1. delegating to self.meta.render_field which
-    # 2. calls field.widget
-    # this bypasses that by making self.widget a method with the same interface as widget.__call__
-    def widget(self, field, param_extensions=None, **kwargs):
-        return govuk_text_input_field_widget(self, field, param_extensions=param_extensions, **kwargs)
-
 
 class SMSCode(GovukTextInputField):
+    input_type = "tel"
```
samuelhwilliams, line 257: shouldn't this just be a number input type? they're not putting in a telephone number, it's a 2fa code right?

leohemsted: `number` is specifically recommended against cos of how browsers render it (it adds a little up/down arrow to increment/decrement which is obviously useless for our case). https://design-system.service.gov.uk/components/text-input/

"tel" means it pops up a phone keypad on mobile, which is better than a full qwerty keyboard for entering a numerical code

app/main/forms.py
leohemsted: I know I refactored all this out to the top level class anyway, but just to clarify, there are three places you can define `param_extensions`:

#### class level
```python
Class MyField(GovukFrontendField):
    param_extensions = {"field": "value"}
```
#### instantiation level
```python
Class MyForm(Form):
    my_field = MyField(param_extensions={"field": "value"}
```
#### render level
```jinja2
{% call form_wrapper() %}
      {{ form.name(param_extensions={"field": "value"}) }}
{% endcall %}
```

Values from all three of these are merged gracefully using the merge_jsonlike function. The priority order is render overrides instantiation which overrides class level variables

leohemsted: Also I share your views about the function being gross, I did briefly look at moving it to return a new deep copy of the merged dict but it's a bit complex to reason about (it does some wild stuff like appending lists to each other if their items differ but if the items are the same then it doesn't duplicate them, stuff like that? it's well tested but i didnt quite understand the exact list of requirements, and didn't want to touch that cos it looked pretty complex.

samuelhwilliams: > The priority order is render overrides instantiation which overrides class level variables

This is the line I'm challenging. That's what it _should_ be - but in the cases where you're using merge_jsonlike in the field `__init__`, the attributes specified in the class `param_extensions` aren't overriden by instantiation params.

app/utils/govuk_frontend_field.py
leohemsted: i think that something with Mixin is reasonable - it implies that you _need_ to specify a base Field type as well to have something that handles basic wtforms functionality.

i'd probs lose the `abstract` as that's not useful - you'll find out based on looking at existing examples that you need to supply a component name. maybe just `GovukFrontendWidgetMixin` then

app/main/forms.py
leohemsted: can you give an example? this line you've highlighted is showing `self.param_extensions` (the class params) being overriden by `param_extensions` (the parameter to `__init__`), as intended.

however, i removed all the param extensions from all subclasses in https://github.com/alphagov/notifications-admin/pull/4444/commits/aba1c1b705809648400c0a424ef22ce6d191edf2 - there's only two places calling `merge_jsonlike` now, `SMSCode.__call__` and `GovukFrontendField.__init__`.

app/main/forms.py
leohemsted: probs worth a comment to remove confusino

tombye (commented):
Wouldn't say I have as deep an understanding of the code structure to comment on it but I found it readable at least (probs leave that to @samuelhwilliams' review). Added a few comments though and, though it doesn't break anything yet, I do think we should use `self.govuk_frontend_component_name` instead of `checkbox` in `NestedFieldMixin` here: https://github.com/alphagov/notifications-admin/blob/5890c296f87281a5a0304f978bd88a360693c3e0/app/main/forms.py#L462

That could sit outside of this PR though.

app/main/forms.py
leohemsted: SMSCode.__call__ doesn't need to exist anyway, i'll remove that 🚮

samuelhwilliams: 💀 yep yep yep, you're right, it's fine by the end of the PR. Reviewing commit-by-commit has been a bit painful because lots of things in earlier commits get overwritten/fixed by later ones.

app/utils/govuk_frontend_field.py
```diff
@@ -1,6 +1,36 @@
...
+        """
+        pass
+
+    def prepare_params(self, **kwargs):
+        """
+        Should return a dictionary that will be passed through to the macro as `params` for use within the component
+        """
+        return {}
+
+    # self.__call__ renders the HTML for the field by:
```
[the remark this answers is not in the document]
leohemsted: Yeah, good call. This comment being confusing was what spurred this whole PR in the first place. fixed.

tombye (approved):
All comments have been addressed so 👍🏻

PR #4445, opened 2022-11-11
Generate param_extensions in init

samuelhwilliams:
We are editing a dictionary in the __init__ method for EmailBrandingChooseLogoForm, which is stored on the class. This can lead to data being mutated between requests on the same worker.

Let's build a whole new param_extensions dict every time we instantiate the form, rather than injecting into an existing dict.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2218,17 +2218,17 @@ class EmailBrandingChooseLogoForm(StripWhitespaceForm):
         "Choose a logo for your emails",
         choices=tuple((key, value["label"]) for key, value in BRANDING_OPTIONS_DATA.items()),
         image_data={key: value["image"] for key, value in BRANDING_OPTIONS_DATA.items()},
-        param_extensions={
-            "classes": "govuk-radios--inline",
-            "fieldset": {"legend": {"classes": "govuk-fieldset__legend--l", "isPageHeading": True}},
-        },
     )
 
     def __init__(self, service, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.branding_options.param_extensions = {
+            "classes": "govuk-radios--inline",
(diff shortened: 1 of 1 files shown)


app/main/forms.py
```diff
@@ -2218,17 +2218,17 @@ class EmailBrandingChooseLogoForm(StripWhitespaceForm):
...
     def __init__(self, service, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.branding_options.param_extensions = {
+            "classes": "govuk-radios--inline",
+            "fieldset": {"legend": {"classes": "govuk-fieldset__legend--l", "isPageHeading": True}},
+        }
         if not service.email_branding_id:
             self.branding_options.param_extensions["hint"] = {
-                "html": (f"{service.name} branding is not set up yet."),
+                "html": f"{service.name} branding is not set up yet.",
```
leohemsted, line 2231: I think this should be passed in from the template, when we render. Then it goes straight into the `widget` call, and never gets assigned to `some_field.param_extensions`
```
{{
  form.branding_options(
    param_extensions={'hint': {'html': current_service.name + branding is not set up yet."}}
  )
}}
```
https://github.com/alphagov/notifications-admin/blob/master/app/templates/views/service-settings/branding/add-new-branding/email-branding-choose-logo.html#L33

leohemsted (approved):
i think this is fine for now to go in. I've left what I think our ultimate fix should be in the comments, but it's 4:45 on a friday so lets not stay late.

also, while we still have param_extensions everywhere I might look into if we can override `bind` to do a deep copy when binding the field to a form

app/main/forms.py
samuelhwilliams: That's a good idea worth pursuing 👍

PR #4446, opened 2022-11-13
Bump Flask/Flask-Login/Werkzeug

samuelhwilliams:
## Summary
Moves us away from deprecated Flask/Flask-Login implementation details onto the new supported interface (mostly `g` for Flask), so that we can then update our Flask/Flask-Login/Werkzeug packages to newer versions.

Good to review commit-by-commit.

Updates most of the tests to use the `client.login` function rather than manually patching `get_user`. The `login` function patches this as needed by default, as well as patching a number of other functions as needed so let's just use `client.login` to be consistent and centralise that logic from now on.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,5 @@
 import os
 import pathlib
-from functools import partial
 from time import monotonic
 
 import jinja2
@@ -15,7 +14,6 @@ from flask import (
     session,
     url_for,
 )
-from flask.globals import _lookup_req_object, _request_ctx_stack
 from flask_login import LoginManager, current_user
 from flask_wtf import CSRFProtect
 from flask_wtf.csrf import CSRFError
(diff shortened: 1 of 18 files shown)


samuelhwilliams:
Deployed to preview manually and ran all of the functional tests, which have also passed (https://concourse.notify.tools/teams/notify/pipelines/apps/jobs/functional-tests-preview/builds/367).

app/__init__.py
```diff
@@ -289,49 +282,47 @@ def make_session_permanent():
...
-            service_id = session.get("service_id")
+    if request.view_args:
+        service_id = request.view_args.get("service_id", session.get("service_id"))
+    else:
+        service_id = session.get("service_id")
 
-        if service_id:
+    if service_id:
+        try:
+            g.current_service = Service(service_api_client.get_service(service_id)["data"])
```
quis, line 297: I know you’ve copied this from the existing code, but would be nicer as this (and matches `Organisation.from_id(org_id)` further down): 
```suggestion
            g.current_service = Service.from_id(service_id)
```

PR #4447, opened 2022-11-14
Allow a letter branding in the pool for an organisation to be made the default

klssmith:
This change allows one of the letter brandings in the pool for an organisation (or no branding) to be set as the default.

If the organisation doesn't originally have no branding as the default, we show a "Use no branding as default instead" link under the current default. Each non-default brand has a "Make default" button under it. For consistency, this is all very similar to the code for email branding except that the no branding option replaces the GOVUK option.

We show a confirmation message if changing the default brand to or from no branding, but no message if you are changing the brand from one set brand to another.

## Page with no branding as the default branding
Screenshot 2022-11-14 at 15 54 43

## Page with a set branding as the default
Screenshot 2022-11-14 at 15 53 55

## Confirmation message before changing from no branding to a brand
Screenshot 2022-11-14 at 15 55 03

## Confirmation message before changing from a set brand to no branding
Screenshot 2022-11-14 at 15 54 28

### DIFF
--- a/app/templates/flash_messages.html
+++ b/app/templates/flash_messages.html
@@ -2,11 +2,19 @@
 {% with messages = get_flashed_messages(with_categories=true) %}
   {% if messages %}
     {% for category, message in messages %}
-      {% if category in ['cancel', 'delete', 'suspend', 'resume', 'remove', 'revoke this API key', 'stop broadcasting'] %}
+      {% if category in [
+        'cancel',
+        'delete',
+        'suspend',
+        'resume',
+        'remove',
+        'revoke this API key',
+        'stop broadcasting',
+        'make this email branding the default',
+        'make this letter branding the default',
+      ] %}
         {% set delete_button_text = "Yes, {}".format(category) %}
       {% elif category == 'try again' %}
-        {% set delete_button_text = category|capitalize %}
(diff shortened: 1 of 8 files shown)


app/templates/flash_messages.html
```diff
@@ -2,12 +2,20 @@
...
+        'revoke this API key',
+        'stop broadcasting',
+        'make this email branding the default',
+        'make this letter branding the default',
+      ] %}
         {% set delete_button_text = "Yes, {}".format(category) %}
       {% elif category == 'try again' %}
         {% set delete_button_text = category|capitalize %}
-      {% elif category == 'yes, make this email branding the default' %}
-        {% set delete_button_text = category|capitalize %}
```
samuelhwilliams, line 10: Ah, good tidy up 👍

samuelhwilliams (approved):
Good work on spotting and updating to use `organisation.update` instead of the client directly.

We've copy/pasted quite a bit of code from the email branding view handlers, but I think you're right to do that here as trying to generalise that would probably become overly complex. So think we're fine to take the slight duplication for now 👍 

Test coverage looks good too.

Have run through the flows locally and it all looks good, nice stuff.

PR #4448, opened 2022-11-15
Fix redaction box not covering all of hidden text

idavidmcdonald:
I suspect this broke in the recent font change but our current redaction box is not fully covering the word 'hidden' and the bottom part of it sticks out slightly. Whilst it does not expose anything sensitive, it is not quite right.

I've manually chosen a value for the CSS which does cover all the text.

I've also checked and this is not an issue for email templates.

The only bit I'm not sure on is whether a similar change from `0.35em` to `0.25em` is required for the bit of CSS below for `*:focus + p &`. I couldn't figure out what it was for but was interested as the value of 0.35em is the same so could be related.

## Before
image

## After
image

## After on a multiline message
image

## After with multiple placeholders
image

## After on small screen size
image

## Proof that email was not affected by this issue
image

### DIFF
--- a/app/assets/stylesheets/components/placeholder.scss
+++ b/app/assets/stylesheets/components/placeholder.scss
@@ -48,7 +48,7 @@
   top: 0.1em;
 
   .sms-message-wrapper & {
-    box-shadow: inset 0 -0.35em 0 0 govuk-shade(govuk-colour("light-grey"), 7%);
+    box-shadow: inset 0 0.05em 0 0 govuk-shade(govuk-colour("light-grey"), 7%), inset 0 -0.25em 0 0 govuk-shade(govuk-colour("light-grey"), 7%);
   }
 
   *:focus + p & {


app/assets/stylesheets/components/placeholder.scss
```diff
@@ -48,7 +48,7 @@
   top: 0.1em;
 
   .sms-message-wrapper & {
-    box-shadow: inset 0 -0.35em 0 0 govuk-shade(govuk-colour("light-grey"), 7%);
+    box-shadow: inset 0 -0.25em 0 0 govuk-shade(govuk-colour("light-grey"), 7%);
```
quis, line 51: This same problem was fixed for emails in https://github.com/alphagov/notifications-admin/pull/4438

I didn’t think to check if there was different styling for hidden placeholders in text messages. Nice work screenshotting all the things.

This very slight tweak will make it match emails exactly, where we added a second inner shadow to get the vertical alignment of the black box with with the text looking balanced:
```suggestion
    box-shadow: inset 0 0.05em 0 0 govuk-shade(govuk-colour("light-grey"), 7%), inset 0 -0.25em 0 0 govuk-shade(govuk-colour("light-grey"), 7%);
```

quis (approved):

PR #4452, opened 2022-11-16
Show when a webauthn key was last used

leohemsted:
This was inspired by me having three keys all called "Unknown key" on prod. I still don't know what one of them is 😱 

When a user logs in, work out which key they used to log in, and pass through that key's id to the API so it can set a `logged_in_at` timestamp against that key.

Then, on the security keys page we can show when each key was last used to log in.

image

-------------

- [x] https://github.com/alphagov/notifications-admin/pull/4451
- [x] https://github.com/alphagov/notifications-api/pull/3638

### DIFF
--- a/app/main/views/webauthn_credentials.py
+++ b/app/main/views/webauthn_credentials.py
@@ -99,8 +99,8 @@ def webauthn_complete_authentication():
     user_id = session["user_details"]["id"]
     user_to_login = User.from_id(user_id)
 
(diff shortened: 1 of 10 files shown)


samuelhwilliams (approved):
Nice 👍  Tested locally, all looks good

quis:
Added a commit with some small visual tweaks:
image

PR #4453, opened 2022-11-17
Bump govuk frontend to v3.14.0

tombye:
Bumps GOVUK Frontend to v3.14.0 so it has parity with the version of GOVUK Frontend Jinja Macros, our templating library, we currently use[^1].

Part of https://trello.com/c/n3OI2lt6/39-bump-govuk-frontend-from-v3-to-latest-version-v44

## Notes for reviewers

Bumping to v3.14.0 doesn't break anything but does bring in some features and remove the need for some hacks we wrote for GOVUK Frontend features we wanted ahead of time.

If you want to understand all the changes, go commit by commit but it's also ok to just check in your browser.

Worth saying v3.12.0 introduced changes to the way links look but the work to include this is complex enough that I've put it in a separate story: https://trello.com/c/jDtuwHih/116-change-links-in-admin-app-to-new-design-system-style so isn't in these changes.

[^1]: We use v1.5.1 of GOVUK Frontend Jinja Macros which has parity with v3.14.0 of GOVUK Frontend, as defined in [this comparison table of versions](https://github.com/LandRegistry/govuk-frontend-jinja#compatibility).

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -238,7 +238,7 @@ class GovukPasswordField(GovukTextInputFieldMixin, PasswordField):
 
 class GovukEmailField(GovukTextInputFieldMixin, EmailField):
     input_type = "email"
-    param_extensions = {"attributes": {"spellcheck": "false"}}  # email addresses don't need to be spellchecked
+    param_extensions = {"spellcheck": False}  # email addresses don't need to be spellchecked
 
 
 class GovukSearchField(GovukTextInputFieldMixin, SearchField):
--- a/app/assets/stylesheets/components/banner.scss
+++ b/app/assets/stylesheets/components/banner.scss
@@ -3,16 +3,14 @@
 .banner-default {
 
   @include govuk-font(19, $weight: bold);
-  // TODO: replace with `$govuk-success-colour` when GOV.UK Frontend version >= 3.10.0
-  color: govuk-colour("green");
+  color: $govuk-success-colour;
   display: block;
   padding: govuk-spacing(3);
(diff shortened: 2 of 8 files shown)


app/main/forms.py
```diff
@@ -238,7 +238,7 @@ class GovukPasswordField(GovukTextInputFieldMixin, PasswordField):
 
 class GovukEmailField(GovukTextInputFieldMixin, EmailField):
     input_type = "email"
-    param_extensions = {"attributes": {"spellcheck": "false"}}  # email addresses don't need to be spellchecked
+    param_extensions = {"spellcheck": False}  # email addresses don't need to be spellchecked
```
leohemsted, line 241: nice. of course cos we were using govuk_frontend_jinja==1.5.1, we already had that code change in so could have done this already. But good to do it now. 👍

leohemsted (approved):
Looks good, lots of sensible code changes and getting rid of lots of TODOs 👍

samuelhwilliams:
Hopefully you'll also need to update the test that checks govuk-frontend and govuk-frontend-jinja versions 🤞

tombye:
> be good to pin it now to 3.14.x of govuk-frontend rather than 3.0.0 <= x < 4.0.0

@samuelhwilliams had a go this in https://github.com/alphagov/notifications-admin/pull/4453/commits/ee43b851ba14d185ca471b0b46dbd39bd1fb961b, is that what you were thinking of?

samuelhwilliams:
@tombye test update looks good 👌

PR #4457, opened 2022-11-18
Clear cache when updating an organisation

klssmith:
This clears the cache when updating an organisation for two scenarios that were missed previously:

**Clear services cache when updating org's crown status**
Updating an organisation's crown status also changes the crown status [for all of its services](https://github.com/alphagov/notifications-api/blob/b21868aa0c82147e3b94928a37733de5efd24e26/app/dao/organisation_dao.py#L116-L117).

This means we should be clearing the services cache too, which we hadn't been doing.

**Clear branding pool caches if org type is set to NHS**
If we update an organisation's type in api to be an NHS org, we also add the NHS branding to the email and letter branding pools. This means that we should clear the pool caches in this case.

### DIFF
--- a/app/main/views/organisations.py
+++ b/app/main/views/organisations.py
@@ -342,14 +342,13 @@ def edit_organisation_crown_status(org_id):
     )
 
     if form.validate_on_submit():
-        organisations_client.update_organisation(
-            current_organisation.id,
-            crown={
-                "crown": True,
-                "non-crown": False,
-                "unknown": None,
-            }.get(form.crown_status.data),
-        )
+        crown_data = {
+            "crown": True,
+            "non-crown": False,
+            "unknown": None,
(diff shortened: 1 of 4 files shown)


app/main/views/organisations.py
```diff
@@ -342,14 +342,13 @@ def edit_organisation_crown_status(org_id):
...
-                "unknown": None,
-            }.get(form.crown_status.data),
-        )
+        crown_data = {
+            "crown": True,
+            "non-crown": False,
+            "unknown": None,
+        }.get(form.crown_status.data)
+
+        current_organisation.update(crown=crown_data, delete_services_cache=True)
```
leohemsted, line 351: looks nicer to use the current_org abstraction too 👍

leohemsted (approved):

PR #4456, opened 2022-11-17
More form refactors

leohemsted:
A handful of the last refactors i've had on my mind for forms.py. Cleans up a couple of the mutability issues we've had lately, and a little bit of DRYing.

i'll stop refactoring now or i'll be here all month

PR #4458, opened 2022-11-18
Email branding choose alt text

leohemsted:
- [ ] https://github.com/alphagov/notifications-admin/pull/4456
- [ ] https://github.com/alphagov/notifications-api/pull/3644

Just two commits in this branch vs the refactor branch - adding alt text (and adding it to the platform admin page).
615ef21684a1c6cd06949ecd7c93991c91432ffc
36f4ba9318cc83ef97a9878b86c6faf782150bb3

For now, it defaults to null. Also, it now sends `None` across rather than empty string for name, logo etc if they're not specified (needs API changes to be merged or that'll break stuff)


app/main/views/email_branding.py
```diff
@@ -74,10 +74,13 @@ def update_email_branding(branding_id, logo=None):
         updated_logo_name = permanent_email_logo_name(logo, session["user_id"]) if logo else None
 
         try:
+            alt_text = form.name.data
```
klssmith, line 77: Do you reckon it would it be worth setting `alt_text` to be `form.text.data or form.name.data`? We don't use the name for the alt text if there is text, so this would mean that we'd store the value that will be used.

klssmith (approved):

PR #4463, opened 2022-11-21
Email branding preview

samuelhwilliams:
## Summary

* removes the separate 'name logo' page, which has been integrated into the 'preview branding' page (this one).
* updates the `/_email` endpoint to allow specifying brand/logo/banner attributes through query params, rather than requiring an email brand in the DB.
* hooks up all of the email branding flow pages to pass their form data through URL query params
* finally, adds the "preview email branding" page and renders a sample email with the appropriate brand

## Caveats
A future PR by @leohemsted will add the 'name/alt text your logo' input

## Examples
image

image

image


app/main/views/index.py
```diff
@@ -66,35 +66,40 @@ def design_content():
...
+    brand_type = "govuk"
     branding_style = request.args.get("branding_style", None)
+    brand_name = None
+    brand_text = None
+    brand_colour = None
+    brand_logo = None
 
     if branding_style == FieldWithNoneOption.NONE_OPTION_VALUE:
-        branding_style = None
+        pass
```
leohemsted, line 77: this `pass` feels like a smell to me. there's also a fair bit of duplication in this fn. how would you feel about something like this - is there some weird gotcha i've missed (there might well be)?

```
branding_style = request.args.get('branding_style', 'govuk')
if branding_style == 'custom':
  source = request.args
elif branding_style == 'govuk':
  source = {"brand_type": "govuk"}
else:
  source = email_branding_client.get(...)

brand_type = source.get('type')
[...]

return make_response(str(HTMLEmailTemplate(...)))
```

i think we can probably get away with only calling the make_response once rather than having that inside an if statement based on request.args. if request.args is empty then we'll handle that anyway in the existing code too as far as i can tell.

leohemsted: other thought: this may be a great candidate for splitting out into multiple routes that set up the source dict and call in to one shared sub function that sets up the template and returns the rendered template

```
/_email/govuk
/_email/custom?type='org_banner'&....
/_email/<uuid:email_branding_id>
```
then we can have `/_email` redirect to `/_email/govuk` to keep it really clear what's happening

app/main/views/service_settings.py
```diff
@@ -1431,26 +1432,45 @@ def email_branding_upload_logo(service_id):
...
-        logo_object["ResponseMetadata"]["HTTPHeaders"]["content-type"]
-        + f";base64,{base64.b64encode(logo_object['Body'].read()).decode('utf-8')}"
+@main.route(
+    "/services/<uuid:service_id>/service-settings/email-branding/when-you-use-this-branding", methods=["GET", "POST"]
+)
+@user_has_permissions("manage_service")
+def email_branding_confirm_upload_logo(service_id):
+    email_branding_data = _email_branding_flow_query_params(request)
+    if "type" not in email_branding_data or "logo" not in email_branding_data:
+        abort(400)
```
leohemsted, line 1442: this isn't a great flow. this bounces you to a "Something went wrong" page with no way to get back to where you were.

If there's no logo i think it should redirect to `main.email_branding_choose_logo`
if there's no type i think it should redirect to `main.email_branding_choose_banner_type`

app/main/views/service_settings.py
```diff
@@ -1431,26 +1432,45 @@ def email_branding_upload_logo(service_id):
...
+        "views/service-settings/branding/add-new-branding/confirm.html",
+        back_link=url_for(
+            ".email_branding_upload_logo",
+            service_id=service_id,
+            **_email_branding_flow_query_params(request, logo=None),
+        ),
+        email_branding_data=email_branding_data,
     )
 
-    return render_template("views/service-settings/branding/add-new-branding/name-logo.html", logo_data=logo_data)
```
leohemsted, line 1447: we should delete the html template too

app/main/views/index.py
```diff
@@ -66,35 +66,40 @@ def design_content():
...
+
+    elif branding_style == "custom":
+        brand_type = request.args["type"]
+        brand_text = request.args.get("text")
+        brand_colour = request.args.get("colour")
+        brand_logo = (
+            f"https://{current_app.config['LOGO_CDN_DOMAIN']}/{request.args.get('logo')}"
+            if request.args.get("logo")
+            else None
+        )
```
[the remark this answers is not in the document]
samuelhwilliams: I'd rather remove `brand_name` from this altogether, as far as I can tell it's not ever used/rendered?

app/main/views/service_settings.py
```diff
@@ -1436,17 +1438,46 @@ def email_branding_confirm_upload_logo(service_id):
     pass
 
 
+def _email_branding_flow_query_params(request, **kwargs):
```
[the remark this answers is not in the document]
samuelhwilliams: We definitely need a way to drop keys (not just have them with `None` values) - or they'll still get rendered into query params with eg `type=None` (or we have to filter them out somewhere else)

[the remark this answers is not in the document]
samuelhwilliams: Updated in ad17a5357056eb888fe80e1591aa9b137476d9be

app/main/views/service_settings.py
samuelhwilliams: Done in [6c1ee58](https://github.com/alphagov/notifications-admin/pull/4463/commits/6c1ee583d4231b3cede697b984470e61fdbf00d5)

app/main/views/service_settings.py
```diff
@@ -1412,15 +1412,17 @@ def email_branding_upload_logo(service_id):
...
+                **_email_branding_flow_query_params(request, logo=logo_filename),
             )
         )
 
-    previous_page = request.args.get("previous_page", "add-banner").lower()
-    if previous_page == "choose-colour":
-        back_link = url_for(".email_branding_choose_banner_colour", service_id=service_id)
+    email_branding_data = _email_branding_flow_query_params(request)
+    if email_branding_data.get("type") == "org_banner":
+        email_branding_data.pop("colour", None)  # Remove 'colour' if present
```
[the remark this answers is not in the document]
samuelhwilliams: Done in e3c2695fc96589392989f43f8e148024ed807f1c

app/main/views/index.py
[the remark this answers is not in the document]
samuelhwilliams: Done in ab07269e54fdb87320e6c781f36d350b1016a4ff

app/main/views/index.py
samuelhwilliams: Nice idea with the `source` refactoring - that's done in b2d85170558a5c2a7558f53c998d90ca8a3b0b62.

Going to leave the multi-route thing as that could also spill over into `/_letters` and I don't want to tackle it now. Doesn't feel super necessary.

app/main/views/index.py
```diff
@@ -66,35 +66,25 @@ def design_content():
...
-            else None
-        )
-        govuk_banner = branding_type in ["govuk", "both"]
-        brand_banner = branding_type == "org_banner"
-        brand_name = email_branding["name"]
+    elif branding_style == "custom":
+        source = request.args
+
+    elif branding_style is not None:
+        source = email_branding_client.get_email_branding(branding_style)["email_branding"]
```
leohemsted, line 78: ah sorry, one small thing I've noticed. if `branding_style` is passed in, but is not any of `__NONE__`, `govuk`, or `{an existing branding UUID}`, then we get a 404 error that i think is thrown by this line as it tries to fetch the email branding but doesn't handle 404 responses.

This test I wrote quickly has two failing types
```python
@pytest.mark.parametrize(
    "params",
    [
        {},
        {"branding_style": None}, 
        {"branding_style": "govuk"},
        {"branding_style": "__NONE__"}, 

        {"branding_style": ""},  # failing, can be reached at `/_email?branding_style=`
        {"branding_style": "1"},  # failing but worryingly used in the test above
    ],
)
def test_displays_govuk_branding(client_request, params):

    page = client_request.get("main.email_template", **params, _test_page_title=False)

    assert page.select_one("a")["href"] == "https://www.gov.uk"
```
i wonder if its best to match UUIDs based on a regex or other casting to a UUID inside a try catch or something, and only calling the API if we know it's a UUID. the likelyhood of us calling with a UUID that doesnt match a branding in the DB so 404s is low enough that I think that's okay.

leohemsted (commented):
one more comment on the _email thing. thanks for your perseverence!

app/main/views/service_settings.py
```diff
@@ -1452,14 +1452,22 @@ def email_branding_confirm_upload_logo(service_id):
         # Can't create the branding yet - need to extend this page to accept and handle the logo name/alt text
         pass
 
+    # Translate new brand form data into query params expected by the `/_email` endpoint
```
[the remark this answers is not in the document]
samuelhwilliams: I don't necessarily agree that we want to implicitly tie those two interfaces together but let's discuss separately if needed.

app/main/views/index.py
samuelhwilliams: Updated in [7fcac5f](https://github.com/alphagov/notifications-admin/pull/4463/commits/7fcac5fd0a9f5f3afc9d5bfb3f5df2a9cdf77ffa)

leohemsted (approved):
good stuff - don't forget to squash!

PR #4464, opened 2022-11-23
Remove the reply-to address from request to go live Zendesk tickets

klssmith:


samuelhwilliams (approved):

app/templates/support-tickets/go-live-request.txt
```diff
@@ -27,8 +27,6 @@ Letters in next year: {{ service.volume_letter|format_thousands }}
 Consent to research: {{ service.consent_to_research|format_yes_no }}
 Other live services for that user: {{ user.live_services|format_yes_no }}
 
-Service reply-to address: {{ service.default_email_reply_to_address or "not set" }}
```
idavidmcdonald, line 30: Is it possible to get a request to go live where the email reply to address has not been set? ie. is it still value to be told if there is no reply to address set?

I'm wondering whether we should say if no email reply to address is set if that is the case, and then otherwise show nothing?

klssmith: We make you enter a reply-to address if you enter anything other than 0 to the "How many emails do you expect to send in the next year?" question. So you can go live without one, but only if you've said you won't send emails.

idavidmcdonald: Ah, that seems like a reasonable protection, all good :) Thanks

idavidmcdonald (approved):

PR #4465, opened 2022-11-23
Bump GOVUK Frontend to 4.2.0

tombye:
Bump GOVUK Frontend to 4.2.0 (the latest version GOVUK Frontend Jinja Macros has parity with[^1])

Part of https://trello.com/c/n3OI2lt6/39-bump-govuk-frontend-from-v3-to-latest-version-v42

## Notes for reviewers

These changes are intended to be reviewed commit-by-commit.

[^1]: See [the table of compatibility table on GOVUK Frontend Jinja's repo](https://github.com/LandRegistry/govuk-frontend-jinja#compatibility).

### DIFF
--- a/app/assets/javascripts/esm/all.mjs
+++ b/app/assets/javascripts/esm/all.mjs
@@ -13,6 +13,7 @@ import Details from 'govuk-frontend/govuk/components/details/details';
 import Button from 'govuk-frontend/govuk/components/button/button';
 import Radios from 'govuk-frontend/govuk/components/radios/radios';
 import ErrorSummary from 'govuk-frontend/govuk/components/error-summary/error-summary';
+import SkipLink from 'govuk-frontend/govuk/components/skip-link/skip-link';
 
 // Modules from 3rd party vendors
 import morphdom from 'morphdom';
@@ -59,8 +60,13 @@ function initAll (options) {
     new Radios($radio).init()
   })
 
+  // There will only every be one error-summary per page
   var $errorSummary = scope.querySelector('[data-module="govuk-error-summary"]')
   new ErrorSummary($errorSummary).init()
+
+  // There will only ever be one skip-link per page
+  var skipLink = scope.querySelector('[data-module="govuk-skip-link"]')
+  new SkipLink(skipLink).init()
 }
 
 // Create separate namespace for GOVUK Frontend.
@@ -68,6 +74,9 @@ var Frontend = {
   "Header": Header,
   "Details": Details,
   "Button": Button,
+  "Radios": Radios,
+  "ErrorSummary": ErrorSummary,
+  "SkipLink": SkipLink,
   "initAll": initAll
 }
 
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/esm/all.mjs
```diff
@@ -59,15 +60,23 @@ function initAll (options) {
     new Radios($radio).init()
   })
 
+  // There will only every be one error-summary per page
   var $errorSummary = scope.querySelector('[data-module="govuk-error-summary"]')
   new ErrorSummary($errorSummary).init()
+
+  // There will only ever be one skip-link per page
+  var skipLink = scope.querySelector('[data-module="govuk-skip-link"]')
+  new SkipLink(skipLink).init()
```
samuelhwilliams, line 64: Hmm .. Trying to test this new functionality to add a tabindex to the related element and I can't get it to trigger 🤔

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -545,7 +545,7 @@ def test_service_name_change_fails_if_new_name_fails_validation(
         _expected_status=200,
     )
     assert not mock_update_service.called
-    assert error_message in page.select_one("span.govuk-error-message").text
+    assert error_message in page.select_one("p.govuk-error-message").text
```
samuelhwilliams, line 548: Maybe we don't need to specify the element type and can just do `page.select_one(".govuk-error-message")`?

app/assets/javascripts/esm/all.mjs
tombye: I see it add `tabindex="-1"` to `<main>` locally. Did you `npm install` or `npm ci` and `npm run build` beforehand?

tests/app/main/views/service_settings/test_service_settings.py
tombye: Great suggestion. That commit now uses just the class.

app/assets/javascripts/esm/all.mjs
samuelhwilliams: oooo ok, I was testing it on a page that has two skip links 😱

samuelhwilliams: (the top-level one that skips to `main` works - the second one that skips to `file-preview` (eg blah.csv table)` doesn't work. But I guess that's because we only activate the first skip link on each page

tombye: Good spot! 💯 I adjusted our JS that intialises the skiplink JS so it doesn't assume only one per page and corrected a fragment identifier in another one, in the page for errors in your CSV. All in https://github.com/alphagov/notifications-admin/pull/4465/commits/4158b07e7ff6f071b1fcf95a7253fb08ea6b1891

Now feeling silly for writing the comment `// There will only ever be one skip-link per page`.

app/templates/views/check/column-errors.html
```diff
@@ -144,7 +144,7 @@ <h1 class='banner-title' data-notify-module="track-error" data-error-type="Missi
         ) }}
       {% endif %}
     </div>
-    <a href="#content" class="govuk-link govuk-link--no-visited-state back-to-top-link">Back to top</a>
```
samuelhwilliams, line 147: 👀 good spot

samuelhwilliams (approved):
👍  I've had a look and play and this all seems good to me. Hard to say with 100% certainty though...

PR #4466, opened 2022-11-23
Show service history even if service archived

samuelhwilliams:
We removed the 'Delete this service' link when services are already archived, but accidentally hid the 'Service history' link at the same time. That link should always be visible.

Updates the CSS to remove left padding from the 'Service history' link if it's the first child, otherwise it's oddly-indented.

## CSS change for single link
| Before | After |
|---|---|
| image | image


tests/app/main/views/service_settings/test_service_setting_permissions.py
```diff
@@ -173,16 +173,16 @@ def test_service_setting_links_displayed_for_active_services(
...
     assert link["href"] == link_url
 
 
-def test_service_setting_links_not_displayed_for_archived_services(
+def test_service_setting_delete_link_not_displayed_for_archived_services(
     get_service_settings_page,
     service_one,
 ):
-    link_texts = ["Delete this service", "Service history"]
+    link_texts = ["Delete this service"]
```
leohemsted, line 185: minor nitpick:

can we instead make this test select just assert that the only link is `Service history`? probably just use the same selector as above (`page.select(".page-footer-link")`) and asserting there's just one, and it says Service History

app/templates/views/service-settings.html
```diff
@@ -454,23 +454,19 @@ <h2 class="heading-medium">Platform admin settings</h2>
 
     {% endif %}
 
-    {% if current_service.active and (current_service.trial_mode or current_user.platform_admin) %}
       <p class="top-gutter-1-2">
-        <span class="page-footer-link page-footer-delete-link-without-button">
-          <a class="govuk-link govuk-link--destructive" href="{{ url_for('.archive_service', service_id=current_service.id) }}">
+        {% if current_service.active and (current_service.trial_mode or current_user.platform_admin) %}
+          <a class="govuk-link govuk-link--destructive page-footer-link page-footer-delete-link-without-button" href="{{ url_for('.archive_service', service_id=current_service.id) }}">
```
leohemsted, line 459: i wouldn't mind @tombye 's input on if we want to put classes here or if there's a stylistic choice behind putting a single element in a span. I don't really have big thoughts myself.

leohemsted (commented):
looks fine, but i'd want tom's input, especially since he's looking at links right now

tests/app/main/views/service_settings/test_service_setting_permissions.py
samuelhwilliams: That seems like a slightly weaker assertion. I've updated the test in 6236a47ed08f80999e3b256c8ea39735879f3348

tombye (commented):
Sorry for the late reply, I missed the notification.

I've been chatting with @leohemsted about this and, sorry Sam, have done a bit of a u-turn in thinking about the changes to the page footer HTML.

I no longer think we should remove the span, or the `page-footer-delete-link-without-button` class. I don't know why we have wrapping spans but that's how we do it everywhere else and the classes are used in other places for the styles they apply rather than their name so, until we can refactor the page footer to make more sense, consistency is more important.

Sorry again for the extra work Sam but I do think keeping within what we've got for now makes more sense.

samuelhwilliams:
@tombye no worries :)

I've pushed an update which removes those changes then, so hopefully we'll now keep that consistency :)

leohemsted (approved):

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -4677,7 +4677,7 @@ def test_archive_service_prompts_user(
     client_request.login(user)
 
     settings_page = client_request.get("main.archive_service", service_id=SERVICE_ONE_ID)
-    delete_link = settings_page.select(".page-footer-link a")[0]
+    delete_link = settings_page.select("a.page-footer-link")[0]
```
leohemsted: dont think this'll work anymore actually and should go back to just `.page-footer-link` or `.page-footer-link a`

PR #4467, opened 2022-11-24
bump utils to v58

leohemsted:
uses brand_alt_text rather than brand_name in the email template code

Note that when we pass in params, currently, we're only passing in type, colour and logo. `text`, `name` and `alt_text` are only populated in the email_template endpoint if it's called with a UUID for an existing branding option


tombye (approved):
Looks straightforward to me 👍🏻

PR #4468, opened 2022-11-24
Refactor email preview view to use `EmailBranding` model

quis:
This:
- makes for less manual introspection of dictionaries
- lets us encapsulate some of the logic around branding better

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -76,22 +76,19 @@ def email_template():
         "content": render_template("example-email.md"),
     }
 
-    if not bool(request.args):
-        resp = make_response(str(HTMLEmailTemplate(template)))
-    else:
-        resp = make_response(
-            str(
-                HTMLEmailTemplate(
-                    template,
-                    govuk_banner=branding.has_govuk_banner,
-                    brand_text=branding.text,
-                    brand_colour=branding.colour,
-                    brand_logo=branding.logo_url,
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):

app/main/views/index.py
```diff
@@ -66,48 +59,36 @@ def design_content():
 
 @main.route("/_email")
 def email_template():
-    branding_style = request.args.get("branding_style", "govuk")
+    branding_style = request.args.get("branding_style")
 
     if not branding_style or branding_style in {"govuk", FieldWithNoneOption.NONE_OPTION_VALUE}:
-        source = {"brand_type": "govuk"}
+        branding = EmailBranding.from_id(None)
```
leohemsted, line 65: i think what could potentially be even nicer is a class method for the govuk branding (so we dont need to think about the ID=None abstraction at all)

```python
class EmailBranding:
  @classmethod
  def govuk_branding(cls):
    return cls.from_id(None)

...
source = EmailBranding.govuk_branding()
```

leohemsted (approved):

app/main/views/index.py
quis: Nice, added in https://github.com/alphagov/notifications-admin/pull/4468/commits/a9601f009f0873cbf1fbc57eec8b9235290ec7d6

PR #4470, opened 2022-11-25
Platform admin set alt text

leohemsted:
image


app/navigation.py
```diff
@@ -102,8 +101,9 @@ class HeaderNavigation(Navigation):
             "platform_admin_reports",
             "platform_admin_returned_letters",
             "platform_admin_splash_page",
+            "platform_admin_create_email_branding",
```
samuelhwilliams, line 104: optional nit: tThese are not quite alphabetical 🙃

app/main/views/email_branding.py
```diff
@@ -53,6 +53,7 @@ def platform_admin_update_email_branding(branding_id, logo=None):
 
     form = AdminEditEmailBrandingForm(
         name=email_branding.name,
+        alt_text=email_branding.alt_text,
         text=email_branding.text,
         colour=email_branding.colour,
         brand_type=email_branding.brand_type,
```
samuelhwilliams, line 59: We can just replace this whole block with `form = AdminEditEmailBrandingForm(obj=email_branding)`

leohemsted: niiiice. did not know that was a thing.

app/navigation.py
leohemsted: yeah, was intentional cos i wanted create and update email branding next to each other :can_of_worms:

app/main/forms.py
```diff
@@ -1629,6 +1630,28 @@ def validate_name(self, name):
...
+            err_msg = "Must have exactly one of alt_text and text"
+            if self.alt_text.data and self.text.data:
+                self.alt_text.errors.append(err_msg)
+                self.text.errors.append(err_msg)
+                return False
+
+            if not (self.alt_text.data or self.text.data):
+                self.alt_text.errors.append(err_msg)
+                self.text.errors.append(err_msg)
+                return False
```
samuelhwilliams, line 1651: optional nit: these can probably just be a single `if` statement; they do the same thing?

app/templates/views/email-branding/manage-branding.html
```diff
@@ -31,7 +34,8 @@
       {% call form_wrapper() %}
         <div class="form-group">
           <div style='margin-top:15px;'>{{form.name}}</div>
-          <div style='margin-top:15px;'>{{form.text}}</div>
+          <div style='margin-top:15px;'>{{ form.alt_text(param_extensions={'hint': {'text': 'must provide one of alt_text and text'}}) }}</div>
```
samuelhwilliams, line 37: optional nit: we might want to change these to `govuk-!-margin-top-x` 🤔

samuelhwilliams (commented):
Last commit can probably be squashed as a fixup as well

app/main/forms.py
leohemsted: changed to separate messages in https://github.com/alphagov/notifications-admin/pull/4470/commits/abcad9a639d056a3a70c740a7ebb7b05cd09fe70

app/templates/views/email-branding/manage-branding.html
leohemsted: did some experimenting and they don't need to be in divs at all (but i added the class to the form-group div instead)

🚮

leohemsted: [e2b531c](https://github.com/alphagov/notifications-admin/pull/4470/commits/e2b531c6e8824673c570fa058687ee6f053f2355)

app/main/forms.py
```diff
@@ -1627,7 +1627,10 @@ class AdminPreviewBrandingForm(StripWhitespaceForm):
 
 class AdminEditEmailBrandingForm(StripWhitespaceForm):
     name = GovukTextInputField("Name of brand")
-    text = GovukTextInputField("Text")
+    text = GovukTextInputField("Logo Text", param_extensions={"hint": {"text": "Text that appears beside the logo"}})
```
samuelhwilliams, line 1630: Temp content or not, `Logo Text` probably doesn't need the capitalisation on the `T`. Same for alt text and the hints.

leohemsted: No, I think this was intentional. CC @karlchillmaid

leohemsted: This isn't temp content now, this is final. its just the service setting page stuff which karl is still playing with.

(I don't mind if karl or saima have a look just to double check capitalisation and stuff)

app/main/forms.py
```diff
@@ -1644,6 +1647,27 @@ def validate_name(self, name):
...
+
+            if self.alt_text.data and self.text.data:
+                self.text.errors.append("If you enter logo text you do not need Alt Text")
+                self.alt_text.errors.append("Must be empty")
+                return False
+
+            if not (self.alt_text.data or self.text.data):
+                self.text.errors.append("If you do not need logo text you must enter Alt Text")
+                self.alt_text.errors.append("Cannot be empty")
+                return False
```
samuelhwilliams, line 1667: I think these errors are a bit weird. We probably only want one of the fields to error in each case.

If they provide both, then add `If you enter logo text you do not need alt text` to the `alt_text` field only.
If they don't provide either, then add `You must have alt text if you do not have any logo text` to `alt_text`

samuelhwilliams (changes requested):
I've tried creating a logo from scratch and if I provide a name and logo text but no alt text, I get a 500 🤔

app/main/forms.py
samuelhwilliams: oh ok. The other fields here don't have title caps? It just feels/looks inconsistent.

app/main/forms.py
samuelhwilliams: Although if content/design have come up with this, then ok 👍

leohemsted: interesting thought. one of my driving thoughts behind the error messages was that you probably are thinking about logo text, and what you need to do with alt text follows from that. so maybe we assume you entered the logo text field correctly, and the alt text is the only one that shows the error?

app/main/forms.py
karlchillmaid: No, cap case was not intentional. I forgot to raise it on Friday because we were focused on the validation errors.

```suggestion
    text = GovukTextInputField("Logo text", param_extensions={"hint": {"text": "Text that appears beside the logo"}})
```

app/main/forms.py
```diff
@@ -1627,7 +1627,10 @@ class AdminPreviewBrandingForm(StripWhitespaceForm):
 
 class AdminEditEmailBrandingForm(StripWhitespaceForm):
     name = GovukTextInputField("Name of brand")
-    text = GovukTextInputField("Text")
+    text = GovukTextInputField("Logo Text", param_extensions={"hint": {"text": "Text that appears beside the logo"}})
+    alt_text = GovukTextInputField(
+        "Alt Text", param_extensions={"hint": {"text": "Text for people who cannot see the logo"}}
```
karlchillmaid, line 1632:
```suggestion
        "Alt text", param_extensions={"hint": {"text": "Text for people who cannot see the logo"}}
```

app/main/forms.py
karlchillmaid: I like the idea of showing only 1 error.

Current thinking:

`Cannot be empty if you do not need/enter logo text`
and
`Must be empty if you need/enter logo text`

Will give it some more thought today and confirm later.

karlchillmaid: Everyone happy with this?

`Cannot be empty if you do not have logo text`

`Must be empty if you enter logo text`

samuelhwilliams: I am 🙌  thanks!

leohemsted: fixed in 64da20b

samuelhwilliams (approved):
Nice 👍

PR #4471, opened 2022-11-25
Email branding query arg refactor

leohemsted:
Some small refactors around how we handle query params.

By renaming `type` to `brand_type` in the admin endpoints, we keep that consistent with the model they'll eventually populate, and also the email_preview endpoint.

By using url_for to generate query params for the email preview, we can avoid constructing query params ourselves, and we can also take advantage of url_for stripping None values to tidy up our code a bit

### DIFF
--- a/app/templates/components/branding-preview.html
+++ b/app/templates/components/branding-preview.html
@@ -3,9 +3,7 @@
 {% endmacro %}
 
 {% macro custom_email_branding_preview(preview_settings) %}
-  {% set query_params = (preview_settings | urlencode) %}
-  {% set query_params = 'branding_style=custom' + ('&' + query_params if query_params else '') %}
-  <iframe src="{{ url_for('main.email_template') }}?{{ query_params }}" class="branding-preview" scrolling="no"></iframe>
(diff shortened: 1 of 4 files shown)


app/templates/components/branding-preview.html
```diff
@@ -3,9 +3,7 @@
 {% endmacro %}
 
 {% macro custom_email_branding_preview(preview_settings) %}
-  {% set query_params = (preview_settings | urlencode) %}
-  {% set query_params = 'branding_style=custom' + ('&' + query_params if query_params else '') %}
-  <iframe src="{{ url_for('main.email_template') }}?{{ query_params }}" class="branding-preview" scrolling="no"></iframe>
+  <iframe src="{{ url_for('main.email_template', branding_style='custom', **preview_settings) }}" class="branding-preview" scrolling="no"></iframe>
```
samuelhwilliams, line 6: lol why didn't I do this initially 🤔 🤦  wonder what was going through my mind

samuelhwilliams (approved):

PR #4472, opened 2022-11-25
Email branding alt text flow

leohemsted:
Includes error messages, content, flash messages, etc.

Does not include checking name is unique, getting a new name if the name does conflict. (it'll currently just 5xx). Doing this in a separate PR so we dont have big long dependency chains between this and api etc

 image


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -4103,6 +4103,57 @@ def test_GET_email_branding_set_alt_text_redirects_on_missing_query_params(
...
 
 
+@pytest.mark.parametrize(
+    "alt_text, expected_error",
+    [
+        ("", "Error: Cannot be empty"),
+        ("My First Logo", "Error: Do not include the word ‘logo’ in your alt text"),
+    ],
+)
+def test_POST_email_branding_set_alt_text_shows_error(client_request, service_one, alt_text, expected_error):
```
samuelhwilliams, line 4113: This commit can probably get squashed down into the last one

app/main/views/service_settings.py
```diff
@@ -1460,6 +1460,11 @@ def email_branding_set_alt_text(service_id):
             created_by_id=current_user.id,
             **email_branding_data,
         )
+
+        flash(
+            "You’ve changed your email branding. Send yourself an email to make sure it looks OK.", "default_with_tick"
+        )
```
samuelhwilliams, line 1466: This flash is slightly weird because the new branding hasn't actually been applied yet and I think it's currently lined up as a separate story 🤔

leohemsted: hmmm, that's true. i didnt want to just bounce without a flash and i didnt want to write content that literally only we will see when testing a page that isn't hooked up yet

samuelhwilliams (approved):

PR #4476, opened 2022-11-29
Fix back_links when dropping out of new brand flow

samuelhwilliams:
If a user bails out of the new email branding flow, we take them to the page where they can submit a zendesk ticket with details on the brand they want. This page has a back link, and it should take the user back to the correct place in the previous flow. We need to pass through URL query params and the correct back links to build the right URL.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1294,11 +1294,21 @@ def email_branding_something_else(service_id):
         flash(THANKS_FOR_BRANDING_REQUEST_MESSAGE, "default")
         return redirect(url_for(".service_settings", service_id=current_service.id))
 
+    branding_options = ChooseEmailBrandingForm(current_service)
+    if branding_options.something_else_is_only_option:
+        back_link = url_for(".service_settings", service_id=current_service.id)
+    else:
+        back_link = url_for(
+            request.args.get("back_link", ".email_branding_request"),
+            service_id=current_service.id,
+            **_email_branding_flow_query_params(request),
+        )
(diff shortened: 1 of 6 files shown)


app/main/views/service_settings.py
```diff
@@ -1294,11 +1294,21 @@ def email_branding_something_else(service_id):
         flash(THANKS_FOR_BRANDING_REQUEST_MESSAGE, "default")
         return redirect(url_for(".service_settings", service_id=current_service.id))
 
+    branding_options = ChooseEmailBrandingForm(current_service)
+    if branding_options.something_else_is_only_option:
+        back_link = url_for(".service_settings", service_id=current_service.id)
```
CrystalPea, line 1299: I know you just moved existing behaviour here, but I wonder if this line could be improved, like so:

```
back_link = url_for(
            request.args.get("back_link", ".service_settings"),
            **_email_branding_flow_query_params(request),
        )
```

Then even if something else was the only option at the start, if the user went through self-service flow, which is most likely, they will be taken back to the flow.

CrystalPea (commented):
Nice one 👏🏼 . 

One comment that I'd like to discuss before this goes in ☺️

app/main/views/service_settings.py
samuelhwilliams: Good idea :) I have done a patch in 78c9cfced9fa2c59d49dbf38e4fe5e0a490f91f3

CrystalPea (approved):
:shipit:

PR #4478, opened 2022-11-30
Link up upload your logo journey

CrystalPea:
When this is deployed our users will be able to upload their own branding.

Done:

- links changed, back links adjusted where necessary
- checked permissions
- updated tests


leohemsted:
I know your PR is relatively simple right now, but I was trying to work out which pages remain and how you get there (and if we could clean up any other end points or query params as unreachable).

My head was getting massively worked up so here's me attempting to write down the various branding options, when they show, and what we do if the user selects it. This was mainly for my benefit trying to work out what on earth happens with all this, but I did find a bug with the query params too - that one it would be good to sort out now.

# govuk
### When do we show it?
Service is part of a central org that doesn't have a default branding of its own, and the service doesn't have `govuk` as its current branding
### Where do we redirect to on submit
`.email_branding_govuk`. Confirmation page, that then sets the branding and redirects to settings ✅ 

# nhs
### When do we show it?
Service is part of an NHS org (even if that org has its own default branding), and the service doesn't have `nhs` as its current branding
> this isn't the same behaviour as govuk. if an org has default branding, why is nhs logo okay but govuk crown not? was that a conscious decision?
### Where do we redirect to on submit
`.email_branding_nhs`. Confirmation page, that then sets the branding and redirects to settings ✅ 

# specific branding pool options
### When do we show it
When the org has brandings available in the pool (excluding the one the service has already set)
### Where do we redirect to on submit
`.email_branding_pool_option`. Confirmation page, that then sets the branding and redirects to settings ✅ 

# govuk_and_org
### When do we show it
Service is part of a central org that doesn't have a default branding of its own, and the org doesn't have a branding pool either. (Aside: can an org ever have default branding that isn't in the org pool?)

### Where do we redirect to on submit
`.email_branding_choose_logo` (while passing through `branding_choice=govuk_and_org`). We don't explicitly check for this in code, however, we do this for all central govts selecting `govuk_and_org`, `organisation` or `something_else`.

choose_logo doesn't care about branding_choice, but passes it through in case user ends up on gov identity path.

> BUG AHOY! branding_type is not preserved as a query param if you are on `/choose-logo?branding_choice=govuk_and_org` and  then click "Use your own logo" and get to the no banner/coloured banner page. If you click back, you'll go back to the choose_logo page, but you'll loose that branding_choice variable. you could go all the way to the preview page and still hit back to get to this screen
> I think, even if we fix that, we should put an else tag in the `government-logo-branding-request.txt` zendesk template to say "We aren't sure whether this user wanted govuk_and_org or organisation" or something.

# organisation
### When do we show it
Service is not NHS and the service branding is not set to a specific option. (IE: if the branding is unset, or is manually set to that of the org default, dont show this). 

### Where do we redirect to on submit
exact same behaviour govuk_and_org

# something else
### When do we show it
always!

### Where do we redirect to on submit
is service in central org? then see govuk_and_org
is service nhs/local/etc? then `.email_branding_choose_banner_type` (email-branding/add-banner. we should rename the endpoint to be consistent with the URL or vice versa)


--------------------

phew, i wrote down a lot but it's still really gnarly and complex. i do definitely understand it a little bit more for having gone through and tested it all locally.

my main thoughts on potential ways to unknot some of this. We don't need to do these now (so im not leaving this as a formal review right now) but I wanted to get them written down somewhere before I forget.

Thought 1. Do we need to ask for govuk_and_org/organisation at the first step? Or can we ask that _after_ the user has gone on to the govuk identity flow, selected that option. That will remove lots of complexity from the logic about which options go in to email-branding radio form, and avoid us having to pass around the `branding_choice` query parameter. 

Thought 2. i got really confused because we have a page called `email_branding_something_else` (`/service-settings/email-branding/something-else`) which does NOT correlate to the something else option at all. Ways of accessing it:
* saying you want a banner, but don't know the hex code.
* getting to the logo upload page, but not having a file to upload

Can we rename that endpoint/template/url so it appears less connected to the something else option in the first form? Is there any other situation when a user might want to just free text?

app/main/views/service_settings.py
```diff
@@ -1269,20 +1255,6 @@ def email_branding_nhs(service_id):
...
-
-        flash(THANKS_FOR_BRANDING_REQUEST_MESSAGE, "default")
-        return redirect(url_for(".service_settings", service_id=current_service.id))
-
-    return render_template("views/service-settings/branding/email-branding-organisation.html")
-
-
 @main.route("/services/<uuid:service_id>/service-settings/email-branding/something-else", methods=["GET", "POST"])
 @user_has_permissions("manage_service")
 def email_branding_something_else(service_id):
```
leohemsted, line 1260: i know you haven't changed this function - but on line 1301 this sets a back_link from `request.args` - this is never populated in any of the places we link to this endpoint from, so we could remove this.

Then in the template, if `branding_options.something_else_is_only_option == False`, we could just set the back link to `email_branding_request` at that point.

(i'm also not in love with creating a form just so we can call that helper method on it in a template - could possibly go `if list(get_email_choices(current_service)) == []` but that's a bit weird too. maybe the form is the lesser of two evils.

CrystalPea:
>  (Aside: can an org ever have default branding that isn't in the org pool?)

when email branding is set as as an org default branding, it is also added to the pool. I guess if it was later removed from that pool, it wouldn't be a part of it? I remember we planned to add a constraint, but couldn't for some reason.

CrystalPea:
> is service in central org? then see govuk_and_org
is service nhs/local/etc? then .email_branding_choose_banner_type (email-branding/add-banner. we should rename the endpoint to be consistent with the URL or vice versa)

This is not correct. If service is in central org, they will see `.email_branding_choose_logo`.
If it's not central it will go to `.email_branding_choose_banner_type` indeed.

As to URLs, that is one for @karlchillmaid I guess ☺️

quis:
> > Do we need to ask for govuk_and_org/organisation at the first step? Or can we ask that after the user has gone on to the govuk identity flow, selected that option. That will remove lots of complexity from the logic about which options go in to email-branding radio form, and avoid us having to pass around the branding_choice query parameter.
> 
> I like that, saves us from lots of potential flakiness. What do you reckon @quis ?

I’d like us to keep these options where they are please.

app/main/views/service_settings.py
CrystalPea: We use the `back_link` request arg to get back to relevant views when users get to something_else page from banner colour page or logo upload page.

app/main/views/service_settings.py
```diff
@@ -1461,6 +1436,9 @@ def email_branding_upload_logo(service_id):
 @user_has_permissions("manage_service")
 def email_branding_set_alt_text(service_id):
     email_branding_data = _email_branding_flow_query_params(request)
+    # we use this key to keep track of user choices through the journey but we don't use it to save the branding
+    del email_branding_data["branding_choice"]
```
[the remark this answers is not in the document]
CrystalPea: Done 👍🏼

app/main/views/service_settings.py
leohemsted: ah, you're right! that was added/fixed by sam after i wrote this comment https://github.com/alphagov/notifications-admin/pull/4476

leohemsted (approved):
Good work on this, glad to see it get through!

PR #4480, opened 2022-12-02
De-duplicate email/letter branding views; enable 'add to pool' for letter branding

samuelhwilliams:
WIP


app/main/views/service_settings.py
```diff
@@ -990,13 +991,28 @@ def service_set_email_branding(service_id):
 
 
 @main.route(
-    "/services/<uuid:service_id>/service-settings/set-email-branding/add-to-branding-pool-step", methods=["GET", "POST"]
+    "/services/<uuid:service_id>/service-settings/set-<notification_type>-branding/add-to-branding-pool-step",
```
samuelhwilliams, line 1016: We are likely to change this URL - may want to add some backwards-compatibility when we do (probably just for release, then we can strip it out immediately).

app/main/views/service_settings.py
```diff
@@ -990,13 +991,28 @@ def service_set_email_branding(service_id):
...
-    email_branding_name = email_branding["name"]
+def service_set_branding_add_to_branding_pool_step(service_id, notification_type):
+    branding_id = request.args.get("branding_id")
+    notification_type = notification_type.lower()
+    notification_type_name = f"{notification_type}_branding"
+
+    if notification_type == "email":
+        # TODO: Temporary check for backwards-compatibility. Remove any time if you see this.
+        if not branding_id:
+            branding_id = request.args.get("email_branding_id")
```
samuelhwilliams, line 1028: Just to cover the one release in case anyone is mid flow. This is super unlikely, and since it's platform admins only we might just decide it's not worth having this at all.

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -3651,6 +3651,108 @@ def test_service_set_email_branding_add_to_branding_pool_step_choices_yes_or_no(
         )
 
 
+def test_get_service_set_letter_branding_add_to_branding_pool_step(
```
samuelhwilliams, line 3654: I haven't combined these tests because it feels like it might make them more complex. There's some duplication between the two tests, but I generally think that's a little bit more OK than in the main code itself.

Combining them feels like it would make them less readable.

app/main/views/service_settings.py
```diff
@@ -969,139 +971,152 @@ def set_rate_limit(service_id):
...
+        )
+
+    elif notification_type == "letter":
+        form = AdminSetLetterBrandingForm(
+            all_branding_options=AllLetterBranding().as_id_and_name,
+            current_branding=current_service.letter_branding_id,
+        )
+
+    else:
+        raise NotFound
```
klssmith, line 992: It has the same effect, but this (and the other places like this) could use `abort(404)` which is what we use in the other places where someone has entered an invalid value in a URL - just for consistency

app/main/views/service_settings.py
```diff
@@ -969,139 +971,152 @@ def set_rate_limit(service_id):
...
+    branding_id = request.args.get("branding_id")
+    notification_type = notification_type.lower()
+    branding_type = f"{notification_type}_branding"
+
+    if notification_type == "email":
+        # TODO: Temporary check for backwards-compatibility. Remove any time if you see this.
+        if not branding_id:
+            branding_id = request.args.get("email_branding_id")
+
+        branding = email_branding_client.get_email_branding(branding_id)[branding_type]
```
klssmith, line 1030: We can use the `EmailBranding` model and call `EmailBranding.from_id(branding_id)` (and can make a similar change for the letter branding in the `elif`)

app/templates/views/service-settings/preview-branding.html
```diff
@@ -1,18 +1,26 @@
...
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      {{ letter_branding_preview(form.branding_style.data) }}
+      {% if notification_type == 'email' %}
+        {{ email_branding_preview(form.branding_style.data) }}
+      {% elif notification_type == 'letter' %}
+        {{ letter_branding_preview(form.branding_style.data) }}
+      {% else %}
+        {# Raise an exception so we can catch this #}
+        {{ 0 / 0 }}
```
klssmith, line 21: Given that the view function that calls this template is checking for cases where notification_type is not email or letter and raising an error, how would this line be reached?

app/main/views/service_settings.py
klssmith: I agree we could get away without this since the chance of anyone using the feature as it's being deployed is extremely low and it's platform admin only

app/main/views/service_settings.py
samuelhwilliams: Yeah, it may be more consistent, though abort(404) always upsets my IDE a bit as it doesn't recognise that it terminates execution of the rest of block 🥲 

But I'll change it and get over it...

app/templates/views/service-settings/preview-branding.html
samuelhwilliams: Currently it wouldn't - this is just defensive for if we update the code/logic and perhaps forget to update this template.

app/main/views/service_settings.py
samuelhwilliams: Is that a better/preferable way of doing it? Or just shorthand?

quis: It’s better. It returns an [`EmailBranding`](https://github.com/alphagov/notifications-admin/blob/35df317e4bca5219c26b1a38b65554bbc244b370/app/models/branding.py#L25) object instead of a `dict`.

app/main/views/service_settings.py
```diff
@@ -969,139 +975,152 @@ def set_rate_limit(service_id):
...
-            and email_branding_id
-            and email_branding_id not in current_service.email_branding_pool.ids
-        ):
+        branding_id = form.branding_style.data
+        if branding_id == "__NONE__":  # This represents the email GOV.UK brand
+            branding_id = None
+
+        # in addition to updating the letter branding we want the option of adding it to the
+        # letter branding pool if desirable
+        if current_service.organisation and branding_id and branding_id not in service_branding_pool.ids:
```
klssmith, line 1105: This check can be simplified slightly since `service_branding_pool` returns a pool object regardless of whether the service belongs to an organisation or not: 
```
if branding_id and branding_id not in service_branding_pool.ids:
```

The comments above this line would apply to emails too, so you could remove reference to letters

app/main/views/service_settings.py
```diff
@@ -969,139 +975,152 @@ def set_rate_limit(service_id):
...
+
+        branding = EmailBranding.from_id(branding_id)
+        add_brandings_to_pool = organisations_client.add_brandings_to_email_branding_pool
+
+    elif notification_type == "letter":
+        branding = LetterBranding.from_id(branding_id)
+        add_brandings_to_pool = organisations_client.add_brandings_to_letter_branding_pool
+
+    else:
+        raise NotFound()
```
klssmith, line 1042:
```suggestion
        abort(404)
```

app/main/views/service_settings.py
```diff
@@ -969,139 +975,152 @@ def set_rate_limit(service_id):
...
+    branding_type = f"{notification_type}_branding"
+
+    if notification_type == "email":
+        service_branding_pool = current_service.email_branding_pool
+
+    elif notification_type == "letter":
+        service_branding_pool = current_service.letter_branding_pool
+
+    else:
+        raise NotFound()
```
klssmith, line 1094:
```suggestion
        abort(404)
```

app/templates/views/service-settings/preview-branding.html
```diff
@@ -1,18 +1,26 @@
...
+  <h1 class="heading-large">Preview {{ notification_type }} branding</h1>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      {{ letter_branding_preview(form.branding_style.data) }}
+      {% if notification_type == 'email' %}
+        {{ email_branding_preview(form.branding_style.data) }}
+      {% elif notification_type == 'letter' %}
+        {{ letter_branding_preview(form.branding_style.data) }}
+      {% else %}
+        {# Raise a DivideByZero exception in case we add a new thing and don't update the template #}
```
klssmith, line 20: It seems a bit strange to have a zero division error in an unreachable code path in case we make a mistake in future. Is there a way we could use a python test to guard against future mistakes instead?

app/templates/views/service-settings/preview-branding.html
klssmith: It seems a bit strange to have a zero division error in an unreachable code path in case we make a mistake in future. Is there a way we could use a python test to guard against future mistakes instead?

samuelhwilliams: Will just remove it

app/main/views/service_settings.py
samuelhwilliams: But we only want to redirect to the 'add to pool' step if they _do_ belong to an organisation - otherwise there's no pool to add them to, right?

klssmith: Ah yes, of course. Ignore my previous comment

klssmith (approved):

PR #4481, opened 2022-12-02
Update link styles

tombye:
Updates the links using the new link styles:
- [pull request explaining their design](https://github.com/alphagov/govuk-frontend/pull/2183)
- [release note explaining the new classes and Sass mixins you can use to implement them](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0)

Story: https://trello.com/c/jDtuwHih/116-change-links-in-admin-app-to-new-design-system-style

The old style changed colour on hover. The new one thickens the underline so doesn't rely on colour (which is a WCAG win[^1])

This 'just works' for any normal links, like in our guidance, but we have quite a few custom links so the majority of these changes are making them work.

## Our custom links

|Example|Description|Notes on changes|
|---|---|---|
|image|Navigation links, including the different side navigations but also the service navigation at the top of pages|Uses the `govuk-link--no-underline` class provided as part of the new styles so we can delete any CSS we wrote to remove underlines by default|
|image|Big number links on the dashboard|Uses the `govuk-link--inverse` class to make them white for all states, except focus, so we could remove CSS we wrote for this. Only the text gets an underline so we needed to turn underlines off for the parent element and use the `govuk-link-decoration` and `govuk-link-hover-decoration` mixins provided as part of the new styles to apply it to the text alone.|
|image|Pill links on dashboardy pages|Same as big numbers. Uses the `govuk-link--underline` class for colour and only the last line of text gets an underline.|
|image|Dashboard big count link|Only the text after the number is underlined so we use the same approach as big numbers and pills, turning it off on the parent element and on for a specific child.|
|image|'Show more' link on dashboard|Was hacked to make the text look like a link when it was a `<span>` so required a rewrite to make it a real link so it could inherit the new styles.|
|image|'Edit' links on letter template page|Used the `govuk-link--inverse` class to make them white for all states, except focus, so could remove CSS we wrote for this.|
|image|Big links on broadcast tour pages|Used the `govuk-link--inverse` class to make them white for all states, except focus, so could remove CSS we wrote for this.|
|image|Links in the blue section of the product page|Used the `govuk-link--inverse` class to make them white for all states, except focus, so could remove CSS we wrote for this. Required some extra CSS sadly, to let the new styles override some existing ones.|

## Other fixes

I also rolled a few other fixes for issues with links spotted along the way:
- changes to the hover colour for our custom 'destructive' links
- fixes for how some links look when clicked in Safari (which doesn't fire a focus event like other browsers so you don't see the focus styles by default)
- removal of some redundant styles for `<details>` elements which interfered with the new styles

## Notes for reviewers

This pull request is made to be reviewed commit by commit. It's all been tested across our browser matrix and in forced-colour mode. The CSS is a bit complex in some places too so it's ok to just review that the changes make sense.

[^1]: [WCAG success criteria 1.4.1](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html) states that information shouldn't be conveyed by colour alone.

### DIFF
--- a/app/assets/stylesheets/components/show-more.scss
+++ b/app/assets/stylesheets/components/show-more.scss
@@ -1,40 +1,88 @@
-%show-more,
+// container for the link which
+// - contains the absolutely positioned child pseudo elements
+// - sets text size and so line height
+// - allows alignment and vertical spacing
 .show-more {
-
   @include govuk-font(16);
-  display: block;
-  padding: 0 0;
-  margin: govuk-spacing(3) 0 govuk-spacing(3) 0;
+  position: relative; // contain all
   text-align: center;
-  border-top: 1px solid $govuk-border-colour;
+  margin: govuk-spacing(1) 0 govuk-spacing(5) 0;
+}
+
+.show-more__link {
+
+  // create the separator line the link sits in front of
+  &::before {
+
+    content: "";
+    position: absolute;
+    top: 11px;
+    left: 0;
+    width: 100%;
+    height: 0;
+    border-top: 1px solid $govuk-border-colour;
+
+  }
+
+  // create a click area matching the size of the containing div + 16px extra height
+  &::after {
+
+    content: "";
+    position: absolute;
+    top: govuk-spacing(1) * -1;
+    left: 0;
+    bottom: govuk-spacing(1) * -1;
+    width: 100%;
+
+  }
+
+  // make the link text sit in front of the separator line, with a gap on either side
+  span {
+
+    position: relative; // required for z-index
+    z-index: 1; // place above ::after pseudo-element
+    padding: 0 10px;
+    background: govuk-colour('white');
+
+  }
 
   &:focus {
 
-    border-color: transparent;
-    /* override default focus style to increase top yellow box-shadow */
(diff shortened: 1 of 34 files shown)


quis:
I’ve had a click around the app locally and everything looks good 🥳

app/assets/stylesheets/components/show-more.scss
```diff
@@ -1,40 +1,88 @@
-%show-more,
+// container for the link which
+// - contains the absolutely positioned child pseudo elements
+// - sets text size and so line height
+// - allows alignment and vertical spacing
 .show-more {
```
samuelhwilliams, line 5: With this component, we don't get a thick underlink on the 'Show more' text when hovered, which seems at odds all of the other links in Notify. Should we add this too, or is that an expected exception?

samuelhwilliams: Similarly, it doesn't look like the big-numbers/pills don't have the heavy underline on hover (not obvious but this **is** the hover state - underline cuts through the 'y'):

image

samuelhwilliams (approved):
Have also clicked around and can't see any links without nice new styling 👍 

Styling changes look good to my backend eyes 🎉

PR #4483, opened 2022-12-02
Remove custom hover state from tour links

quis:
Originally these made the background of the link slightly lighter.

When we moved to the Design System colour pallette we used the new hover colour, which is a darker blue that clashes with the light blue.

Since we now have the new link styles with a thicker underline on hover we don’t need to differentiate with colour, solving the clashing problem.

— | Before | After
---|---|---
Default | image | image
Hover | image | image


tombye (approved):
The changes look good to me so once CI passes 👍🏻

PR #4485, opened 2022-12-05
remove misleading aria-expanded attribute on New template button

Adam-C-Mitchell:
Remove "aria-expanded=false" from the "New template" button when adding templates. The attribute is unnecessary on button elements.

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -299,7 +299,7 @@
     this.nothingSelectedButtons = $(`
       <div id="nothing_selected">
         <div class="js-stick-at-bottom-when-scrolling">
-          <button type="button" class="govuk-button govuk-button--secondary govuk-!-margin-right-3 govuk-!-margin-bottom-1" value="add-new-template">
+          <button type="button" class="govuk-button govuk-button--secondary govuk-!-margin-right-3 govuk-!-margin-bottom-1" value="add-new-template" ${!this.$singleNotificationChannel ? 'aria-expanded="false"' : ''}>
             New template
           </button>
           <button type="button" class="govuk-button govuk-button--secondary govuk-!-margin-bottom-1" value="add-new-folder" aria-expanded="false">New folder</button>
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -299,7 +299,7 @@
     this.nothingSelectedButtons = $(`
       <div id="nothing_selected">
         <div class="js-stick-at-bottom-when-scrolling">
-          <button type="button" class="govuk-button govuk-button--secondary govuk-!-margin-right-3 govuk-!-margin-bottom-1" value="add-new-template" aria-expanded="false">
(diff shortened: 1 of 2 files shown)


jonathan-owens-gds (approved):

gds-gharper (approved):

quis:
I think this might not be appropriate for regular Notify, where clicking ‘New template’ reveals radio buttons to choose the type of template.

@tombye can you advise?

gds-gharper (commented):
Ah, I'd missed that this was a generic Notify file, we may have to wait for full decoupling in that case

tombye:
Chris is right, that attribute is there for our non-broadcasts version, where it expands a dialog-like thing.

We could have some branching logic in the JS to support this attribute if we needed this change before decoupling. It would only need an extra class in the HTML and a few lines in the JS to handle the alternative scenario.

tombye:
@gds-gharper that all makes sense. A test for this in `tests/javascripts` should suffice rather than a smoketest, seeing as it's a change in how the JS changes the DOM so ok to be scoped to those tests.

gds-gharper:
> @gds-gharper that all makes sense. A test for this in `tests/javascripts` should suffice rather than a smoketest, seeing as it's a change in how the JS changes the DOM so ok to be scoped to those tests.

Yes apologies, I meant just a test rather than a smoke test (blame fingers typing too quickly before thinking),  I had linked Adam specifically to https://github.com/alphagov/notifications-admin/blob/main/tests/javascripts/templateFolderForm.test.js in a call.

tombye:
@jonathan-owens-gds the 'New Template' button isn't a toggle and the whole pattern for new templates/folders or moving templates/folders doesn't match any of the patterns `aria-expanded` is commonly used with.

This pattern is instead a bit of a mash up of a dialog and a disclosure widget. It's not quite a dialog because it doesn't restrict tabbing to itself or render all other content inert, but it does open new content when clicked, move focus to it and return focus back when cancelled. Likewise, as you say, it does expand content but the shift in focus and removal of the original button means it isn't quite a disclosure either.

This mix up of patterns came about by us tweaking the HTML and content of the original version, based on guidance from accessibility audits. We didn't have `aria-expanded` on the buttons originally but added it when auditors asked, to indicate that new content would be appear when clicked. Because the button is removed when clicked, we decided to make the following changes to give users the information they needed to understand what happened and how to proceed:
1. we shift focus to the new region labelled "Choose template type"
2. we label the radios 'New template', to match the button text
3. we add a description to the region saying how to submit their choice or cancel the action "Press continue to confirm selection or cancel to close"

The response to this pattern has been mixed. We've had auditors mention its inconsistency with disclosure widgets but we've also had auditors use it to complete tasks, notably creating alerts templates, without comment.

We have some stories planned for next year to look at other components in admin that expand and shift focus so likely we'll look at this one again at the same time.

Adam-C-Mitchell:
Have amended to conditionally use the aria-expanded attribute dependent upon whether we're running in Notify or Emergency Alerts (i.e., Single Notification Channel, which is also what we're using to determine whether the 'New template' button should be a link or an expanding component), and added a test to verify for the latter scenario (There is already a test that expects aria-expanded to exist and to be false if data-channel is null).

jonathan-owens-gds:
cc: @tombye @quis

tombye (approved):
These changes look good to me 👍🏻

PR #4492, opened 2022-12-08
Make failed messages links fill their block

tombye:
The link for failed messages on the main dashboard page doesn't fill it's container and it should.

This fixes it.

## Before

failure_message_link_focused_before

## After

failure_message_link_focused_after


quis (approved):
Tested locally, looks good 👍🏻

PR #4494, opened 2022-12-09
Let organisation users see the templates for their services

quis:
We want to let organisation users approve go live requests for teams in their organisation.

As part of reviewing these requests, we think that being able to see the templates a team has set up will be important.

For a user of an organisation viewing a service, this pull request:
- makes the ‘Templates’ navigation item visible
- gives them permission to view the various pages under the ‘Templates’ navigation item

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -44,7 +44,7 @@ form_objects = {
 
 
 @main.route("/services/<uuid:service_id>/templates/<uuid:template_id>")
-@user_has_permissions()
+@user_has_permissions(allow_org_user=True)
 def view_template(service_id, template_id):
     template = current_service.get_template(template_id)
     template_folder = current_service.get_template_folder(template["folder"])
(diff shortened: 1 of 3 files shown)


app/main/views/templates.py
```diff
@@ -785,7 +785,7 @@ def redact_template(service_id, template_id):
 
 
 @main.route("/services/<uuid:service_id>/templates/<uuid:template_id>/versions")
-@user_has_permissions("view_activity")
+@user_has_permissions(allow_org_user=True)
```
leohemsted, line 788: I'm not 100% sure on which users don't have `view_activity` (See dashboard) in practice, but is this ok? those users who cant see the dashboard currently cant view this page at the moment.

that said, i can't figure out why this check would be useful especially since we allow those users to access the specific version files. assume this was just a bug

leohemsted (approved):

app/main/views/templates.py
quis: We don’t restrict who can see the link, so I’m assuming a bug/oversight too:
https://github.com/alphagov/notifications-admin/blob/ced9ef0b11d32945d5a10318058e815fe2055168/app/templates/views/templates/template.html#L64

PR #4498, opened 2022-12-12
Set org default branding if conditions are met

leohemsted:
1. The user set organisation on the very first page
2. the org doesn't have a default set yet
3. the org isn't central govt (keep govuk as default in that case)
4. the org has no other live services (so we dont change any settings for them)

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1199,6 +1199,7 @@ def email_branding_request(service_id):
                     ".email_branding_choose_banner_type",
                     service_id=current_service.id,
                     back_link=".email_branding_request",
+                    branding_choice=branding_choice,
                 )
             )
 
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1504,7 +1504,7 @@ def email_branding_set_alt_text(service_id):
         )
 
         if _should_set_default_org_branding(branding_choice):
-            current_service.organisation.update(email_branding_id=new_email_branding.id)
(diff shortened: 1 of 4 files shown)


app/main/views/service_settings.py
```diff
@@ -1465,6 +1484,9 @@ def email_branding_set_alt_text(service_id):
             current_service.organisation.id, [new_email_branding["id"]]
         )
 
+        if _should_set_default_org_branding(branding_choice):
+            current_service.organisation.update(email_branding_id=new_email_branding["id"])
```
klssmith, line 1488: Since we will potentially update some services here, I think we'd want to call this with `delete_services_cache=True` instead of the default argument of `False`

app/main/views/service_settings.py
```diff
@@ -1446,7 +1465,7 @@ def email_branding_set_alt_text(service_id):
 
     if form.validate_on_submit():
         # we use this key to keep track of user choices through the journey but we don't use it to save the branding
-        del email_branding_data["branding_choice"]
+        branding_choice = email_branding_data.pop("branding_choice")
```
klssmith, line 1468: I didn't manage to get this change to work for me locally. I found that `branding_choice` had a value of `None`, so the default org branding wasn't being set when I'd expect it to be. `brand_type` was in `email_branding_data` but with a value of either `org` or `org_banner`

leohemsted: Good spot, thanks! Think I'd been busy hacking data in the url by the time i wrote this. i've fixed this in d63cade (and tested it locally)

klssmith (approved):

PR #4501, opened 2022-12-12
Reflect selection of ’GOV.UK and `organisation name`’ throughout new branding journey

quis:
If someone selects ‘GOV.UK and `organisation name`’ in the first part of the journey this should be respected when:
- they preview the branding they have upload
- they save the branding they have upload

To achieve this we need to make sure we are passing the info on via the query string.

Unfortunately we do call this different names so the code is a bit of a combination of both:

— | variable name | variable value
---|---|---
UI code | `branding_choice` | `govuk_and_org`
Database | `brand_type` | `both`

When we auto-generate the `name` of the branding this also needs to reflect the choice of GOV.UK and organisation name.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1343,9 +1343,22 @@ def email_branding_enter_government_identity_logo_text(service_id):
 @service_belongs_to_org_type("central")
 def email_branding_choose_logo(service_id):
     form = EmailBrandingChooseLogoForm()
+    branding_choice = request.args.get("branding_choice")
 
     if form.validate_on_submit():
         if form.branding_options.data == "org":
+
+            if branding_choice == "govuk_and_org":
+                return redirect(
+                    url_for(
+                        ".email_branding_upload_logo",
+                        service_id=current_service.id,
+                        branding_choice=branding_choice,
+                        brand_type="both",
+                        back_link=".email_branding_choose_logo",
+                    )
+                )
+
             return redirect(
                 url_for(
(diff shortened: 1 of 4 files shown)


app/main/views/service_settings.py
```diff
@@ -1343,9 +1343,22 @@ def email_branding_enter_government_identity_logo_text(service_id):
...
     if form.validate_on_submit():
         if form.branding_options.data == "org":
+
+            if branding_choice == "govuk_and_org":
+                return redirect(
+                    url_for(
+                        ".email_branding_upload_logo",
+                        service_id=current_service.id,
+                        branding_choice=branding_choice,
+                        brand_type="both",
```
leohemsted, line 1357: as mentioned on slack, does it still make sense to separate branding_choice and brand_type? The alternative is to update the choice to be `('both', '[GOV.UK](http://gov.uk/) and {org.name}')`.

That way, brand_type is a subset of branding_choice and we can just pass one value around.

Happy for that to happen later on in a separate PR.

leohemsted (approved):
Im okay with this for now, but a couple of ideas. One for a future PR (renaming the form option), one that we might be able to do now (moving the govuk_and_org -> both mapping from the form selection to the branding creation moment) if it's worth doing. i'm not super fussed if we dont do either right now tho

CrystalPea (commented):
I just managed to look at this PR, I know it has been merged 😅 . Good work, I have one question though :D

PR #4507, opened 2022-12-13
Check for valid HTML as part of running the tests

quis:
HTMl is a forgiving language, so valid markup is not always needed in order for a page to render properly.

But invalid markup can:
- cause problems with spacing or layout where elements are improperly nested (for example mismatched closing tags, or `<p>` inside `<p>`)
- cause problems for users of assistive technology, if element names are misspelled (for example `<thread>` will be ignored whereas `<thead>` has  semantic meaning)
- by symptomatic of code that is hard to follow because it’s not indented properly

This commit adds a step to `client_request.get()` which checks for valid HTML. If invalid HTML is found then an exception will be raised.

It only runs on HTML documents. Almost all of our pages are HTML, the only exception being the email template, which is nominally XHTML (and hard to validate because of all the tricks we use to get around email client rendering issues).

Examples of output:

image

image

### DIFF
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -49,9 +49,11 @@
 {%- endmacro %}
 
 {% macro row_group(id=None) %}
+  </tbody>
   <tbody class="table-row-group" {% if id %}id="{{id}}"{% endif %}>
     {{ caller() }}
   </tbody>
+  <tbody>
 {%- endmacro %}
 
 {% macro settings_row(if_has_permission='') -%}
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -1,4 +1,4 @@
-{% macro mapping_table(caption='', field_headings=[], field_headings_visible=True, caption_visible=True, equal_length=False) -%}
+{% macro mapping_table(caption='', field_headings=[], field_headings_visible=True, caption_visible=True, equal_length=False, output_tbody_tags=True) -%}
   <table class="table table-font-xsmall">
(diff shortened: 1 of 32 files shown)


app/templates/components/table.html
```diff
@@ -49,9 +49,11 @@
 {%- endmacro %}
 
 {% macro row_group(id=None) %}
+  </tbody>
   <tbody class="table-row-group" {% if id %}id="{{id}}"{% endif %}>
     {{ caller() }}
   </tbody>
+  <tbody>
```
samuelhwilliams, line 56: This looks weird, and even looking at it on the page I can't quite understand why we want this? It's inserting blank `<tbody></tbody>` entries between every service. Why is this better? 🤔 

## Before
image

## After
image

samuelhwilliams (commented):
Has clearly identified some problematic HTML and will stop us adding any more.

Across a few sampled test runs locally (not super reliable data of course) it seems to increase test suite runtime by about 15% (from ~17.5s to ~20s.5), which is quite a bit, but probably this is still worth it.

One comment on something that I don't understand, but otherwise think this is good.

app/templates/components/table.html
quis: That was a bit of a hack to get around nested `<tbody>`s not being valid HTML. I’ve redone it in a less hacky way in [826dd08](https://github.com/alphagov/notifications-admin/pull/4507/commits/826dd085d23bce346ba2170ab7adca128daf8c48)

samuelhwilliams: Ah, nice 👍

samuelhwilliams (approved):

PR #4509, opened 2022-12-14
Refactor counting conditionals in formatters

quis:
There are a bunch of unecessary `elif`s and `else`s which can be made simpler with appropriate use of `return`


leohemsted (approved):
🚮 ✨

PR #4510, opened 2022-12-14
Fix strings which Black has left broken in multiple pieces

quis:
When Black decides that a string split across multiple lines can be squeezed onto one line it doesn’t join the definitions into a single definition.

For example this:
```python
(
    "a"
    "b"
)
```

Becomes:
```python
("a" "b")
```

This pull requests fixes these occurrences so they look like:
```python
("ab")
```

The brackets are then redundant, so again we manually fix them to look like
```python
"ab"
```


samuelhwilliams:
Can we squash these down into a single commit and then add that commit's hash to `.git-blame-ignore-revs`?

quis:
@samuelhwilliams done (see 13422af and 7204141)

samuelhwilliams (approved):
Nice ✨

PR #4513, opened 2022-12-15
move service_settings into subfolder

leohemsted:
it's a pretty big file. we already have a subfolder for testing so lets make a subfolder for the endpoints too.

This'll make it easier for us to figure out where to put other files if we move endpoints out from service_settings into their own files

note: git blame still works fine - see https://github.com/alphagov/notifications-admin/blame/move-service-settings/app/main/views/service_settings/index.py


app/main/__init__.py
```diff
@@ -46,3 +45,6 @@
     verify,
     webauthn_credentials,
 )
+from app.main.views.service_settings import (  # noqa isort:skip
```
CrystalPea, line 48: interesting, is there a reason we care about linting for this enough to use noqa?

CrystalPea (approved):
Nice one, good preparation for more clean up.

One small thing that surprised me commented on below:

app/main/__init__.py
leohemsted: good question - it's that in those files all individually import the `main` and/or `no_cookie` blueprints that are defined above this on lines 3 and 4. isort tries to move the imports to the top of the file.

see https://github.com/alphagov/notifications-admin/pull/2870/commits/23bbfb955f5c0767b0024a7af95b1bfc424041a7

leohemsted: Actually, this is no longer the case - I tested it out and it doesn't change anything. Removed! 🎉

CrystalPea: Ahh, nice one! *dancing panda*

PR #4514, opened 2022-12-15
Say ‘or something else’ if there are lots of items in the branding pool

quis:
If there are lots of items in the pool, it’s handy to flag that _Something else_ is a separate, special option, without making you scan the whole list.

We can do this by putting a divider ahead of it.

We don’t need to do this when the list is very short, which is also the case where there are no items in the pool and all the choices are magic options.

The divider is created by implementing [Radio items with a text divider](https://design-system.service.gov.uk/components/radios/#radio-items-with-a-text-divider) from the GOV.UK Design System.

Before | After
---|---
image | image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2005,7 +2005,10 @@ class ChooseEmailBrandingForm(ChooseBrandingForm):
 
     def __init__(self, service):
         super().__init__()
-        self.options.choices = tuple(OrderedSet(branding.get_email_choices(service)) | {self.FALLBACK_OPTION})
+        choices = OrderedSet(branding.get_email_choices(service))
+        if len(choices) > 2:
+            choices = choices | {GovukRadiosField.Divider("or")}
+        self.options.choices = tuple(choices | {self.FALLBACK_OPTION})
 
 
 class ChooseLetterBrandingForm(ChooseBrandingForm):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2001,7 +2001,7 @@ class ChooseBrandingForm(StripWhitespaceForm):
 
 
 class ChooseEmailBrandingForm(ChooseBrandingForm):
-    options = RadioField("Choose your new email branding")
+    options = GovukRadiosField("Choose your new email branding")
 
     def __init__(self, service):
         super().__init__()
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -687,7 +687,21 @@ class GovukRadiosField(GovukFrontendWidgetMixin, RadioField):
     govuk_frontend_component_name = "radios"
 
     class Divider(str):
-        pass
+        """
+        Behaves like a normal string but can be used instead of a `(value, label)`
+        pair as one of the items in `GovukRadiosField.choices`, for example:
+
+            numbers = GovukRadiosField(choices=(
+                (1, "One"),
+                (2, "Two"),
+                GovukRadiosField.Divider("or")
+                (3, "Three"),
+            ))
+
+        When rendered it won’t appear as a choice the user can click, but instead
+        as text in between the choices, as per:
+        https://design-system.service.gov.uk/components/radios/#radio-items-with-a-text-divider
+        """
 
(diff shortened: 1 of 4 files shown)


app/main/forms.py
```diff
@@ -1999,7 +2029,7 @@ class ChooseLetterBrandingForm(ChooseBrandingForm):
     def __init__(self, service):
         super().__init__()
 
-        self.options.choices = tuple(OrderedSet(list(branding.get_letter_choices(service)) + [self.FALLBACK_OPTION]))
+        self.options.choices = tuple(list(branding.get_letter_choices(service)) + [self.FALLBACK_OPTION])
```
tombye, line 2032: Why was the `OrderedSet` removed here? This may be obvious to a Python dev but I'm not clear the reason.

app/main/forms.py
```diff
@@ -1985,11 +2001,25 @@ def something_else_is_only_option(self):
...
+            },
+        }
+    )
 
     def __init__(self, service):
         super().__init__()
-        self.options.choices = tuple(OrderedSet(branding.get_email_choices(service)) | {self.FALLBACK_OPTION})
+        choices = OrderedSet(branding.get_email_choices(service))
+        if len(choices) > 2:
+            choices = choices | {GovukRadiosField.Divider("or")}
```
tombye, line 2021: This use of the bitwise OR operator seems kind of magic. At least in how it adds the "or" 2nd from last. If this is obvious to a Python dev, feel free to ignore but if not could we have a comment explaining it?

app/main/forms.py
```diff
@@ -686,7 +686,23 @@ class GovukCollapsibleNestedCheckboxesField(NestedFieldMixin, GovukCollapsibleCh
 class GovukRadiosField(GovukFrontendWidgetMixin, RadioField):
     govuk_frontend_component_name = "radios"
 
+    class Divider(str):
```
tombye, line 689: I'd quite like a comment explaining why to use a class here to make dividers a kind of variant type. Thanks for adding the comments on the internals though. Really helps reading it.

tombye (approved):
Overall this looks good to me so marking as approved. I left a few comments from reading it around explaining a few things but nothing critical and may well just be my understanding as a someone who doesn't write much Python.

app/main/forms.py
quis: This was a mistake – I didn’t mean to change the letter branding form yet. I’ve squashed it out.

app/main/forms.py
quis: `|` is the bitwise or operator when used with integers. When used with two sets it is the [union](https://docs.python.org/3/library/stdtypes.html#frozenset.union) operator, returning a set containing elements of both sets.

For example:
```python
{1, 2} | {2, 3} == {1, 2, 3}
```

This is similar to using `+` with lists:
```python
[1, 2] + [2, 3] == [1, 2, 2, 3]
```

Unfortunately the `+` operator doesn’t work for sets:
```python
{1, 2} + {2, 3}
>> TypeError: unsupported operand type(s) for +: 'set' and 'set'
```

Probably wouldn’t add a comment for this since it’s a core part of the language.

app/main/forms.py
quis: Docstring added in [676c51a](https://github.com/alphagov/notifications-admin/pull/4514/commits/676c51a63f58d2f98e94d386940ef3733294db16)

The alternative would be to interpret strings as dividers, but that feels a bit more hacky and brittle, eg:
```python
numbers = GovukRadiosField(choices=(
    (1, "One"),
    (2, "Two"),
    "or",
    (3, "Three"),
))
```

app/main/forms.py
tombye: From discussion outside of this PR, I was missing how this line works with the next one. I thought the union on this line somehow added the Divider instance 2nd from last but it actually just bolts it on the end and the next line adds a final option (`self.FALLBACK_OPTION`). No magic after all, just Python.

PR #4343, opened 2022-09-05
Revert "Revert "Update geodata for local authorities and electoral wards to the latest versions""

quis:
In the original attempt to update the source data to the latest versions we accidentally switched to using files which only covered Great Britain, not the whole of the UK: https://github.com/alphagov/notifications-admin/pull/4318/files#diff-8c8fde3505da9c209b3b71805e3c8b5ba985c9673e29cf4dbc5ee7ef94a7bee9L193-R197

This meant we had to roll back the change in alphagov/notifications-admin#4333

This pull request:
- Fixes the source data to use files for the whole of the UK, not just Great Britain
- Re-applies the fixes from https://github.com/alphagov/notifications-admin/pull/4318/commits/696d1bf568c68eaa61b3ffcc7ecdbee033646331
- Adds some extra test cases to make sure we have data for local authorities and electoral wards from all nations of the UK

image

PR #4515, opened 2022-12-15
Add missing source files

quis:
These were missed out when updating the data sources in https://github.com/alphagov/notifications-admin/pull/4343/files

They were missed because committing `.csv` files to the admin app requires `git add --force` to override the precautions taken in our `.gitignore` file.


samuelhwilliams:
It looks like we list these exceptions explicitly in `.gitignore` for other cases - should we add these there as well?

quis:
Unfortunately a blanket `!app/broadcast_areas/source_files/` doesn’t override the `*.csv` declaration.

I’ve added the files we have now to `.gitignore` in 3b9a19ac9, which hopefully is better than nothing

samuelhwilliams (approved):

PR #4516, opened 2022-12-16
Bump utils to bring in new version of pyproj

klssmith:
This bumps utils to version 60.0.0 which brings in a breaking change to how we use Polygons. As a result of this, we need to regenerate the broadcast areas.


leohemsted (approved):
did you need to do much manual work to rebuild the sqlite db? is there anything useful we can write down in a readme we can add in the `broadcast_areas` folder for if we need to regenerate again in the future?

klssmith:
It was just a case of running `./app/broadcast_areas/create-broadcast-areas-db.py` and then committing the changes

klssmith:
I've added an extra commit to document that

PR #4517, opened 2022-12-16
Tidy up inbound sms code

klssmith:
This ensures we always use the new api route that was added for inbound SMS and simplifies the `add_sms_sender` since it's no longer used for inbound SMS.


CrystalPea (approved):
Nice one 👏🏼 ✨

PR #4518, opened 2022-12-16
Fix a variety of service and org name XSS scripts

leohemsted:
The first example we found was:

specifically this can be seen when deleting a template - if you name your template `<script>alert(1)</script>` for example, that html will be rendered and the alert will happen

jinja treats all variables as unsafe unless we declare them safe (Note that all jinja macros are considered to return fully formed HTML, thus are declared as safe internaly automatically). To mark a variable as containing HTML you wish to render, you pass it through the `safe` filter `{{ foo | safe }}`. This is what the govuk-frontend-jinja does to any html you pass in through the html property.

When showing a delete prompt we use this html field to wrap the name of the thing with a visually hidden span so it is read out by screen readers but not visible to the user (we do this because the name's already above in the `body` variable normally).

However, we had an [XSS](https://owasp.org/www-community/attacks/xss/) vulnerability where the name of the template isn't escaped, and then it's marked as safe. It's not safe, as it's untrusted user input.

To fix this, rather than concatenating the string in python inside the macro invocation, we can create the string in html in a set block above. By calling `{{ thing }}`, jinja automatically escapes any HTML contained within.

--------------
All these are fixed, also, Tom and I have gone through all the places where we pass html in to govuk frontend components and ensured that we:

* are escaping user provided values by taking advantage of jinja's escape-by-default `{{foo}}` syntax
* only use html where necessary, using text by default
* are declaring html with user provided data in templates rather than in forms (which also avoids weird form state issues)

## Screenshots of pages where XSS vulnerabilities were found

|||
|-|-|
| image | image | 
image | image |

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2050,26 +2050,7 @@ class GovernmentIdentityLogoForm(StripWhitespaceForm):
     logo_text = GovukTextInputField(
         "Enter the text that will appear in your logo",
         validators=[DataRequired("Cannot be empty")],
-        param_extensions={
-            "label": {
-                "isPageHeading": True,
-                "classes": "govuk-label--l",
-            },
-            "hint": {
-                "html": (
-                    "This is usually the full name of your organisation.<br/><br/>For example, {organisation_name}"
-                )
-            },
-        },
     )
-
-    def __init__(self, organisation=None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-        organisation_name = organisation.name if organisation else "Department of Education"
-        self.logo_text.param_extensions["hint"]["html"] = self.logo_text.param_extensions["hint"]["html"].format(
-            organisation_name=organisation_name
-        )
 
 
 class EmailBrandingChooseLogoForm(StripWhitespaceForm):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1728,42 +1728,18 @@ class AddLetterBrandingOptionsForm(StripWhitespaceForm):
 
 
 class AdminSetBrandingAddToBrandingPoolStepForm(StripWhitespaceForm):
-    def __init__(self, *args, org_name, service_name, **kwargs):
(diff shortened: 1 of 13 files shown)


app/main/forms.py
```diff
@@ -1728,42 +1728,18 @@ class AddLetterBrandingOptionsForm(StripWhitespaceForm):
...
-            },
-            "items": [
-                {"hint": {}},
-                {"hint": {}},
-            ],
-        }
-        self.add_to_pool.param_extensions["items"][0]["hint"]["html"] = Markup(
-            f"""
-            <ul class="govuk-list govuk-hint govuk-!-margin-bottom-0">
-                <li>Apply this branding to ‘{service_name}’</li>
```
quis, line 1751: Instead of deleting completely could do `from notifications_utils.formatters import escape_html` and then `{escape_html(service_name)}`

tombye (approved):
All looks good to me. I think @quis added a comment that needs answering but I'm happy to approve as-is 👍🏻

quis (approved):
Happy to merge as-is to fix the problem, then do a partial revert with `escape_html` to put the hint text back

app/main/forms.py
leohemsted: we could - i moved this definition to [the template instead](https://github.com/alphagov/notifications-admin/pull/4518/files/ca87d0560c8c2bdc763a5ee2a3f3f75eb3725fbd#diff-6fe4bb1bc913cd5676f7d899d6a32efe454d9a8a76d1d1d58ccd823809803779R23) where we can just use jinja's builtin {{}} to do the escaping. Is there an obvious advantage from keeping this html defined in python? I know the `escape_html` is a bit complex, does jinja's escaping handle the same edge cases?

it definitely feels much nicer for me to have the html defined within an html template.

quis: Ah sorry, I missed that you moved it. All good, and the HTML is better in the Jinja files where possible.

PR #4519, opened 2022-12-19
Fix backlinks to notification dashboard

samuelhwilliams:
The notifications dashboard shows an overview of how many notifications are in each state, and below that a table containing links to the latest notifications in each state.

When clicking a notification, the page shown has a back link that returns the user to the notifications dashboard. However, that link didn't take the user back to the previous filtered state (eg only failed notifications), which could be confusing. This patch updates the links so that they correctly record and return the user to the filtered dashboard view.


leohemsted (approved):
simple and sensible

PR #4520, opened 2022-12-19
Update content when setting service name for central government

samuelhwilliams:
Updates the content for creating a service and changing a service's name in line with the content docs provided/design decisions made.

I'm wondering if we can combine some of this content or remove some of the disparities as I'm not sure it's necessary to have such different words based on the org type, but as it stands this is what we're going for.

Ticket: https://trello.com/c/x1KbCk2U/127-add-more-explanation-in-ui-about-what-a-service-is-and-how-to-name-it

# Creating a new service

## Before
| Local | All others |
|---|---|
| image | image |

## After
| Central | Local | All others |
|---|---|---|
| image | image |  image |

# Changing the name of an existing service

## Before
| Local | All others |
|---|---|
| image | image |

## After

| Central | Local | All others |
|---|---|---|
| image | image |  image |


saimaghafoor:
Thanks @samuelhwilliams , looks great to me.

quis (approved):
Looks good to me

PR #4521, opened 2022-12-19
Remove old csv logic

samuelhwilliams:
While fixing the backlinks for viewing notifications I spotted that this CSV endpoint seems to no longer be in use (replaced by `download_notifications_csv`). So this attempts to clean that up and rename it as what it actually is a - helper for the refreshing dashboard ajax blocks.

Athena says the removed route hasn't been used:
image

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -192,14 +192,14 @@ def view_notifications(service_id, message_type=None):
 @main.route("/services/<uuid:service_id>/notifications/<template_type:message_type>.json", methods=["GET", "POST"])
 @user_has_permissions()
 def get_notifications_as_json(service_id, message_type=None):
-    return jsonify(get_notifications(service_id, message_type, status_override=request.args.get("status")))
+    return jsonify(get_notifications(service_id, message_type))
 
 
 @main.route(
     "/services/<uuid:service_id>/notifications/<template_type:message_type>.csv", endpoint="view_notifications_csv"
 )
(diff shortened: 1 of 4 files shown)


app/main/views/jobs.py
```diff
@@ -191,17 +190,11 @@ def view_notifications(service_id, message_type=None):
 @main.route("/services/<uuid:service_id>/notifications.json", methods=["GET", "POST"])
 @main.route("/services/<uuid:service_id>/notifications/<template_type:message_type>.json", methods=["GET", "POST"])
 @user_has_permissions()
-def get_notifications_as_json(service_id, message_type=None):
-    return jsonify(get_notifications(service_id, message_type, status_override=request.args.get("status")))
+def _get_notifications_dashboard_partials_data_as_json(service_id, message_type=None):
```
quis, line 193: Not sure I love the naming here:
- I think of the dashboard as the page with the three blue boxes. I’d call this page the notifications page
- I can’t find other examples where we use underscore prefixing to mean an function which serves partials

Given those points I’d go for something more like:
```suggestion
def get_notifications_page_partials_as_json(service_id, message_type=None):
```

samuelhwilliams: Nice, will rename. I suppose I was using the underscore prefix to indicate that it wasn't a view intended for public consumption (ie semi internal) moreso than one for partials.

quis (approved):
Nice clean up

PR #4524, opened 2022-12-19
Create `OnOffSettingForm` and `YesNoSettingForm` instead of `ServiceOnOffSettingForm`

quis:
We also use `ServiceOnOffSettingForm` this form on the user profile page, so it’s not strictly for service settings. We will also use it on organisations in the future, so it makes sense for it to be more generic.

We often use the `OnOffSettingForm` with ‘Yes’ and ‘No’ as the default choices, so this pull request also makes a reusable thing for doing that.


samuelhwilliams (approved):
👍

PR #4525, opened 2022-12-20
Letter branding svg upload

leohemsted:
image

viewable by platform admins for now. it's linked in properly, albeit the form still requires you to put something in the zendesk form to make validation easier.

when a file is uploaded, it's put on s3 with a temp tag (the same way we do for email and letter brandings currently).


app/utils/branding.py
```diff
@@ -32,9 +32,5 @@ def get_letter_choices(service):
         for branding in service.letter_branding_pool.excluding(service.letter_branding_id):
             yield (branding.id, branding.name)
 
-    elif (
-        service.organisation
-        and not service.is_nhs
-        and (not service.letter_branding_id or service.letter_branding_id != service.organisation.letter_branding_id)
-    ):
+    if service.organisation and not service.letter_branding_pool:
```
quis, line 35: Not for now but we should change this to `not service.letter_branding_pool.excluding(<NHS branding>)` in the future, because NHS organisations will always have a pool of at least 1 item (the NHS branding)

app/main/views/service_settings/letter_branding.py
```diff
@@ -14,6 +14,22 @@
...
+
+    We've got a variety of query parameters we want to pass around between pages. Any params that are passed in, we
+    should pass through to ensure that back links continue to work the whole way through etc. In addition, we use the
+    branding_choice param (from the letter_branding_request page) later on so need to pass that through.
+
+    To set a new value:
+        _email_branding_flow_query_params(request, branding_choice='organisation')
+
+    To remove a value:
+        _email_branding_flow_query_params(request, branding_choice=None)
```
quis, line 28:
```suggestion
    To set a new value:
        _letter_branding_flow_query_params(request, branding_choice='organisation')
    To remove a value:
        _letter_branding_flow_query_params(request, branding_choice=None)
```

quis (approved):
Think this is good to go as a building block, rather than trying to do everything in one PR.

app/utils/branding.py
leohemsted: good to know - does that also apply to emails?

PR #4529, opened 2022-12-20
Create schedule-emails-and-text-messages.html

saimaghafoor:
New page in guidance called 'Schedule emails and text messages'.

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -292,6 +292,14 @@ def letter_specification():
     )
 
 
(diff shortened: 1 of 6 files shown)


samuelhwilliams:
Hey hey 👋 

I've pushed a couple of commits that add the code side of this work.

Here's the page as it stands:
image

I also added a bullet link to it from the top-level guidance page as that feels right?
image

samuelhwilliams (approved):

PR #4534, opened 2022-12-21
remove all unused imports from jinja templates

leohemsted:
i hacked together a script that parses all jinja files, identifies macro declarations, then finds any files that import those macros but don't use them

the only unused import i didn't remove was from the accessibility statement because then it wanted me to update the last reviewed date for the unit test

the script wasn't perfect, there were some false positives when words like `select` and are sometimes a macro, and are sometimes just the word select. but all the things i removed were legitimate

```python
import re
import pathlib
import logging


all_macros = {}
# read all files to find macro definitions
for template_path in pathlib.Path("app/templates").glob("**/*.html"):
    with open(template_path, "r") as f:
        lines = f.readlines()
    for line in lines:
        if match := re.search(r"\{%\s+macro\s+(.+?)\(", line):
            all_macros[match.group(1)] = 0

# read all files again to find usage
for template_path in pathlib.Path("app/templates").glob("**/*.html"):
    with open(template_path, "r") as f:
        lines = f.readlines()

    imports = set()
    used_macros = set()
    defined_macros = set()

    for line in lines:
        line = line.strip()

        # find imports. theres a bit of nuance - they can look as complex as
        # {% from "components/page-footer.html" import a, b, c, d with context -%}
        if match := re.search(r"""\{%\s*from\s+['"](.+?)['"]\s+import\s+(.+?)( with context)?\s*-?%\}""", line):
            imported_things = [x.strip() for x in match.group(2).split(",")]
            for imported_thing in imported_things:
                if imported_thing not in all_macros and "govuk_frontend_jinja/components" in line:
                    # make sure we dont have unused frontend macros (that have no definition in our code)
                    all_macros[imported_thing] = 0
                imports.add(imported_thing)
        elif match := re.search(r"\{%\s+macro\s+(.+?)\(", line):
            defined_macros.add(match.group(1))
        # its not a macro definition OR an import. just a normal template line
        else:
            # split into tokens on spaces, commas, brackets etc
            tokens = re.split(r"[%{\s,(]", line)
            for token in tokens:
                if token in all_macros:
                    used_macros.add(token)
                    all_macros[token] += 1

    if unused_macros := imports - used_macros:
        logging.error(f" {template_path} unused: {' '.join(unused_macros)}")
    if undeclared_macros := used_macros - defined_macros - imports:
        logging.error(f" {template_path} undeclared {' '.join(undeclared_macros)}")

for macro in all_macros:
    if all_macros[macro] == 0:
        # this is okay if its a govuk-jinja component as we dont import them directly (but render them as a widget)
        logging.error(f"globally unused macro {macro}")
```

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -1,5 +1,4 @@
 {% extends "govuk_frontend_jinja/template.html"%}
-{% from "components/banner.html" import banner %}
 {% from "components/cookie-banner.html" import cookie_banner %}
 
 {% block headIcons %}
--- a/app/templates/components/radios.html
+++ b/app/templates/components/radios.html
@@ -1,12 +1,7 @@
-{% from "components/select-input.html" import select, select_list, select_nested, select_wrapper, select_input %}
+{% from "components/select-input.html" import select, select_nested, select_input %}
 
 {% macro radios(field, hint=None, disable=[], option_hints={}, hide_legend=False, inline=False) %}
   {{ select(field, hint, disable, option_hints, hide_legend, input="radio", inline=inline) }}
-{% endmacro %}
-
-
-{% macro radio_list(options, child_map, disable=[], option_hints={}) %}
-  {{ select_list(options, child_map, disable, option_hints, input="radio") }}
 {% endmacro %}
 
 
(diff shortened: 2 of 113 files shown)


samuelhwilliams:
If this is cleanup we think is valuable, then can/should we add a test to make sure things don't start creeping back in?

leohemsted:
@samuelhwilliams see https://github.com/alphagov/notifications-admin/pull/4533 - im still scratching my head to figure out why chris's diff isn't exactly the same as mine tho

quis:
@leohemsted I’ve pulled out the removing-the-unused-macros bit into https://github.com/alphagov/notifications-admin/pull/4556

PR #4535, opened 2022-12-21
Add test tracking all app routes

samuelhwilliams:
Maybe we want this, maybe we don't. Scratched an itch as I thought about it.

If we want this it'd probably be good to centralise some of the flask commands framework, which is currently just a copy/paste of what we have in the API.

---

Whenever we remove a route from the admin app, we should think about how this might impact on any users who've saved that route (eg in a bookmark) and try to put a redirect in place. This makes it easier to immediately spot when a route has been removed, as well as reducing the burden on remembering this team decision around redirects.

This patch adds a test to remind us to think about this. It checks against a cached list of all the admin app routes. We add a helper command to generate this automatically so that the dev burden is low.

Example failure:
image


quis:
Generally I think this is worth trying out. It could be a pain having to remember to add to the list every time a route is added, but we can see how it goes. Would be interested in seeing what others think as well.

samuelhwilliams:
Yeah I agree about the potential annoyance given we probably add routes more often than remove them. That feels like the biggest negative of this approach. Hopefully the command makes that easier, but still feels potentially like extra faff.

leohemsted:
so the process is:

* added a route: just need to add that route to the file
* removed a route: need to think about redirects if relevant, etc

if there are only new routes and nothing removed, we could just update the json and carry on with our day? possibly this could be a pre-commit hook too for tidiness? 🤔

samuelhwilliams:
@leohemsted I've updated the test to automatically save the routes rather than fail on them - it's possibly a little bit weird, but more dev-friendly?

I've also added a pre-commit hook. This is nice for enforcement, but booting up pytest is pretty slow so having this happen on every commit is possible more annoying than useful?

leohemsted:
i definitely agree that having `make test` modify files on disk is a bit weird - i think i hadn't fully completed my thought and then explained it a bit badly. i was imagining usage being more how we use, eg, `isort`. when we run `make test` it uses `isort --check-only` to just return true or false. But we use `isort` without args on pre-commit to fix things automatically.

we could then distinguish the two modes by changing the command to have a separate call to `save-app-routes` with a `--remove-old-routes` flag or something? The command could load the old file, work out the new file, and use the flag to work out if to append or to remove the old stuff too.

Then we could just run `flask command save-app-routes` in the pre-commit, and in the description in the test failure we could put a note saying "if you intend to remove the old route and are confident a redirect isn't suitable, run `flask save-app-routes --remove-old-routes`. Or something like that?

also fyi `pytest tests/app/test_routes.py::test_record_changes_in_routes` should be a bit faster than collecting the entire suite and using -k to filter on test name

samuelhwilliams:
Yeah good point - have refactored that a bit.

leohemsted (approved):
i think this looks real tidy now with the nice commit hook too. worth trying out - we can always iterate or remove entirely if it's a real pain but i think it's a nice sensible step

PR #4539, opened 2022-12-23
Letter branding set name

leohemsted:
image

This is a bit barebones to make reviewing and deploying a bit easier. A future PR will add things like setting the name, setting org default, renaming the filename to not be "temp", etc.

### DIFF
--- a/app/templates/views/service-settings/branding/add-new-branding/letter-branding-set-name.html
+++ b/app/templates/views/service-settings/branding/add-new-branding/letter-branding-set-name.html
@@ -1,0 +1,38 @@
+{% extends "withnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/branding-preview.html" import custom_letter_branding_preview %}
+{% from "components/error-summary.html" import errorSummary %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+
+{% block service_page_title %}
+  Preview your letter branding
+{% endblock %}
+
+{% block backLink %}
+  {{ govukBackLink({"href": back_link}) }}
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ page_header('Preview your letter branding') }}
+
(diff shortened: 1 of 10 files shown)


app/templates/views/service-settings/branding/add-new-branding/letter-branding-set-name.html
```diff
@@ -0,0 +1,38 @@
...
+  Preview your letter branding
+{% endblock %}
+
+{% block backLink %}
+  {{ govukBackLink({"href": back_link}) }}
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ page_header('Preview your letter branding') }}
```
CrystalPea, line 19: Title and header are the same, we could set it as a variable so that we have single source of truth, like here:

https://github.com/alphagov/notifications-admin/blob/main/app/templates/views/service-settings/branding/add-new-branding/email-branding-choose-logo.html#L7

tests/app/main/views/service_settings/test_letter_branding_requests.py
```diff
@@ -556,6 +556,55 @@ def test_GET_letter_branding_set_name_redirects_if_temp_filename_not_provided(cl
     )
 
 
+def test_POST_letter_branding_set_name_shows_error(client_request, service_one):
```
CrystalPea, line 559: Test name could be a bit more descriptive, like: `test_POST_letter_branding_set_name_shows_error_if_field_left_empty`

CrystalPea (commented):
Good work, great to see this nearing completion ☺️ .

I left a couple comments. Also I did not click approve yet as I couldn't test it on my local due to issues with running Notify locally (docker issues - will look at them tomorrow, but we can also have a short call where you give me a quick demo so I could approve? :) )

CrystalPea (approved):
It works as is, just we need to remember to handle those TODOs, including ensuring names are unique.

klssmith:
The logos are being saved in the database with the temporary prefix (`temp-{user_id}_`) which is going to risk us running into the same bug with email branding where logos that platform admin users have uploaded get deleted.

leohemsted:
@klssmith yeah, it's on my radar and i've got a PR on the way that'll solve that, i just wanted to keep the PRs small.

PR #4541, opened 2022-12-28
Email an organisation’s users when one of its service requests to go live

quis:
If an organisation can approve its own go-live requests then we should let them know when there’s a new service that wants to go live.

Depends on:
- [x] https://github.com/alphagov/notifications-admin/pull/4538
- [x] https://github.com/alphagov/notifications-api/pull/3682

### DIFF
--- a/tests/app/main/views/service_settings/test_service_settings.py
+++ b/tests/app/main/views/service_settings/test_service_settings.py
@@ -1786,32 +1786,16 @@ def test_should_be_able_to_request_to_go_live_with_no_organisation(
 def test_request_to_go_live_is_sent_to_organiation_if_can_be_approved_by_organisation(
     client_request,
     mocker,
-    single_reply_to_email_address,
-    single_letter_contact_block,
-    single_sms_sender,
     mock_get_organisations_and_services_for_user,
     mock_get_service_organisation,
-    mock_get_service_templates,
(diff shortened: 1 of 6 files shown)


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -1776,6 +1776,49 @@ def test_should_be_able_to_request_to_go_live_with_no_organisation(
...
+    mock_notify_users_of_request_to_go_live_for_service,
+    service_one,
+    organisation_one,
+    can_approve_own_go_live_requests,
+    expected_call_args,
+):
+    organisation_one["can_approve_own_go_live_requests"] = can_approve_own_go_live_requests
+    mocker.patch("app.organisations_client.get_organisation", return_value=organisation_one)
+
+    for channel in {"email", "sms", "letter"}:
```
klssmith, line 1807: The test passes without mocking the volumes here.

There are also some of the mock fixtures that could be removed - `single_reply_to_email_address`, `single_letter_contact_block`, `single_sms_sender`, `mock_get_service_templates`, `mock_get_users_by_service` and `mock_get_invites_without_manage_permission`.

klssmith (approved):
This worked nicely when I tried it locally 👍

PR #4542, opened 2022-12-29
Only show ‘not set up’ message when user has chosen organisation branding

quis:
If the user has chosen their organisation’s branding but we don’t have it then we should let them know that they’re going to need to set it up.

I can’t remember if we did this and it changed to talk about the service<sup>1</sup> or if it was never like this.

Anyway, talking about the service’s branding not being set up doesn’t make sense when the user has pick an option labelled with the name of their organisation, so let’s fix it.

Before | After
---|---
image | image

***

1. `git blame` got me back as far as https://github.com/alphagov/notifications-admin/commit/7dafb91fb3bca38248236849e0b6b9fce7ecada4#diff-626679b981f01eb2e5471518344e7d131ef28b01b8d0aafbf2bc31d2f704ccabR35-R39 which doesn’t explain why the line was added

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -1420,6 +1420,7 @@ def email_branding_choose_logo(service_id):
             "views/service-settings/branding/add-new-branding/email-branding-choose-logo.html",
             form=form,
             branding_options=form,
+            branding_choice=branding_choice,
         ),
         400 if form.errors else 200,
     )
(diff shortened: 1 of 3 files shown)


CrystalPea:
The line was added because it was in the design screenshots: https://docs.google.com/presentation/d/1xCCVccjtZe-W-6AAIHKHXwQSghphO3np1zzqqu3r-44/edit?usp=sharing

I just didn't catch that it should refer to the org branding, not service branding.

quis:
@CrystalPea got it, always good to understand why something is the way it is 😄 

Are you ok to approve this change then?

CrystalPea (approved):

PR #4543, opened 2022-12-29
Tweak email branding journey to make it smoother

quis:
If we jump people straight into the new upload journey from the settings page they don’t have much context about what they are doing.

This happens when:
- they don’t have an organisation
- or there is nothing in the pool for their organisation
- and we can’t show any magic options

This commit shows them the preview of the current branding, which also gives them the ‘You should change your branding’ line if they are using GOV.UK.

They should only go into the upload journey once they’ve pressed the continue button.

***

Then the banner page needs to change so that it still makes sense in context. ‘Adding’ a banner doesn’t feel quite right:
- adding it to what
- and in addition to what?

This commit rewords the page to be about how the branding should appear, which feels better.

I’ve swapped the order of the radios because yes/no is more intuitive than no/yes, even if no will be the more likely option.

Words by @karlchillmaid 

Before | After
---|---
image | image
— | image
image | image
image | image

### DIFF
--- a/app/assets/stylesheets/components/radios-with-images.scss
+++ b/app/assets/stylesheets/components/radios-with-images.scss
@@ -1,5 +1,6 @@
 .notify-radios-with-images {
   display: flex;
+  margin-top: govuk-spacing(3);
 }
 
 .notify-radios-with-images__item {
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2113,33 +2113,33 @@ class EmailBrandingChooseLogoForm(StripWhitespaceForm):
 
 class EmailBrandingChooseBanner(Form):
     BANNER_CHOICES_DATA = {
-        "org": {
-            "label": "No banner",
-            "image": {
-                "url": asset_fingerprinter.get_url("images/branding/org.png"),
-                "alt_text": 'An example of an email with the heading "Your logo" in blue text on a white background.',
-                "dimensions": {"width": 404, "height": 454},
-            },
-        },
         "org_banner": {
-            "label": "Coloured banner",
+            "label": "Yes",
             "image": {
                 "url": asset_fingerprinter.get_url("images/branding/org_banner.png"),
                 "alt_text": "An example of an email with a logo on a blue banner.",
                 "dimensions": {"width": 404, "height": 454},
             },
         },
+        "org": {
+            "label": "No",
+            "image": {
+                "url": asset_fingerprinter.get_url("images/branding/org.png"),
+                "alt_text": 'An example of an email with the heading "Your logo" in blue text on a white background.',
+                "dimensions": {"width": 404, "height": 454},
+            },
+        },
     }
 
     banner = GovukRadiosWithImagesField(
-        "Add a banner to your logo",
+        "Does your logo appear on a coloured background?",
         choices=tuple((key, value["label"]) for key, value in BANNER_CHOICES_DATA.items()),
         image_data={key: value["image"] for key, value in BANNER_CHOICES_DATA.items()},
     )
 
 
 class EmailBrandingChooseBannerColour(StripWhitespaceForm):
-    hex_colour = HexColourCodeField("Choose a colour for your banner", validators=[DataRequired()])
+    hex_colour = HexColourCodeField("Choose a background colour", validators=[DataRequired()])
 
 
 class EmailBrandingAltTextForm(StripWhitespaceForm):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2125,7 +2125,7 @@ class EmailBrandingChooseBanner(Form):
             "label": "No",
             "image": {
                 "url": asset_fingerprinter.get_url("images/branding/org.png"),
-                "alt_text": 'An example of an email with the heading "Your logo" in blue text on a white background.',
+                "alt_text": "An example of an email with a logo on a clear background.",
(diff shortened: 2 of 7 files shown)


karlchillmaid (changes requested):
Content looks good.

Just one change please. Can we make the alt text for the no-banner image something like:

An example of an email with a logo on a transparent/clear/white background.

quis:
@karlchillmaid nice, changed in d5f3d183d90aca83752ac8eb41a35eff458bd970

karlchillmaid (approved):

samuelhwilliams:
> This commit shows them the preview of the current branding, which also gives them the ‘You should change your branding’ line if they are using GOV.UK.

Has this line since been removed?

quis:
@samuelhwilliams it’s there but it’s conditional so you won’t always see it:

https://github.com/alphagov/notifications-admin/blob/afd63706f0f8e38dc414c1912eeab415a9174f74/app/templates/views/service-settings/branding/email-branding-options.html#L32-L36

samuelhwilliams:
@quis interesting... I removed the organisation from a service locally but I'm not seeing that content. Inside `current_service.needs_to_change_email_branding` I'm getting `self.organisation == Organisation(<None>)` which has a type of `central`. Possibly something to do with how I hacked that service together?

---

So, this gets a bit sticky with all of the different ways through the possible pages. This change breaks the back link from the "Does your logo appear on a coloured background?" page - but only when travelling back from beyond it (eg the "Choose a background colour") page. In that case, you get taken to the "Choose a logo for your emails" which you don't see on the way forward.

If we _want_ to fix this super edge case, we could apply a patch like the one below. Will need to update/add some tests around the flow as well.

```diff
diff --git a/app/main/views/service_settings/index.py b/app/main/views/service_settings/index.py
index 5b16a8336..bd9cc2919 100644
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -1201,6 +1201,7 @@ def email_branding_request(service_id):
                 service_id=current_service.id,
                 back_link=".email_branding_request",
                 branding_choice="something_else",
+                only_something_else=1,
             )
         )
 
@@ -1524,6 +1525,7 @@ def email_branding_set_alt_text(service_id):
     if form.validate_on_submit():
         # we use this key to keep track of user choices through the journey but we don't use it to save the branding
         branding_choice = email_branding_data.pop("branding_choice")
+        email_branding_data.pop("only_something_else", None)
 
         new_email_branding = EmailBranding.create(
             alt_text=form.alt_text.data,
@@ -1578,7 +1580,10 @@ def _email_branding_flow_query_params(request, **kwargs):
 
     These values can get passed to the `/_email` endpoint to generate a preview of a new brand.
     """
-    return {k: kwargs.get(k, request.args.get(k)) for k in ("brand_type", "branding_choice", "colour", "logo")}
+    return {
+        k: kwargs.get(k, request.args.get(k))
+        for k in ("brand_type", "branding_choice", "colour", "logo", "only_something_else")
+    }
 
 
 @main.route("/services/<uuid:service_id>/service-settings/email-branding/add-banner", methods=["GET", "POST"])
@@ -1606,16 +1611,24 @@ def email_branding_choose_banner_type(service_id):
             )
 
     org_type = current_service.organisation_type
-    back_link_fallback = ".email_branding_choose_logo" if org_type == "central" else ".service_settings"
+
+    if request.args.get("only_something_else"):
+        back_view = ".email_branding_request"
+    else:
+        back_view_fallback = ".email_branding_choose_logo" if org_type == "central" else ".service_settings"
+        back_view = request.args.get("back_link", back_view_fallback)
+
+    back_link = url_for(
+        back_view,
+        service_id=current_service.id,
+        **_email_branding_flow_query_params(request),
+    )
+
     return (
         render_template(
             "views/service-settings/branding/add-new-branding/email-branding-choose-banner.html",
             form=form,
-            back_link=url_for(
-                request.args.get("back_link", back_link_fallback),
-                service_id=current_service.id,
-                **_email_branding_flow_query_params(request),
-            ),
+            back_link=back_link,
         ),
         400 if form.errors else 200,
     )
diff --git a/tests/app/main/views/service_settings/test_email_branding_requests.py b/tests/app/main/views/service_settings/test_email_branding_requests.py
index 37b1e9533..4fc96b3ef 100644
--- a/tests/app/main/views/service_settings/test_email_branding_requests.py
+++ b/tests/app/main/views/service_settings/test_email_branding_requests.py
@@ -249,6 +249,7 @@ def test_email_branding_request_page_redirects_to_choose_banner_type_page_if_som
             service_id=SERVICE_ONE_ID,
             back_link=".email_branding_request",
             branding_choice="something_else",
+            only_something_else=1,
         ),
     )
```

CrystalPea (commented):
Default back link for `email_branding_something_else` view could be simplified now.

quis:
@samuelhwilliams you’re probably not seeing that line because `Service.organisation_type` is set to `central`. `Organisation.organisation_type` will take precedence if the service has an organisation, per: https://github.com/alphagov/notifications-admin/blob/41e211d365c66dffc269f95b61733e937b7488d4/app/models/service.py#L477-L479

quis:
@samuelhwilliams I think I’ve figured out the back links in 6932638

samuelhwilliams (approved):
nice 👍

PR #4544, opened 2022-12-30
Add NewRelic to dev+preview

samuelhwilliams:
We add basic APM instrumentation for NewRelic. This will inject tracing
headers that filter through to any other instrumented apps and send them
to NewRelic so that we can see how requests travel through our services
and how long each span takes. This should help us highlight slower areas
of the codebase which we can then investigate and optimise.

We only send APM information - not logs - here, and we heavily limit the
information sent so that we hopefully aren't gathering much/any PII. We
also only deploy to dev and preview at the moment, so there's no risk of
this code gathering real user data for now. This will allow us to safely
investigate the tool and see whether it's something useful for us.

Required env vars:
`NEW_RELIC_LICENCE_KEY`

### DIFF
--- a/application.py
+++ b/application.py
@@ -1,3 +1,13 @@
+import os
+
+if (environment := os.getenv("NOTIFY_ENVIRONMENT")) in {"development", "preview"} and os.getenv(
+    "NEW_RELIC_ENABLED"
+) == "1":
+    import newrelic.agent
+
+    # Expects NEW_RELIC_LICENSE_KEY set in environment as well.
+    newrelic.agent.initialize("newrelic.ini", environment=environment, ignore_errors=False)
+
 from flask import Flask
 
 from app import create_app
--- a/newrelic.ini
+++ b/newrelic.ini
@@ -1,0 +1,220 @@
+# ---------------------------------------------------------------------------
+
+#
+# This file configures the New Relic Python Agent.
+#
+# The path to the configuration file should be supplied to the function
+# newrelic.agent.initialize() when the agent is being initialized.
+#
+# The configuration file follows a structure similar to what you would
+# find for Microsoft Windows INI files. For further information on the
+# configuration file format see the Python ConfigParser documentation at:
+#
+#    http://docs.python.org/library/configparser.html
+#
+# For further discussion on the behaviour of the Python agent that can
+# be configured via this configuration file see:
+#
+#    http://newrelic.com/docs/python/python-agent-configuration
+#
+
+# ---------------------------------------------------------------------------
+
+# Here are the settings that are common to all environments.
+
+[newrelic]
+
+# You must specify the license key associated with your New
+# Relic account. This key binds the Python Agent's data to your
+# account in the New Relic service.
+# license_key = <provided in environment via NEW_RELIC_LICENSE_KEY>
+
+# The application name. Set this to be the name of your
+# application as you would like it to show up in New Relic UI.
+# The UI will then auto-map instances of your application into a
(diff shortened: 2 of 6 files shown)


newrelic.ini
```diff
@@ -0,0 +1,220 @@
...
+app_name = local.notify-admin
+monitor_mode = true
+
+[newrelic:test]
+app_name = test.notify-admin
+monitor_mode = false
+
+[newrelic:preview]
+app_name = preview.notify-admin
+monitor_mode = false
```
leohemsted, line 210: should this be true?

```suggestion
monitor_mode = true
```

application.py
```diff
@@ -1,3 +1,10 @@
+import os
+
+if os.getenv("NOTIFY_ENVIRONMENT") in {"development", "local", "preview"}:
```
leohemsted, line 3: hmmm i'd rather newrelic is always imported and we rely on its config to enable/disable it per env ideally to keep this file simple, but appreciate that this is only a prototype and so i'm fine with this for now

application.py
```diff
@@ -1,3 +1,10 @@
...
+    import newrelic.agent
+
+    newrelic.agent.initialize("newrelic.ini")
```
leohemsted, line 6: i think we need to pass the environment in here to appropriately enable/disable the tracing?

```suggestion
    newrelic.agent.initialize("newrelic.ini", os.getenv("NOTIFY_ENVIRONMENT"))
```

newrelic.ini
```diff
@@ -0,0 +1,220 @@
...
+#
+# The application environments. These are specific settings which
+# override the common environment settings. The settings related to a
+# specific environment will be used when the environment argument to the
+# newrelic.agent.initialize() function has been defined to be either
+# "development", "test", "staging" or "production".
+#
+
+[newrelic:local]
+app_name = local.notify-admin
```
leohemsted, line 201: i think this should be development? i don't think we use local anywhere
```suggestion
[newrelic:development]
app_name = development.notify-admin
```

leohemsted (commented):
not sure about a couple of bits

newrelic.ini
samuelhwilliams: No, I want it off for everything by default - then we enable it in dev and preview at the bottom

application.py
samuelhwilliams: I think when we productionise this yes, but for now I didn't want to affect prod codepaths at all.

application.py
samuelhwilliams: I was choosing to use the `NEW_RELIC_x` env vars for now, eg `NEW_RELIC_ENVIRONMENT`.

newrelic.ini
samuelhwilliams: Ah sure, we can use development 👍

newrelic.ini
leohemsted: this is at the bottom?

samuelhwilliams: Oh sorry 😬

application.py
samuelhwilliams: 1) No, it doesn't crash if you omit the LICENSE_KEY or most settings. Config errors we can hard-fail on, we will (which I had done through a config var but will now hardcode).
2) It will initialise with some default values, which is why I want to have the default monitor_mode as off so that the risk of anything happening if we're misconfigured is lower.
3) Have ripped out all env vars except the license key.

leohemsted (approved):
happy to leave this collecting data for a bit. keen to see what it could show us!

PR #4546, opened 2022-12-30
Link to user profile for platform admins

samuelhwilliams:
On the team members page of a service, add a link on the name to take platform admins to the user's profile page so that we can jump to their other services or see other profile information.

By keeping it the same black colour I'm hoping to make it relatively inconspicuous - available but not distracting. Another option might be to have a 'View profile' link beneath the 'Change details' link, but that felt crowded.

image

### DIFF
--- a/app/templates/views/manage-users.html
+++ b/app/templates/views/manage-users.html
@@ -25,28 +25,28 @@
         <div class="govuk-grid-row">
           <div class="govuk-grid-column-three-quarters">
             <h2 class="user-list-item-heading" title="{{ user.email_address }}">
-              {% if current_user.platform_admin and not user.is_invited_user %}
-              <a
-                class="govuk-link govuk-link--text-colour"
-                href="{{ url_for('.user_information', user_id=user.id) }}">
+              {%- if user.name -%}
+                <span class="heading-small live-search-relevant">
+                  {% if current_user.platform_admin and not user.is_invited_user %}
+                    <a class="govuk-link govuk-link--text-colour" href="{{ url_for('.user_information', user_id=user.id) }}">
+                      {{ user.name }}
+                    </a>
+                  {% else %}
+                    {{ user.name }}
+                  {% endif %}
+                </span>&ensp;
+              {%- endif -%}
+              <span class="hint">
+              {%- if user.status == 'pending' -%}
+                <span class="live-search-relevant">{{ user.email_address }}</span> (invited)
+              {%- elif user.status == 'cancelled' -%}
+                <span class="live-search-relevant">{{ user.email_address }}</span> (cancelled invite)
+              {%- elif user.id == current_user.id -%}
+                <span class="live-search-relevant">(you)</span>
+              {% else %}
+                <span class="live-search-relevant">{{ user.email_address }}</span>
               {% endif %}
-                {%- if user.name -%}
-                <span class="heading-small live-search-relevant">{{ user.name }}</span>&ensp;
-                {%- endif -%}
-                <span class="hint">
-                {%- if user.status == 'pending' -%}
-                  <span class="live-search-relevant">{{ user.email_address }}</span> (invited)
-                {%- elif user.status == 'cancelled' -%}
-                  <span class="live-search-relevant">{{ user.email_address }}</span> (cancelled invite)
-                {%- elif user.id == current_user.id -%}
-                  <span class="live-search-relevant">(you)</span>
-                {% else %}
-                  <span class="live-search-relevant">{{ user.email_address }}</span>
-                {% endif %}
-                </span>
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
@tombye is it weird to have this 'link' in black text? Should we just make it blue and normal?

leohemsted:
i'm not sure about this - can you show screenshots of what it looks like for other users (also includes an email address) and invited users (greyed out)

app/templates/views/manage-users.html
```diff
@@ -25,20 +25,28 @@ <h1 class="heading-medium">
...
-              {%- elif user.status == 'cancelled' -%}
-                <span class="live-search-relevant">{{ user.email_address }}</span> (cancelled invite)
-              {%- elif user.id == current_user.id -%}
-                <span class="live-search-relevant">(you)</span>
-              {% else %}
-                <span class="live-search-relevant">{{ user.email_address }}</span>
+              {% if current_user.platform_admin %}
+              <a
+                class="govuk-link govuk-link--text-colour"
+                href="{{ url_for('.user_information', user_id=user.id) }}">
```
leohemsted, line 31: also this won't work for invited users i believe (as `InvitedUser.id` points to a different database table). i'm not sure if it'll work if that email is already part of notify 🤔.

tests/app/main/views/test_manage_users.py
```diff
@@ -433,6 +433,24 @@ def test_manage_users_page_shows_member_auth_type_if_service_has_email_auth_acti
...
+    mock_get_invites_for_service,
+    mock_get_template_folders,
+    user_is_platform_admin,
+):
+    active_user_with_permissions["platform_admin"] = user_is_platform_admin
+    client_request.login(active_user_with_permissions)
+    page = client_request.get("main.manage_users", service_id=service_one["id"])
+    entry = page.select_one("h2.user-list-item-heading")
+
+    assert (entry.select_one("a") is not None) == user_is_platform_admin
```
leohemsted, line 451: i think this test should be a bit more robust. we should be checking the href location too. might be simpler to split it in to two tests in that case

app/templates/views/manage-users.html
samuelhwilliams: 🤔 removed the link for invited users.

app/templates/views/manage-users.html
```diff
@@ -25,20 +25,28 @@ <h1 class="heading-medium">
...
-              <span class="hint">
-              {%- if user.status == 'pending' -%}
-                <span class="live-search-relevant">{{ user.email_address }}</span> (invited)
-              {%- elif user.status == 'cancelled' -%}
-                <span class="live-search-relevant">{{ user.email_address }}</span> (cancelled invite)
-              {%- elif user.id == current_user.id -%}
-                <span class="live-search-relevant">(you)</span>
-              {% else %}
-                <span class="live-search-relevant">{{ user.email_address }}</span>
+              {% if current_user.platform_admin and user.status not in ['pending', 'cancelled'] %}
```
leohemsted, line 28: i _think_ this can be `and user.is_active` (and the same on the closing tag block on line 47) ofc

samuelhwilliams: Are you telling me to change this or just saying it _could_ be that?

samuelhwilliams: (I chose this because it's more closely related to the logic we use elsewhere in the template to 'detect' invited users)

leohemsted: ah so actually i didn't realise something sneaky about this code that i don't love.

the checks above look at `user.status` which is only present on invited users. this can be `accepted`, `pending` or `cancelled` - but we'll never get `accepted` here as we just show pending invites. if it's an accepted invite we'll just have their actual user object rather than the old invite object.

if they're a real user, they have `user.state` instead, which can be `inactive`, `active`, or `pending`. I don't think pending's relevant.

Either way, we're interested in "is this user a `User` or an `InvitedUser`", we don't really care what the state/status is beyond that. We could technically say `if user.state is not None` or something like that, but I don't love that - it might be worth putting some properties on the BaseUser class perhaps?

```python
class BaseUser:
    @property
    def is_invited_user(self):
        return self.__class__ == InvitedUser
    ....
```

quis:
This is OK design-wise. I don’t love when we do things that make the page look different for platform admin users, but I can’t think of a more elegant way to do this. Black link text should be alright because the underline signifies a link to any visual user – blue text alone isn’t sufficient. Might look better if it was just the name (not name and email address) which made up the link.

samuelhwilliams:
@quis have done that in c74a7db73. Small edge case is that I suppose if a user doesn't have a name (somehow?) then we wouldn't be able to click them - but I would've hoped that should not be able to happen. The template suggests it's possible though.

PR #4547, opened 2022-12-30
Show notification processor to platform admins

samuelhwilliams:
Adds `via <provider>` to the delivery status section - only visible to platform admins.

### DIFF
--- a/app/templates/partials/notifications/status.html
+++ b/app/templates/partials/notifications/status.html
@@ -11,6 +11,9 @@
     {% if status_url %}
       </a>
     {% endif %}
+    {% if current_user.platform_admin and notification.sent_by %}
+      via {{ notification.sent_by | format_provider }}
+    {% endif %}
     {% if sent_with_test_key %}
       (test)
     {% endif %}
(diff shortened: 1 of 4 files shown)


leohemsted (approved):

app/templates/partials/notifications/status.html
```diff
@@ -11,6 +11,9 @@
     {% if status_url %}
       </a>
     {% endif %}
+    {% if current_user.platform_admin and notification.sent_by %}
+      via {{ notification.sent_by | format_provider }}
```
idavidmcdonald, line 15: Optional: Channeling my inner @karlchillmaid but I feel like we could use 'by' rather than 'via' here to use a slightly plainer english term. But this is maybe just personal preference

karlchillmaid: I know this is merged now, but I agree with David.

samuelhwilliams: https://github.com/alphagov/notifications-admin/pull/4550

PR #4550, opened 2023-01-04
By not via

samuelhwilliams:


idavidmcdonald (approved):
Thanks Sam

karlchillmaid:
`by` works fine for `delivered` and `not delivered`, but doesn’t quite sound right for `sending`.

It also doesn’t sound right for `Failed`, `Technical failure` and `Phone not accepting messages right now`. I’m not sure the original `via` is better, though.

What are we trying to communicate here, and to whom?

There’s a risk that it sounds like we’re saying messages failed _because of_ the provider, not that the provider tried to send them, but could not.

samuelhwilliams:
Something to bear in mind: only platform admins will see the 'via/by/whatever <provider>'. It's meant to be so _we_ can quickly see what provider is the issue and maybe tie it together with support requests. We're not naming providers to our users.

karlchillmaid (approved):
Thanks for adding the rationale, @samuelhwilliams.

PR #4551, opened 2023-01-04
Fix the HTML for lists of templates

klssmith:
**Remove template lists from `<nav>`s**

We removed the `<nav>` around template list items on the templates page, but there are a couple of other places where we show a template list that still needed updating - the copy template page and the page to choose a template to reply to a received text message with. Changing the tag used doesn't affect the appearance of the pages.

**Stop using list tags for templates**

This replaces list tags with navs when displaying a list of templates in the following places:
- The main templates page when the user doesn't have `manage_templates` permission
- The pages to copy a template
- The page to pick a template to use to reply to a received text message

This should keep the pages looking the same, though there were a couple of places where we were had forgotten to add the `govuk-list` class to the list of templates so makes the templates display as expected (without visible bullet points).

---

**Templates list before (the bullet points should not have been visible)**

Screenshot 2023-01-04 at 14 27 38

**Templates list after**

Screenshot 2023-01-04 at 14 29 26

### DIFF
--- a/app/templates/views/templates/_template_list.html
+++ b/app/templates/views/templates/_template_list.html
@@ -24,10 +24,6 @@
   <div id="template-list" class="{{ 'govuk-!-margin-top-1' if (not show_template_nav and not show_search_box) else 'govuk-!-margin-top-6' }}">
     {% set checkboxes_data = [] %}
 
-    {% if not current_user.has_permissions('manage_templates') %}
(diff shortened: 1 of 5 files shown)


tombye (approved):
Looks good to me 👍🏻 and changes a few pages the card didn't mention 🚀

PR #4552, opened 2023-01-04
Let organisation users see all template folders

quis:
We already let organisation users see all templates in their service: https://github.com/alphagov/notifications-admin/pull/4494

However if the template is in a folder then the user won’t be able to see it, since the user also needs permission to see the folder in order to see the template.

This defeats the objective of letting organisation users see all templates.

This commit effectively gives organisation users permission to see all folders in a service, and by extension all templates no matter which folder they are in.

Before | After
---|---
image | image

***

There’s a small amount of refactoring which is in a second commit.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -882,10 +882,22 @@ class BasePermissionsForm(StripWhitespaceForm):
         return set(self.permissions_field.data)
 
     @classmethod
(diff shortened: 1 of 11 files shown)


samuelhwilliams (approved):
Tested locally, works 👍 

Nice bit of refactoring.

PR #4555, opened 2023-01-04
allow users of archived services to create new services

leohemsted:
Remove check_inactive_service
---------

we added this as a safeguard to prevent people modifying inactive services. However, when a user hits an endpoint that is associated with a service we decorate that service with the `user_has_permissions` decorator.

Among the various checks the `user_has_permissions` decorator does is `user.belongs_to_service` - this checks in the list of services returned from the api in the get_user_by_id blob. In the API we explicitly only return services that are active in that list[^1]. It also later checks that any permissions match (if that endpoint requires specific permissions to hit for that service). Again, the API explicitly filters to only return permissions for active services when returning the user[^2].

Note that platform admins bypass both these checks (unless `restrict_admin_usage` is set).

This means that there's no way a normal user can ever enter the view code at all with an inactive service, be it for a GET or a POST. The only risks are if we don't clear the user cache when updating a service - however we explicitly clear user cache when archiving a service[^3].

This means we can remove the check_inactive_service, as we're confident this code is never tripped legitimately. The only time it can ever be reached is if the user is in an endpoint that has a different decorator (specifically `@user_is_logged_in`) so doesn't look at service permissions etc. Previously we were blocking POST updates on pages with these decorators, and causing bugs where users couldn't create new services or update their name/password/etc if they'd recently viewed an archived service, but now we have fixed that bug properly.

[^1]: https://github.com/alphagov/notifications-api/blob/ef555db14afa96ba0122341154b00d14edf9a98d/app/models.py#L179
[^2]: https://github.com/alphagov/notifications-api/blob/ef555db14afa96ba0122341154b00d14edf9a98d/app/dao/permissions_dao.py#L65-L71
[^3]: https://github.com/alphagov/notifications-admin/blob/b6a16bebea65ca59360261b1b1ee8aab58e54897/app/main/views/service_settings/index.py#L382-L386

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -232,15 +232,20 @@ class User(BaseUser, UserMixin):
         if org_id:
             return self.belongs_to_organisation(org_id)
 
+        # if theres no org id, we're dealing with a service endpoint. The service has already been set on
+        # current_service from `load_service_before_request` which happens before this decorator
+        from app import current_service
+
+        if not current_service.active:
+            return False
+
         if not permissions and self.belongs_to_service(service_id):
             return True
 
-        if any(self.permissions_for_service(service_id) & set(permissions)):
+        if self.permissions_for_service(service_id) & set(permissions):
             return True
 
-        from app.models.service import Service
-
-        return allow_org_user and self.belongs_to_organisation(Service.from_id(service_id).organisation_id)
+        return allow_org_user and self.belongs_to_organisation(current_service.organisation_id)
 
     def permissions_for_service(self, service_id):
         return self._permissions.get(service_id, set())
--- a/app/notify_client/__init__.py
+++ b/app/notify_client/__init__.py
@@ -1,4 +1,4 @@
-from flask import abort, has_request_context, request
+from flask import has_request_context, request
 from flask_login import current_user
 from notifications_python_client import __version__
 from notifications_python_client.base import BaseAPIClient
@@ -39,28 +39,6 @@ class NotifyAdminAPIClient(BaseAPIClient):
         headers["X-B3-TraceId"] = request.request_id
         headers["X-B3-SpanId"] = request.span_id
         return headers
-
-    def check_inactive_service(self):
-        # this file is imported in app/__init__.py before current_service is initialised, so need to import later
-        # to prevent cyclical imports
-        from app import current_service
-
-        # if the current service is inactive and the user isn't a platform admin, we should block them from making any
-        # stateful modifications to that service
-        if current_service and not current_service.active and not current_user.platform_admin:
-            abort(403)
-
-    def post(self, *args, **kwargs):
-        self.check_inactive_service()
-        return super().post(*args, **kwargs)
-
-    def put(self, *args, **kwargs):
-        self.check_inactive_service()
-        return super().put(*args, **kwargs)
-
-    def delete(self, *args, **kwargs):
-        self.check_inactive_service()
-        return super().delete(*args, **kwargs)
 
 
 class InviteTokenError(Exception):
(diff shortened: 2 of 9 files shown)


app/notify_client/__init__.py
```diff
@@ -41,25 +41,42 @@ def _add_request_id_header(headers):
...
+        """
         # this file is imported in app/__init__.py before current_service is initialised, so need to import later
         # to prevent cyclical imports
         from app import current_service
 
-        # if the current service is inactive and the user isn't a platform admin, we should block them from making any
-        # stateful modifications to that service
         if current_service and not current_service.active and not current_user.platform_admin:
+            current_app.logger.warning(
+                f"{current_user.id} tried to perform update on inactive service {current_service.id}"
```
leohemsted, line 63: annoyingly we can't see what api method we tried to call to get here without doing gross traceback introspection or similar

app/main/views/add_service.py
```diff
@@ -68,8 +68,12 @@ def add_service():
         )
         if error:
             return _render_add_service_page(form, default_organisation_type)
-        if len(service_api_client.get_active_services({"user_id": session["user_id"]}).get("data", [])) > 1:
 
+        # set that as current service immediately or when we create the example template we'll still be checking
+        # the user's previous service to see if that is inactive.
+        g.current_service = Service.from_id(service_id)
```
leohemsted, line 74: i really cannot think of a viable way to thoroughly test this in a unit test.

the problem:

two calls below (new_service.force_permission and _create_example_template) call `check_inactive_service`, as they should! you wouldn't want someone to do either of those on an archived service.

the quick solution:

update the session to reflect that as soon as the service has been created without errors, we're now interested in this service, not wherever the user was previously

quis:
Note also that it’s not possible to use `@user_is_logged_in` on a service-or-organisation-specific endpoint because of
https://github.com/alphagov/notifications-admin/blob/b6a16bebea65ca59360261b1b1ee8aab58e54897/tests/app/main/test_permissions.py#L248-L257

tests/app/main/test_permissions.py
```diff
@@ -92,6 +92,18 @@ def test_services_pages_that_org_users_are_allowed_to_see(
     assert mock_get_service.called is organisation_checked
 
 
+# this in effect asserts that a user cannot modify an inactive service,
+# since the API returns empty permissions for an inactive service
+def test_user_with_empty_permissions_cannot_do_anything(
+    client_request,
+    api_user_active,
+):
+    api_user_active["permissions"] = {SERVICE_ONE_ID: []}
```
quis, line 101: Is there not situation where a real user will have empty permissions? For example if none of the permission checkboxes are ticked when inviting someone?

leohemsted: yes you're right - and org users would also always get through the `user_has_permissions(allow_org_user=True)` check even if the service was inactive ,so i added an explicit check in https://github.com/alphagov/notifications-admin/pull/4555/commits/849c9fa186aa0686145f00bafebb1f5b711c856b

app/models/user.py
```diff
@@ -212,6 +212,8 @@ def platform_admin(self):
         return self._platform_admin and not session.get("disable_platform_admin_view", False)
 
     def has_permissions(self, *permissions, restrict_admin_usage=False, allow_org_user=False):
+        from app.models.service import Service
```
quis, line 215: Think you can do
```suggestion
        from app import current_service
```

…then we are only hitting the API/Redis once per request context

leohemsted: good shout. needs me to fix a couple of tests but worth doing

quis (approved):

PR #4558, opened 2023-01-05
Show whichever set of daily limits is the applicable one

quis:
Assumes that `email_message_limit`, `sms_message_limit`, `letter_message_limit` will all be reduced from 999,999,999 in one go

### DIFF
--- a/app/templates/views/service-settings.html
+++ b/app/templates/views/service-settings.html
@@ -292,14 +292,19 @@
       {% else %}
         <h2 class="heading-medium top-gutter-0">Your service is live</h2>
 
-        <p class="govuk-body">
-          You can send up to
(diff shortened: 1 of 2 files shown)


samuelhwilliams (commented):
I'd maybe rather we just do a concrete switch in one go rather than have conditional logic that could show different things for different services (theoretically - don't expect it would happen in reality).

But this should only be short lived so I don't mind it as-is. We'll rip out everything about base message limits as cleanup shortly after release.

quis:
@samuelhwilliams I have:
- added tests for this change
- added https://github.com/alphagov/notifications-admin/pull/4562 as a cleanup PR

samuelhwilliams (approved):

PR #4561, opened 2023-01-06
Add personalisation and optional content guidance

karlchillmaid:
This PR adds new guidance pages for:

- Personalise your message content
- Optional content

It also removes content about personalisation and optional content from the current Formatting page.


app/main/views/index.py
```diff
@@ -276,6 +276,21 @@ def guidance_letter_branding():
...
+    )
+
+
+@main.route("/using-notify/guidance/optional-content")
+def guidance_optional_content():
+    return render_template(
+        "views/guidance/optional-content.html",
+        navigation_links=using_notify_nav(),
+    )
+
```
karlchillmaid, line 293:
```suggestion
@main.route("/using-notify/guidance/optional-content")
def guidance_optional_content():
    return render_template(
        "views/guidance/optional-content.html",
        navigation_links=using_notify_nav(),
    )


@main.route("/using-notify/guidance/personalisation")
def guidance_personalisation():
    return render_template(
        "views/guidance/personalisation.html",
        navigation_links=using_notify_nav(),
    )

```

tests/app/test_navigation.py
```diff
@@ -215,6 +216,7 @@
             "organisation_trial_mode_services",
             "organisations",
             "performance",
+            "personalisation",
```
samuelhwilliams, line 219: Same with this, should be `guidance_personalisation`

app/templates/views/guidance/personalisation.html
```diff
@@ -0,0 +1,53 @@
...
+
+  <h1 class="heading-large">Personalisation</h1>
+
+  <p class="govuk-body">You can send personalised messages using a single template.</p>
+
+  <p class="govuk-body">To personalise a message, use double brackets to add a placeholder to your content. For example:</p>
+
+  <div class="govuk-inset-text">
+    
+    <p class="govuk-body">Hello ((first name)), your reference is ((reference number)).</p>
```
samuelhwilliams, line 22: The content on this page looks good, but I wonder if we want to add highlighting to the personalisation so that it looks the same as it does when editing a template?

Currently the guidance looks like this:
image


And editing templates looks like this:
image


Any thoughts @karlchillmaid  / @quis?

app/main/views/index.py
```diff
@@ -276,6 +276,21 @@ def guidance_letter_branding():
...
+
+@main.route("/using-notify/guidance/personalisation")
+def guidance_personalisation():
+    return render_template(
+        "views/guidance/personalisation.html",
+        navigation_links=using_notify_nav(),
+    )
+
```
samuelhwilliams, line 293: One more tiny bit - we have a test that checks our formatting is consistent, and it expects two blank lines between the end of one function and the start of another. We only have one here.

💅

samuelhwilliams:
And then finally, we need to update `tests/route-list.json` to include the following lines:

```
    "/using-notify/guidance/optional-content",
    "/using-notify/guidance/personalisation",
```

app/main/views/index.py
karlchillmaid:
```suggestion


```

app/templates/views/guidance/optional-content.html
```diff
@@ -0,0 +1,44 @@
+{% extends "content_template.html" %}
+
+{% from "components/service-link.html" import service_link %}
```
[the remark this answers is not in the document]
karlchillmaid:
```suggestion
```

app/templates/views/guidance/personalisation.html
karlchillmaid: I like this suggestion.

How would I add highlighting to the placeholders?

If we do that here, we should consider adding highlighting to the guidance on the edit template page too.

tests/app/test_navigation.py
```diff
@@ -206,6 +206,7 @@
             "old_service_dashboard",
             "old_terms",
             "old_using_notify",
+            "optional content",
```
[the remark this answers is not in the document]
karlchillmaid:
```suggestion
            "guidance_optional_content",
```

tests/app/test_navigation.py
karlchillmaid:
```suggestion
            "guidance_personalisation",
```

tests/app/test_navigation.py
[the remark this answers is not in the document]
karlchillmaid: @samuelhwilliams I should move this (and guidance_personalisation) so they’re grouped with all the other `guidance_` items, right?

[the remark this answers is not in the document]
samuelhwilliams: yes please :)

app/templates/views/guidance/personalisation.html
karlchillmaid: @samuelhwilliams Are you happy if we revisit this later?

One of the next PRs is about changes to the formatting guidance, and this could happen as part of that.

samuelhwilliams: Yes of course

samuelhwilliams (approved):
Looks all good to me now - nicely done 👍

PR #4563, opened 2023-01-09
Add new daily limits to bulk sending guidance

quis:
We removed these temporarily while we migrated from the old, non-channel-specific limit.

I’ve made these dynamic so that if we ever change the values we know to update the guidance.

# Before

image

# After

image

***

Depends on:
- [ ] https://github.com/alphagov/notifications-api/pull/3696


samuelhwilliams (approved):
nice 💯

PR #4564, opened 2023-01-09
Fix error message for per-channel rate limits

samuelhwilliams:
Show the correct limit and message noun on admin pages when we hit one of the new rate limits.

## Before
image

## After
image


app/formatters.py
```diff
@@ -415,6 +415,22 @@ def message_count(count, template_type):
...
+        message_limit = min(message_limit, service.email_message_limit)
+    elif template_type == "sms":
+        message_limit = min(message_limit, service.sms_message_limit)
+    elif template_type == "letter":
+        message_limit = min(message_limit, service.letter_message_limit)
+
+    if message_limit == service.message_limit:
+        template_type = "message"
+
+    return f"{format_thousands(message_limit)} {message_count_noun(message_limit, template_type)}"
```
quis, line 431:
```suggestion
    return message_count(message_limit, template_type)
```

quis (approved):

PR #4567, opened 2023-01-10
Remove setting of template process type

quis:
Originally we had the idea that we could set certain templates (for example two factor codes) as ‘priority’ and try to deliver them more quickly.

We only ever let platform admins decide whether a template was priority.

Since then we discovered that:
- Celery 3 was slower to deliver messages from an often-empty queue (which the priority queue was) so this feature had the opposite effect to what was intended
- Celery 4 is so fast the it obviates the need for a priority queue
- We have so many templates that it’s beyond the capacity of our team to review which ones should or shouldn’t be ‘priority’

Removing the setting from the admin app is a first step towards getting rid of the feature entirely. It’s a useful one to start with because it cleans up the template editing interface for platform admin users.

Before | After 
---|---
image | image


samuelhwilliams:
There's a couple of hanging references to `process_type` that we can probably drop as well - one is `tests.__init__` and one in `service_api_client`

samuelhwilliams:
And `test_should_show_page_for_one_template` has a line `assert "priority" not in str(page.select_one("main"))` that I suspect is redundant now too.

PR #4473, opened 2022-11-25
Set name based on alt-text

leohemsted:
- [x] https://github.com/alphagov/notifications-admin/pull/4472
- [x] https://github.com/alphagov/notifications-api/pull/3651

See [0631d3d](https://github.com/alphagov/notifications-admin/pull/4473/commits/0631d3df53836802b575e4fe0f49627edd21158a). Given an alt text "My Department", checks if that exists as a name in the DB, and if so, returns a new name along the lines of `"My Department (alternate 1)"`.

PR #4572, opened 2023-01-12
Letter branding - enforce unique name and set organisation default

leohemsted:
Bring these in parity with email branding[^1][^2]. There's one slight change in that there's no check to see if the org is central or not. But the rest is the same.

[^1]: #4473
[^2]: #4498

### DIFF
--- a/app/main/views/service_settings/letter_branding.py
+++ b/app/main/views/service_settings/letter_branding.py
@@ -155,6 +155,19 @@ def letter_branding_upload_branding(service_id):
     )
 
 
+def _should_set_default_org_letter_branding(branding_choice):
+    # 1. the user has chosen ‘[organisation name]’ in the first page of the journey
+    user_chose_org_name = branding_choice == "organisation"
+    # 2. and the organisation doesn’t have default branding already
+    org_doesnt_have_default_branding = current_service.organisation.letter_branding_id is None
+    # 3. and the organisation has no other live services
+    no_other_live_services_in_org = not any(
+        service.id != current_service.id for service in current_service.organisation.live_services
+    )
+
+    return user_chose_org_name and org_doesnt_have_default_branding and no_other_live_services_in_org
+
+
 @main.route("/services/<uuid:service_id>/service-settings/letter-branding/set-name", methods=["GET", "POST"])
 def letter_branding_set_name(service_id):
     letter_branding_data = _letter_branding_flow_query_params()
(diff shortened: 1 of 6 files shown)


app/main/views/service_settings/letter_branding.py
```diff
@@ -179,9 +192,8 @@ def letter_branding_set_name(service_id):
             current_service.organisation.id, [new_letter_branding.id]
         )
 
-        # TODO: implement _should_set_default_org_branding for letter branding
-        # if _should_set_default_org_letter_branding(letter_branding_data["branding_choice"]):
-        #     current_service.organisation.update(letter_branding_id=new_letter_branding.id, delete_services_cache=True)
+        if _should_set_default_org_letter_branding(letter_branding_data["branding_choice"]):
+            current_service.organisation.update(letter_branding_id=new_letter_branding.id, delete_services_cache=True)
 
         flash("You’ve changed your letter branding.", "default_with_tick")
```
samuelhwilliams, line 198: Not related to your PR but this caught my eye - do we normally write these in 2nd person? 🤔

samuelhwilliams (approved):
This PR works as expected 👍

When running through locally I noticed that we still ask the user to explain their branding when choosing 'something else' but then don't seem to do anything with that, because we walk them through creating their own. Are we planning to remove that?

image


Couple of comments, mostly just as an aside.

app/main/views/service_settings/letter_branding.py
leohemsted: hmmm... sometimes? 

https://github.com/alphagov/notifications-admin/blob/main/app/main/views/service_settings/index.py#L721

sometimes not?

https://github.com/alphagov/notifications-admin/blob/main/app/main/views/service_settings/index.py#L390

samuelhwilliams: yeah I only found that 1 similar example, all the rest seem passive/to not mention 'you'

leohemsted:
As to the something else box, this is currently tied to the fact this new flow is currently only platform admin, and normal users need somewhere to go - we'll sort this out in: https://trello.com/c/tLSwxmig/165-connect-new-letter-flow

app/main/views/service_settings/letter_branding.py
samuelhwilliams: maybe just something for Karl to be aware of, I think validation messages (which maybe tangentially includes these) is up for a full review at some point soon

samuelhwilliams:
makes sense, cool

app/main/views/service_settings/letter_branding.py
```diff
@@ -167,7 +167,9 @@ def letter_branding_set_name(service_id):
 
     if form.validate_on_submit():
         # TODO: Handle name already existing
```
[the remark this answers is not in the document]
leohemsted: yup, squashed

app/main/views/service_settings/letter_branding.py
leohemsted: hmmm i'm not sure if validation messages covers happy-path stuff (it's probably reasonable to though, as these are still short information snippets about something you've done)

PR #4574, opened 2023-01-13
Archive email branding

CrystalPea:
In this PR:
- new page for viewing an email branding (more detail in the commit)
- let admin users archive a branding

TODO:
- don't show archive link when branding already archived

Screenshots:

Screenshot 2023-01-13 at 17 32 34

Screenshot 2023-01-13 at 17 32 50

Screenshot 2023-01-13 at 17 33 07

### DIFF
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -99,6 +99,12 @@ def platform_admin_update_email_branding(branding_id, logo=None):
     )
 
 
+@main.route("/email-branding/<uuid:branding_id>/archive", methods=["GET", "POST"])
+@user_is_platform_admin
+def platform_admin_archive_email_branding():
+    pass
+
+
 @main.route("/email-branding/create-government-identity/logo", methods=["GET", "POST"])
 @user_is_platform_admin
 def create_email_branding_government_identity_logo():
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -1,7 +1,7 @@
 import os
 from io import BytesIO
 
-from flask import abort, current_app, redirect, render_template, request, url_for
+from flask import abort, current_app, flash, redirect, render_template, request, url_for
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
 
@@ -32,8 +32,19 @@ def email_branding():
     )
 
 
-@main.route("/email-branding/<uuid:branding_id>/edit", methods=["GET", "POST"])
-@main.route("/email-branding/<uuid:branding_id>/edit/<logo>", methods=["GET", "POST"])
+@main.route(
+    "/email-branding/<uuid:branding_id>/edit", methods=["GET", "POST"], endpoint="platform_admin_update_email_branding"
+)
+@main.route(
+    "/email-branding/<uuid:branding_id>/edit/<logo>",
+    methods=["GET", "POST"],
+    endpoint="platform_admin_update_email_branding",
+)
+@main.route(
+    "/email-branding/<uuid:branding_id>/archive",
+    methods=["GET"],
+    endpoint="platform_admin_confirm_archive_email_branding",
+)
 @user_is_platform_admin
 def platform_admin_update_email_branding(branding_id, logo=None):
     email_branding = EmailBranding.from_id(branding_id)
@@ -87,6 +98,9 @@ def platform_admin_update_email_branding(branding_id, logo=None):
         if not form.errors:
             return redirect(url_for(".email_branding", branding_id=branding_id))
 
+    if request.endpoint == "main.platform_admin_confirm_archive_email_branding":
+        flash("Are you sure you want to archive this email branding?", "archive")
+
     return (
         render_template(
             "views/email-branding/manage-branding.html",
@@ -99,7 +113,7 @@ def platform_admin_update_email_branding(branding_id, logo=None):
     )
 
 
-@main.route("/email-branding/<uuid:branding_id>/archive", methods=["GET", "POST"])
+@main.route("/email-branding/<uuid:branding_id>/archive", methods=["POST"])
 @user_is_platform_admin
 def platform_admin_archive_email_branding():
     pass
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -115,8 +115,11 @@ def platform_admin_update_email_branding(branding_id, logo=None):
 
 @main.route("/email-branding/<uuid:branding_id>/archive", methods=["POST"])
 @user_is_platform_admin
-def platform_admin_archive_email_branding():
-    pass
+def platform_admin_archive_email_branding(branding_id):
+    # TODO: if branding used by active services, don't archive it.
+
+    email_branding_client.archive_email_branding(branding_id=branding_id)
+    return redirect(url_for(".email_branding"))
 
 
 @main.route("/email-branding/create-government-identity/logo", methods=["GET", "POST"])
--- a/app/main/views/email_branding.py
+++ b/app/main/views/email_branding.py
@@ -33,18 +33,33 @@ def email_branding():
 
 
 @main.route(
-    "/email-branding/<uuid:branding_id>/edit", methods=["GET", "POST"], endpoint="platform_admin_update_email_branding"
-)
-@main.route(
-    "/email-branding/<uuid:branding_id>/edit/<logo>",
-    methods=["GET", "POST"],
-    endpoint="platform_admin_update_email_branding",
+    "/email-branding/<uuid:branding_id>", methods=["GET", "POST"], endpoint="platform_admin_view_email_branding"
 )
 @main.route(
     "/email-branding/<uuid:branding_id>/archive",
     methods=["GET"],
     endpoint="platform_admin_confirm_archive_email_branding",
 )
+def platform_admin_view_email_branding(branding_id):
+    email_branding = EmailBranding.from_id(branding_id)
+
+    if request.endpoint == "main.platform_admin_confirm_archive_email_branding":
+        if email_branding.is_used_by_orgs_or_services:
+            flash("This email branding is in use. You cannot delete it.")
+        else:
+            flash("Are you sure you want to delete this email branding?", "delete")
+
+    return render_template(
+        "views/email-branding/view-branding.html",
+        email_branding=email_branding,
+        cdn_url=current_app.config["LOGO_CDN_DOMAIN"],
+        branding_orgs=email_branding.organisations,
+        branding_services=email_branding.services,
+    )
+
+
+@main.route("/email-branding/<uuid:branding_id>/edit", methods=["GET", "POST"])
+@main.route("/email-branding/<uuid:branding_id>/edit/<logo>", methods=["GET", "POST"])
 @user_is_platform_admin
 def platform_admin_update_email_branding(branding_id, logo=None):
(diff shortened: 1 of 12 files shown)


app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,77 @@
...
+				<a class="govuk-link"
+					href="{{ url_for('.platform_admin_update_email_branding', branding_id=email_branding.id) }}">
+					Edit this branding</a>
+			</span>
+			&emsp;
+			<!-- Should not be shown if email branding already archived -->
+			<span class="page-footer-link page-footer-delete-link-without-button bottom-gutter-2-3">
+				<a class="govuk-link govuk-link--destructive"
+					href="{{ url_for('.platform_admin_confirm_archive_email_branding', branding_id=email_branding.id) }}">
+					Archive this branding</a>
```
samuelhwilliams, line 71: Are we sure we want to use the "archive" word here? Everywhere else it looks like we use "Delete" 🤔

I do sort of prefer "archive" - because that's what we're actually doing - but it is inconsistent.

app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,77 @@
...
+
+{% block platform_admin_content %}
+
+<div class="govuk-grid-row bottom-gutter">
+	<div class="govuk-grid-column-full">
+		{{ page_header(email_branding.name) }}
+
+		{{ email_branding_preview(email_branding.id) }}
+
+		<h2 class="heading-medium">Organisations using this branding as their default</h2>
```
leohemsted, line 25: is it useful to show "Organistions that have this branding in their pool" or similar?

Otherwise a branding that has been added, but isn't linked to an org default or currently active for a service, won't show any links on this page

app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,77 @@
...
+{% endblock %}
+
+{% block platform_admin_content %}
+
+<div class="govuk-grid-row bottom-gutter">
+	<div class="govuk-grid-column-full">
+		{{ page_header(email_branding.name) }}
+
+		{{ email_branding_preview(email_branding.id) }}
+
```
leohemsted, line 24: other ideas for potentially useful content on this page

`Created by {{ email_branding.created_by.email_address }} on {{ email_branding.created_at | format_date }}`

`Last updated at {{ email_branding.updated_at | format date }}`

Feel free to add one/both/neither/other!

app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,77 @@
...
+					<a class="govuk-link govuk-link--no-visited-state browse-list-hint" href="{{ url_for('.service_settings', service_id=service.id) }}">
+						{{ service.name }}
+					</a>
+				</li>
+				{% endfor %}
+			</ul>
+			{% else %}
+			<p class="hint">No services use this branding.</p>
+			{% endif %}
+		</nav>
```
leohemsted, line 58: it;s quite hard visually to distinguish the end of this list from the sticky footer since they're both just lines with links on, and the Edit this branding and service links are styled the same.

Could we add, perhaps, a `hr`? Stealing from the org branding pool page

```html
<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible govuk-!-margin-bottom-4" />
```

(happy to take other ideas, this was just the first page i found that separates this stuff nicely)

Here's a mockup:
image

app/main/views/email_branding.py
```diff
@@ -104,6 +130,13 @@ def platform_admin_update_email_branding(branding_id, logo=None):
     )
```
leohemsted, line 130: https://github.com/alphagov/notifications-admin/pull/4574/files#diff-b838f01bb7b2e3334c72bd5389e18ffe44871f9cd3beefbb227303b00a97892cR119

i cant comment on the right line (line 119) but when an admin updates a logo, it would be nice to redirect back to the summary page rather than just the platform admin list now

leohemsted (commented):
Had a play around with this locally, a few comments on some bits or pieces but im really positive about having a nice summary page for each branding.

it looks like it'll be relatively straightforward to make this view letters-friendly too 👍

app/main/views/email_branding.py
CrystalPea: Good idea, will do 👍🏼

app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,77 @@
...
+
+
+{% block backLink %}
+{{ govukBackLink({ "href": url_for('.email_branding') }) }}
+{% endblock %}
+
+{% block platform_admin_content %}
+
+<div class="govuk-grid-row bottom-gutter">
+	<div class="govuk-grid-column-full">
```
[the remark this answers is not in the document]
CrystalPea: ok, I can change them to spaces - didn't realise we had a preference as a team.

app/templates/views/email-branding/view-branding.html
CrystalPea: Nice ones! Will add.

app/templates/views/email-branding/view-branding.html
CrystalPea: I was thinking about it. It wouldn't be excessively hard to add: adding it to the query, and adding a section for it here.

I just wonder what the use case might be? Since we can still archive a branding if it's in a pool, but not actively used by any org / service. Maybe just audit? What do you think?

app/templates/views/email-branding/view-branding.html
CrystalPea: Ohh, good idea, I will add it.

app/templates/views/email-branding/view-branding.html
CrystalPea: Ohh that's an interesting one, I will ask @karlchillmaid .

I went with archive as, like you say, it's more accurate, and I didn't want future admins to be worried about images not appearing in emails etc.

karlchillmaid: _‘Be consistent, not uniform’_

If what we’re doing here is archiving something (so that it can be retrieved), then I think it’s ok to use ‘Archive‘ instead of ‘Delete’.

Perhaps it’s a good time to start adding some more of our rationale to this page on the wiki: https://github.com/alphagov/notifications-manuals/wiki/How-to-label-links-and-headings

app/templates/views/email-branding/view-branding.html
leohemsted: yeah, more for auditing purposes (esp if we're trying to determine if a new branding is a duplicate). i dont really mind if we add it - it would be neat but not essential i think, i'll leave it up to you.

app/main/views/email_branding.py
CrystalPea: This one done ☺️

app/templates/views/email-branding/view-branding.html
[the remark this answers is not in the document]
CrystalPea: Done ☺️

app/templates/views/email-branding/view-branding.html
CrystalPea: Hi Karl, thanks a lot for that!

So this one is tricky one: after we archive it, we do not plan to "unarchive" it, but we are also not deleting it, because we need to keep the logo to serve it when someone opens an old email where that logo is embedded, if that makes sense.

app/templates/views/email-branding/view-branding.html
CrystalPea: Makes sense. I think let's leave it for now, but I am not averse to adding it later / when we're doing the same flow for letter branding.

app/templates/views/email-branding/view-branding.html
CrystalPea: Done now ☺️

app/templates/views/email-branding/view-branding.html
CrystalPea: Done now ☺️

app/templates/views/email-branding/view-branding.html
samuelhwilliams: > If what we’re doing here is archiving something (so that it can be retrieved), then I think it’s ok to use ‘Archive‘ instead of ‘Delete’.

Well, I bring it up because on other pages we say 'Delete' but actually are archiving rather than deleting it. But here we are archiving it and calling it 'Archive'. So I think we're breaking the 'be consistent' thing.

I'm not anti this saying "archive" at all - but just flagging our options:

1) Call this archive and accept the inconsistency?
2) Call this archive and update the other places that say delete to 'archive'
3) Call this delete and be consistent

karlchillmaid: Thanks @CrystalPea and @samuelhwilliams for taking the time to interrogate this properly.

The key thing here is that the item cannot be retrieved or ‘unarchived’ by the user.

I think `Delete` is the better option because – as far as the user is concerned – that’s what they’re doing.

Yes, we actually keep the file so for a specific use case, but as the user cannot undo the action or interact with the file anymore, `Delete` seems more appropriate than `Archive`.

That was the rationale behind the other pages that use `Delete`, and I think it applies here too.

I know this doesn’t address the risk Pea raised about future admins worrying that they’re going to break old emails. But if that were the case, I would expect us to explain that in the `Are you sure?` flash message. If we’re really worried about this we could always add some content to that message – but maybe we hold off doing that until we know that it causes issues for team members?

app/main/views/email_branding.py
```diff
@@ -37,6 +37,32 @@ def email_branding():
...
+    "/email-branding/<uuid:branding_id>/archive",
+    methods=["GET"],
+    endpoint="platform_admin_confirm_archive_email_branding",
+)
+def platform_admin_view_email_branding(branding_id):
+    email_branding = EmailBranding.from_id(branding_id)
+
+    if request.endpoint == "main.platform_admin_confirm_archive_email_branding":
+        if email_branding.is_used_by_orgs_or_services:
+            flash("This branding is used and so it can't be archived.")
```
karlchillmaid, line 53:
```suggestion
            flash("This branding is in use and cannot be archived.")
```

app/templates/views/email-branding/view-branding.html
```diff
@@ -0,0 +1,95 @@
...
+                        <li class="browse-list-item">
+                            <a class="govuk-link govuk-link--no-visited-state browse-list-hint"
+                                href="{{ url_for('.organisation_settings', org_id=organisation.id) }}">
+                                {{ organisation.name }}
+                            </a>
+                        </li>
+                    {% endfor %}
+                </ul>
+            {% else %}
+                <p class="hint">This branding is not a default for any organisations.</p>
```
karlchillmaid, line 39:
```suggestion
                <p class="hint">No organisations use this branding as their default.</p>
```

Part of me wonders whether we could get away with something as simple as `None` for this and the services list below – but let’s stick with this for now.

app/templates/views/email-branding/view-branding.html
leohemsted: I agree with karl - consistent behaviour with, eg, deleting a template is a useful parallel to draw. i'll make the change

app/main/views/email_branding.py
leohemsted: Added this, but changed archive for delete per discussion above

leohemsted (approved):

PR #4576, opened 2023-01-16
Make letter branding live

leohemsted:
Best reviewed commit by commit. There's also NHS work that needs to go in, but that can be a separate PR.

### Screenshots
-------------
| what | looks like |
|---------|------|
| Now uses govuk frontend radio components | image |
| if something else is the only option doesn't show radios | image |
| freetext zendesk entry is now a separate page | image |

-----------

Most of this I tried to make as much stuff re-used between this and the email branding flow as possible


tests/app/main/views/service_settings/test_letter_branding_requests.py
```diff
@@ -163,81 +163,6 @@ def test_letter_branding_request_redirects_to_branding_preview_for_a_branding_po
...
-        ticket_type="question",
-        user_name="Test User",
-        user_email="test@user.gov.uk",
-        org_id=organisation_id,
-        org_type="central",
-        service_id=SERVICE_ONE_ID,
-    )
-    mock_send_ticket_to_zendesk.assert_called_once()
-    assert normalize_spaces(page.select_one(".banner-default").text) == (
-        "Thanks for your branding request. We’ll get back to you within one working day."
```
CrystalPea, line 237: Don't we need this test anymore? It feels it would be good to test that it creates a Zendesk ticket on submit.

app/main/views/service_settings/letter_branding.py
```diff
@@ -44,6 +44,16 @@ def _letter_branding_flow_query_params(**kwargs):
 def letter_branding_request(service_id):
     form = ChooseLetterBrandingForm(current_service)
     from_template = request.args.get("from_template")
+
+    if form.something_else_is_only_option and request.method == "POST":
```
CrystalPea: What happens if something else is an only option and the method is GET? Should we serve users some extra content, like we do for email branding?

CrystalPea (commented):
Nice one 👏🏼 

I like that you did some clean up as a part of this, less work for later :).

I left a couple comments.

tests/app/main/views/service_settings/test_letter_branding_requests.py
leohemsted: we have `test_POST_letter_branding_something_else_creates_zendesk_ticket` but maybe its worth taking the richer assert from this test

app/main/views/service_settings/letter_branding.py
leohemsted: | letter branding | non-gov email | gov email when it shouldn't be |
| image | image | image |

i dont think we need to replicate the third column, because that's not a concept that exists for letters. its fine for letters not to have branding if they don't want.

tests/app/main/views/service_settings/test_letter_branding_requests.py
leohemsted: fixed in 4e4271a7dd11b22ca747bf619c06bf13a8db9539

app/main/views/service_settings/letter_branding.py
CrystalPea: Ahhh, I thought there is more guidance for emails, for some reason, sorry! 😳

CrystalPea (approved):

PR #4578, opened 2023-01-17
preview current letter branding when changing it

leohemsted:
only shows if you have existing letter branding

image

Note that this is separate to the other letter branding work, and can go in before OR after the other stuff

### DIFF
--- a/tests/app/main/views/service_settings/test_letter_branding_requests.py
+++ b/tests/app/main/views/service_settings/test_letter_branding_requests.py
@@ -82,6 +82,9 @@ def test_letter_branding_request_page_when_no_branding_is_set(
 
     assert normalize_spaces(page.select_one("main p").text) == "Your letters currently have no branding."
 
+    # no preview if no existing branding
+    assert not page.select_one("iframe")
+
     button_text = normalize_spaces(page.select_one(".page-footer button").text)
     assert button_text == "Continue"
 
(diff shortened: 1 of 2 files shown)


tests/app/main/views/service_settings/test_letter_branding_requests.py
```diff
@@ -106,6 +109,11 @@ def test_letter_branding_request_page_when_branding_is_set_already(
     page = client_request.get(".letter_branding_request", service_id=SERVICE_ONE_ID)
     assert normalize_spaces(page.select_one("main p").text) == "Your letters currently have HM Government branding."
 
+    letter_preview = page.select_one("iframe")
+    letter_preview_url = letter_preview.get("src")
+    letter_preview_query_args = parse_qs(urlparse(letter_preview_url).query)
+    assert letter_preview_query_args == {"branding_style": [fake_uuid]}
```
quis, line 115: Not sure why you’re not asserting against the whole URL here, but this still does the job

quis (approved):

tests/app/main/views/service_settings/test_letter_branding_requests.py
leohemsted: Good point - it was because I copy and pasted from another assertion, but we can just make this into a neater one-liner that is more obvious what it's checking while testing the entire URL too.

38a9b7d

PR #4582, opened 2023-01-17
Standardise how we do `IntegerField`s 

quis:
Using `ForgivingIntegerField` meant if you paste a number that contains commas or spaces (common for numbers in the hundreds of thousands) it will still be accepted. But we weren’t using it everywhere. This PR:
- makes it used wherever we ask for integers
- renames it to replace `GovukIntegerField` to make it clear that it’s the one canonical way to ask for integers

Also added thousands separators to some hardcoded numbers to make them easier to read.

### DIFF
--- a/app/formatters.py
+++ b/app/formatters.py
@@ -479,3 +479,8 @@ def format_auth_type(auth_type, with_indefinite_article=False):
         return f"{indefinite_article} {auth_type.lower()}"
 
     return auth_type
+
+
+def sentence_case(sentence):
+    first_word, rest_of_sentence = (sentence + " ").split(" ", 1)
+    return f"{first_word.title()} {rest_of_sentence}"[:-1]
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -30,7 +30,6 @@ from wtforms import (
     FieldList,
     FileField,
     HiddenField,
-    IntegerField,
     PasswordField,
 )
 from wtforms import RadioField as WTFormsRadioField
@@ -279,10 +278,6 @@ class GovukDateField(GovukTextInputFieldMixin, DateField):
     pass
 
 
-class GovukIntegerField(GovukTextInputFieldMixin, IntegerField):
-    pass
-
-
 class SMSCode(GovukTextInputField):
     # the design system recommends against ever using `type="number"`. "tel" makes mobile browsers
     # show a phone keypad input rather than a full qwerty keyboard.
@@ -300,7 +295,7 @@ class SMSCode(GovukTextInputField):
             self.data = InsensitiveDict.make_key(valuelist[0])
 
 
-class ForgivingIntegerField(GovukTextInputField):
+class GovukIntegerField(GovukTextInputField):
 
     #  Actual value is 2,147,483,647 but this is a scary looking arbitrary number
     POSTGRES_MAX_INT = 2_000_000_000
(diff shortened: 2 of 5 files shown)


samuelhwilliams (commented):
This change looks nice and I like it in theory 🙌 

But... I think this PR has some unintended side effects on the error messages we show now. Just the example I found straight away - it's possible there are other places that have changed in more impactful ways.

For example, before:
image

And after:
image

I think both of these error messages are rubbish for the given form, but still we probably shouldn't be changing the validation output here. Esp as @tombye is gathering validation messages right now to audit them.

quis:
@samuelhwilliams good point, lots of the error messages for these fields were unhelpful for non-empty, non-integer input.

I’ve amended them in 9303f35 and 4251bf3 to make more sense, following the guidance in the [Design System](https://design-system.service.gov.uk/components/error-message/) – some examples:

> Number of messages must be a whole number

> Percentage must be a whole number

> Days of retention must be a whole number

app/main/forms.py
```diff
@@ -301,8 +301,8 @@ class GovukIntegerField(GovukTextInputField):
     #  Actual value is 2,147,483,647 but this is a scary looking arbitrary number
     POSTGRES_MAX_INT = 2_000_000_000
 
-    def __init__(self, label=None, *, things, **kwargs):
-        self.things = things
+    def __init__(self, label=None, *, prefix_for_format_error, **kwargs):
+        self.prefix_for_format_error = prefix_for_format_error
```
samuelhwilliams, line 305: This new name is super cumbersome. I'd prefer either the singular `thing`, which I feel includes a 'percentage' just fine, or maybe `what`.

app/main/forms.py
```diff
@@ -1128,23 +1130,23 @@ def __init__(self, *args, **kwargs):
 class AdminServiceSMSAllowanceForm(StripWhitespaceForm):
     free_sms_allowance = GovukIntegerField(
         "Numbers of text message fragments per year",
-        things="text message fragments",
+        prefix_for_format_error="Number of text message fragments",
```
samuelhwilliams, line 1133: Minor nit but given my history with text processing I'd like to see this change: given these are fragments of text, we should provide them with the first letter lowercase (unless it's a proper noun, but it never is so far).

Then the function presenting the fragment should capitalise the first letter if it wants to display it at the start of the sentence.

This is a bit annoying because Python doesn't have a built-in for capitalising the first letter of a string nicely (`.capitalize()` has the side-effect of forcing everything else to lowercase), so we need a tiny helper probably.

But it then means we don't have to update all of the callers if we decide to re-write the content message and end up with the fragment in the middle of the sentence.

app/main/forms.py
quis: I’ve dropped the renaming commit

app/main/forms.py
quis: I’ve written a helper and used it in [660d083](https://github.com/alphagov/notifications-admin/pull/4582/commits/660d0839a46a0930addc6d7cfc857f498515f9f1)

app/formatters.py
```diff
@@ -479,3 +479,8 @@ def format_auth_type(auth_type, with_indefinite_article=False):
         return f"{indefinite_article} {auth_type.lower()}"
 
     return auth_type
+
+
+def sentence_case(sentence):
+    first_word, rest_of_sentence = (sentence + " ").split(" ", 1)
+    return f"{first_word.title()} {rest_of_sentence}"[:-1]
```
samuelhwilliams, line 486: This should be fine for us but does still have the fun edge case of breaking for bicapitalised words (eg McGonagall, iPhone, etc).

samuelhwilliams (approved):
Nice 🎉

PR #4588, opened 2023-01-23
User email branding - new logo client

samuelhwilliams:
Update the user-facing email branding flows to use the new logo_client.

---

We also switch tack here and change the logo client to store logos (both temporary and permanent) in their existing paths. This is because of some trickiness around letter logos, which have their expected path hard-coded in utils templates. We don't have the time to tackle this right now, so will use S3 object tagging to handle object lifecycles instead of more explicit/obvious object key pathing.


app/s3_client/logo_client.py
```diff
@@ -42,7 +44,7 @@ def _slugify(self, text: str) -> str:
         return text.lower()
 
     def _get_temporary_logo_key(self, logo_file_name: str, logo_type: LOGO_TYPES) -> str:
-        return self.TEMPORARY_LOGO_PATHS[logo_type].format(filename=logo_file_name)
+        return self.LOGO_PATHS[logo_type].format(filename=logo_file_name)
```
leohemsted, line 47: is it worth re-adding the `temp-` prefix if we aren't able to separate out the folders any more? a minor thing but will help debugging a little bit

app/s3_client/logo_client.py
```diff
@@ -104,7 +107,9 @@ def save_permanent_logo(
             logo_file_name=logo_key, logo_type=logo_type, logo_key_extra=logo_key_extra
         )
 
-        self._get_object(permanent_logo_key).copy_from(CopySource="{}/{}".format(self.bucket_name, temporary_logo_key))
+        self._get_object(permanent_logo_key).copy_from(
+            CopySource="{}/{}".format(self.bucket_name, temporary_logo_key), TaggingDirective="REPLACE"
```
leohemsted, line 111: can we stick a code comment here explaining that this strips all tags? since i had to google that to find out what it did and what the options are

app/s3_client/logo_client.py
samuelhwilliams: Sure 👍

leohemsted (approved):

PR #4589, opened 2023-01-23
Fix margins on branding

leohemsted:
Before: the sticky bottom nav margin completely overlaps the last item in the list, meaning it's impossible to click that list

After: the items in the list all have more consistent styling with other pages, and crucially, a `margin-bottom: 15px` that prevents them from overlapping with the sticky nav.
| before | after |
|-|-|
| image | image |

### DIFF
--- a/tests/app/main/views/test_email_branding.py
+++ b/tests/app/main/views/test_email_branding.py
@@ -16,7 +16,7 @@ def test_email_branding_page_shows_full_branding_list(client_request, platform_a
     client_request.login(platform_admin_user)
     page = client_request.get(".email_branding")
 
-    links = page.select(".message-name a")
+    links = page.select(".email-brand a")
     brand_names = [normalize_spaces(link.text) for link in links]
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_email_branding.py
```diff
@@ -16,7 +16,7 @@ def test_email_branding_page_shows_full_branding_list(client_request, platform_a
     client_request.login(platform_admin_user)
     page = client_request.get(".email_branding")
 
-    links = page.select(".message-name a")
+    links = page.select(".email-brand a")
```
quis, line 19: Think you can do this then you don’t need to introduce an extra class
```suggestion
    links = page.select(".browse-list-item a")
```

leohemsted: it was like this already so the live search had a class to update, however, we're already using generic selectors for live search in plenty of other pages so i've updated in 96ed2a7e6

PR #4596, opened 2023-01-25
Letter branding logo client updates

samuelhwilliams:
Update the letter branding flows (both platform admin and user-facing) to use the new logo_client interface to S3.

We don't update the path that letter logos are stored. I tried to, but the path is hard-coded into notifications-utils templates and we only store partial filenames in the DB rather than the full path, too, so we don't have an easy switch migration path for DB values either. So it's out of scope.

What this still gets us is: temporary letter logos will be tagged and automatically lifecycle-managed by S3. And we can remove the less consistent s3_logo_client interface with a central one that works for both email and letter branding logos.


app/main/views/letter_branding.py
```diff
@@ -127,30 +125,34 @@ def create_letter_branding(logo=None):
...
+                    logo_type="letter",
+                    logo_key_extra=letter_branding_details_form.name.data,
+                )
                 letter_branding = LetterBranding.create(
-                    filename=db_filename, name=letter_branding_details_form.name.data
+                    filename=letter_filename_for_db_from_logo_key(permanent_logo_key),
+                    name=letter_branding_details_form.name.data,
                 )
 
-                upload_letter_svg_logo(logo, db_filename, current_user.id)
```
leohemsted, line 152: what happens to the uploaded logo if there's an error with the name being duplicate? i feel like we discussed this before but cant remember what our conclusion was. its a bit awkward that it's slightly diff to how the user facing flow silently fixes your branding name, while this one raises an error

app/utils/branding.py
```diff
@@ -34,3 +36,10 @@ def get_letter_choices(service):
 
     if service.organisation and not service.letter_branding_pool:
         yield ("organisation", service.organisation.name)
+
+
+def letter_filename_for_db_from_logo_key(logo_key):
+    # For letters, in the DB we store the filename without the extension.
+    # Note: this is to maintain backwards compatibility and does deviate from what we do for EmailBranding.
+    # But we aren't taking on this migration effort right now. Ideally this might be different.
```
leohemsted, line 44: minor: this comment reads a bit like a git commit, and I'm not sure it adds masses of context. Maybe something a little terser like:

> For letters, we store just the filename without extension - note this is different to EmailBranding where we store a full filepath
> TODO: store full filepath with folder and extension

leohemsted (commented):
one comment and a couple of super minor bits, but generally looks good

app/main/views/letter_branding.py
samuelhwilliams: If that happens then, yes, we end up with a logo in S3 that wouldn't ever get cleaned up or used. I suppose we could wrap the 'create brand' call in a try/except and clean up the logo, but it's such a minor edge case that it doesn't feel _that_ important (especially as an admin-only flow).

Or we could update this to try to silently fix the name as well and avoid that edge case?

leohemsted: id rather we didnt silently fix the name for platform admins because that's a good indicator to us that htere are two brandings that have the same name and we should have better names than just "alternate 1"

if it does happen though, what will end up happening if the platform admin changes the name and saves again? will anything strange happen if it tries to copy the file but it already exists?

samuelhwilliams: I've added the "get unique name" step to the admin flow.

leohemsted (approved):

PR #4601, opened 2023-01-30
Allow services to enable email auth

samuelhwilliams:
Content/design doc: https://docs.google.com/document/d/1PTZD3nwaj1H9ijMaYVrEkChP_sZ1iyAsCv5rQzCo2FE/edit#heading=h.akt2c46d6o07

Service managers can request to allow their users to login via a magic email link instead of a text message as a form of 2FA. At the moment this goes through Zendesk, requiring support action from us, but we believe this should be able to be self-served.

This changes the `set-auth-type` change to a form which allows managers to decide between text message and magic link auth.

Enabling email auth does not change users' sign in to actually use email links - so service managers will still need to go to the team members page and update permissions individually from there. A future patch will provide an immediate page where the manager can update permissions in bulk.

Disabling email auth _does_ automatically change any users of the service who sign-in with emails to sign in with text messages instead. Anyone signing in with webauthn is unaffected.

image

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -761,6 +761,14 @@ def service_set_channel(service_id, channel):
 @user_has_permissions("manage_service")
 def service_set_auth_type(service_id):
     form = SetAuthTypeForm(sign_in_method=current_service.sign_in_method)
+
+    if current_service.sign_in_method == SIGN_IN_METHOD_TEXT_OR_EMAIL:
+        if users_without_phone_numbers := [user for user in current_service.active_users if not user.mobile_number]:
+            return render_template(
+                "views/service-settings/disable-email-auth-blocked.html",
+                users_without_phone_numbers=users_without_phone_numbers,
+            )
+
     if form.validate_on_submit():
         if (
             current_service.sign_in_method == SIGN_IN_METHOD_TEXT_OR_EMAIL
@@ -783,6 +791,9 @@ def service_set_auth_type(service_id):
 @main.route("/services/<uuid:service_id>/service-settings/set-auth-type/confirm", methods=["GET", "POST"])
 def service_confirm_disable_email_auth(service_id):
     if current_service.sign_in_method != SIGN_IN_METHOD_TEXT_OR_EMAIL:
+        return redirect(url_for(".service_set_auth_type", service_id=service_id))
+
+    if any(not user.mobile_number for user in current_service.active_users):
         return redirect(url_for(".service_set_auth_type", service_id=service_id))
 
     if request.method == "POST":
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -59,6 +59,7 @@ from app.main.forms import (
     ServiceReplyToEmailForm,
     ServiceSmsSenderForm,
     ServiceSwitchChannelForm,
+    SetAuthTypeForm,
     SMSPrefixForm,
     YesNoSettingForm,
 )
@@ -70,6 +71,7 @@ from app.models.branding import (
     LetterBranding,
 )
 from app.utils import DELIVERED_STATUSES, FAILURE_STATUSES, SENDING_STATUSES
+from app.utils.constants import SIGN_IN_METHOD_TEXT_OR_EMAIL
 from app.utils.user import (
     user_has_permissions,
     user_is_gov_user,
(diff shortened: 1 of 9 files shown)


samuelhwilliams:
@karlchillmaid I'm wondering if we want a flash message when the setting is changed?

klssmith:
When I tried to change the auth type for a service to SMS from email there was an error because a user didn't have a phone number set

samuelhwilliams:
@quis should we just link directly to the user's page for this list?

image

app/templates/views/service-settings/disable-email-auth-blocked.html
```diff
@@ -0,0 +1,38 @@
...
+        You cannot switch off the email link sign-in method for the whole team until these members add a mobile phone number:
+      </p>
+      <ul class="govuk-list govuk-list--bullet">
+      {% for user in users_without_phone_numbers %}
+        <li>{{ user.name }}</li>
+      {% endfor %}
+      </ul>
+      <p class="govuk-body">
+        If you need to change an individual’s sign-in method, go to the
+        <a class="govuk-link" href="{{ url_for('.manage_users', service_id=current_service.id) }}">team members</a>
```
klssmith, line 33:
```suggestion
        <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('.manage_users', service_id=current_service.id) }}">team members</a>
```

app/main/views/service_settings/index.py
```diff
@@ -755,14 +757,59 @@ def service_set_channel(service_id, channel):
...
+        return redirect(url_for(".service_settings", service_id=service_id))
+
     return render_template(
         "views/service-settings/set-auth-type.html",
+        form=form,
     )
 
 
+@main.route("/services/<uuid:service_id>/service-settings/set-auth-type/confirm", methods=["GET", "POST"])
+def service_confirm_disable_email_auth(service_id):
```
klssmith, line 792: Should this route have the `@user_has_permissions("manage_service")` decorator too? We don't link to it directly, so it's low risk, but someone could still potentially change a setting they shouldn't by visiting the page directly.

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -4552,27 +4553,173 @@ def test_send_files_by_email_contact_details_does_not_update_invalid_contact_det
...
-        service_id=SERVICE_ONE_ID,
-    )
+class TestSetAuthType:
+    def test_page_requires_manage_settings_permission(
+        self,
+        client_request,
+        service_one,
+        active_user_with_permissions,
+    ):
+        active_user_with_permissions["permissions"][service_one["id"]].remove("manage_settings")
```
klssmith, line 4563: Optional: You could use the `active_user_view_permissions` fixture instead of removing permissions

klssmith:
I noticed when trying this that you can invite someone to log in with an email link, then switch the auth type to SMS and when that person accepts the invite their auth type will still be email. That's how things work at the moment with the existing code too, so I don't think we need to change anything here, but might be one we think about changing in another story.

app/main/views/service_settings/index.py
samuelhwilliams: Good idea 👍

samuelhwilliams:
It's a good spot 👍  Let's record a ticket for that so we can keep track of and hopefully prioritise it (I've asked Leo where to put said ticket)

klssmith (approved):
Looks good to me, but I'll let Karl and Chris address the questions to them

samuelhwilliams:
`Save` is what we have in the content doc, but now you point it out I agree - let's see what Karl says.

karlchillmaid:
`Confirm` makes more sense than `Save`.

I think usually when we ask ‘Are you sure?’, the button text is something like ’Yes, delete’ or ‘Yes, remove’. _Possibly_ with visually hidden text that describes the thing again for accessibility. 

So, in this case we could go with `Yes, remove`

karlchillmaid:
@samuelhwilliams a flash message makes sense to me, but I don’t think this is something we currently do when most other settings are toggled on/off.

We’re a bit inconsistent, because we do display flash messages for branding requests: for example: `Thanks for your branding request. We’ll get back to you within one working day.`

But, as I say, we don’t currently show flash messages for most other things. 😞

PR #4602, opened 2023-01-30
Fix New Relic function path

klssmith:
The `organisation_dashboard` view function was moved to a new file in e9d8803. The New Relic config needed to be updated too since it still had the old path to the file, and the error this gave was stopping the app from starting up.


samuelhwilliams (approved):
Aw ew. That's rubbish! Good find.

PR #4605, opened 2023-01-30
Remove unexpected argument when calling `create_service_template`

klssmith:
This function no longer takes queue type.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -289,7 +289,7 @@ def _add_template_by_type(template_type, template_folder_id):
 
     if template_type == "letter":
         blank_letter = service_api_client.create_service_template(
-            "New letter template", "letter", "Body", current_service.id, "Main heading", "normal", template_folder_id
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
I haven't looked at the code too much but is it easy for us to add a unit test to this bit of code so we would catch something like this in the future?

Open to not adding one if it isn't easy to test though...

samuelhwilliams (approved):
Agree it'd be nice to add a test - could be done in a separate followup PR if needed though.

klssmith:
@idavidmcdonald I think a test would definitely be useful, but it didn't look particularly straightforward to add one in - I could merge this then look separately since I'm not sure how long it would take me to write a test.

idavidmcdonald:
I'll leave you to judge Katie, thanks :)

leohemsted (approved):
oops ignore, was checked out to wrong commit 🤦

klssmith:
Turns out it wasn't really difficult to test this after all, so I've added one here https://github.com/alphagov/notifications-admin/pull/4610

PR #4606, opened 2023-01-30
Add a central 'find anything' page

samuelhwilliams:
Uses the new `find-by-uuid` API endpoint for platform admins to provide a single central search form that can take any UUID and send a platform admin to the best page we have to view that record.

As part of this change, we pull the two other 'platform admin lookup' pages onto a single page so that we don't have 3 pages for looking up records. This page is fairly unoptimised and shows 3 search boxes. The existing ones (service and user) redirect to their existing pages, while the find-by-uuid takes the user to the actual record page where appropriate.

---

Could do with fixing the alignment of the search box for the UUID form when there's an error 🤔 

## Look at it
image

## Fixing error summaries with inline search inputs
The next 4 commits create a new GOV.UK Frontend-derived component that we can use to display search inputs with an inline submit button. This addresses a styling issue where, when those inputs have errors attached, the error would push the input down but not the related search button. See below (bad) and the fixed (better) screenshots. There remains a tiny styling nit in that, in the error state, the left hand edge of the button doesn't line up with the left-hand edge of the buttons above in the non-error state. But I think this is fine (and I don't know how to fix it really).

It also makes the layout better on smaller screens, see bad on mobile vs better on mobile.

This new component is then applied to the find-services, find-users, and find-uuid forms - all platform-admin only.

### Bad
image

### Better
image

### Bad on mobile
image

### Better on mobile
image


samuelhwilliams:
@karlchillmaid any chance of some input on words for this? :)

karlchillmaid:
@samuelhwilliams 

Is there a reason the page is called `Find` rather than `Search`?

For accessibility, can we add some hidden text to the 3 `Search` buttons to help differentiate them?

I’ll make some suggestions for other small edits.

samuelhwilliams:
@karlchillmaid 

> Is there a reason the page is called Find rather than Search?

No - not really. The old pages were 'Find service by name' and 'Find user by email' so I just copied that. I haven't put much (any) thought into the content to be honest!

> For accessibility, can we add some hidden text to the 3 Search buttons to help differentiate them?

Definitely 👍

karlchillmaid:
Starter for ten.

I renamed the page ‘Search’ because:
* it feels consistent with the green buttons
* ‘Find’ on its own (without a description of what you could find) felt a little lost

Content in square brackets should be given the `visually hidden` class.

I’ve tried dropping the ‘or partial’ hint – what do you think?

># Search
>
>Find services by name or service ID
>Search [by service name or service ID]
>
>Find users by email address
>Search [by email address]
>
>Find anything by UUID
>Search [by UUID]

karlchillmaid:
Validation errors should echo the rest of the content, but we could include ‘full or partial’ in there?

samuelhwilliams:
@tombye would you mind glancing over the FE changes here as/when you have time?

app/templates/views/platform-admin/find-by-uuid.html
```diff
@@ -0,0 +1,26 @@
+{% extends "views/platform-admin/_base_template.html" %}
```
leohemsted, line 1: can we delete this file? this isn't used by anything

app/main/forms.py
```diff
@@ -274,7 +287,12 @@ class GovukEmailField(GovukTextInputFieldMixin, EmailField):
 
 class GovukSearchField(GovukTextInputFieldMixin, SearchField):
     input_type = "search"
-    param_extensions = {"classes": "govuk-!-width-full"}  # email addresses don't need to be spellchecked
+    param_extensions = {"classes": "govuk-!-width-full"}
+
+
+class GovukInlineButtonSearchField(GovukInlineButtonTextInputFieldMixin, SearchField):
+    input_type = "search"
+    param_extensions = {"button": {"classes": "govuk-!-width-full notify-input--inline-button", "text": "Search"}}
```
leohemsted, line 295: this `text` is never used as we define custom html to pass in to param_extensions when we actually use this class in search.html
```suggestion
    param_extensions = {"button": {"classes": "govuk-!-width-full notify-input--inline-button"}}
```

app/main/views/platform_admin.py
```diff
@@ -619,6 +641,65 @@ def clear_cache():
...
+                ".conversation", "notification_id", with_service_id=True, extra={"_anchor": f"n{uuid_}"}
+            ),
+        }
+        if not (spec := url_for_data.get(result["type"])):
+            raise KeyError(f"Don't know how to redirect to {result['type']}")
+
+        url_for_kwargs = {"endpoint": spec.endpoint, **spec.extra}
+
+        if spec.param:
+            url_for_kwargs[spec.param] = uuid_
```
leohemsted, line 693: i cant help but feel that this isn't very helpful in those (two?) cases where there's no specific URL to link to - complaints and inbound numbers

tests/app/test_navigation.py
```diff
@@ -223,13 +223,13 @@
             "organisations",
             "performance",
             "platform_admin",
+            "platform_admin_archive_email_branding",
+            "platform_admin_create_email_branding",
+            "platform_admin_confirm_archive_email_branding",
```
leohemsted, line 228: very minor but dont think these email branding lines should have moved up? keeping the email branding views grouped together in the list would be nice

leohemsted (commented):
i really like the combined page, and the new uuid page seems decent - i've got some reservations about the new frontend macro override though 🤔

app/templates/govuk_frontend_jinja_overrides/templates/components/inline-button-input/macro.html
```diff
@@ -0,0 +1,107 @@
+{#
+  An override of the input component from v1.5.1 of govuk_frontend_jinja, adding an inline button next to the input
```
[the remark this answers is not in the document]
samuelhwilliams: I know where you're coming from, although I feel it's slightly overstated. This macro does some custom alignment and construction but then still hands off to core GOV.UK Design System components and govuk_frontend_jinja macros the power the actual key fragments.

If you have a less duplicative way I'm all ears, but I think this is it. This is consistent and vaguely a pattern we already have. I think it's ok for us to have some custom components.

We use the similar pattern in a few places already (that I probably should have included in the patch tbh), such as https://www.notifications.service.gov.uk/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553/notifications/email?status=sending,delivered,failed. And places like https://www.notifications.service.gov.uk/organisations or https://www.notifications.service.gov.uk/email-branding probably _should_ use them rather than not have a button visible at all.

app/main/forms.py
samuelhwilliams: I think I'm OK keeping this nonetheless, as this field could be re-used and without it the button will just have no text at all, which doesn't feel like a sensible default.

app/main/views/platform_admin.py
samuelhwilliams: I agree those are much _less_ useful - but it would at least let you know _what kind_ of thing you're dealing with. It feels wrong to just say "no, Notify doesn't have anything like that" and could be a bit more misleading than this implementation is unhelpful.

tests/app/test_navigation.py
samuelhwilliams: Meh, alphabetical vs grouped-by-semantics is going to be a back-and-forth debate on the team if we don't enforce one. Think I've got in the habit of alphabetising as we do that in a few other places, but don't recall why I made this change now so I'll undo it.

app/main/views/platform_admin.py
samuelhwilliams: I could go either way though...

app/templates/govuk_frontend_jinja_overrides/templates/components/inline-button-input/macro.html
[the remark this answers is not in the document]
samuelhwilliams: But perhaps the more salient point is, I'm not really aware of a better way to render both of the components in the same row without using a custom macro, as the button needs to be nested inside a lot of stuff that the plain macro provides in order to address the alignment issue mentioned in the main description.

tests/app/test_navigation.py
leohemsted: i only raised it cos the new line you'd added isn't alphabetical 😅 

that said i agree with you - that if we don't enforce stuff then we shouldn't worry too much, and i also reckon it'd be too onerous for us to try and draw up some catch-all ordering that will cover all the cases and nuance and it's better to just not worry too much about it generally and just try and generally:

* keep things in their current positions where possible to reduce git blame churn
* add new things in sensible places (whether its best placed alphabetically/semantic i think just leave to the author)

samuelhwilliams: I think I undid the change but maybe not. I will though 👍

app/main/views/platform_admin.py
leohemsted: lets keep them for now -  think it's still somewhat useful, even if you'll need to then open up a conduit to find out more. i suspect it'll be very rare that we have a complaint uuid anyway. maybe inbound number id is a bit more plausible for us to end up with on our clipboard

app/templates/govuk_frontend_jinja_overrides/templates/components/inline-button-input/macro.html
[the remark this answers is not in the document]
samuelhwilliams: @tombye has a better way so he's going to put a commit up and I'll tidy this all up (read: remove it) afterwards 👍

leohemsted (approved):
think this looks really good now 👍

PR #4607, opened 2023-01-31
added new Additional area to cover 0,0 Eastings/Northings

gds-gharper:
This is to deal with cells which have no current location and allow us to target them.


app/broadcast_areas/create-broadcast-areas-db.py
```diff
@@ -7,8 +7,8 @@
 from pathlib import Path
 
 import geojson
-from notifications_utils.formatters import formatted_list
-from notifications_utils.polygons import Polygons
+from emergency_alerts_utils.formatters import formatted_list
```
idavidmcdonald, line 10: These should point to `notifications.utils` as we don't have `emergency_alerts_utils` available to this codebase

idavidmcdonald (approved):
Seems to follow the existing patterns and the tests pass

I can't vouch that the coordinates you've picked are correct but I assume they will be

I'm hoping there is no issue with the functional tests as this does change the pages they click through, but hopefully they are doing smart things already and picking explicitly rather than based on the order of things on the page... we will see when this gets merged...

gds-gharper:
> Seems to follow the existing patterns and the tests pass
> 
> I can't vouch that the coordinates you've picked are correct but I assume they will be
> 
> I'm hoping there is no issue with the functional tests as this does change the pages they click through, but hopefully they are doing smart things already and picking explicitly rather than based on the order of things on the page... we will see when this gets merged...

That's not necessarily the case, I guess we'll find out in a few minutes, but the integration tests were explicitly relying on England to be the first country (which is what I had to change to fix them).  It's something I've put a ticket in our backlog to change, I'm not a huge fan of brittle tests.  Right now they're still brittle but I didn't want to make this changeset too large for what is essentially a data change.

PR #4609, opened 2023-02-01
Add admin page to view a single letter branding

samuelhwilliams:
Partner PR for https://github.com/alphagov/notifications-api/pull/3715

Duplicates the 'view email branding' view, but for letter brandings.

---

A copy of
https://github.com/alphagov/notifications-admin/commit/b0f61c13372a94e4a1e5c2a1eafe22a9b1382f4b but for letter branding.

This page has:
- branding preview
- list of orgs with the branding as a default
- list of services  using the branding
- edit link

---

This doesn't give much thought to de-duplicating/refactoring because we have https://trello.com/c/AazU7YzJ/185-look-through-branding-code-and-see-what-can-be-refactored to do that. This brings us (closer to) feature parity between the two branding types though.

I explicitly don't take archiving letter brandings here and leave that to be handled separately, if we want to.

### DIFF
--- a/app/main/views/letter_branding.py
+++ b/app/main/views/letter_branding.py
@@ -35,6 +35,7 @@ def letter_branding():
 
 
 @main.route("/letter-branding/<uuid:branding_id>", methods=["GET", "POST"])
+@user_is_platform_admin
 def platform_admin_view_letter_branding(branding_id):
     letter_branding = LetterBranding.from_id(branding_id)
 
--- a/app/main/views/letter_branding.py
+++ b/app/main/views/letter_branding.py
@@ -34,6 +34,20 @@ def letter_branding():
     )
 
 
+@main.route("/letter-branding/<uuid:branding_id>", methods=["GET", "POST"])
+def platform_admin_view_letter_branding(branding_id):
+    letter_branding = LetterBranding.from_id(branding_id)
+
+    return render_template(
+        "views/letter-branding/view-branding.html",
+        letter_branding=letter_branding,
+        cdn_url=current_app.config["LOGO_CDN_DOMAIN"],
+        logo=permanent_letter_logo_name(letter_branding.filename, "svg"),
+        branding_orgs=letter_branding.organisations,
+        branding_services=letter_branding.services,
+    )
+
+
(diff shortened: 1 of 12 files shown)


app/main/views/letter_branding.py
```diff
@@ -34,6 +34,20 @@ def letter_branding():
     )
 
 
+@main.route("/letter-branding/<uuid:branding_id>", methods=["GET", "POST"])
```
klssmith, line 37: We're not using the `POST`, so could remove it until it's needed unless we think it will be used soon.

This doesn't have the `@user_is_platform_admin` decorator - that matches the equivalent email branding route, but maybe that was an oversight that email branding doesn't have it since these are platform admin only routes

app/templates/views/letter-branding/view-branding.html
```diff
@@ -0,0 +1,91 @@
...
+{% endblock %}
+
+{% block platform_admin_content %}
+
+<div class="govuk-grid-row bottom-gutter">
+    <div class="govuk-grid-column-three-quarters">
+        {{ page_header(letter_branding.name) }}
+
+
+        {% if logo %}
```
klssmith, line 23: I can't think of a situation where `if logo` could be false - is this check needed?

tests/conftest.py
```diff
@@ -2649,6 +2649,16 @@ def _create_letter_branding(filename, name, created_by_id):
     return mocker.patch("app.letter_branding_client.create_letter_branding", side_effect=_create_letter_branding)
 
 
+@pytest.fixture(scope="function")
+def mock_get_letter_branding(mocker, fake_uuid):
```
klssmith, line 2653: This fixture mocks the same endpoint as the [`mock_get_letter_branding_by_id`](https://github.com/alphagov/notifications-admin/blob/876134ae146b9bcb590a713ae4106ae40a2499ec/tests/conftest.py#L2455-L2464) fixture

You could maybe just use that one to avoid adding another

klssmith (approved):

CrystalPea:
This has already been merged, but there is one more thing, that is design-related.

For email branding, we show preview embedded in an example email. I think it would be good to do the same for letters. It would look better.

samuelhwilliams:
Nice idea. I will put an update.

PR #4611, opened 2023-02-02
Show letter logo as full preview

samuelhwilliams:
## Before
image

## After
image

### DIFF
--- a/app/templates/views/letter-branding/view-branding.html
+++ b/app/templates/views/letter-branding/view-branding.html
@@ -1,6 +1,7 @@
 {% extends "views/platform-admin/_base_template.html" %}
 {% from "components/page-header.html" import page_header %}
 {% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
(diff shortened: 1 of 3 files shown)


app/templates/views/letter-branding/view-branding.html
```diff
@@ -16,12 +17,10 @@
 {% block platform_admin_content %}
 
 <div class="govuk-grid-row bottom-gutter">
-    <div class="govuk-grid-column-three-quarters">
+    <div class="govuk-grid-column-full">
         {{ page_header(letter_branding.name) }}
 
-        <div id="logo-img">
-          <img src="https://{{ cdn_url }}/{{ logo }}"/>
```
quis, line 23: Should be able to remove these 2 lines now:
https://github.com/alphagov/notifications-admin/blob/b3701e6fed08cc0d08e8a2c4ffe7870c11ce1284/app/main/views/letter_branding.py#L116-L117

quis (commented):
I’d like to see a test for this, even if it doesn’t break any existing ones

quis (approved):

PR #4612, opened 2023-02-06
Update pricing pages

saimaghafoor:
Remove letters and sms content (to be placed on individual pages)


klssmith:
## Screenshots

### Updated pricing page
Screenshot 2023-02-08 at 09 47 43

### Text message pricing page
Screenshot 2023-02-08 at 09 48 15

### Letter pricing page
Screenshot 2023-02-08 at 09 48 32

### Trial mode page before
Screenshot 2023-02-08 at 09 49 36

### Trial mode page after (no changes to the content, but the menu and URL have changed)
Screenshot 2023-02-08 at 09 48 45

saimaghafoor:
@klssmith On the pricing page, could we add a heading 'Emails' and then 'It’s free to send emails through Notify.' at the end. This was missed off and needs including!

klssmith:
@saimaghafoor no problem, I've updated the page to look like this:


Screenshot 2023-02-08 at 10 37 23

saimaghafoor:
Looks good to go :)

app/main/views/index.py
```diff
@@ -211,19 +211,11 @@ def who_can_use_notify():
 
 @main.route("/trial-mode")
 @main.route("/features/trial-mode")
+@main.route("/using-notify/trial-mode")
 def trial_mode():
     return redirect(url_for(".trial_mode_new"), 301)
```
samuelhwilliams, line 216: I wonder if we should drop this and move these redirects to the main redirect handler at the bottom of this file? Then the new view can just be called `trial_mode` rather than `trial_mode_new`

app/main/views/pricing.py
```diff
@@ -13,6 +13,14 @@
 def pricing():
     return render_template(
         "views/pricing/index.html",
+        navigation_links=pricing_nav(),
+    )
+
+
+@main.route("/pricing/text-messages")
+def text_message_pricing():
```
samuelhwilliams, line 21: Might be nice to name this (and similar) `pricing_text_messages` so that they're all grouped together in our alphabetised routes/tests?

CrystalPea (approved):
👍🏼

app/main/views/pricing.py
karlchillmaid: I prefer `pricing_text_messages`

app/main/views/sub_navigation_dictionaries.py
```diff
@@ -26,14 +26,28 @@ def pricing_nav():
...
+                    "name": "Text messages",
+                    "link": "main.text_message_pricing",
+                },
+                {
+                    "name": "Letters",
+                    "link": "main.letter_pricing",
+                },
+                {
+                    "name": "Trial mode",
+                    "link": "main.trial_mode_new",
```
samuelhwilliams, line 40: I don't think I understand why this is part of the pricing nav. It seems quite conspicuous, and nothing on the actual page it shows talks about pricing at all. 🤔

app/templates/views/pricing/index.html
```diff
@@ -24,8 +21,8 @@ <h1 class="heading-large">Pricing</h1>
 
   <p class="govuk-body">GOV.UK Notify is free to use unless you:</p>
   <ul class="govuk-list govuk-list--bullet">
-    <li>exceed your <a class="govuk-link govuk-link--no-visited-state" href="#text-messages">free text message allowance</a></li>
-    <li>send <a class="govuk-link govuk-link--no-visited-state" href="#letters">letters</a></li>
+    <li>exceed your <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.text_message_pricing') }}">free text message allowance</a></li>
+    <li>send <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.letter_pricing') }}">letters</a></li>
   </ul>
 
   <p class="govuk-body">You’ll only pay for the additional text messages or letters that you send. There’s no monthly charge, no setup fee and no procurement cost.</p>
```
karlchillmaid, line 28: There’s some repetition here about paying for text messages and letters.

Can we pick one of the 2?

Also, can we make sure we include some version of ‘It’s free to send emails through Notify.’

Emails don’t need their own page like text messages and letters, so we should explicitly call out that they’re free on the main pricing page.

karlchillmaid: Or, flip it...

>It’s free to send emails through Notify. There’s also no monthly charge, no setup fee and no procurement cost.
>
>You will only have to pay if you:
>
>* exceed your free text message allowance
>* send letters

app/templates/views/pricing/index.html
```diff
@@ -40,230 +37,7 @@ <h1 class="heading-large">Pricing</h1>
 
   <p class="govuk-body">When you add a new service it will start in <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.trial_mode_new') }}">trial mode</a>.</p>
 
-  <h2 class="heading-medium" id="emails">Emails</h2>
+  <h2 class="heading-medium">Emails</h2>
   <p class="govuk-body">It’s free to send emails through Notify.</p>
```
karlchillmaid, line 42: Should we incorporate this into the intro?

Would that be better than having a 1 line section with its own heading?

I don’t feel strongly either way – maybe see how it looks.

app/templates/views/pricing/index.html
klssmith: There is some content about emails now, but I added after I'd taken the screenshots and forgot to update them

app/templates/views/pricing/index.html
samuelhwilliams: My reaction was also that it looks a bit odd standing alone at the bottom. Might be nice to integrate it further up if possible.

app/templates/views/pricing/letters.html
```diff
@@ -0,0 +1,57 @@
+{% extends "content_template.html" %}
+
+{% from "components/content-metadata.html" import content_metadata %}
+{% from "components/table.html" import mapping_table, row, text_field, row_heading %}
+
+{# Used by the content_template.html layout, prefixes the "navigation" accessible name #}
+{% set navigation_label_prefix = 'Pricing information' %}
+
+{% block per_page_title %}
+  Letters
```
karlchillmaid, line 10: This h1 should include ‘pricing’ somehow.

app/templates/views/pricing/text-messages.html
```diff
@@ -0,0 +1,210 @@
...
+{% from "components/content-metadata.html" import content_metadata %}
+{% from "components/table.html" import mapping_table, row, text_field, field %}
+{% from "components/live-search.html" import live_search %}
+{% from "govuk_frontend_jinja/components/details/macro.html" import govukDetails %}
+
+{# Used by the content_template.html layout, prefixes the "navigation" accessible name #}
+{% set navigation_label_prefix = 'Pricing information' %}
+
+{% block per_page_title %}
+Text messages
```
karlchillmaid, line 11: This h1 should include ‘pricing’ somehow.

karlchillmaid:
`trial-mode.html`, line 37: do we need to change this now that organisations can approve their own requests-to-go-live? Or are we still piloting that feature?

app/templates/views/pricing/index.html
karlchillmaid: @samuelhwilliams I suggested a couple of ways to do this in another comment thread.

app/templates/views/pricing/index.html
karlchillmaid: Yeah – I think it feels a bit lost at the bottom as there’s not enough to say about emails to justify giving it its own heading.

My preference is the second suggestion in this thread, but I’ll leave it to Saima to sort it. 🙂

app/main/views/pricing.py
klssmith: This is now updated

klssmith:
@karlchillmaid We're still piloting the feature to let orgs approve their own go-live request, so I don't think we need to update that content yet - for most services it will still be us reviewing and approving the request

app/main/views/sub_navigation_dictionaries.py
karlchillmaid: For something so important, trial mode info has always felt a little bit lost.

Restructuring our content and navigation hasn’t fully fixed that (yet). It doesn’t really sit with the new Using Notify content – nor is it a feature.

Our plan is to align it more closely with the onboarding journey. An important part of which is understanding Notify’s payment model: what you can send, when you can send it, and when you’re going to need to pay.

You’re right that the current Trial mode page doesn’t say much about pricing and payment – but future iterations will change that.

app/main/views/index.py
klssmith: I've made this change (it's in a new commit at the end) 👍

app/templates/views/pricing/text-message-pricing.html
```diff
@@ -0,0 +1,210 @@
...
+  <p class="govuk-body">When a service has used its annual allowance, it costs {{ sms_rate }} pence (plus VAT) for each text message you send.</p>
+  <p class="govuk-body">You’ll use more free messages, or pay more for each message, if you send <a class="govuk-link govuk-link--no-visited-state" href="#long-text-messages">text messages longer than 160 characters</a>.</p>
+  <p class="govuk-body">You may also use more free messages, or pay more for each message, if you:</p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>use certain <a class="govuk-link govuk-link--no-visited-state" href="#symbols">signs and symbols</a></li>
+    <li>use <a class="govuk-link govuk-link--no-visited-state" href="#accents">accents and accented letters</a></li>
+    <li>send text messages to <a class="govuk-link govuk-link--no-visited-state" href="#international-numbers">international numbers</a></li>
+  </ul>
+
+  <p class="govuk-body">See <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.how_to_pay') }}">how to pay</a>.
```
karlchillmaid, line 34: Not sure we need this on this page anymore.

```suggestion
```

app/templates/views/pricing/index.html
```diff
@@ -38,232 +34,6 @@ <h1 class="heading-large">Pricing</h1>
...
-        {% call row() %}
-          {% call row_heading() %} {{ sheets }} {% endcall %}
-          {{ text_field(second + ' + VAT') }}
-          {{ text_field(first + ' + VAT') }}
-          {{ text_field(international + ' + VAT') }}
-        {% endcall %}
-      {% endfor %}
-    {% endcall %}
-  </div>
+  <p class="govuk-body">When you add a new service it will start in <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.trial_mode') }}">trial mode</a>.</p>
```
karlchillmaid, line 37:
```suggestion
  <p class="govuk-body">When you add a new service it will start in <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.trial_mode') }}">trial mode</a>.</p>
  
  <p class="govuk-body">Find out <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.how_to_pay') }}">how to pay</a>.
```

app/templates/views/pricing/letter-pricing.html
```diff
@@ -0,0 +1,57 @@
...
+  <p class="govuk-body">The cost of sending a letter depends on the postage you choose and how many sheets of paper you need.</p>
+
+  <p class="govuk-body">Prices include:</p>
+  <ul class="govuk-list govuk-list--bullet">
+     <li>paper</li>
+     <li>postage</li>
+     <li>double-sided colour printing</li>
+     <li>C5 size envelopes with an address window</li>
+  </ul>
+
```
karlchillmaid, line 32: I borrowed this line from the `Upload a letter` page, because ought to use it somewhere prominent - and this seems like a sensible place.

```suggestion

<p class="govuk-body">Letters must be 10 pages or less (5 double-sided sheets of paper).</p>

```

karlchillmaid (approved):
Content for the pages looks good.

PR #4614, opened 2023-02-07
Toggle email auth for team members in bulk

samuelhwilliams:
After a service manager enables email auth for the service, show them a
page listing all active (and invited) users for the service so they can
toggle on/off email auth for many users at once. Otherwise they have to
go to the team members page and do each user individually, which will be
a bit painful.

## Todo
- See what JS can be shared/refactored between setAuthTypeForm/emailBrandingOptionsForm/templateFoldersForm


quis:
Interaction design on this looks good to me 👍🏻

tombye (approved):
This all looks good to me 👍🏻 

I've put in some suggestions for changes to the JS, mainly to fix issues that were in the code ported across already so excuse the extra work.

I've read the Python bits and they all seem logical/sensible though I don't write much Python so you might want another opinion. I'm approving this anyway as leaving that up to you.

PR #4619, opened 2023-02-08
Rename some branding-related elements

CrystalPea:
This is so their names better reflect what those elements do.

Also move a couple things around.

Review commit by commit. I appreciate commits could be more atomic for this, I hope it's still manageable to read through as is.


app/main/views/service_settings/email_branding.py
```diff
@@ -196,7 +196,7 @@ def email_branding_request(service_id):
 def email_branding_request_government_identity_logo(service_id):
     branding_choice = request.args.get("branding_choice")
     return render_template(
-        "views/service-settings/branding/email-branding-create-government-identity-logo.html",
+        "views/service-settings/branding/add-new-branding/email-branding-create-government-identity-logo.html",
```
samuelhwilliams, line 199: We have the word `branding` here three times, seems excessive? Could the folder just be called `new` or `add` or `create`. add and new feel redundant, and branding feels redundant too?

app/main/views/service_settings/email_branding.py
```diff
@@ -90,11 +92,18 @@ def email_branding_request(service_id):
 @main.route("/services/<uuid:service_id>/service-settings/email-branding/pool", methods=["GET", "POST"])
 @user_has_permissions("manage_service")
 def email_branding_pool_option(service_id):
+    # TODO: remove this view, it's temporary
+    return redirect(url_for("email_branding_option_preview", service_id=service_id), code=301)
+
+
+@main.route("/services/<uuid:service_id>/service-settings/email-branding/option-preview", methods=["GET", "POST"])
```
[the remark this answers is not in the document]
CrystalPea: I think it is useful - as we also have new branding preview. And we want to signify it's not that, but a preview before you set branding to an existing option.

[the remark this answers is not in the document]
CrystalPea: @karlchillmaid what do you reckon?

app/main/views/service_settings/email_branding.py
CrystalPea: I quite like how the folder describes the journey it contains. It reads very naturally when browsing through the folder structure 😳

samuelhwilliams: It seems verbose to me, but I guess we're both just discussing personal preferences so can probably leave it as-is.

karlchillmaid: My preference is for folder structures and/or URLs to be:

* descriptive
* concise
* free from repetition

When designing this sort of thing, I think we should consider the information contained in the whole thing, rather than trying to make each individual part work in isolation. The path tells a story, and means we can remove any repetition of common words that are already covered further up the folder structure.

So, In this example, I’d go for something like:

```views/service-settings/branding/new/government-identity-logo.html```

I don’t feel strongly either way about `add` vs `new` - we should check if there’s a precedent elsewhere. So long as we’re consistent, I’m happy.

app/main/views/service_settings/email_branding.py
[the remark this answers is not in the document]
karlchillmaid: I think I need a little more context before I can advise on this one.

I think you both make valid points. As I said in my other comment - it’s important for us to be concise.

However we should be careful about having near-duplicate (or very similar) URLs for different things.

samuelhwilliams (approved):

app/main/views/service_settings/email_branding.py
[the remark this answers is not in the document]
CrystalPea: Ok, I checked, and for the other page, header is `Preview your email branding`, but url is `/email-branding/when-you-use-this-branding`.

So the URLs would be quite different, but the URL for the option preview page would be very similar to the header for new branding preview page 🤷🏼

app/main/views/service_settings/email_branding.py
CrystalPea: Ok, I will change it to `/new/`.

PR #4623, opened 2023-02-13
Allow non-emergency 3-digit contact numbers

samuelhwilliams:
NHS COVID Pass want their 'send a file' contact phone number to be 119, but we currently require numbers to be 5 digits or more. This is clearly not the right restriction to have, as 119 is legitimately the contact point for covid enquiries.

Let's allow 3-digit (and technically 4-digit numbers) numbers as points of contact, but specifically exclude emergency numbers that don't seem reasonable to use.


tests/app/main/forms/test_service_contact_details_form.py
```diff
@@ -95,3 +95,28 @@ def test_form_phone_number_validation_fails_with_invalid_phone_number_field(noti
...
         assert form.errors["phone_number"] == ["Must be a valid phone number"]
+
+
+@pytest.mark.parametrize(
+    "short_number, allowed",
+    (
+        ("119", True),
+        ("999", False),
+        (" 999 ", False),
+        ("112", False),
```
quis, line 106: `("(9)99", False)` and `("1-1-2", False)` will cause this test to fail (they will be allowed when they probably(?) shouldn’t be)

quis (approved):

PR #4629, opened 2023-02-16
Update the left-hand navigation menu

karlchillmaid:
This PR updates the left-hand navigation menus for Notify’s static guidance content.

We’re creating a single left-hand menu that:
* always shows the 3 section landing pages - Features, Pricing, Using Notify
* only shows the sub-pages for the selected section

- [x] move file structure to reflect path heirarchy
- [x] restyle subnav links to be blue (rather than black) ~NOT DONE, ASK TOM TO HELP~
- [x] look at `guidance_` prefix for the routes, see if we can/should get rid of that
- [x] make sure all internal links in ./admin/app/templates/views/guidance are url_for rather than hard coded URLs


samuelhwilliams:
First notes from a quick glance through of the actual pages.

## Performance page

This doesn't seem well suited to displaying in a smaller column.

image

## Security page

Bit of an aside, but have noticed that a number of pages have a 'last updated / next review due at' date and wonder if these would make sense for us to show on the security page 🤔

## Using Notify
The main page here has inconsistent capitalisation. The first 'help and guidance' bullets are capitalised but the 'more information' ones aren't. I'm not sure if this is intentional or not. It definitely might be intentional.

image

## Headings
The three left-nav main headings are 'Features', 'Pricing', and 'Using Notify'. In the footer these all match links - but they come under their own headings. 'Features' is under 'About Notify', 'Pricing' is under 'Pricing and payment', and 'Using Notify' is under 'Using Notify'. Also, 'Trial mode' is under 'Using Notify' in the footer but then on the new page 'Trial mode' is under the 'Pricing' left-nav heading. All of the other footer links seem to tie up OK. Are we looking at the footer separately?

app/main/views/index.py
```diff
@@ -372,15 +372,16 @@ def guidance_upload_a_letter():
 @main.route("/trial-mode", endpoint="old_trial_mode")
 @main.route("/features/trial-mode", endpoint="old_trial_mode")
 @main.route("/using-notify/trial-mode", endpoint="old_trial_mode")
+@main.route("/using-notify/guidance", endpoint="old_guidance_index")
 def old_page_redirects():
     redirects = {
         "main.old_roadmap": "main.roadmap",
         "main.old_terms": "main.terms_of_use",
-        "main.information_security": "main.guidance_index",
```
samuelhwilliams, line 379: I know this is a sensible change based on what it was, but I wonder if this should just go to the security page?

app/main/views/sub_navigation_dictionaries.py
```diff
@@ -49,11 +61,23 @@ def pricing_nav():
...
             ],
         },
+        {
+            "name": "Using Notify",
+            "link": "main.using_notify",
+        },
     ]
 
 
 def using_notify_nav():
```
samuelhwilliams, line 76: It looks like we're making a lot of the navs broadly consistent. I wonder if we should have a single nav dictionary and return 'views' of it with each of these functions (basically just drop the sub-nav lists for all of the 'other' top-level dicts)?

samuelhwilliams: I think the code to do this might be a bit gross compared to the minimal gains it gives us though.

karlchillmaid:
Thanks @samuelhwilliams 

Here’s think rationale – Happy to be challenged on any of these 😄 

1. Performance page – we discussed this with @tombye and he’s going to look at fixes for this page. This will probably come in as a separate PR so that we can merge this now. If you think the layout bug is a blocker to this PR being merged we can see if Tom can prioritise working on this sooner.

2. I’m not against date-stamping the security page, but I don’t want us to end up in a situation where we can make a case for date stamping _every_ page. I guess the question we need to answer is ‘What makes this more like Pricing, Roadmap, Privacy or Accessibility and less like Features, Who can use Notify, How to Pay, Billing details or any of the guidance pages?

3. Capitalisation: the first example doesn’t have a lead-in line, the second one does. The first is a bit of a break from the style guide, but it avoids us having to rewrite the links in a convoluted way just to make them part of a natural English sentence. At some point I’d like to redesign this page completely to get away from the long bulleted list which duplicates the nav. Saima and I are thinking about helpful ways to group and introduce the links – but that’s for later.

4. The footer is going to change completely as a separate story. Changing it now to match the LH nav seems like a lot of extra effort given that a) we don’t think the footer gets a lot of use, b) they’ll only be out of sync for a few days.

app/main/views/index.py
karlchillmaid: Would that mean changing it to:

```"main.information_security": "main.features_security",```

?

samuelhwilliams:
@karlchillmaid all good replies and roughly what I expected you'd say 👍

I think the performance page looks a bit bad but if we've got that work in the backlog to address it (and tom/chris are happy with this going out) then I'm fine with it too 👍

app/main/views/index.py
samuelhwilliams: I think `main.guidance_security`, but yes.

quis:
Happy for the performance page to go out as-is. Not convinced it should sit in the left hand nav but don’t want to block this getting merged.

app/main/views/sub_navigation_dictionaries.py
karlchillmaid: @leohemsted if we don’t want to revisit the approach, can I merge?

app/main/views/index.py
karlchillmaid: Done

samuelhwilliams (approved):

PR #4631, opened 2023-02-21
Update the header menu

karlchillmaid:
This PR moves the nav definition out of `admin_template.html` and into `navigation.py`. This means we can use logic to make the code more concise.

## New header menu links

The first 4 links in the header menu should be consistent, whether or not you’re signed in to Notify. The aim is to make the header links work more like a global nav for Notify.

### Signed out

* Support
* Features
* Pricing
* Using Notify
* [Platform admin]
* **Sign in**

### Signed in

* Support
* Features
* Pricing
* Using Notify
* [Platform admin]
* **Your profile**


app/navigation.py
```diff
@@ -128,6 +127,44 @@ class HeaderNavigation(Navigation):
...
+
+        nav_items = [
+            {"href": url_for("main.support"), "text": "Support", "active": self.is_selected("support")},
+            {"href": url_for("main.guidance_features"), "text": "Features", "active": self.is_selected("features")},
+            {"href": url_for("main.guidance_pricing"), "text": "Pricing", "active": self.is_selected("pricing")},
+            {
+                "href": url_for("main.guidance_using_notify"),
+                "text": "Using Notify",
+                "active": self.is_selected("using-notify"),
+            },
```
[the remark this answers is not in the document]
leohemsted: So our python code is formatted automatically 🪄 (using [black](https://github.com/psf/black)), so while this looks differente to the other ones due to line length, there's actually nothing to do here - we've decided as a team "if it passes black's formatting rules, it's good enough"

quis:
Once this has gone in we should update the header in the docs to match – at the moment it looks like this:
image

app/navigation.py
```diff
@@ -128,6 +144,44 @@ class HeaderNavigation(Navigation):
...
+        from app import current_user
+
+        nav_items = [
+            {"href": url_for("main.support"), "text": "Support", "active": self.is_selected("support")},
+            {"href": url_for("main.guidance_features"), "text": "Features", "active": self.is_selected("features")},
+            {"href": url_for("main.guidance_pricing"), "text": "Pricing", "active": self.is_selected("pricing")},
+            {
+                "href": url_for("main.guidance_using_notify"),
+                "text": "Using Notify",
+                "active": self.is_selected("using-notify"),
```
quis, line 157: This one isn’t working – I don’t need *Using Notify* highlighted when I’m in that page or its sub-pages

karlchillmaid: Highlighting _Using Notify_ when in that page or its sub-pages is consistent with the way _Features_, _Pricing_ and _Support_ behave.

leohemsted: wait, we _don't_ need that highlighted? 🤔 is that because it's already highlighted on the left hand nav? should the top nav ever be highlighted for using features/pricing/using notify guidance pages

karlchillmaid: I think it should. To create a relationship between the header and LH nav and to show you which ‘section’ of Notify you’re in.

If we decide not to do this, then that’s a change from what we’ve been doing for the last couple of years.

quis: _Using Notify_ in the **header** should be highlighted it:
- you are on the Using Notify page
- you are on one of its sub pages

_Using Notify_ in the **left hand nav** should be highlighted if
- you are on the Using Notify page

***

In my original comment I think I meant to write _see_ not _need_, sorry for the confusion

karlchillmaid: Thank you @quis!

leohemsted: on _Using Notify_ page:

* header highlighted
* left hand nav highlighted

on _Api documentation_ (using notify sub page):

* header highlighted
* _Using notify_ not highlighted in the left hand nav 
image
image

seems to be working as expected 👍

tombye (approved):
Looks good to me. It needs a rebase (which I'll do now) but once that's done 👍🏻

PR #4633, opened 2023-02-23
Harmonise the guidance navigation with the service navigation

quis:
Making the service navigation and the guidance navigation the same not only has the benefit of consistency but:
- the larger font size is a bit easier to read
- the larger focus area is easier to click

***

– | Before | After
---|---|---
Signed out | Screenshot 2023-02-23 at 10 30 16 | Screenshot 2023-02-23 at 10 30 28
Signed in | Screenshot 2023-02-23 at 10 30 06 | Screenshot 2023-02-23 at 10 29 56
Focus area | image | image


samuelhwilliams (approved):
Looks nice 👍

PR #4635, opened 2023-02-23
Tidy up old redirects

leohemsted:
A little simpler than having the routes _and_ the dict within the endpoints, is to use the `defaults` property of flask routes to populate a redirect_to variable. there's a little nuance and an obscure werkzeug config flag we have to set to stop it doing weird redirects to the wrong place though


klssmith (approved):
This looks good. Took me a while to understand how `application.url_map.redirect_defaults ` works, but I can see we're returning the expected `301` status now that it's set to `False`

app/main/views/index.py
```diff
@@ -19,6 +20,9 @@
 from app.main.views.sub_navigation_dictionaries import features_nav, using_notify_nav
 from app.models.branding import EmailBranding
 
+redirects = Blueprint("redirects", __name__)
+main.register_blueprint(redirects)
```
samuelhwilliams, line 24: 🙌

PR #4637, opened 2023-02-24
Refactor branding code

CrystalPea:
This is not an exhaustive refactor, but since it was meant to be time-capped, and since it contains a bug fix, I decided that it's enough for this time / PR.

Feel free to suggest more, but I suggest that should happen in a separate PR.

Also I was not sure where to put the endpoints that are shared between email branding and letter branding flows. For now I left them in service_settings/email_branding.py

Please review commit by commit.

Trello ticket: https://trello.com/c/AazU7YzJ/185-look-through-branding-code-and-see-what-can-be-refactored


app/main/views/service_settings/letter_branding.py
```diff
@@ -57,20 +57,20 @@ def letter_branding_options(service_id):
...
                     ".letter_branding_option_preview", service_id=current_service.id, branding_option=branding_choice
                 )
             )
-
-        return redirect(
-            url_for(
-                ".letter_branding_upload_branding",
-                service_id=current_service.id,
-                **_letter_branding_flow_query_params(branding_choice=branding_choice),
+        else:
```
samuelhwilliams, line 69: I think I personally would prefer this not indented/in a final else, but it doesn't matter too much I guess.

app/main/views/service_settings/letter_branding.py
```diff
@@ -167,8 +167,9 @@ def letter_branding_option_preview(service_id):
         return redirect(url_for(".service_settings", service_id=current_service.id))
 
     return render_template(
-        "views/service-settings/branding/letter-branding-pool-option.html",
+        "views/service-settings/branding/branding-option-preview.html",
```
samuelhwilliams: nice 🎉

app/templates/views/service-settings/branding/branding-option-preview.html
```diff
@@ -3,17 +3,18 @@
...
+{% set page_title = "Change {} branding".format(branding_type) %}
 
 {% block service_page_title %}
-  {{  page_title }}
+{{ page_title }}
 {% endblock %}
 
 {% block backLink %}
   {{ govukBackLink({
-    "href": url_for('.letter_branding_request', service_id=current_service.id)
```
samuelhwilliams, line 16: It might be nicer to just pass a `back_link` variable in here from Python which has already done the url_for stuff 🤔

app/main/views/service_settings/letter_branding.py
CrystalPea: Why is that? I like it in an else, as it happens if all the others don't, and that makes it explicit.

But I might now know something that you know ☺️

app/templates/views/service-settings/branding/branding-option-preview.html
CrystalPea: Ohh, good idea ☺️

app/main/views/service_settings/letter_branding.py
samuelhwilliams: Partly because I prefer less indentation.

It's sort of more explicit, but also a little bit semantically different. If the final option is inside an `else` block, then the fallback state for `form.validate_on_submit` becomes "render this template". If it's not inside the `else` block, then we are terminating the `if form.validate_on_submit` with a redirect. It's a bit more defensive against future changes. With the code as it stands, it doesn't make a difference because we know that all of the branches of the `if` explicitly `return`. But if one of them didn't, we'd fall out of the validated form flow altogether and end up back in the basic "get page template render".

CrystalPea: ok, that makes sense, I will change it 👍🏼

samuelhwilliams (approved):
👍

PR #4645, opened 2023-03-08
Fix link colour for new service

samuelhwilliams:
We've had a link styling regression where the 'Now go to your dashboard' link when creating a new service is in blue on a blue background (ie basically not visible).

Let's invert the link colour so it's visible again.

## Before
image

## After
image

### DIFF
--- a/app/templates/partials/tour.html
+++ b/app/templates/partials/tour.html
@@ -31,7 +31,7 @@
           Notify delivers the message
         </p>
         {% if help == '3' %}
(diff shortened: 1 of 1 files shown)


idavidmcdonald (approved):
Based on the screenshots, looks good. I'll approve and if any more nuance is needed for the CSS classes then can be followed up subsequently.

tombye (approved):
Looks good to me 👍🏻

PR #4646, opened 2023-03-08
Make performance page less wonky

tombye:
Work to fix visual issues with the performance page.

https://trello.com/c/engQbh1r/282-stop-performance-page-being-wonky

This doesn't make it align exactly across all widths of page but I've been working to the main ones from analytics, which are shown below. Worth noting we can't fix columns overlapping in the first table on the smallest screen but I'm making the assumption that user is on mobile and so can rotate to a wider view (called 'landscape' below).

I'm only showing the top table because the others, and their totals, all work so required much less work and so I'm leaving scrutiny of them to local testing.

### Before

perf_page_desktop_before

### After

perf_page_desktop_after

### Before

perf_page_mobile_portrait_320px_before

### After

perf_page_mobile_portrait_320px_after

### Before

perf_page_mobile_portrait_390px_before

### After

perf_page_mobile_portrait_390px_after

### Before

perf_page_mobile_landscape_568px_before

### After

perf_page_mobile_landscape_568px_after

### Before

perf_page_mobile_landscape_667px_before

### After

perf_page_mobile_landscape_667px_after

## The changes

These changes:
- make the main content column as wide as it can be (it was constrained to an 'ideal sentence width')
- add a `performance` class to scope styles to just this page
- style the content blocks that show totals for a table to be 'table-like' and to line each total up with the data it totals below
- change the HTML to wrap content in paragraphs where it was just `<div>`s, to add some semantics
- reduce the first columns on smaller screens to give more space to the other columns

They don't:
- make it line up perfectly on all screen sizes
- stop the content of some table cells overlapping (I've assumed users seeing this can rotate their device for a wider screen if needed)

## Notes for reviewers

There's a ton of detail in the initial commit about reasons for taking the approach I did. It's also worth working through commit-by-commit.

## Testing

I've tested these changes on the [service manual browser matrix](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2022) + IE11 (because we're still working out whether to drop support for it or not).

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -73,6 +73,7 @@ $govuk-new-link-styles: true;
 @import './views/history';
 @import './views/cookies';
 @import './views/webauthn';
+@import './views/performance';
 
 // TODO: break this up
 @import './app';
--- a/app/assets/stylesheets/views/performance.scss
+++ b/app/assets/stylesheets/views/performance.scss
@@ -1,0 +1,118 @@
+// Styles specific to the performance page
+
+$gutter-with-14px-text: 1.696em;
+$gutter-with-16px-text: 1.484em;
+$gutter-with-19px-text: 1.25em;
+
+.totals__all,
+.totals__set-type {
+  margin: 0; // cancel the vertical margins from using <p> tags
+}
+
+// Separate 'grand total' from others in vertical stack of totals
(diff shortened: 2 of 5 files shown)


CrystalPea (approved):
Nice one :shipit:

PR #4650, opened 2023-03-08
Say ‘Untitled’ instead of ’New’ letter template

quis:
When you add a letter template we give it a default name of ‘New letter template’.

Then when you go to your templates pages you see:
- a link saying ‘New letter template’
- a button saying ‘New template’

In research we’ve seen people confused about how to add another template, and how to get back to the one they’ve previously been working on. We think this is because we’re using the same word (new) to refer to different things.

‘Untitled’ is the word used by Microsoft Windows, Google Docs, Mac OS and probably others to mean ‘thing that exists but hasn’t been named yet’.

This won’t rename any existing templates.

Before | After 
---|---
image | image

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -289,7 +289,7 @@ def _add_template_by_type(template_type, template_folder_id):
 
     if template_type == "letter":
         blank_letter = service_api_client.create_service_template(
-            name="New letter template",
+            name="Untitled letter template",
             type_="letter",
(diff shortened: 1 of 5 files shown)


samuelhwilliams (approved):
Looks good.

I think it would be fine to add a migration that updates existing templates, but fine not to as well. Looks like we have just over 4k templates with the name 'New letter template'.

PR #4651, opened 2023-03-10
Letter template attach pages button and view

CrystalPea:
Adds an "Attach pages" button to the template page. It links to the "Attach pages" page, that contains guidance and a "Choose file" upload button. All hidden behind a feature flag.

Questions: is letter specification right for this use case? Since the attachment doesn't need to have a first page with address block.

### Button:
Screenshot 2023-03-10 at 13 57 07

### New page:
Screenshot 2023-03-10 at 13 30 41

Story card: https://trello.com/c/R5LHg26V/291-add-attach-pages-letter-template-page

### DIFF
--- a/app/templates/views/templates/attach-pages.html
+++ b/app/templates/views/templates/attach-pages.html
@@ -6,7 +6,7 @@
 {% set page_title = "Attach pages" %}
 
 {% block service_page_title %}
-	{{ page_title }}
+    {{ page_title }}
 {% endblock %}
 
 {% block backLink %}
@@ -17,35 +17,35 @@
 
 {% block maincolumn_content %}
 <div class="govuk-grid-row">
-	<div class="govuk-grid-column-five-sixths">
-		{{ page_header(page_title) }}
+    <div class="govuk-grid-column-five-sixths">
+        {{ page_header(page_title) }}
 
-		<p class="govuk-body">Upload a PDF and we’ll print it as part of your letter.</p>
-		<p class="govuk-body">
-			In total, your letter must be 10 pages or less (5 double-sided sheets of paper).
-			You can attach up to 9 pages.
-		</p>
-		<p class="govuk-body">
-			Attached pages will start on the next available page.
-			If you want to start your attachment on a new sheet of paper, add 2 line breaks
-		</p>
-		<p class="govuk-body">
-			Your file must meet our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_upload_a_letter') }}">letter
-				specification</a>.
-		</p>
-		<p class="govuk-body">Page size and layout: A4 portrait (210 × 297 mm)</p>
-		<p class="govuk-body">Maximum file size: 2 MB</p>
-		<p class="govuk-body">Your letter must be 10 pages or less (5 double-sided sheets of paper).</p>
+        <p class="govuk-body">Upload a PDF and we’ll print it as part of your letter.</p>
+        <p class="govuk-body">
+            In total, your letter must be 10 pages or less (5 double-sided sheets of paper).
+            You can attach up to 9 pages.
+        </p>
+        <p class="govuk-body">
+            Attached pages will start on the next available page.
+            If you want to start your attachment on a new sheet of paper, add 2 line breaks
+        </p>
+        <p class="govuk-body">
+            Your file must meet our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_upload_a_letter') }}">letter
+                specification</a>.
+        </p>
+        <p class="govuk-body">Page size and layout: A4 portrait (210 × 297 mm)</p>
+        <p class="govuk-body">Maximum file size: 2 MB</p>
+        <p class="govuk-body">Your letter must be 10 pages or less (5 double-sided sheets of paper).</p>
 
-		<div class="govuk-body">
-			{{ file_upload(
-			form.file,
-			allowed_file_extensions=['pdf'],
-			action=url_for('main.letter_template_attach_pages', service_id=current_service.id, template_id=template_id),
-			button_text='Choose file',
-			show_errors=False
-			)}}
-		</div>
-	</div>
+        <div class="govuk-body">
+            {{ file_upload(
+            form.file,
+            allowed_file_extensions=['pdf'],
+            action=url_for('main.letter_template_attach_pages', service_id=current_service.id, template_id=template_id),
+            button_text='Choose file',
+            show_errors=False
+            )}}
+        </div>
+    </div>
 </div>
 {% endblock %}
--- a/app/templates/views/templates/attach-pages.html
+++ b/app/templates/views/templates/attach-pages.html
@@ -23,19 +23,11 @@
         <p class="govuk-body">Upload a PDF and we’ll print it as part of your letter.</p>
         <p class="govuk-body">
             In total, your letter must be 10 pages or less (5 double-sided sheets of paper).
-            You can attach up to 9 pages.
-        </p>
-        <p class="govuk-body">
-            Attached pages will start on the next available page.
-            If you want to start your attachment on a new sheet of paper, add 2 line breaks
         </p>
         <p class="govuk-body">
             Your file must meet our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_upload_a_letter') }}">letter
                 specification</a>.
         </p>
-        <p class="govuk-body">Page size and layout: A4 portrait (210 × 297 mm)</p>
-        <p class="govuk-body">Maximum file size: 2 MB</p>
-        <p class="govuk-body">Your letter must be 10 pages or less (5 double-sided sheets of paper).</p>
 
         <div class="govuk-body">
             {{ file_upload(
--- a/app/templates/views/templates/attach-pages.html
+++ b/app/templates/views/templates/attach-pages.html
@@ -1,0 +1,51 @@
+{% extends "withnav_template.html" %}
+{% from "components/file-upload.html" import file_upload %}
+{% from "components/page-header.html" import page_header %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+
+{% set page_title = "Attach pages" %}
+
+{% block service_page_title %}
+	{{ page_title }}
+{% endblock %}
+
+{% block backLink %}
+{% if not error %}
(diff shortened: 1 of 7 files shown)


app/templates/views/templates/template.html
```diff
@@ -44,6 +44,21 @@
...
+      "text": "Attach pages",
+      "href": url_for(
+          '.letter_template_attach_pages',
+          service_id=current_service.id,
+          template_id=template.id
+        ),
+      "classes": "govuk-button--secondary govuk-!-margin-right-3"
+      }) }}
+    </div>
+  {% endif %}
```
samuelhwilliams: This button still looks a little cramped underneath, do we maybe want a bit more padding/margin on the bottom?

tests/app/main/views/test_templates.py
```diff
@@ -776,6 +776,47 @@ def test_view_letter_template_does_not_display_send_button_if_template_over_10_p
...
+    page = client_request.get(
+        "main.view_template",
+        service_id=SERVICE_ONE_ID,
+        template_id=fake_uuid,
+        _test_page_title=False,
+    )
+
+    buttons = [b for b in page.select(".govuk-button--secondary") if normalize_spaces((b).text) == "Attach pages"]
+
+    if template_page_count < 10 and template_type == "letter":
```
samuelhwilliams, line 811: I'm a bit tempted to say that it might be nice to split these out into separate tests. I don't _hate_ having one conditional check around some assertions but having two makes me think that we're testing quite different things/cases. Having them separate and with clear names around the good/bad thing under test might be better. I don't mind a bit of repetition/duplication in test cases.

app/templates/views/templates/attach-pages.html
```diff
@@ -0,0 +1,51 @@
+{% extends "withnav_template.html" %}
+{% from "components/file-upload.html" import file_upload %}
+{% from "components/page-header.html" import page_header %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+
+{% set page_title = "Attach pages" %}
+
+{% block service_page_title %}
+	{{ page_title }}
```
samuelhwilliams, line 9: omg it's an actual tab character 😱  haven't seen one of those in a while...

samuelhwilliams: Looks like all of this file is tabs which is pretty unusual for our project.

app/templates/views/templates/template.html
CrystalPea: You are right, the prototype has a bit more padding under the button, I will add some more padding 👍🏼

tests/app/main/views/test_templates.py
```diff
@@ -776,6 +776,64 @@ def test_view_letter_template_does_not_display_send_button_if_template_over_10_p
     assert page.select_one("h1", {"data-error-type": "letter-too-long"})
 
 
+@pytest.mark.parametrize("template_page_count", [6, 10])
+@pytest.mark.parametrize("template_type", ["letter", "email", "sms"])
+def test_view_letter_template_has_attach_pages_button_if_template_below_10_pages_long(
```
samuelhwilliams, line 781: I think there are some sticky edge cases that you should think about here as epic lead (but potentially could be handled separately). If the template is 9 pages long with the blank personalisation it could hit 10 pages with personalisation injected (esp if some of the personalisation is eg an address that spans multiple lines). How do we want to handle that?

I think the button disappearing altogether is maybe weird - possibly we show show the button but it flags an error saying 'you can't attach pages as the letter would be too long', or the button is greyed out, or something. But disappearing altogether might be odd?

app/templates/views/templates/template.html
```diff
@@ -44,6 +44,21 @@
     {% include 'views/templates/_template.html' %}
   </div>
 
+  {% if template.template_type == "letter" and page_count < 10 and show_attach_pages_button %}
+    <div class="bottom-gutter-1-2">
```
samuelhwilliams, line 48: Sorry, I've only just realised - do we still want to use this class? GOV.UK Design System classes I'm more used to using for this sort of thing are `govuk-!-margin-bottom-[1-9]`. @tombye do you have thoughts?

tests/app/main/views/test_templates.py
CrystalPea: Hmm on one side yes, but I think it's actually easier to look through as a single test. It has so much set up that stays the same regardless of which case we look at. I'd keep it as a single test.

samuelhwilliams:
I think on the whole this looks good 👍  A couple of niggly bits and some sidebar questions, but I think broadly this is in a good place. Once everything's addressed please re-review tag me :)

tests/app/main/views/test_templates.py
CrystalPea: > If the template is 9 pages long with the blank personalisation it could hit 10 pages with personalisation injected (esp if some of the personalisation is eg an address that spans multiple lines). How do we want to handle that?

That feels a bit out of scope. Also we don't do anything like this when letter template is exactly 10 pages long - we don't warn users it can easily go over based on placeholders.
I think as long as there is a space to add an attachment, we should offer that option, and then if the whole would breach 10 pages upon sending the letter due to placeholder values, we will throw an error then.

> I think the button disappearing altogether is maybe weird - possibly we show show the button but it flags an error saying 'you can't attach pages as the letter would be too long', or the button is greyed out, or something. But disappearing altogether might be odd?

Hmm... We could show a message instead a button, like: "this letter template is at maximum length, no space left for an attachment". If we do that it will make sense to split those test cases you mentioned, indeed. I will ask @karlchillmaid about the right content.

app/templates/views/templates/template.html
CrystalPea: I just copied what was used already in that file for other bottom items 😳 . If we change it, I think it would be good to change it in all the files that use it at once, which is out of scope for this PR.

tests/app/main/views/test_templates.py
samuelhwilliams: Also, one edge case you definitely need to have in mind is what happens if someone edits the template after attaching some pages and that pushes the page over the 10 pages limit.

app/templates/views/templates/template.html
samuelhwilliams: I'm not particularly bothered if you leave this, then, but I probably don't agree.

If Tom agrees that govuk-!-margin-x is the more correct/preferred thing, then using that means we aren't making the situation worse, and we're being more clear about what we want to be using going forward. We don't have to fix everything all at once before we can start doing things better. I would prefer to use the 'new' way now, and clean up old stuff later.

I'd be fine with leaving this as-is in this PR if you're saying you're happy to go through and replace the bottom-gutter classes separately "soon", but I don't think that's what we're saying

tombye: It would be great if you could use govuk-!-margin-bottom-3` instead, assuming doing so doesn't break anything. There is [a story to remove these classes](https://www.pivotaltracker.com/story/show/171603272) I haven't ported across from pivotal yet, so it will get done but it would be really useful if any new code we introduce could use the override classes from GOVUK Frontend instead, just so we don't increase the work involved.

It's worth noting that the `bottom-gutter-x-x` classes don't just add margin. They also apply a 'clearfix' to the element (see https://github.com/alphagov/notifications-admin/blob/main/app/assets/stylesheets/_grids.scss#L54). This is an old hack to stop any child elements set to float from poking out of the parent. Because of this, I'd try a straight swap but let me know if it breaks anything as I could help fix it.

app/templates/views/templates/attach-pages.html
```diff
@@ -0,0 +1,51 @@
...
+{% block maincolumn_content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-five-sixths">
+        {{ page_header(page_title) }}
+
+        <p class="govuk-body">Upload a PDF and we’ll print it as part of your letter.</p>
+        <p class="govuk-body">
+            In total, your letter must be 10 pages or less (5 double-sided sheets of paper).
+            You can attach up to 9 pages.
+        </p>
```
karlchillmaid, line 27:
```suggestion
        <p class="govuk-body">
            In total, your letter must be 10 pages or less (5 double-sided sheets of paper).
        </p>
```

app/templates/views/templates/attach-pages.html
```diff
@@ -0,0 +1,51 @@
...
+        <p class="govuk-body">
+            Attached pages will start on the next available page.
+            If you want to start your attachment on a new sheet of paper, add 2 line breaks
+        </p>
```
karlchillmaid, line 31:
```suggestion
```

app/templates/views/templates/attach-pages.html
```diff
@@ -0,0 +1,51 @@
...
+        <p class="govuk-body">
+            Your file must meet our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_upload_a_letter') }}">letter
+                specification</a>.
+        </p>
+        <p class="govuk-body">Page size and layout: A4 portrait (210 × 297 mm)</p>
+        <p class="govuk-body">Maximum file size: 2 MB</p>
+        <p class="govuk-body">Your letter must be 10 pages or less (5 double-sided sheets of paper).</p>
```
karlchillmaid, line 38:
```suggestion
```

tests/app/main/views/test_templates.py
CrystalPea: Hmmm... I think the same should happen that we do when they try to edit the template to go over 10 pages by itself. Worth checking if we don't allow them to save such changes, or if we block them when they are sending.

app/templates/views/templates/template.html
CrystalPea: Thanks both, I will try swapping it and let you know what I find.

CrystalPea: All looks good I think.

Screenshot 2023-03-14 at 17 55 04

CrystalPea:
Thanks for content review @karlchillmaid 🙏🏼 

Here is a screenshot of the updated page:

Screenshot 2023-03-14 at 18 12 27

tests/app/main/views/test_templates.py
CrystalPea: I will add a ticket for this.

CrystalPea: Voilla: https://trello.com/c/PqMqV88F/222-for-a-template-with-an-attachment-show-an-error-if-their-total-number-of-pages-is-too-long 😊

samuelhwilliams (approved):

PR #4656, opened 2023-03-16
Add some log statements for CSV uploads

samuelhwilliams:
Should help debug broken uploads if we get support requests.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -131,7 +131,9 @@ def send_messages(service_id, template_id):
     form = CsvUploadForm()
     if form.validate_on_submit():
         try:
+            current_app.logger.info("User %s uploaded %s", current_user.id, form.file.data.filename)
             upload_id = s3upload(service_id, Spreadsheet.from_file_form(form).as_dict, current_app.config["AWS_REGION"])
+            current_app.logger.info("%s persisted in S3 as %s", form.file.data.filename, upload_id)
             file_name_metadata = unicode_truncate(SanitiseASCII.encode(form.file.data.filename), 1600)
             set_metadata_on_csv_upload(service_id, upload_id, original_file_name=file_name_metadata)
             return redirect(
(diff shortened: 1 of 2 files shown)


app/main/views/send.py
```diff
@@ -131,7 +131,9 @@ def send_messages(service_id, template_id):
     form = CsvUploadForm()
     if form.validate_on_submit():
         try:
+            current_app.logger.info("User %s uploaded %s", current_user.id, form.file.data.filename)
```
CrystalPea, line 134: Is there a reason to use this older string interpolation format instead of the newer one? Newer one seems easier to read, as one doesn't have to jump back and forth.

CrystalPea (approved):
Nice one 👏🏼 . Just a question about string interpolation format.

app/main/views/send.py
samuelhwilliams: Great question :)

For logging, generally the 'best practice' is to use this format because of the lazy evaluation of log messages. 'Effort' doesn't have to be expended to calculate the values of the variables unless the log message is actually emitted. The performance impact here is obviously trivial.

A more significant reason to prefer this format is to prepare for integrating something like Sentry or NewRelic for error handling. These tools are able to report on exceptions and log messages, but one of the things they do to identify unique errors is check the message that's been generated. In this case, the thing these tools will check is `User %s uploaded %s"` - so if this was an error, they would all be grouped together into a single report. However if we do the string interpolation upfront (eg using f-strings), then each error would be considered unique on the user id and the filename. They wouldn't be automatically grouped together.

We do have another option for lazy log formatting, which is equally valid with some other pros/cons, which would be to do something like this:

```python
            current_app.logger.info("User {user} uploaded {filename}", extra=dict(user=current_user.id, filename=form.file.data.filename))
```

Note that it's not an f-string. This would still allow sentry/NR/etc to group the messages together as a single error, and is probably more readable. It will also cause the `user` and `filename` to be emitted as key/value pairs in the JSON object. This may or may not be useful/desirable.

PR #4657, opened 2023-03-20
Admin session lifetime

samuelhwilliams:
## Restrict session lifetime for all users, and more restrictive for platform admins
Updated risk assessments have recommended further restricting our admin
session lifetimes. At the moment a session lasts 20 hours, but is
refreshed on every page load and so in theory could last forever if you
use Notify at least once every 20 hours.

We want to restrict lifetimes to 20 hours maximum. For normal users
there is no refreshing required. For platform admins, we also want the
session to expire if the user has been inactive for 30 minutes, with an
absolute maximum lifetime of 20 hours.

---

Also includes a refactor for freezegun to pull in pytest-freezegun, which allows freezing time in fixtures. This means we can stop ignoring `itsdangerous` for freezegun, which I wanted to work for the tests for the above change.

And moves all of the `.json` endpoints to a separate blueprint which will return a 401 if the user's session has expired, as opposed to redirecting to the sign-in page. This is more appropriate for a machine-to-machine API-style view.


quis (commented):
The JSON endpoints wont be hit if someone is inactive (for example tab in the background) on an auto-refreshing page because we started checking [`document.visibilityState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) in https://github.com/alphagov/notifications-admin/pull/2577

I don’t see how the change you’ve made here is applicable to platform admins only. This means it could start causing silent failures for people who have their Notify dashboard open on a big screen in the office, for example.

samuelhwilliams:
Yep, I've (already) messaged Tom to talk about detecting the logout and eg doing a redirect back to the sign-in page but will have to wait until he's back tomorrow.

The permanent session timeout isn't only restricted to platform admins. _All_ accounts will be logged out after 20 hours; that is a new absolute timeout.

Platform admins additionally will be timed out after 30 minutes of inactivity.

app/models/user.py
```diff
@@ -163,6 +163,7 @@ def activate(self):
     def login(self):
         login_user(self)
         session["user_id"] = self.id
+        session["session_start"] = datetime.now(timezone.utc).isoformat()
```
leohemsted, line 166: minor nit: is there any functional difference between `datetime.now(timezone.utc)` and `datetime.utcnow()`? i see we use it in a couple of places but we vastly overwhelmingly use `utcnow` and i'd rather we stick to just one fn across our code if poss

app/notify_flask.py
```diff
@@ -0,0 +1,68 @@
...
+                session["session_start"] = now.isoformat()
+            session["session_expiry"] = self._get_inactive_session_expiry(
+                app, datetime.fromisoformat(session["session_start"])
+            ).isoformat()
+
+        super().save_session(app=app, session=session, response=response)
+
+
+class NotifyAdminFlaskApp(Flask):
+    session_interface = NotifyAdminSessionInterface()
```
leohemsted, line 68: do we need to define this custom class here? can this just be added to the flask app during `app.py::init_app` in the same way as all our other plugins?

leohemsted (commented):
Nice tests

app/models/user.py
samuelhwilliams: There's a big different tbh.

```
In [9]: datetime.datetime.utcnow()
Out[9]: datetime.datetime(2023, 3, 23, 7, 59, 4, 126424)

In [10]: datetime.datetime.utcnow().isoformat()
Out[10]: '2023-03-23T07:59:06.037818'

In [11]: datetime.datetime.now(timezone.utc)
Out[11]: datetime.datetime(2023, 3, 23, 7, 59, 9, 922874, tzinfo=datetime.timezone.utc)

In [12]: datetime.datetime.now(timezone.utc).isoformat()
Out[12]: '2023-03-23T07:59:11.801122+00:00'
```

You are right though, we use `datetime.utcnow()` in a lot more places. But that's probably a bit funky as it means we're always handling naive datetimes. Probably we should prefer datetimes with tzinfo?

Here at least it will mean that the session lifetimes handle DST correctly, I think.

app/notify_flask.py
samuelhwilliams: Yeah we could do that. It feels much the same.

leohemsted: sorry, i'm going to double down on this i think. i'd much rather this was defined in the same place that we set up error handlers, add blueprints, define our login_manager details, jinja environment, etc. Editor integrations aside, that's where I look to understand how our app is configured, we don't have a pattern of expecting things to be defined or configured in application.py, and I don't think this should be separate to those other plugins and extensions I mentioned.

If we want to refactor that to move away from the `init_app` pattern, I'm fine with that (i don't really love that pattern anyway), but I'm unsure if that refactor would be small enough to squeak in to this PR (I might be wrong on that though).

samuelhwilliams: Shrug, sure, if you're hard blocking on this then I'll change it.

app/models/user.py
leohemsted: Oooh interesting, good to know! 

My biggest concern was "will functions like `utc_string_to_aware_gmt_datetime` handle these dates the same", from my basic prodding and probing it appears that the answer is yes, they'll be handled equivalently by that function, which was my big concern as it's used in all of our datetime jinja formatters. Calling `convert_utc_to_bst` on a tz aware timestamp will raise an exception though as it's explicitly expecting a naive timestamp.

I'm still a tiny bit wary of this in general - I don't love us having different types of datetimes floating around that are very hard to distinguish but may behave differently or be surprising when compared/combined/etc. I don't think I'm worried enough to stop anything in this pr going through though, given this timestamp is pretty self contained.

As a quick note, elsewhere, we use `pytz.utc` rather than `timezone.utc` (eg in timezone utils). certainly back in the day datetime's built in timezone support was shocking but I'd like to hope that's improved, and datetimes with those two different UTC representations pass the `__eq__` equivalency.

Note: The DST changeover is a valid concern when handling timestamps in general, but for a UTC timestamp it will (should) be handled correctly regardless of if it's naive or aware, since UTC has no concept of DST.

samuelhwilliams: > Note: The DST changeover is a valid concern when handling timestamps in general, but for a UTC timestamp it will (should) be handled correctly regardless of if it's naive or aware, since UTC has no concept of DST.

Oh duh yeh of course.

samuelhwilliams:
@leohemsted have redone the `.json` endpoints into a separate blueprint that we can return a 401 for when unauthorised rather than redirecting to the sign-in page.

tombye (approved):
I rewrote a lot of the JS originally authored here, to fix some bugs and some problems with the tests so my changes could do with some 👀 but I've given it all another look and happy to 👍🏻 (the JS bits I mean).

app/main/views/sign_in.py
```diff
@@ -82,4 +85,7 @@ def sign_in():
 
 @login_manager.unauthorized_handler
 def sign_in_again():
+    if request.blueprint == JSON_UPDATES_BLUEPRINT_NAME:
+        return abort(HTTPStatus.UNAUTHORIZED)
```
leohemsted, line 89: was gonna ask if this was the best way of doing things rather than having a blueprint specific error handler but saw this is explicitly what they recommend in flask-login 👍

app/__init__.py
```diff
@@ -503,28 +503,27 @@ def setup_blueprints(application):
...
     """
-    from app.main import json_api as json_api_blueprint
+    from app.main import json_updates as json_updates_blueprint
     from app.main import main as main_blueprint
     from app.main import no_cookie as no_cookie_blueprint
     from app.status import status as status_blueprint
 
     main_blueprint.before_request(make_session_permanent)
-    json_api_blueprint.before_request(make_session_permanent)
+    # json_updates_blueprint.before_request(make_session_permanent)
```
leohemsted, line 522: can we get rid of this line?

leohemsted (approved):
just one super minor tweak.

also minor note for the future: it would have been helpful to have all of the renaming `.endpoint_name` to `main.endpoint_name` in a separate commit to make it easier to skip that to make it easier to find the lines you actually changed (the route decorators and flask login code)

app/__init__.py
samuelhwilliams: Oops yes

PR #4660, opened 2023-03-23
Redirect old platform admin search pages

quis:
One more step towards searching everything from one page.

Mainly doing this because my browser autocompletes the old URLs first, and I’d rather be using the new, unified page.


samuelhwilliams:
Ah, I know this is pretty much the opposite of what you've done, but I actually wanted to change these to all use GET instead of POST so that they have nicer URL-sharing and navigation behaviour (can go back in the browser and get the search results without having to re-submit). Changes here: https://github.com/alphagov/notifications-admin/pull/4661. It maybe doesn't help your autocomplete stuff - assuming you can't delete those pages from your history? I know they'd come back as/when you search for services but presumably with a lower priority than the landing search page so possibly that would start to be at the top?

samuelhwilliams (approved):

PR #4661, opened 2023-03-23
Use GET instead of POST for platform admin searches

samuelhwilliams:


quis:
The problem with using `GET` here is that search terms will end up in logs. This might be ok for service names, but it’s definitely not ok for email addresses, which are personally identifiable information.

samuelhwilliams:
Yeah that's a fair point for the user lookup - but with uuid-search and service-search this _should_ be fine, as you say

PR #4666, opened 2023-03-29
Remove date stamp from main Pricing page

karlchillmaid:
It doesn’t really make sense to include the ‘Last updated’ date on the [main pricing page](https://www.notifications.service.gov.uk/pricing), because it doesn’t include any actual prices.

The content metadata for this page is already out of sync with the [text message pricing page](https://www.notifications.service.gov.uk/pricing/text-messages).

Over time this gap will increase and be potentially confusing for users.


tombye (approved):
Makes sense and the changes look good to me 👍🏻

PR #4668, opened 2023-03-30
Make the default email-as-a-password behaviour clearer

karlchillmaid:
The PR updates the content to make the default behaviour for email-as-a-password clearer.

### DIFF
--- a/app/templates/views/guidance/using-notify/send-files-by-email.html
+++ b/app/templates/views/guidance/using-notify/send-files-by-email.html
@@ -23,7 +23,7 @@
   </ul>
 
   <h2 class="heading-medium">Protect personal information about your users</h2>
-  <p class="govuk-body">When a recipient clicks the link in the email you’ve sent them, they have to enter their email address before they can download the file. This security feature is optional.</p>
+  <p class="govuk-body">When the recipient clicks the link in the email, they will be asked to enter their email address. Only someone who knows the recipient’s email address can download the file. This security feature is optional.</p>
   <p class="govuk-body">You can also choose the length of time that a file is available to download.</p>
 
   <h2 class="heading-medium">How to send files by email</h2>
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/using-notify/send-files-by-email.html
```diff
@@ -23,7 +23,7 @@ <h1 class="heading-large">Send files by email</h1>
   </ul>
 
   <h2 class="heading-medium">Protect personal information about your users</h2>
-  <p class="govuk-body">When a recipient clicks the link in the email you’ve sent them, you can ask them to enter their email address. Only someone who knows the recipient’s email address can download the file.</p>
+  <p class="govuk-body">When a recipient clicks the link in the email you’ve sent them, they have to enter their email address before they can download the file. This security feature is optional.</p>
```
karlchillmaid, line 26:
```suggestion
  <p class="govuk-body">When the recipient clicks the link in the email, they will be asked to enter their email address. Only someone who knows the recipient’s email address can download the file. This security feature is optional.</p>
```

saimaghafoor (commented):
Could we say 'they will need to enter their email...' instead of 'they will be asked to enter...'as it's more active?

app/templates/views/guidance/using-notify/send-files-by-email.html
```diff
@@ -23,7 +23,7 @@ <h1 class="heading-large">Send files by email</h1>
   </ul>
 
   <h2 class="heading-medium">Protect personal information about your users</h2>
-  <p class="govuk-body">When a recipient clicks the link in the email you’ve sent them, you can ask them to enter their email address. Only someone who knows the recipient’s email address can download the file.</p>
+  <p class="govuk-body">When the recipient clicks the link in the email, they will be asked to enter their email address. Only someone who knows the recipient’s email address can download the file. This security feature is optional.</p>
```
karlchillmaid, line 26:
```suggestion
  <p class="govuk-body">When the recipient clicks the link in the email, they will need to enter their email address. Only someone who knows the recipient’s email address can download the file. This security feature is optional.</p>
```

samuelhwilliams (approved):

PR #4677, opened 2023-04-06
Refactor `get_errors_for_csv` to be less complex and easier to read

quis:

### DIFF
--- a/app/utils/csv.py
+++ b/app/utils/csv.py
@@ -5,7 +5,7 @@ from app.models.spreadsheet import Spreadsheet
 from app.utils.templates import get_sample_template
 
 
-def get_errors_for_csv(recipients, template_type):  # noqa: C901
+def get_errors_for_csv(recipients, template_type):
 
     errors = []
 
--- a/app/utils/csv.py
+++ b/app/utils/csv.py
@@ -1,5 +1,6 @@
 from notifications_utils.recipients import RecipientCSV
 
+from app.formatters import recipient_count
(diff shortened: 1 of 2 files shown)


app/utils/csv.py
```diff
@@ -7,34 +7,30 @@
 
 def get_errors_for_csv(recipients, template_type):  # noqa: C901
```
samuelhwilliams: We don't really use generators in very many places in the codebase. I don't think turning this function into one is necessarily a complete improvement - they're more annoying to debug and any savings in memory and processing here are negligible.

But if you prefer it, then I would like it to be more obvious at a glance that this function returns a generate. So maybe we can either call this `generate_errors_for_csv` or add a type hint to the return value indicating a generator rather than a list.

quis: What do you think of [7516304](https://github.com/alphagov/notifications-admin/pull/4677/commits/7516304925ac414455e9577d36af28933c7c3c0a)? Also open to just going back to `append`ing to a `list` since that’s maybe introducing more complexity now

samuelhwilliams: Yeah I think we can probably just go back to how it was. We're at most generating 4 errors messages - I don't think this is a particularly impactful area to minimise memory use/allocations. And since we're consuming the full generator immediately we aren't saving any cpu cycles or anything by aborting the processing early.

quis: Done

samuelhwilliams (approved):

PR #4678, opened 2023-04-06
Target even pages with class, not nth-of-type

tombye:
[This pull request on
admin](https://github.com/alphagov/notifications-admin/pull/4648) introduced a visual bug due to its use of
`nth-of-type` assuming `.letter:nth-of-type(even)` would affect even-numbered elements matching this
selector: `div.letter`.

It actually matches even-numbered `div` elements,
filtered by whether they have the `letter` class
or not.

You will be able to match even-numbered pages in
CSS in future (see
https://css-tricks.com/css-nth-of-class-selector/) but until then I've added a class to all pages to
indicate if they are odd or even:

https://github.com/alphagov/notifications-utils/pull/1025

This uses the `page--even` class to fix the bug.

### DIFF
--- a/app/assets/stylesheets/components/letter.scss
+++ b/app/assets/stylesheets/components/letter.scss
@@ -38,7 +38,7 @@ $iso-paper-ratio: 141.42135624%;
     box-shadow: inset 0 0 0 1px $govuk-border-colour;
   }
 
-  &:nth-of-type(even) {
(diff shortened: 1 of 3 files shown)


quis (approved):
Spun up locally, works good

leohemsted (approved):

PR #4681, opened 2023-04-17
Plumbing for Can ask to join a service organisation permission

whpearson:
This sets up the admin interface for updating the can ask to join a service organisation permission which is setup in notification-api.

The PR for that is here https://github.com/alphagov/notifications-api/pull/3769


app/main/views/organisations/index.py
```diff
@@ -442,6 +442,28 @@ def edit_organisation_can_approve_own_go_live_requests(org_id):
     )
 
 
+@main.route("/organisations/<uuid:org_id>/settings/edit-can-ask-to-join-a-service", methods=["GET", "POST"])
+@user_is_platform_admin
+def edit_organisation_can_ask_to_join_a_service(org_id):
+
+    form = YesNoSettingForm(
+        name="Can people ask to join services in this organisation?",
+        enabled="can_ask_to_join_a_service" in current_organisation.organisation_permissions,
```
quis, line 451: Since you’ve written the property on the model you might as well use it 😄 
```suggestion
        enabled=current_organisation.can_ask_to_join_a_service,
```

app/main/views/organisations/index.py
```diff
@@ -442,6 +442,28 @@ def edit_organisation_can_approve_own_go_live_requests(org_id):
...
+    )
+
+    if form.validate_on_submit():
+        if form.enabled.data:
+            current_organisation.update_permissions(["can_ask_to_join_a_service"])
+        else:
+            current_organisation.update_permissions([])
```
quis, line 458: This feels like a bit of a trap. If we add more types of permission in the future we have to remember to update this code. Otherwise when the user removes this one permission they will inadvertently remove all permissions.

quis (commented):
Generally looks really good. Nice work getting up to speed so quick. Just a couple of comments if you don’t mind…

app/main/views/organisations/index.py
quis: Maybe a more robust approach would be to have methods like `Organisation.add_permission()` and `Organisation.remove_permission()`

app/main/views/organisations/index.py
whpearson: True, I didn't have the property for a bit.

app/main/views/organisations/index.py
whpearson: Hmm, true. I'll go with this.

samuelhwilliams (approved):
Nice one - this all looks good and acts as I'd expected locally 🙌 

Good stuff

app/models/organisation.py
```diff
@@ -122,6 +123,7 @@ def __init__(self, _dict):
             self.email_branding_id = None
             self.letter_branding_id = None
             self.can_approve_own_go_live_requests = False
+            self.organisation_permissions = []
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah I think I left this comment before you pushed your change. You have :)

PR #4682, opened 2023-04-19
Move the ‘get ready to send’ button for letter templates

quis:
Before | After
---|---
image | image

In user research we found that this didn’t make it any harder to find.

It has the advantage of:
- being consistent with the position of other buttons on the page which are next to the thing they modify
- being consistent with the appearance of other buttons on the page which are only as wide as the text they contain
- taking up less space, something we’ve lost by adding the sticky ‘Attach pages’ button

### DIFF
--- a/app/templates/views/templates/_letter_template.html
+++ b/app/templates/views/templates/_letter_template.html
@@ -11,7 +11,7 @@
 <div class="template-container {% if current_service.has_permission("extra_letter_formatting") %}template-container--with-attach-pages-button{% endif %}">
   {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) and not letter_too_long %}
     <a href="{{ url_for(".set_sender", service_id=current_service.id, template_id=template.id) }}" class="govuk-link govuk-link--inverse pill-separate-item edit-template-link-get-ready-to-send">
-      Get ready to send<span class="govuk-visually-hidden"> a message using this template</span>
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
This seems to show the 'Change' button weirdly out of position when we see the 'letter too long' error state now 🤔 

image

quis:
@samuelhwilliams should be all fixed now:

image

image

app/templates/views/templates/_letter_template.html
```diff
@@ -6,17 +6,14 @@
...
-        Get ready to send<span class="govuk-visually-hidden"> a message using this template</span>
-      </a>
-    </div>
-  </div>
 {% endif %}
 
 <div class="template-container {% if current_service.has_permission("extra_letter_formatting") %}template-container--with-attach-pages-button{% endif %}">
+  {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) and not letter_too_long %}
+    <a href="{{ url_for(".set_sender", service_id=current_service.id, template_id=template.id) }}" class="govuk-link govuk-link--inverse pill-separate-item edit-template-link-get-ready-to-send">
+      Get ready to send<span class="govuk-visually-hidden"> a message using this template</span>
```
samuelhwilliams, line 14: Maybe this can say letter now instead of message? I missed this when we split out the templates.

quis: Changed in 934991f

PR #4683, opened 2023-04-19
Validate total template length after letter attachment 

samuelhwilliams:
## Commits 1+2
Minor refactors

## Commit 3 
Our print provider can only normally handle letters of 10 or fewer
pages, so let's try to validate that when an attachment is added to a
base template they remain within that limit.

## Show it
image

Ignoring the display format of the error for now as there is a separate ticket about changing this to use an error summary.

Ticket: https://trello.com/c/gjq7kVcv/293-validate-that-letter-template-attachment-10-pages

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -13,7 +13,6 @@ from flask import (
     url_for,
 )
 from flask_login import current_user
-from markupsafe import Markup
 from notifications_python_client.errors import HTTPError
 from notifications_utils import LETTER_MAX_PAGE_COUNT, SMS_CHAR_COUNT_LIMIT
 from notifications_utils.pdf import is_letter_too_long, pdf_page_count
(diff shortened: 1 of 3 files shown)


app/main/views/templates.py
```diff
@@ -967,26 +969,24 @@ def letter_template_attach_pages(service_id, template_id):
...
-                    error=get_letter_validation_error(validation_failed_message, invalid_pages, page_count),
+                    error=get_letter_validation_error(validation_failed_message, invalid_pages, attachment_page_count),
                 )
-            else:
-                raise ex
-        else:
 
-            # TODO in next PR: upload letter to S3
-            pages_content = "1 page" if page_count == 1 else f"{page_count} pages"
+            raise
```
whpearson, line 982: Do you want to raise ex?

app/main/views/templates.py
```diff
@@ -967,26 +969,24 @@ def letter_template_attach_pages(service_id, template_id):
...
 
-            flash(
-                f"You have attached {pages_content} to the end of your letter",
-                "default_with_tick",
-            )
+        template = current_service.get_template(template_id)
```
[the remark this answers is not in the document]
samuelhwilliams: It totally is - good spot, thanks. Will remove.

app/main/views/templates.py
samuelhwilliams: `raise` will by default re-raise the caught exception if nothing else is provided, so I tend to prefer this cleaner syntax. But it's admittedly personal preference.

(there's a linting rule I think we could enable to enforce this but we haven't turned it on - might in the future?)

whpearson (approved):
Looks good now

app/main/views/templates.py
whpearson: I didn't know that behaviour. I don't have strong preferences one way or the other.

PR #4684, opened 2023-04-19
Allow different fields on a template to be updated independently

quis:
This will let us update the `name`, `subject`, and `content` of a template independently.

This doesn’t change any current functionality. In a future pull request we can use this to add a page for renaming letters.

The API seems happy enough to accept a partial object – [the `post_update_template_schema` in the API doesn’t specify any fields as `required`](https://github.com/alphagov/notifications-api/blob/cae3164bfe5bdcdbbbd5117e31f74c3118ef663c/app/template/template_schemas.py#L25-L44). And, as an example, we already don’t update `postage` when we edit a template.

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -184,7 +184,7 @@ class ServiceAPIClient(NotifyAdminAPIClient):
 
     @cache.delete("service-{service_id}-templates")
     @cache.delete_by_pattern("service-{service_id}-template-*")
-    def update_service_template(self, id_, name, content, service_id, subject=None):
+    def update_service_template(self, id_, *, name, content, service_id, subject=None):
         """
         Update a service template.
         """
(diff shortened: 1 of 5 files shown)


app/notify_client/service_api_client.py
```diff
@@ -184,7 +184,7 @@ def create_service_template(self, name, type_, content, service_id, subject=None
 
     @cache.delete("service-{service_id}-templates")
     @cache.delete_by_pattern("service-{service_id}-template-*")
-    def update_service_template(self, id_, name, content, service_id, subject=None):
+    def update_service_template(self, id_, *, name, content, service_id, subject=None):
```
samuelhwilliams: I'd be semi tempted to make them all kwargs. Given the first noun in the method is `service` my brain mildly assumes the first arg would be the service id. But it's the template id.

Either way this is an improvement but I'd also be mildly pro just explicitly having `service_id` then `template_id`.

samuelhwilliams (approved):

app/notify_client/service_api_client.py
quis: Agree, have made all arguments into keyword arguments in [5397390](https://github.com/alphagov/notifications-admin/pull/4684/commits/5397390b26409d2b3fe667ffee53ab2c0cf58e10)

PR #4685, opened 2023-04-20
Display 'last updated at' for org dashboard

samuelhwilliams:
We are going to stop on-demand calculation of up-to-the-second billing data when users visit the org dashboard[^1]. Given this, we need to show an indication of when the data was last generated so that users know if it's a bit out of date.

We'll add a small "Updated at" line to the dashboard.

## example before redesign/words
image

**TODO**: @quis working on the design element so expect the words/location of words to change.

[^1]: https://github.com/alphagov/notifications-api/pull/3775

### DIFF
--- a/app/assets/stylesheets/_app.scss
+++ b/app/assets/stylesheets/_app.scss
@@ -16,7 +16,8 @@
 .column-main {
 
   .heading-large,
-  > .heading-medium {
+  > .heading-medium,
+  > .heading-with-aside--heading-medium {
     margin: govuk-spacing(3) 0 govuk-spacing(4) 0;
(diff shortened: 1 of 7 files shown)


tombye:
I added a commit to redo how the code for the h1 and 'Last updated' works. It includes a bit of tweaking to get their baselines to line up while retaining their position but intends to do it in a way that we could use in future in other places.

tombye (approved):
It's not my area but the Python code is understandable and I added a commit to sort a few things I noticed with the frontend bits. Assuming you're happy with my commit, this looks good to me 👍🏻

PR #4687, opened 2023-04-20
Use flynt to switch to f-strings

samuelhwilliams:
[flynt](https://github.com/ikamensh/flynt) is a project that provides an autoformatter for converting from %-formatted and "".format() strings to modern f-strings.

I've noticed a few PRs lately where we piecemeal convert a few parts of the codebase over to f-strings, so this is an attempt to stop that being a bitty manual process.

The command I used was `flynt .`:
```
Flynt run has finished. Stats:

Execution time:                            2.237s
Files checked:                             286
Files modified:                            87
Character count reduction:                 2966238 (99.99%)

Per expression type:
No old style (`%`) expressions attempted.
`.format(...)` calls attempted:            474/478 (99.2%)
No concatenations attempted.
No static string joins attempted.
F-string expressions created:              437
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -2,3 +2,6 @@
 8a9ee35a0a4d9e1f8a8547af382ff4db630ddd5f
 508d2d2568fa116ab65a5de56284c6bb6c4e46a5
 13422afdfbe9ae9c52100b49d069be4fcdcd3182
+
+# Ignore bulk reformatting to f-strings from flynt
+fa3c94d74db7f03f33837595633ff04b405d1abe
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -405,7 +405,7 @@ def register_errorhandlers(application):  # noqa (C901 too complex)
     def _error_response(error_code, error_page_template=None):
         if error_page_template is None:
             error_page_template = error_code
-        resp = make_response(render_template("error/{0}.html".format(error_page_template)), error_code)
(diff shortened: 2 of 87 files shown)


leohemsted (approved):
curious on what those 4 `.format`s it didnt attempt to convert were?

samuelhwilliams:
```
Not converting code '"default-src 'self' {asset_domain} 'unsafe-inline';""script-src 'self' {asset_domain} *.google-analytics.com 'unsafe-inline' 'unsafe-eval' data:;""connect-src 'self' *.google-analytics.com;""object-src 'self';""font-src 'self' {asset_domain} data:;""img-src 'self' {asset_domain} *.tile.openstreetmap.org *.google-analytics.com"" *.notifications.service.gov.uk {logo_domain} data:;""frame-src 'self' www.youtube-nocookie.com;" . format ( asset_domain = current_app . config [ "ASSET_DOMAIN" ] , logo_domain = current_app . config [ "LOGO_CDN_DOMAIN" ] , )': A variable asset_domain is used multiple times - better not to replace it.
Not converting code '"<{asset_url}>; rel=dns-prefetch, <{asset_url}>; rel=preconnect" . format ( asset_url = f'https://{current_app.config["ASSET_DOMAIN"]}' )': A variable asset_url is used multiple times - better not to replace it.
...
Not converting code '"Send {}" . format ( { "email" : "emails" , "sms" : "text messages" , "letter" : "letters" , } . get ( channel ) )': values starting with '{' are better left not transformed.
```

I think it's these

PR #4688, opened 2023-04-20
Fix icons in high contrast modes

tombye:
This makes our icons work when users have 'high contrast mode' turned on.

https://trello.com/c/iifxHCtS/207-icons-disappear-in-windows-high-contrast-mode

'High contrast mode' mainly means [Windows high contrast mode](https://www.smashingmagazine.com/2022/06/guide-windows-high-contrast-mode/) but it is also a setting in Firefox (see 'Note for reviewers').

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

### Normal

image

### In high contrast mode

image

Note: chevron contrast is lost if user choses a light background

### Normal

image

### In high contrast mode

image

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

Note: Folder's black outline gets lost if user choses a dark background.

#### Normal

image

#### In high contrast mode

image

Note: Folder's black outline gets lost if user choses a dark background.

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

#### Normal

image

#### In high contrast mode

image

### Normal

image

#### In high contrast mode

image

## Notes for reviewers

The screenshots above are from Windows high contrast mode. You can use [Firefox's color modes](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use) to check this too.

I've tested these changes in all the browsers in [our matrix](https://github.com/alphagov/notifications-manuals/wiki/Support-for-browsers,-email-clients-and-assistive-technologies#browsers), and Windows high contrast mode with Edge.

Could do with some 👀 from @quis to make sure everything is positioned correctly and I haven't missed any states.

## Additional

Worth saying that moving the SVGs to be inline:
- saves our users the HTTP requests for those files
- (with the folder icons at least) can also mean one macro can be used for multiple types of icon
- we can delete CSS for the various states because their use of `currentColor` means they just inherit everything from the parent


app/assets/stylesheets/components/vendor/breadcrumbs.scss
```diff
@@ -78,9 +64,31 @@
       }
     }
 
+    // Chevron inserted using a pseudo element, mostly taken from design system's breabcrumbs component
```
quis, line 67:
```suggestion
    // Chevron inserted using a pseudo element, mostly taken from design system's breadcrumbs component
```

tests/app/test_jinja_templates.py
```diff
@@ -20,10 +20,15 @@ def __init__(self, unused_imports, file):
 
 @pytest.mark.parametrize("file", files, ids=[str(path.relative_to(templates_dir)) for path in files])
 def test_for_unused_jinja_imports(client_request, file):
+    # FromImport nodes store their names as strings by default but will use tuples if the import is aliased
```
quis, line 23: Hope this wasn’t too tricky to figure out 😰

quis (approved):
Checked the alignment and different states locally, all looks good to me

tests/app/test_jinja_templates.py
tombye: The error message made it ok actually. Nice one Jinja.

quis: I wrote that error message 😉

PR #4689, opened 2023-04-21
Use `autospec=True` wherever possible

quis:


quis:
Closing this because it’s too much work and there are too many merge conflicts now

PR #4691, opened 2023-04-21
Letter attachment design tweaks

quis:
# Template page

The ‘Attach pages’ button is now sticky and has better spacing around it.

Before | After
---|---
image | image

# Upload attachment page (with error)

The guidance and error banner are now full width (wrapping to a narrower column is only needed for longer passages of text). The page has a back link to match the ‘Upload a letter’ page.

Before | After
---|---
image | image


tombye (approved):
Looks good to me 👍🏻

PR #4692, opened 2023-04-21
Save letter attachments to the database

leohemsted:
Things this PR does:

* saves the attachment to the new bucket in s3. we stick a bunch of stuff in the metadata to keep it available, but i'm not sure how necessary that is given we're also saving everything to the database
* saves an original to the transient bucket in case we need to investigate weird font issues. we already do this for precompiled letters
* saves the details to the database.

Things this PR doesn't do:

* handling errors from the API app - do we expect them to practically happen, or is it okay to display a "something went wrong" page for those. The only one I'm kinda interested in handling separately would be if the template already has an attachment, perhaps.
* significantly overhaul more of the save attachment function. i want to clean up the error handling, make the flow a bit easier to read, remove the calls to pdf_page_count, see if the template page count call can be done in a smarter way, and possibly more

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -100,6 +100,7 @@ from app.notify_client.events_api_client import events_api_client
 from app.notify_client.inbound_number_client import inbound_number_client
 from app.notify_client.invite_api_client import invite_api_client
 from app.notify_client.job_api_client import job_api_client
+from app.notify_client.letter_attachment_client import letter_attachment_client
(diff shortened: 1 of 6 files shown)


samuelhwilliams (commented):
Tested locally and all works as expected. Would be good to handle the error for re-uploading if there's already an attachment - especially since we already get an explicit error code back for it.

Agree some tidying up would be nice once the functionality is all in place - could be done later.

samuelhwilliams (approved):

PR #4698, opened 2023-04-27
Clear templates cache when uploading attachment

samuelhwilliams:
When we attach a PDF to a letter template, we need to invalidate that cache so that the `letter_attachment` attribute on templates can get updated.


leohemsted (approved):
oops, good spot

PR #4699, opened 2023-04-27
Save letter attachment refactor

leohemsted:
This isn't the end of the error work or even the beginning - there's lots to do there on using the govuk component and behaving more consistently with how we handle other file upload errors. But I was doing a little tidy up to the code flow because it was making me mad, and ended up tweaking a couple of bits while I was there 💅

This was just an attempt to make the code a bit more readable by using exceptions as control flow. Previously we were either calling  `_invalid_upload_error`, which would render the page, or we were adding to `forms.file.errors` and just flowing down the function and then at the end we'd turn that in to an error object

The only observable changes this refactor should make are:

* minor tweaks in wording on file size errors (now consistent with errors elsewhere within notify)
* errors where we don't have a title and detail component now have a generic title as the h1 and the error message is now a detail equivalent 

| before | after
|-|-|
| image | image |
|image | image |

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1358,7 +1358,7 @@ class CsvUploadForm(StripWhitespaceForm):
         validators=[
             DataRequired(message="Please pick a file"),
             CsvFileValidator(),
-            FileSize(max_size=10e6, message="File must be smaller than 10Mb"),  # 10Mb
+            FileSize(max_size=10 * 1024 * 1024, message="File must be smaller than 10MB"),
         ],
     )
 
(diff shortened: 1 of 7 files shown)


app/main/forms.py
```diff
@@ -1854,6 +1854,7 @@ class PDFUploadForm(StripWhitespaceForm):
         validators=[
             FileAllowed(["pdf"], "Save your letter as a PDF and try again."),
             DataRequired(message="You need to choose a file to upload"),
+            FileSize(max_size=(2 * 1024 * 1024), message="File must be smaller than 2MB"),
```
samuelhwilliams, line 1857: Technically this is Mebibytes (MiB) rather than Megabytes (MB) so this could be changed to be more consistent with the other FileSize validator.

```suggestion
            FileSize(max_size=(2e6), message="File must be smaller than 2MB"),
```

samuelhwilliams (approved):

PR #4704, opened 2023-05-04
Update link guidance

karlchillmaid:
Updates link guidance in line with the new guidance page: https://www.notifications.service.gov.uk/using-notify/links-and-URLs

This is a short-term update, until we redesign the edit page guidance.

### DIFF
--- a/app/templates/partials/templates/guidance-links.html
+++ b/app/templates/partials/templates/guidance-links.html
@@ -2,9 +2,17 @@
 
 <h2 class="heading-medium">Links and URLs</h2>
 <p class="bottom-gutter-1-3">
-  Always use full URLs, starting with https://
+  If the recipient is not expecting to receive an email from you, write the URL in full, starting with https://
 </p>
 {{ govukInsetText({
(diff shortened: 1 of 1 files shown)


app/templates/partials/templates/guidance-links.html
```diff
@@ -2,9 +2,17 @@
...
 <h2 class="heading-medium">Links and URLs</h2>
 <p class="bottom-gutter-1-3">
-  Always use full URLs, starting with https://
+  If the recipient is not expecting to receive an email from you, write the URL in full, starting with https://
 </p>
 {{ govukInsetText({
   "text": "Apply now at https://www.gov.uk/example",
   "classes": "govuk-!-margin-top-0"})
 }}
+<p class="bottom-gutter-1-3">
```
karlchillmaid, line 11: @samuelhwilliams does this paragraph need the `bottom-gutter-1-3` class? I wasn’t sure.

samuelhwilliams: What you've done is in line with what we do elsewhere on the page, so I think you've done the right thing 👍

PR #4705, opened 2023-05-04
Make a separate Jinja template for viewing each kind of Notify template 

quis:


app/templates/views/templates/_letter_template.html
```diff
@@ -0,0 +1,50 @@
...
+      "href": url_for(
+          '.letter_template_attach_pages',
+          service_id=current_service.id,
+          template_id=template.id
+        ),
+      "classes": "govuk-button--secondary"
+      }) }}
+    </div>
+  </div>
+{% endif %}
```
samuelhwilliams, line 50: missing a line-ending here 🙃

samuelhwilliams (approved):
Only spotted the one styling problem - otherwise looks good 👍 

The "Attach pages" button has lost some spacing between it and the letter preview

image

app/templates/views/templates/_letter_template.html
quis:
```suggestion
{% endif %}

```

PR #4706, opened 2023-05-04
Adding manage attachment page

whpearson:
This is the first bit of work on the manage attachment page

It adds a deletion link that soft deletes the attached page,

The button to upload a different file will be a further bit of work as will displaying the attachment

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -98,7 +98,6 @@ def view_template(service_id, template_id):
         return redirect(url_for(".set_sender", service_id=service_id, template_id=template_id))
 
     page_count = get_page_count_for_letter(template)
-
     return render_template(
         "views/templates/template.html",
         template=get_template(
@@ -118,6 +117,7 @@ def view_template(service_id, template_id):
         letter_too_long=is_letter_too_long(page_count),
         letter_max_pages=LETTER_MAX_PAGE_COUNT,
         page_count=page_count,
+        show_manage_attachment_button="letter_attachment" in template and template["letter_attachment"] is not None,
     )
 
 
@@ -930,7 +930,6 @@ def letter_template_attach_pages(service_id, template_id):
     template = current_service.get_template(template_id)
     form = PDFUploadForm()
     error = {}
-
     if form.validate_on_submit():
         try:
             return _process_letter_attachment_form(service_id, template, form)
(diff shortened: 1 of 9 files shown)


app/main/views/templates.py
```diff
@@ -940,10 +939,67 @@ def letter_template_attach_pages(service_id, template_id):
...
+@main.route("/services/<uuid:service_id>/templates/<uuid:template_id>/attach-pages/delete", methods=["GET", "POST"])
+@user_has_permissions("manage_templates")
+@service_has_permission("extra_letter_formatting")
+def letter_template_remove_pages(template_id, service_id):
+    template = current_service.get_template(template_id)
+    error = {}
+
+    if "letter_attachment" not in template or template["letter_attachment"] is None:
+        abort(404)
+    else:
```
samuelhwilliams, line 974: Might be nice to just drop down an indentation level here as well 👍 

I know some IDEs can be a bit unhappy because they don't recognise that abort(404) stops executing the rest of the function, but I think it's better than having everything else indentend.

app/main/views/templates.py
```diff
@@ -940,10 +939,67 @@ def letter_template_attach_pages(service_id, template_id):
...
+
+        if request.method == "POST":
```
samuelhwilliams, line 976: Hmmm ... I can see we do this in quite a few places, so understand why you might've done this, but does this leave the user open to CSRF attacks? I think we normally use flask-wtforms to protect us from CSRF issues, but I'm not sure if flask can still validate CSRF when we aren't using forms 🤔

samuelhwilliams (commented):
Nice work on this 🙌  tested the functionality locally and all looks good.

A few of optional nits/comments for you to think about and one CSRF-y question. After that I think this is good.

app/main/views/templates.py
whpearson: Thanks for picking this up. Hmm, we are using forms here, the flash confirmation is a form with a CSRF. I've tried to remove the CSRF token (in chrome), and it errors so it is checking it. 

I also couldn't post using curl (removing the CSRF token), so I think we are okay. It looks like all views are checked for csrf automatically. https://flask-wtf.readthedocs.io/en/0.15.x/csrf/

It is hard to know that these invisible things are working as intended.

samuelhwilliams: Ah yes, the banner macro makes a form with CSRF token for the delete flash messages. Nice, this is all good then 👍 

Part of me thinks we should enable CSRF in tests at some point but that's a separate issue.

samuelhwilliams (approved):
Nice work 🙌

PR #4708, opened 2023-05-05
Don’t refer to internal property in permissions code

quis:
Using properties marked as internal (prefixed with an underscore) should be avoided where possible.

This code is nicer if the methods refer to the same properties as the external interface of the class.


whpearson (approved):
Makes sense 👍

PR #4709, opened 2023-05-05
Send s3 url for letter attachment to template preview

CrystalPea:
If letter template has an attachment.

### DIFF
--- a/app/s3_client/s3_letter_upload_client.py
+++ b/app/s3_client/s3_letter_upload_client.py
@@ -89,6 +89,16 @@ def get_letter_metadata(service_id, file_id):
     return LetterMetadata(s3_object["Metadata"])
 
 
+def get_letter_attachment_url(service_id, attachment_id):
+    file_location = get_transient_letter_file_location(service_id, attachment_id)
+    s3 = resource("s3")
+    return s3.meta.client.generate_presigned_url(
+        ClientMethod="get_object",
+        Params={"Bucket": current_app.config["S3_BUCKET_LETTER_ATTACHMENTS"], "Key": file_location},
+        ExpiresIn=3600,
+    )
+
+
 def upload_letter_attachment_to_s3(data, *, file_location, page_count, original_filename):
     # Use of urllib.parse.quote encodes metadata into ascii, which is required by s3.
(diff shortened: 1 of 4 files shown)


app/template_previews.py
```diff
@@ -23,6 +24,11 @@ def from_database_object(cls, template, filetype, values=None, page=None):
             "values": values,
             "filename": current_service.letter_branding.filename,
         }
+        if template["letter_attachment"]:
+            data["template"]["letter_attachment"]["s3_url"] = get_letter_attachment_url(
```
samuelhwilliams, line 28: If we do use presigned URLs then maybe let's just call this `url` rather than `s3_url` as in theory the `s3` part of it doesn't matter

app/s3_client/s3_letter_upload_client.py
```diff
@@ -89,6 +89,16 @@ def get_letter_metadata(service_id, file_id):
     return LetterMetadata(s3_object["Metadata"])
 
 
+def get_letter_attachment_url(service_id, attachment_id):
+    file_location = get_transient_letter_file_location(service_id, attachment_id)
+    s3 = resource("s3")
+    return s3.meta.client.generate_presigned_url(
+        ClientMethod="get_object",
+        Params={"Bucket": current_app.config["S3_BUCKET_LETTER_ATTACHMENTS"], "Key": file_location},
+        ExpiresIn=3600,
```
samuelhwilliams, line 98: `3600` is probably far longer than it needs to be. It shouldn't matter much but can we restrict it even more - 30 seconds or less?

samuelhwilliams (commented):
Pending some conversation with @sakisv on if we are OK with the presigned approach.

app/s3_client/s3_letter_upload_client.py
CrystalPea: Done :). I went with 30 seconds as this is maximum synchronous request time.

app/template_previews.py
CrystalPea: Ok, I changed it.

samuelhwilliams (approved):

CrystalPea:
This is not needed anymore - it actually makes sense to get the file straight in template preview (it already accesses most letter-related buckets) and it's easy to do

PR #4710, opened 2023-05-09
clean static build directores when running npm build

leohemsted:
clean static build directories before rebuilding to avoid any old deprecated artifacts being left over in local directories and confusing/causing unit test failures

notably this would cause errors when we updated our fonts if a specific font file was no longer generated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@babel/core": "7.4.0",
     "@babel/preset-env": "7.4.2",
     "better-npm-audit": "3.7.3",
+    "del": "<7.0.0",
     "gulp": "4.0.2",
     "gulp-add-src": "1.0.0",
     "gulp-babel": "8.0.0",
--- a/tests/app/main/views/test_index.py
+++ b/tests/app/main/views/test_index.py
@@ -223,7 +223,6 @@ def test_terms_page_has_correct_content(client_request):
 
 
 def test_css_is_served_from_correct_path(client_request):
-
     page = client_request.get("main.guidance_api_documentation")  # easy static page
 
     for index, link in enumerate(page.select("link[rel=stylesheet]")):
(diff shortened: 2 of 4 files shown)


muzakh (approved):
lgtm

quis (commented):
Would it be better if we cleaned the fonts and error pages<sup>1</sup> and the images<sup>2</sup> directories as part of the `npm run` tasks?

That way no part of the code could rely on assets which would not be present in a clean install of the app.

1. https://github.com/alphagov/notifications-admin/blob/ed0484437e70a05f51cd4aa456d6f12326c48b39/gulpfile.js#L45-L62
2. https://github.com/alphagov/notifications-admin/blob/ed0484437e70a05f51cd4aa456d6f12326c48b39/gulpfile.js#L174-L180

leohemsted:
@samuelhwilliams been hoping to get back to it but needed to chat to Tom about how to approach this and he's off on hols

package.json
```diff
@@ -34,6 +34,7 @@
     "@babel/core": "7.4.0",
     "@babel/preset-env": "7.4.2",
     "better-npm-audit": "3.7.3",
+    "del": "<7.0.0",
```
leohemsted, line 37: del 7.0 and greater is EMS only rather than commonjs so would need to rewrite our gulpfile and change all our requirements if we want to upgrade

tests/app/main/views/test_index.py
```diff
@@ -389,7 +389,7 @@ def test_font_preload(
 
     preload_tags = page.select('link[rel=preload][as=font][type="font/woff2"][crossorigin]')
 
-    assert len(preload_tags) == 2, "Run `npm run build` to copy fonts into app/static/fonts/"
+    assert len(preload_tags) == 2, "Delete all files from app/static/fonts/ and run `npm run build` to rebuild fonts"
```
quis, line 392:
```suggestion
    assert len(preload_tags) == 2, "Run `npm run build` to rebuild fonts"
```

quis (approved):
Tested locally.

Before:
```
ls -R app/static/ | wc -l
     273
```

After 
```
npm run build 
ls -R app/static/ | wc -l
     125
```

Should help avoid some unexpected weirdness 👍🏻

PR #4712, opened 2023-05-11
Submit support tickets without notify type

samuelhwilliams:
Support/feedback tickets could be either technical or non-technical, but at the moment they all come through as non-technical initially. This can make it a bit harder to know whether something's been triaged.

Let's submit support requests without a notify ticket type. It's then more apparent that no initial triage has occurred and that someone needs to look at the ticket.

Relies on: https://github.com/alphagov/notifications-utils/pull/1034


tests/app/main/views/test_find_users.py
```diff
@@ -338,11 +338,12 @@ def test_archive_user_does_not_create_event_if_user_client_raises_unexpected_exc
     mocker,
     mock_events,
 ):
-    with pytest.raises(Exception):
+    mocker.patch("app.main.views.find_users.user_api_client.archive_user", side_effect=ValueError())
```
whpearson, line 341: Is this a separate bug you found?

samuelhwilliams: yeah

samuelhwilliams: Didn't mean to include it in this PR (sorry), but might as well get it in now that I have 🤷

PR #4713, opened 2023-05-11
Don't 500 if invalid message_type

samuelhwilliams:


whpearson (approved):
+1 for better warnings. Was this a sentry bug?

samuelhwilliams:
Yep, it was 👍

PR #4714, opened 2023-05-11
Update accessibility statement

karlchillmaid:
Update estimated completion dates

@tombye will complete this in late May.


tombye:
This is out of date. It was serving as a useful reminder to review the statement but we did that recently so I'm closing this.

PR #4715, opened 2023-05-12
Fix how copy to clipboard announces updates

tombye:
Fix for this issue:

https://trello.com/c/58NoKaMt/380-copy-to-clipboard-doesnt-work-with-screenreaders

Basically, a fix for some problems with the changes in [the last attempt to fix the fact that users of the NVDA and JAWS screen readers didn't hear anything to confirm something had happened when you clicked the button](https://github.com/alphagov/notifications-admin/pull/3591).

## Problems with the last changes

The existing implementation mixed announcements for its status and the new content of the button when you clicked the button on Voiceover. On NVDA and JAWS no status was announced at all.

I think this was because:
1. when the component updates, all its HTML is replaced rather than changed
2. because all the HTML is replaced, focus is lost so the button is re-focused

Number 1. included replacing a live-region, which I think broke it in NVDA and JAWS (correctly).

Number 2. meant the content of the button and the status update, when it worked, both needed announcing at the same time. In NVDA and JAWS the button always won, probably because focus usually comes from an interaction so should win.

This commit tries to fix these problems by:
- changing the HTML when it updates rather than replacing it
- changing the content and class attribute of the button instead of the whole thing

This seems to fix the issues listed here on Voiceover, NVDA and JAWS.

Final note: I changed the contents of the live region to say 'use button' from 'press button' because you can move to that text and, when you do, 'press button' can sound like what you hear when you move to a button and so is confusing.

## Copy to clipboard with these changes

You start with it looking like this:

copy-to-clipboard-default-state

Clicking the button once changes it to this:

copy-to-clip-board-after-copy

Clicking the button again changes it back to this:

copy-to-clipboard-default-state

[On JAWS, the most used screen reader, that process sounds like this](https://drive.google.com/file/d/1nQg-wJWcNGM6WFCGsnVEDNVzzMGHd24r/view?usp=sharing) (Google Drive link)

## Notes for reviewers

I acknowledge that testing this is hard, because most of our team don't have my familiarity with screen readers. That being the case, I suggest reviewers:
- try using the copy to clipboard component while using a screen reader if they are comfortable doing so
- if not, just review the code changes

### DIFF
--- a/app/assets/javascripts/copyToClipboard.js
+++ b/app/assets/javascripts/copyToClipboard.js
@@ -6,23 +6,57 @@
   Modules.CopyToClipboard = function() {
 
     const states = {
-      'valueVisible': (options) => `
-        <span class="copy-to-clipboard__value">${options.valueLabel ? '<span class="govuk-visually-hidden">' + options.thing + ': </span>' : ''}${options.value}</span>
-        <span class="copy-to-clipboard__notice govuk-visually-hidden" aria-live="assertive">
-          ${options.onload ? '' : options.thing + ' returned to page, press button to copy to clipboard'}
-        </span>
-        <button type="button" class="govuk-button govuk-button--secondary copy-to-clipboard__button--copy">
-          Copy ${options.thing} to clipboard${options.name ? '<span class="govuk-visually-hidden"> for ' + options.name + '</span>' : ''}
-        </button>
-      `,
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
I have tested this with voiceover on mac and it does announce better - nice work 🙌

PR #4716, opened 2023-05-15
convert radios in app/templates/views/manage-users/permissions.html

BlessedDev:
This PR migrates the radio buttons on manage-users/permissions.html page to GOVUK FE as stipulated in [this](https://trello.com/c/clOirfB6/140-migrate-all-remaining-radios-to-govuk-frontend-except-those-for-moving-templates-folders) ticket.
Subsequent PRs will deal with the other files listed on the ticket.
I have attached before and after pictures
permissions-a-before
permissions-a-after
permissions-b-before
permissions-b-a
<img width=
fter" src="https://github.com/alphagov/notifications-admin/assets/32799090/c1f95eb0-1db3-4417-b2a9-5ea8695abb82">
create-api-key-before
/c8b69d93-4338-467e-9152-6f8d873aa884">

The opacity for the GOVUK radio component is currently set to 0.5 which cannot be overridden, hence the hint texts are lighter than before. A separate ticket will be raised to address that issue.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -842,14 +842,31 @@ class AuthTypeForm(StripWhitespaceForm):
 
 
 class BasePermissionsForm(StripWhitespaceForm):
-    def __init__(self, all_template_folders=None, *args, **kwargs):
+    def __init__(self, all_template_folders=None, disable_sms_auth=False, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.folder_permissions.choices = []
+        self.disable_sms_auth = disable_sms_auth
         if all_template_folders is not None:
             self.folder_permissions.all_template_folders = all_template_folders
             self.folder_permissions.choices = [
                 (item["id"], item["name"]) for item in ([{"name": "Templates", "id": None}] + all_template_folders)
             ]
+
+        # In a scenario where there is no mobile number for the user
+        # the option to select sms_auth is disabled
+        if self.disable_sms_auth:
+            self.login_authentication.param_extensions = {
+                "items": [
+                    {
+                        "hint": {
+                            "text": "Not available because this team member has not added a "
+                            "phone number to their profile"
+                        },
+                        "disabled": True,
+                    },
+                    {"checked": True},
+                ]
+            }
 
     folder_permissions = GovukCollapsibleNestedCheckboxesField("Folders this team member can see", field_label="folder")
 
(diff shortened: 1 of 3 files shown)


app/main/forms.py
```diff
@@ -863,6 +863,31 @@ def __init__(self, all_template_folders=None, *args, **kwargs):
...
+                        "text": "Not available because this team member has not added a "
+                        "phone number to their profile"
+                    },
+                    "disabled": True,
+                },
+                {"checked": True},
+            ]
+        },
+        validators=[DataRequired()],
+    )
```
samuelhwilliams, line 889: I think we might be able to have a cleaner approach if we stick to having the single field `login_authentication` for this. 

In the __init__ of the form we can pass in some argument (the `user` maybe?) and, based on that, add the `param_extension` to the field.

All of the other arguments to the `GovukRadiosField` are the same so I think we're just duplicating that in this approach, which feels like it could lead to inconsistency in the future.

We also haven't updated any other parts of the code to know about the `login_authentication_for_no_mobile_number` field (and I don't think we should have to, hence maybe just sticking to `login_authentication`).

What do you think / did you try that and have a problem?

samuelhwilliams: You should be able to find some similar examples in other form `__init__` methods - but let me know if you want to chat about it or pair :)

samuelhwilliams (approved):
Looks good 👍  Nice work

tests/app/main/views/test_manage_users.py
```diff
@@ -491,16 +491,14 @@ def test_manage_users_page_does_not_links_to_user_profile_page_if_user_only_invi
             None,
             """
             Text message code
-            Not available because this team member has not added a
-            phone number to their profile
-        """,
+            """,
```
quis, line 494: It would be nice to keep the test coverage for the hint text.

This might mean updating the selector here:
https://github.com/alphagov/notifications-admin/blob/252b3ad558b3782a89eddc6799a0f9f239f63df8/tests/app/main/views/test_manage_users.py#L529-L531

Instead of:
```python
page.select_one("label[for=login_authentication-0]").text
```

You should be able to do:
```python
page.select_one("label[for=login_authentication-0]").parent.text
```

Based on the HTML which looks like
```html
<div class="govuk-radios__item">
  <input class="govuk-radios__input" id="key_type-0" name="key_type" type="radio" value="normal" disabled="" aria-describedby="key_type-0-item-hint">
  <label class="govuk-label govuk-radios__label" for="key_type-0">
    Live – sends to anyone
  </label>
  <div id="key_type-0-item-hint" class="govuk-hint govuk-radios__hint">
    Not available because your service is in <a class="govuk-link govuk-link--no-visited-state" href="/features/trial-mode">trial mode</a>
  </div>
</div>
```

BlessedDev: Just seen this. Good suggestion, I will address this in a separate PR.

tombye: @BlessedDev put this fix in with https://github.com/alphagov/notifications-admin/pull/4720 so marking this as resolved.

PR #4719, opened 2023-05-16
Update formatting.html

saimaghafoor:
Update formatting page to include guidance for markdown and what it each component can be used for.

### DIFF
--- a/app/main/views/sub_navigation_dictionaries.py
+++ b/app/main/views/sub_navigation_dictionaries.py
@@ -103,7 +103,7 @@ def using_notify_nav():
                     "link": "main.guidance_email_branding",
                 },
                 {
-                    "name": "Formatting",
+                    "name": "Formatting emails and letters",
                     "link": "main.guidance_formatting",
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
I've pulled these changes into https://github.com/alphagov/notifications-admin/pull/4738

samuelhwilliams:
I've taken these changes out of #4738 so we can keep this around. It will need some updating but we can go over that together :)

samuelhwilliams:
@saimaghafoor Here's what this branch looks at for now :)

One thing - I think we need to update 'Formatting' in the sidebar to match the h1 from the page, 'Formatting emails and letters'?

FireShot Capture 001 - Formatting emails and letters – GOV UK Notify - notify localhost

samuelhwilliams:
Did you want to have a go at it or shall we pair? :)

saimaghafoor:
> 

I think defo pair!

samuelhwilliams (approved):

PR #4721, opened 2023-05-16
Make 'Choose different file' functional

whpearson:
This makes the 'choose different file' button work by making it post to the attach-pages with an existing attachment, first archive the old attachment and then upload a new attachment.

It then redirects to the template page, as that is currently what happens when you upload an attachment.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -961,11 +961,13 @@ def letter_template_attach_pages(service_id, template_id):
     )
 
 
-@main.route("/services/<uuid:service_id>/templates/<uuid:template_id>/attach-pages/delete", methods=["GET", "POST"])
+@main.route("/services/<uuid:service_id>/templates/<uuid:template_id>/attach-pages/edit", methods=["GET", "POST"])
 @user_has_permissions("manage_templates")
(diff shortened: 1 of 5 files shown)


app/main/views/templates.py
```diff
@@ -966,6 +966,8 @@ def letter_template_attach_pages(service_id, template_id):
 @service_has_permission("extra_letter_formatting")
 def letter_template_remove_pages(template_id, service_id):
     template = current_service.get_template(template_id)
+    form = PDFUploadForm()
+
```
CrystalPea, line 970: If this view deals with both changing and deleting the attachment, should the name and url change to something like `/manage`, and view name change accordingly?

CrystalPea: For example for reply-to email addresses the url for page where one can change or delete an email address is `/edit`

CrystalPea (commented):
Good work, as per comment view url and name might need a rethink.

Also we should have a test that the new upload button links to the right place.

app/main/views/templates.py
whpearson: I went with /edit as that made sense to me.

CrystalPea (approved):

PR #4722, opened 2023-05-17
Combine platform admin search fields; add organisations

samuelhwilliams:
## Dependencies
 * [ ] https://github.com/alphagov/notifications-api/pull/3804

## Before
image

## After
image

image


app/templates/views/platform-admin/search.html
```diff
@@ -1,68 +1,127 @@
...
+          <h2 class="govuk-heading-m">Users</h2>
+          <nav class="browse-list">
+            <ul>
+            {% for user in users %}
+              <li class="browse-list-item">
+                <a href="{{url_for('main.user_information', user_id=user.id)}}" class="govuk-link govuk-link--no-visited-state browse-list-link">{{ user.email_address }}</a>
+                <p class="browse-list-hint">{{ user.name }}</p>
+              </li>
+              {% if not loop.last %}
+                <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
```
tombye, line 51: Can we make the line between items in CSS instead of a `<hr>`? It doesn't look like `<ul>`s can contain semantic tags other than `<li>`s (see https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element). I'd suggest adding a class to the 'browse-list' and adding the spacing with padding (for spacing above the line) and margin (for spacing below the line) with something like:
```css
.browse-list--tabbed > ul > .browse-list-item {
  @include govuk-responsive-padding(6, "bottom");
  border-bottom: solid 1px $govuk-border-colour;
  @include govuk-responsive-margin(6, "bottom");
}
```

tombye (changes requested):
I found a few minor things but once they're addressed this looks good to me so let me know and I'll approve it 👍🏻

app/templates/views/platform-admin/search.html
samuelhwilliams: I've made it 5/5 as 6 looked big .. but open to your thoughts?

But also a very easy thing to tweak separately so probably not a blocker for now.

# 5/5
image

# 6/6
image

samuelhwilliams:
Thanks for all the comments @tombye - should all be updated now :)

tombye (approved):
Looks good to me 👍🏻

PR #4724, opened 2023-05-17
Make edit letter buttons into secondary buttons

quis:
Before | After
---|---
image | image
image | image


tombye (approved):
Looks good to me 👍🏻

PR #4729, opened 2023-05-23
Migrate conditional radio buttons

BlessedDev:
This PR removes the `radio` and `conditional_radios macros` in app/templates/views/agreement/agreement-accept.html and app/templates/views/organisations/add-gp-organisation.html and fully embraces GOVUK design system.

I have attached before and after photos
add-gp-before-1
add-gp-after-1
add-gp-before-2
add-gp-after-2
add-gp-before-3
add-gp-after-3
add-gp-before-4
add-gp-after-4
agreement-before-1
agreement-after-1
agreement-before-2
agreement-after-2
agreement-before-3
agreement-after-3
agreement-before-4
agreement-after-4

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -881,6 +881,7 @@ class BasePermissionsForm(StripWhitespaceForm):
         ],
         thing="how this team member should sign in",
         validators=[DataRequired()],
+        param_extensions={"fieldset": {"legend": {"classes": "govuk-fieldset__legend--s"}}},
     )
 
     permissions_field = GovukCheckboxesField(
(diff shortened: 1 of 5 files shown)


quis:
I’d prefer if the fieldset legend could stay non-bold. In fact I’d go as far as to say that this should be the default in our app. I’m not sure if this is possible. If it’s not we’ll need to override it every time we make a new set of radio buttons.

BlessedDev:
> I’d prefer if the fieldset legend could stay non-bold. In fact I’d go as far as to say that this should be the default in our app. I’m not sure if this is possible. If it’s not we’ll need to override it every time we make a new set of radio buttons.

I have overridden the default font weight by adding  `govuk-!-font-weight-regular` to the definition of fieldset legend in `GOVUKRadioField.prepare_params`.

It seems to have the desired effect:

accept-agreement-before
agreement-accept-after
accept-agreement-before
agreement-accept-after

PR #4730, opened 2023-05-30
Check for allowed template type on `POST` as well as `GET` requests

quis:
This check stops you viewing the page to edit a template if your service doesn’t have that type of template switched on. It doesn’t stop you actually editing the template if you can craft a post request.

This doesn’t mean anyone can do anything malicious, but the position of the check in the code doesn’t properly express its intention.

This commit moves it nearer the top, the same as in `add_service_template`:
https://github.com/alphagov/notifications-admin/blob/f4ed8e130633183b9fc798db2d95cc48932a2754/app/main/views/templates.py#L559-L570

This means it will now catch both `POST` and `GET` requests.


whpearson (approved):
Makes sense 👍

PR #4732, opened 2023-05-31
Return a `Template` model from the `Service` model

quis:
At the moment when we have a variable like `template` it’s sometimes a `dict` straight from the API and it’s sometimes an instance of a model from utils like `LetterImageTemplate` or `EmailPreviewTemplate`.

This commit makes the model-based approach the default. This is better because:
- it follows the pattern we use for users, services, branding, etc.
- it means we can add useful methods and properties to the model, which are easily reusable across the codebase
- it avoids the looseness of dictionaries where there’s no guarantee that specific keys will be present

There’s still more work to do to remove or prevent any interaction with the underlying `dict`, but this gets us most of the way there.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -92,41 +92,34 @@ def get_example_letter_address(key):
 @main.route("/services/<uuid:service_id>/send/<uuid:template_id>/csv", methods=["GET", "POST"])
 @user_has_permissions("send_messages", restrict_admin_usage=True)
 def send_messages(service_id, template_id):
-    db_template = current_service.get_template_with_user_permission_or_403(template_id, current_user)
-
-    email_reply_to = None
-    sms_sender = None
-
-    if db_template["template_type"] == "email":
-        email_reply_to = get_email_reply_to_address_from_session()
-    elif db_template["template_type"] == "sms":
-        sms_sender = get_sms_sender_from_session()
-
-    if db_template["template_type"] not in current_service.available_template_types:
-        return redirect(
-            url_for(
-                ".action_blocked",
-                service_id=service_id,
-                notification_type=db_template["template_type"],
-                return_to="view_template",
-                template_id=template_id,
-            )
-        )
-
-    template = get_template(
-        db_template,
-        current_service,
+    template = current_service.get_template_with_user_permission_or_403(
+        template_id,
+        current_user,
         show_recipient=True,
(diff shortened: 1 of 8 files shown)


app/main/views/send.py
```diff
@@ -337,25 +330,22 @@ def send_one_off_letter_address(service_id, template_id):
...
             filetype="png",
         ),
-        page_count=get_page_count_for_letter(db_template, session_placeholders),
-        email_reply_to=None,
-        sms_sender=None,
     )
 
+    session_placeholders = get_normalised_placeholders_from_session()
+
+    template.page_count = get_page_count_for_letter(template._template, session_placeholders)
```
samuelhwilliams, line 347: Oof don't like that we have to do this out-of-band in order to get the correct functionality. But not sure what we can do better here right now...

app/main/views/templates.py
```diff
@@ -90,34 +90,32 @@ def as_error_dict(self) -> Dict[str, int]:
...
+        show_recipient=True,
+    )
+    template_folder = current_service.get_template_folder(template.get_raw("folder"))
 
     user_has_template_permission = current_user.has_template_folder_permission(template_folder, service=current_service)
     if should_skip_template_page(template):
         return redirect(url_for(".set_sender", service_id=service_id, template_id=template_id))
 
-    page_count = get_page_count_for_letter(template)
+    template.page_count = get_page_count_for_letter(template._template)
```
samuelhwilliams, line 109: I feel like we need `page_count` to throw an error if this isn't populated manually - the `1` default value feels probably like the wrong thing to do... 🤔

samuelhwilliams (commented):
I appreciate the intent of this PR and it feels like an improvement, but we probably should address the `page_count` thing if we can now and I'm left thinking that something isn't quite right given how often we have to break out to `get_raw` 🤔

But we can get in, pending `page_count` thoughts, if it's blocking you.

quis:
Agree that `template.page_count = ` is bad. I removed it, but split into a separate PR for the sake of smaller diffs: https://github.com/alphagov/notifications-admin/pull/4733

The use of `template.get_raw(…)` is something we can hopefully factor out over time.

samuelhwilliams (approved):
This looks good. I've not exhaustively tested it but I've had a flick through some flows and clearly the tests all pass 🤞

PR #4740, opened 2023-06-07
Remove use of `OrderedDict`

quis:
Dictionaries are ordered in Python now:

> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

– https://docs.python.org/3/library/stdtypes.html?highlight=dict#mapping-types-dict

Therefore we don’t need to use a library to get this behaviour.


app/main/views/organisations/index.py
```diff
@@ -382,7 +380,7 @@ def edit_organisation_domains(org_id):
         try:
             organisations_client.update_organisation(
                 org_id,
-                domains=list(OrderedDict.fromkeys(domain.lower() for domain in filter(None, form.domains.data))),
+                domains=list(dict.fromkeys(domain.lower() for domain in form.domains.data if domain)),
```
samuelhwilliams, line 383:
```suggestion
                domains=[domain.lower() for domain in form.domains.data if domain],
```

samuelhwilliams (approved):
There is still an OrderedDict in `test_should_update_guestlist` but otherwise I think this is good

app/main/views/organisations/index.py
quis: I tried that but it wants unique values. Could leave as-is, or replace with `OrderedSet` to make it clearer, what do you think?

PR #4741, opened 2023-06-07
Add govuk frontend nested radios

BlessedDev:
This PR migrates the nested radio buttons in `app/templates/views/templates/_move_to.html` to GOVUK Frontend.
This involved creating a new nested-radios component which based on the GOVUK radio component as the latter doesn't have a nested radios functionality.

It is still a work in progress

### DIFF
--- a/app/assets/stylesheets/components/checkboxes.scss
+++ b/app/assets/stylesheets/components/checkboxes.scss
@@ -74,29 +74,6 @@ $govuk-checkboxes-size: 40px;
 
 }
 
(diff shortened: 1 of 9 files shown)


leohemsted:
can we close this now that we have https://github.com/alphagov/notifications-admin/pull/4751?

BlessedDev:
This is no longer needed as [this](https://github.com/alphagov/notifications-admin/pull/4751) PR now deals with nested radios.

PR #4744, opened 2023-06-13
Fix colour contrast of `disabled` radio button text

quis:
We have noticed that when radio buttons have the `disabled` attribute set<sup>1</sup>, the text does not meet WCAG standards<sup>2</sup> for colour contrast. This is true for the label<sup>3</sup>, and worse for the hint text<sup>4</sup>:

image

This is caused by some CSS in GOV.UK Frontend<sup>5</sup>.

I propose:
- using the secondary text colour for
  - the radio button itself
  - the label text
  - the hint text
- using a light grey fill to give an additional visual cue that the choice can’t be filled in by clicking it

Which looks like: 

image

Colour alone isn’t used here to differentiate – there is still the semantics of the `disabled` attribute itself, the different cursor behaviour, and the content of the hint text.

***

1. Using `disabled` is not often the best way to do things (and is not documented in the Design System) but sometimes it’s appropriate
2. WCAG 2.0 AA requires a contrast ratio of 4.5:1 for ‘normal text’
3. `$govuk-text-colour` with `opacity: .5` applied computes to a value of  `#858585`, which gives a contrast ratio of 3.69:1 against a white background
4. `$govuk-secondary-text-colour` with `opacity: .5` applied computes to a value of  `#a8acae`, which gives a contrast ratio of 2.28:1 against a white background
5. https://github.com/alphagov/govuk-frontend/blob/fbed2b59889aff35fd1c65f1cd5ad368a5e4ddfe/packages/govuk-frontend/src/govuk/components/radios/_index.scss#L137

***

Card: https://trello.com/c/7Q6pMovq/381-fix-contrast-ratio-of-text-and-hint-for-disabled-radio-buttons

### DIFF
--- a/app/assets/stylesheets/components/radios.scss
+++ b/app/assets/stylesheets/components/radios.scss
@@ -115,3 +115,16 @@
     margin-right: 15px;
   }
 }
+
+.govuk-radios__input:disabled + .govuk-radios__label,
+.govuk-radios__input:disabled ~ .govuk-hint {
+  opacity: 1;
+  color: $govuk-secondary-text-colour;
+}
+
+.govuk-radios__input:disabled + .govuk-radios__label {
+  &:before {
+    border-color: $govuk-secondary-text-colour;
+    background-color: govuk-colour("light-grey");
+  }
+}


quis:
For context, this is how the page looked before the move to using the CSS from GOV.UK Frontend:

Screenshot 2023-06-14 at 09 48 43

quis:
Here are a bunch of ideas for how the disabled radio button could also meet the [non-text constrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) criteria.

They are all a bit odd/jarring in some way, so I don’t think worth the small benefit, given that the information is already conveyed in the hint text.

– | A | B | C
---|---|---|---
1 | Screenshot 2023-06-16 at 13 59 11 | Screenshot 2023-06-16 at 13 59 19| Screenshot 2023-06-16 at 13 59 40
2 | Screenshot 2023-06-16 at 13 59 52 | Screenshot 2023-06-16 at 14 20 09|  Screenshot 2023-06-16 at 14 23 14
3 | Screenshot 2023-06-16 at 14 23 43 | Screenshot 2023-06-16 at 14 25 50| Screenshot 2023-06-16 at 14 35 25
4 | Screenshot 2023-06-16 at 14 40 13 | Screenshot 2023-06-16 at 16 23 11| Screenshot 2023-06-16 at 16 26 57
5 | Screenshot 2023-06-16 at 16 33 33 | Screenshot 2023-06-16 at 16 34 16| Screenshot 2023-06-16 at 16 34 42
6 | Screenshot 2023-06-16 at 16 38 21 | Screenshot 2023-06-16 at 16 41 14| Screenshot 2023-06-16 at 16 43 04
7 | Screenshot 2023-06-16 at 17 00 07 | Screenshot 2023-06-16 at 17 01 50| Screenshot 2023-06-16 at 17 04 57
8 | Screenshot 2023-06-16 at 17 07 57 | |

PR #4746, opened 2023-06-13
Preview notifications with attachment

leohemsted:
We have the template-preview so that it renders the attachment, when the template preview PR[^1] is merged this will immediately work for the check page before you hit save.

When this PR is merged, this will also roll out stitching the attatchment ot the letter for the png preview/download a pdf button you can see after your notification has been saved. (whether via API, upload, job, etc).

Previously we went via the API[^2] to fetch this preview, however, we can refactor this to go directly to the template-preview app using the TemplatePreview class - this is consistent with how all the other PNG/PDF previews are rendered on the admin app.

The only remaining usage of the previously mentioned API endpoint[^2] is for precompiled letters - the API is the only place with the logic to know which bucket a letter pdf file will be in based on the status/key type/etc. This could be a candidate for a future refactor 😉 

[^1]: https://github.com/alphagov/notifications-template-preview/pull/737
[^2]: https://github.com/alphagov/notifications-api/blob/main/app/template/rest.py#L235

### DIFF
--- a/app/main/views/notifications.py
+++ b/app/main/views/notifications.py
@@ -34,6 +34,7 @@ from app import (
 )
 from app.main import json_updates, main
 from app.notify_client.api_key_api_client import KEY_TYPE_TEST
+from app.template_previews import TemplatePreview
 from app.utils import (
     DELIVERED_STATUSES,
     FAILURE_STATUSES,
@@ -182,7 +183,6 @@ def view_notification(service_id, notification_id):  # noqa: C901
 @main.route("/services/<uuid:service_id>/notification/<uuid:notification_id>/cancel", methods=["GET", "POST"])
 @user_has_permissions("view_activity", "send_messages")
 def cancel_letter(service_id, notification_id):
-
     if request.method == "POST":
         try:
             notification_api_client.update_notification_to_cancelled(current_service.id, notification_id)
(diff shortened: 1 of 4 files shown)


app/main/views/notifications.py
```diff
@@ -207,6 +207,10 @@ def get_preview_error_image():
 @main.route("/services/<uuid:service_id>/notification/<uuid:notification_id>.<letter_file_extension:filetype>")
 @user_has_permissions("view_activity", "send_messages")
 def view_letter_notification_as_preview(service_id, notification_id, filetype, with_metadata=False):
+    notification = notification_api_client.get_notification(service_id, notification_id)
+    if not notification["template"]["is_precompiled_letter"]:
+        return TemplatePreview.from_notification(notification, filetype=filetype, page=request.args.get("page"))
```
samuelhwilliams, line 212: I don't know if I agree that the latency saving here is worth fragmenting the data flow and having the mental overhead of "ok if its precompiled then go here but if not then go there", it feels a weird thing for admin to be thinking about, but ... 🤷😢

samuelhwilliams (approved):
👍 

I get the attachment as part of the PDF when downloading a sent notification now

app/main/views/notifications.py
leohemsted: yeah i definitely wrestled with this. 

my main motivation for this was that this function is the only place that goes admin->api->template-preview, everywhere else where PDFs/PNGs are handled on the admin app goes direct to template preview. also that function in API is absolutely terrifying too (one endpoint that handles precompiled, templated, happy path, unhappy path with red overlaps, pngs and PDFs) so i was quite keen to move away from that to reduce the complexity on the API (and also I think the `TemplatePreview.from_notification` felt relatively clean and logicial and makes it marginally easier to find wherever we call template preview when searching through code

PR #4748, opened 2023-06-14
Add categories to Zendesk tickets we create

quis:
We’re trying to add categories to all the Zendesk tickets we get. When these tickets are ones that we create (rather than ones where the user is providing the content of the ticket) we can do this automatically because we already know what the ticket is about.

### DIFF
--- a/app/main/views/service_settings/branding.py
+++ b/app/main/views/service_settings/branding.py
@@ -52,6 +52,7 @@ def create_email_branding_zendesk_ticket(detail=None):
         org_id=current_service.organisation_id,
         org_type=current_service.organisation_type,
         service_id=current_service.id,
+        ticket_categories=["notify_email_letter_branding"],
     )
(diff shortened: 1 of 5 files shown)


klssmith:
The value that we pass into `ticket_categories` shouldn't be human readable one you see in the drop down menu - there will be a "code" version of it in Zendesk we'll need to double check, though it's probably going to be something like 'branding_requests_for_emails_and_letters'

samuelhwilliams:
From testing a ticket looks lik the new ones are:

"notify_email", "notify_email_letter_branding", "notify_go_live_request", "notify_sign_in_issue", "notify_sms"

samuelhwilliams (approved):
Could we define these in `constants.py` and use values from there? Be nice to keep things with meaning in one place. Possibly/ideally using an enum.

I know that file is fairly empty right now, but will hopefully slowly move towards having things in there (we have a similar file in api as well)

PR #4754, opened 2023-06-16
Only show request to go live form if the service has organisation

whpearson:
This is to stop services going live without signing the MOU etc

### DIFF
--- a/app/templates/views/make-service-live.html
+++ b/app/templates/views/make-service-live.html
@@ -34,7 +34,7 @@
           </li>
         {% endfor %}
       </ul>
-
+      {% if organisation %}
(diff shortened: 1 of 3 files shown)


app/templates/views/make-service-live.html
```diff
@@ -44,6 +44,9 @@
         }) }}
         {{ page_footer('Confirm') }}
       {% endcall %}
+      {% else %}
+         The service needs an organisation first before it can be made live.
```
samuelhwilliams, line 48: I think only platform admins will see this, right? If so, might be nice to add a link straight to the service settings page to add an org?

samuelhwilliams (approved):

PR #4755, opened 2023-06-16
Cache page count for templated letters in Redis

quis:
Getting the page count for a letter means:
- the admin app calls template preview
- template preview either
  - renders the HTML to a PDF
  - gets the cached version of the above from S3
- template preview counts the number of pages in the PDF
- template preview returns the count to the admin app

This is quick, but probably not as quick as going straight to Redis. Which is what this commit does.

We can’t do this for letters which are personalised, because the personalisation might change the page count of the letter.

### DIFF
--- a/app/utils/templates.py
+++ b/app/utils/templates.py
@@ -5,7 +5,9 @@ from notifications_utils.template import (
 )
 from notifications_utils.template import LetterImageTemplate as UtilsLetterImageTemplate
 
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):

quis:
On production this reduced the page load time for a one page letter with no attachments from ~450ms to ~130ms

PR #4759, opened 2023-06-22
Hide/restrict 'make services live' permission

samuelhwilliams:
The ability for organisation users to make services live is guarded by the 'can approve own go live requests' org-level permission. Let's hide and restrict editing of the 'can make services live' org user permission based on the org-level permission.

| x | Team members | Edit user |
|---|---|---|
| Current state / new state when org has permission | image | image |
| New state when org does not have permission | image | image |

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -51,7 +51,7 @@ from wtforms.validators import (
     Regexp,
 )
 
-from app import asset_fingerprinter
+from app import asset_fingerprinter, current_organisation
 from app.formatters import (
     format_auth_type,
     format_thousands,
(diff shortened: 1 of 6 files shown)


whpearson (approved):
Changes tested locally. Toggling the feature flag made the permissions  visible and editable as expected.

Would we expect a user to be able to modify/remove their own permissions because currently they can't. I can see arguments both ways.

samuelhwilliams:
> Would we expect a user to be able to modify/remove their own permissions because currently they can't. I can see arguments both ways.

We don't allow service users to edit their own permissions, so I've followed that pattern here for now. We can revisit it easily in the future as/when we need to.

PR #4760, opened 2023-06-23
Set org user permissions on invite

samuelhwilliams:
* [x] To merge after: https://github.com/alphagov/notifications-api/pull/3830

---

Adds a form to the 'invite user to org' page which allows setting permissions (currently just whether the user can make services live).

Also show these permissions for invited users on the org team members page.

And finally, respect those invited org user permissions when the invite is actually accepted (currently it's just hard-coded to give them the permission).

Some refactoring, including adding an error summary to the invite org user page.

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -666,11 +666,8 @@ class InvitedOrgUser(BaseUser):
         return self._permissions
 
     @permissions.setter
-    def permissions(self, permissions):
-        if isinstance(permissions, list):
-            self._permissions = permissions
-        else:
-            self._permissions = [p for p in permissions.split(",") if p]
+    def permissions(self, permissions: list[str]):
+        self._permissions = permissions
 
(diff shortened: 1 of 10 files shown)


app/models/user.py
```diff
@@ -659,6 +661,22 @@ def accept_invite(self):
...
+    @property
+    def permissions(self):
+        return self._permissions
+
+    @permissions.setter
+    def permissions(self, permissions):
+        if isinstance(permissions, list):
+            self._permissions = permissions
+        else:
+            self._permissions = [p for p in permissions.split(",") if p]
```
quis, line 673: When would it happen that the permissions are a comma-separated string? If so should we have a test for this coercion?

samuelhwilliams: Mm I did copy this from the service permissions equivalent, but fair point to say we don't need it. Will rip it out 👍

PR #4762, opened 2023-06-27
Update dates for oustanding issues

tombye:
Makes the following changes:
1. update the issue with the time/date picker to reflect that work has begun
2. bump the date on the collapsable checkboxes issue to be the next piece of work we pick up
3. don't change the issue with our forms validation as we are still planning out the work
4. move all other issues up to fit into the remaining time


saimaghafoor (approved):
Looks good to go :)

PR #4763, opened 2023-06-27
Expand DAC acronym in accessibility statement

tombye:
This used to be shown in the full form higher up in the statement, so the acronym could be used. This is no longer the case and there are now no other uses so this should just be expanded.


quis (approved):
🦅 👀

PR #4767, opened 2023-06-30
Viewing an attachment on the manage-attachment page

whpearson:
This simply adds the looping logic to the manage-attachment page

We may want to pull it out to it's own template so it can be reused for the validation failure page.

Added a simple test for the endpoint

goes with https://github.com/alphagov/notifications-template-preview/pull/739

### DIFF
--- a/app/templates/views/templates/manage-attachment.html
+++ b/app/templates/views/templates/manage-attachment.html
@@ -40,8 +40,13 @@
     )}}
     </div>
     <div class="template-container">
-        <div class="letter page--odd">
-        </div>
+
+        {% for page_number in page_numbers %}
+          <div class="letter page--{{ loop.cycle('odd', 'even') }}{% if loop.first %} page--first{% endif %}{% if loop.last %} page--last{% endif %}">
+            <img src="{{ letter_attachment_image_url }}?page={{ page_number }}" alt="" loading="{{ 'eager' if (page_number==1) else 'lazy' }}">
+          </div>
+        {% endfor %}
+
     </div>
     <div class="govuk-!-margin-bottom-2">
       <div class="js-stick-at-bottom-when-scrolling">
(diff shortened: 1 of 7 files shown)


app/templates/views/templates/manage-attachment.html
```diff
@@ -40,8 +40,13 @@ <h1 class="banner-title">{{ error.title }}</h1>
     )}}
     </div>
     <div class="template-container">
-        <div class="letter page--odd">
-        </div>
+
+        {% for page_number in page_numbers %}
+          <div class="letter page--{{ loop.cycle('odd', 'even') }}{% if loop.first %} page--first{% endif %}{% if loop.last %} page--last{% endif %}">
```
quis, line 45: Think this would be better wrapped up in a class, but happy to save that for a subsequent refactor.

Technically whether the first page of the attachment is `page--odd` or `page--even` depends on how many pages are in the template. This would be easier to do after a refactor though, so doesn’t need to change now.

tests/app/main/views/test_templates.py
```diff
@@ -1513,6 +1513,25 @@ def test_should_show_preview_letter_templates(
...
+    response = client_request.get_response(
+        "no_cookie.view_letter_attachment_preview",
+        service_id=service_id,
+        attachment_id=attachment_id,
+    )
+
+    assert response.get_data(as_text=True) == "foo"
+    assert mocked_preview.call_args[0][0] == attachment_id
+
+
```
quis, line 1534: Should probably add a test for:
- correct number of images are shown on the page
- at least the first image has the correct `src` attribute

quis (changes requested):

app/templates/views/templates/manage-attachment.html
whpearson: But we aren't showing the rest of the template here, does that matter?

whpearson: Yeah I think saving it for a refactor makes sense.

quis (approved):

PR #4768, opened 2023-07-03
Add create-droplet/deploy scripts

samuelhwilliams:
At the moment we deploy the API app by building a CloudFoundry droplet and then copying that droplet to preview/staging/prod (broadly equivalent to building a docker image and deploying that). In our other apps, though, we build the droplet in each environment as deployments happen. This is both slower and in theory less reproducible, since each environment could have issues building the app or build things differently.

This copies across the relevant scripts from the API so that we can build the droplet once and promote it through each environment.


whpearson (approved):
Changes make sense, it would be good to get things the same.

PR #4774, opened 2023-07-12
Don't explicitly list every SASS file in gulpfile

quis:
Instead, use globs to just grab everything.

Note that we don't need `/**/*.scss` since main.scss imports everything (and we get errors if we glob all the smaller components).

***

This implements https://github.com/alphagov/notifications-admin/pull/4764 but without the changes to the Javascript build.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -149,11 +149,7 @@ const javascripts = () => {
 
 
 const sass = () => {
-  return src([
-      paths.src + '/stylesheets/main*.scss',
-      paths.src + '/stylesheets/map.scss',
-      paths.src + '/stylesheets/print.scss'
-    ])
+  return src(paths.src + '/stylesheets/*.scss')
     .pipe(plugins.prettyerror())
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -149,11 +149,7 @@ const javascripts = () => {
 
 
 const sass = () => {
-  return src([
-      paths.src + '/stylesheets/main*.scss',
-      paths.src + '/stylesheets/map.scss',
-      paths.src + '/stylesheets/print.scss'
-    ])
+  return src(paths.src + '/stylesheets/*.scss')
```
samuelhwilliams, line 152: I think this will pull in a bunch of scss files that we don't currently (the `_`-prefixed ones), which would lead to duplicated CSS wouldn't it?

quis: By convention SASS won’t compile things which are named with an underscore prefix.

Tested here:
image

samuelhwilliams (approved):

PR #4775, opened 2023-07-13
Defer loading of crests and insignia

quis:
Currently this app tries to load the images for government crests and insignia at startup time.

This means that if the files cannot be found then the app will crash on startup. This typically happens when the app is restarting because it’s noticed some changes (for example by checking out a different branch).

The files may be missing because the Gulp task to copy them into the static folder has not run yet.

Previously this wasn’t a problem because we didn’t delete the old files before copying the new ones. But as of https://github.com/alphagov/notifications-admin/pull/4710 we clean the whole directory before re-copying the assets.

This commit changes the app to not look for the images until it needs them. By this time the Gulp task should have had a chance to finish cleaning the folder and copying the images into place.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -79,7 +79,7 @@ from app.main.validators import (
 )
 from app.models.branding import (
     GOVERNMENT_IDENTITY_SYSTEM_COLOURS,
-    GOVERNMENT_IDENTITY_SYSTEM_CRESTS_OR_INSIGNIA,
+    get_government_identity_system_crests_or_insignia,
 )
 from app.models.feedback import PROBLEM_TICKET_TYPE, QUESTION_TICKET_TYPE
 from app.models.organisation import Organisation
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
I don't see any crashing behaviour if `app/static/images/branding/insignia/` is empty - I just see a form with no choices:

image

I actually _would_ rather it did crash on startup as that would mean it becomes undeployable without the assets there. Maybe we can settle for adding a test that asserts the form has the expected choices? The tests run in concourse after generating the assets so it should always be good to pass there.

Given that it doesn't crash on startup currently, I'm OK with this change in theory but a test might be nice (if the test fails locally then probably just print a message saying "run `npm run build"` or somehing?)

quis:
@samuelhwilliams you have to delete the folder, not just the contents to cause a crash. Which is what the Gulp task does here: https://github.com/alphagov/notifications-admin/blob/20fe9b60ebb24bf02a3e6c6bca6984764c38585b/gulpfile.js#L242

There’s already a test which makes sure all the files are present here (the can’t generate the hashes unless it can read the files): https://github.com/alphagov/notifications-admin/blob/20fe9b60ebb24bf02a3e6c6bca6984764c38585b/tests/app/main/views/test_email_branding.py#L822-L864

samuelhwilliams (approved):

PR #4779, opened 2023-07-20
Render QR codes in letters - or rather pull in utils that do it

CrystalPea:
Needs this to be merged to work: https://github.com/alphagov/notifications-template-preview/pull/744


samuelhwilliams (approved):
This looks a bit broken without the template-preview changes but given the feature isn't released yet, this should be fine 👍 

Let's just make sure we merge them at a similar time.

PR #4780, opened 2023-07-20
Preview invalid attachment upload

CrystalPea:
Both when there is a pre-existing attachment for that template, and if there is not.

Needs https://github.com/alphagov/notifications-template-preview/pull/747 to be merged in first

Screenshots:

### no pre-existing attachment, content outside printable area 
Screenshot 2023-07-21 at 15 03 15

### no pre-existing attachment, different error that doesnt need overlay
Screenshot 2023-07-20 at 16 28 55

### with pre-existing attachment, content outside printable area 
Screenshot 2023-07-21 at 15 04 32

Story ticket: https://trello.com/c/eisYQprO/373-show-preview-of-new-attachment-when-validation-fails

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -944,7 +944,7 @@ def letter_template_attach_pages(service_id, template_id):
                 template=template,
                 error=error,
                 letter_attachment_image_url=letter_attachment_image_url,
-                page_numbers=[*range(1, attachment_page_count + 1)],
+                page_numbers=_get_page_numbers(attachment_page_count),
             ),
             400 if error else 200,
         )
@@ -963,7 +963,7 @@ def letter_template_attach_pages(service_id, template_id):
         template=template,
         service_id=service_id,
         letter_attachment_image_url=letter_attachment_image_url,
-        page_numbers=list(range(1, attachment_page_count + 1)),
+        page_numbers=_get_page_numbers(attachment_page_count),
         error=error,
     )
 
@@ -1019,7 +1019,7 @@ def letter_template_edit_pages(template_id, service_id):
             service_id=service_id,
             attachment_id=template.attachment.id,
         ),
-        page_numbers=list(range(1, template.attachment.page_count + 1)),
+        page_numbers=_get_page_numbers(template.attachment.page_count),
         error=error,
     )
 
@@ -1151,3 +1151,7 @@ def view_invalid_letter_attachment_as_preview(service_id, file_id):
         return TemplatePreview.from_invalid_pdf_file(pdf_file, page, is_an_attachment=True)
     else:
         return TemplatePreview.from_valid_pdf_file(pdf_file, page)
+
+
+def _get_page_numbers(page_count):
+    return [*range(1, page_count + 1)]
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -919,7 +919,7 @@ def letter_template_attach_pages(service_id, template_id):
     form = PDFUploadForm()
     error = {}
     letter_attachment_image_url = None
-    attachment_page_count = None
+    attachment_page_count = 0
     if form.validate_on_submit():
         upload_id = uuid.uuid4()
         try:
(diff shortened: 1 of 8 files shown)


app/main/views/templates.py
```diff
@@ -922,37 +918,55 @@ def letter_template_attach_pages(service_id, template_id):
...
         except LetterAttachmentFormError as e:
             error = e.as_error_dict()
+            attachment_page_count = error.get("attachment_page_count", 0)
+            letter_attachment_image_url = (
+                url_for(
+                    "no_cookie.view_invalid_letter_attachment_as_preview",
+                    service_id=service_id,
+                    file_id=upload_id,
+                ),
+            )[0]
```
samuelhwilliams, line 936: This looks like a mistake? We're making a one element tuple and then getting them first element from it.

I think the hanging comma was a typo that turned this into a tuple, and then `[0]` was a workaround for it. But we should just be able to remove it

```suggestion
            attachment_page_count = error.get("attachment_page_count", 0)
            letter_attachment_image_url = url_for(
                "no_cookie.view_invalid_letter_attachment_as_preview",
                service_id=service_id,
                file_id=upload_id,
            )
```

app/main/views/templates.py
```diff
@@ -922,37 +918,55 @@ def letter_template_attach_pages(service_id, template_id):
 
     form = PDFUploadForm()
     error = {}
+    letter_attachment_image_url = None
+    attachment_page_count = None
```
samuelhwilliams, line 922: The use of this variable throughout the function is a bit confusing. Basically every time we set it we do 'or 0' - should we just set it to 0 upfront?

app/main/views/templates.py
```diff
@@ -922,37 +918,55 @@ def letter_template_attach_pages(service_id, template_id):
...
     if not template.attachment:
+        attachment_page_count = attachment_page_count or 0
         return (
             render_template(
                 "views/templates/attach-pages.html",
                 form=form,
                 template=template,
                 error=error,
+                letter_attachment_image_url=letter_attachment_image_url,
+                page_numbers=[*range(1, attachment_page_count + 1)],
```
samuelhwilliams, line 950: We are building this slightly funky list of numbers in two different places in two different ways.

What do you think about just passing in `attachment_page_count` and doing `{% for page_number in range(1, attachment_page_count + 1) %}` in the template? 🤔

app/s3_client/s3_letter_upload_client.py
```diff
@@ -89,6 +98,12 @@ def get_letter_metadata(service_id, file_id):
     return LetterMetadata(s3_object["Metadata"])
 
 
+def get_attachment_pdf_and_metadata(service_id, file_id):
+    s3_object = get_letter_s3_object(service_id, file_id, bucket_name="S3_BUCKET_TRANSIENT_UPLOADED_LETTERS")
+    pdf = s3_object["Body"].read()
+    return pdf, LetterAttachmentMetadata(s3_object["Metadata"])
```
samuelhwilliams, line 104: 🤔  this is an interesting wrapper object. I don't know if I understand what it's really giving us over a standard dict - other than a classname?

I can see why we have `LetterMetadata` - is does some decoding behind the scenes. I'm guessing you've done this to try to be consistent, but I think it's probably fine to drop?

samuelhwilliams (approved):
Functionally this all looks good - nice work 🙌 

I've got a couple of minor code comments for you to review and consider :)

app/main/views/templates.py
CrystalPea: Ahh, good catch, I was wondering why it's being wrapped in a tuple and somehow missed that we were the ones explicitly doing it 🙈

app/main/views/templates.py
CrystalPea: Ohh, good point, `0` is Falsey, so this line `attachment_page_count = attachment_page_count or template.attachment.page_count` will still work as expected, duh 😅 . Good spot, changed now!

app/main/views/templates.py
CrystalPea: Ahh good point, I abstracted getting a page count into a helper method now 😁  (check https://github.com/alphagov/notifications-admin/pull/4780/commits/d27667a4a2f47a5c851f81bd1833622389f0f358)

app/s3_client/s3_letter_upload_client.py
CrystalPea: You are right, it is superfluous, thanks for spotting! I removed it now.

PR #4784, opened 2023-07-24
Automatically prefix `Error: ` to page titles

samuelhwilliams:
And add a test that will try to flag any forms passed into templates that won't benefit from this automatic handling. Passing a form into a `render_template` call with an underscore will deactivate the test/check for that form, allowing us to pass in forms that won't (reasonably) ever show errors (like search input forms).

There are a few pages that display multiple forms which can each render errors. For these pages, we add an explicit override to the test and handle prefixing the page title with `Error: ` manually.


samuelhwilliams:
@tombye I'd like to think we could also automatically inject the error summary by adding a block that renders the error summary for `form` by default. I think we should do that as a separate ticket, as it needs a bit of exploration work to confirm it's possible, and then a bunch of validation to make sure it's all good.

tests/utils.py
```diff
@@ -1,8 +1,33 @@
...
+        context = call.args[2]
+        for key, value in context.items():
+            if isinstance(value, Form) and key != "form":
+                if key.startswith("_") and not value.errors:
+                    # Forms passed in with a leading prefix are excluded from this check; the underscore implies the
+                    # form won't throw errors. If it doesn't have errors then that's good; but if it does let's flag
+                    # it.
+                    continue
+
+                raise ValueError(
```
tombye, line 31: Really like this being documented here 🥇.

tests/utils.py
```diff
@@ -13,7 +13,13 @@ def __lt__(self, other):
 
 def check_render_template_forms(calls: list[mock.call]):
     for call in calls:
+        template = call.args[1]
         context = call.args[2]
+
+        # Specific exclusions to this check that a developer is manually overriding due to explicit handling.
```
tombye, line 19: Another 🥇 for how this makes it clear which templates are exceptions to the rule.

tombye (approved):
I can't see any problems and really enjoying seeing how this work benefits both users using screen readers (massive improvement already) and developers, through making our rules explicit and documented in the codebase 👍🏻

leohemsted:
@samuelhwilliams yeah we already have the `errorSummary` component that renders the error summary for an arbitrary form - i'd love to make that included in the admin_template by default. There's definitely a few different usages of it (and inconsistencies around if it's above or below page_header) but i think it shouldn't be too onerous to move it out.

agree it's a sep ticket though cos we'd need to make sure we dont end up with broken or duplicate summaries anywhere (manually test every form? 😱)

PR #4787, opened 2023-07-26
Open up letter attachment feature to all users sending letters

CrystalPea:
By removing the check for extra letter formatting flag on attachment related views.

### DIFF
--- a/tests/app/main/views/test_templates.py
+++ b/tests/app/main/views/test_templates.py
@@ -781,53 +781,9 @@ def test_view_letter_template_does_not_display_send_button_if_template_over_10_p
     assert page.select_one("h1", {"data-error-type": "letter-too-long"})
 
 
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -781,53 +781,9 @@ def test_view_letter_template_does_not_display_send_button_if_template_over_10_p
...
-    mock_get_service_templates,
-    mock_get_template_folders,
-    single_letter_contact_block,
-    mock_has_jobs,
-    active_user_with_permissions,
-    mocker,
-    fake_uuid,
-    template_type,
-):
-    service_one["permissions"] = ["extra_letter_formatting"]
```
quis, line 797: I think it’s better to delete this line rather than the whole test. It’s still useful to check that the button is on the page (for letter templates only)

quis (approved):

PR #4791, opened 2023-07-28
Restrict negative values for rate limits

samuelhwilliams:
Not too fussed about the error text (but we can customise it as needed) because it's just a platform admin form. But feels like we should at least stop people being able to enter negative numbers.

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -47,8 +47,10 @@ from wtforms.validators import (
     DataRequired,
     InputRequired,
     Length,
+    NumberRange,
     Optional,
     Regexp,
+    StopValidation,
 )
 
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
@karlchillmaid if you've got any interest in this error message I'm happy to take some words from you to use for the error message :) It's a platform-admin only form though so no worries if you have better things to do for now.

quis:
Should we also prevent floats while we’re here?

karlchillmaid:
> 

`Number must be equal to or greater than 0`?

CrystalPea (approved):

samuelhwilliams:
@quis I think we already do?

image

quis (approved):

PR #4793, opened 2023-07-28
Add error summaries to base templates

samuelhwilliams:
By inserting a new block that populates error summaries automatically, we can make sure we are consistently displaying these without any developer effort (in most cases, ie where there's a single form on the page).

If a page has multiple forms, the developer will need to override the `errorSummary` block (as they do with the errorPrefix block) and build the errorSummary manually using the macro.

Although we're adding this to the base templates, we add an opt-in step so that we can review and enable error summaries on a page-by-page basis, as some error messages are not well suited to the summary display at the moment.


BlessedDev (approved):
Looks good to me, it establishes a pattern and works as intended.

PR #4795, opened 2023-07-30
Add error summaries to login registration

BlessedDev:
This PR follows on from the work done [here](https://github.com/alphagov/notifications-admin/pull/4793) and enables error summaries for the following:

`register-from-org-invite`, `register-from-invite` , `sign-in` , `two-factor-sms` .

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -570,6 +570,15 @@ class RegisterUserForm(StripWhitespaceForm):
 
 
 class RegisterUserFromInviteForm(RegisterUserForm):
+    custom_field_order = (
+        "name",
+        "mobile_number",
+        "password",
+        "service",
+        "email_address",
+        "auth_type",
+    )
+
     def __init__(self, invited_user):
         super().__init__(
             service=invited_user.service,
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -560,7 +560,7 @@ class LoginForm(StripWhitespaceForm):
 
 
 class RegisterUserForm(StripWhitespaceForm):
-    name = GovukTextInputField("Full name", validators=[DataRequired(message="Cannot be empty")])
+    name = GovukTextInputField("Full name", validators=[NotifyDataRequired(thing="your full name")])
     email_address = make_email_address_field()
     mobile_number = international_phone_number()
     password = make_password_field()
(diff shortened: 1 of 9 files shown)


samuelhwilliams:
I think it's worth a chat with @karlchillmaid or @saimaghafoor to check the error messages that these forms will throw, and potentially get new words for them. Some of the default messages like " Cannot be empty" probably need to be reworded to make sense in an error summary.

samuelhwilliams:
As we rollout error summaries to pages using the `withoutnav_template` base template, there are some layout issues that crop up (too much space between the summary and the heading):

image

This is because those pages use the GOVUK Elements heading class, which has spacing above the header, rather than the GOVUK Frontend heading class, which doesn't.

So we need to update headings on those pages to use the GOVUK Frontend header classes.

This effectively means changing from `heading-large` to `govuk-heading-l` (or equivalents). Which will take us to this:

image

Can we make those changes and check spacing on all of these pages? :) All of the pages edited here use `withoutnav_template`, so will be affected.

BlessedDev:
> 



> As we rollout error summaries to pages using the `withoutnav_template` base template, there are some layout issues that crop up (too much space between the summary and the heading):
> 
> image
> This is because those pages use the GOVUK Elements heading class, which has spacing above the header, rather than the GOVUK Frontend heading class, which doesn't.
> 
> So we need to update headings on those pages to use the GOVUK Frontend header classes.
> 
> This effectively means changing from `heading-large` to `govuk-heading-l` (or equivalents). Which will take us to this:
> 
> image
> Can we make those changes and check spacing on all of these pages? :) All of the pages edited here use `withoutnav_template`, so will be affected.

Yes that is a good catch. I will update the header classes.

BlessedDev:
@samuelhwilliams 
I have now updated the headings classes to `govuk-heading-l`.

register-from-org-invite:

register-from-org-invite-after

register-from-invite:

register-from-invite-after

two-factor-sms:

two-factor-sms-after

signin:

signin-after

samuelhwilliams:
These look good except for `register-from-invite`, where we still have a `Cannot be empty` that should be a more specific message.

BlessedDev:
> These look good except for `register-from-invite`, where we still have a `Cannot be empty` that should be a more specific message.
@samuelhwilliams 
oops, missed that one. I have updated the error message.
register-from-invite-after

samuelhwilliams (approved):

PR #4797, opened 2023-07-31
Add some more error summaries

samuelhwilliams:
This patch enables error summaries on 4 more pages, and tweaks some of the error messages as appropriate.

It also adds a new base form `OrderableFieldsForm` in 60f614170f99beed8f343002d225cf1128aa5c92 that allows us to override the order that fields are iterated through on a form. WTForms normally orders them according to when they are created in the code, but with our form inheritance structure sometimes the creation order differs from the display order and is not easy to correct. This gives us a way to explicitly set the field iteration order in line with the display order. We use this because errors are iterated through in field order, and the error summary should show errors in the same order as the fields appear on the page. "Invite a team member" was the page that required the fix in this PR.

---

image

image

image

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -256,11 +256,11 @@ def international_phone_number(label="Mobile number"):
     return InternationalPhoneNumber(label, validators=[DataRequired(message="Cannot be empty")])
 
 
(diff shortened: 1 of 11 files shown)


BlessedDev (approved):
Looks good to me

PR #4803, opened 2023-08-02
Remove fixed statuspage issues from a11y statement

tombye:
I noticed some of the accessibility issues raised against our statuspage have been fixed by their (Atlassian I think) developers 👏🏻. This removes those issues. 

## Our status page ‘subscribe to updates’ link does not contain programmatically discernible link text – this fails success criterion 1.3.1: info and relationships

The link is now a button (which matches its visual design) but the same questions apply. The text is now:
- visible in the page
- available to programs, like assistive tech, through accessibility APIs and HTML parsing
- being announced by screen readers, the most obvious assistive tech for discerning whether what’s in the HTML is usable

## Our status page subscribe to updates form fields do not contain an explicit label – this fails success criterion 1.3.1: info and relationships

This was fixed in 2022, as noted by @kr8n3r in:

https://github.com/alphagov/paas-product-pages/pull/143

## Our status page subscribe to updates forms do not indicate to users that they are expandable or collapsible – this fails success criterion 4.1.2: name, role, value

From reading this criterion, I’m confident to think that the button saying it’s expanded or collapsed, which it now does, is enough to pass this. I also checked that its state is now:
- available to programs, like assistive tech, through accessibility APIs and HTML parsing
- being announced by screen readers, the most obvious assistive tech for discerning whether what’s in the HTML is usable

## Our status page subscribe to updates form close link (x) is not descriptive enough for some users to determine its function or purpose – this fails success criterion 2.4.4: link purpose, in context

The link is now a button but the same questions apply. It now has a ‘Close subscribe form’ accessible name which is:
The link text is now:
- available to programs, like assistive tech, through accessibility APIs and HTML parsing
- being announced by screen readers, the most obvious assistive tech for discerning whether what’s in the HTML is usable

## Notes about research done

To check these issues, I did the following:
- checked that the visuals matched the semantics
- checked the browsers were communicating the right accessible names and state to the accessibility APIs by checking in Chrome and Firefox devtools
- checked the user experience in: 
   - Chrome on Windows 10 with the NVDA and JAWS screen readers
   - Chrome and Safari on MacOS with the Voicover screen reader

## Also included

Rolled into this is a change to the issue we have with extra spaces being announced in our emails with screen readers. This is now only observable in NVDA and not JAWS.


samuelhwilliams (approved):
I have not done a deep examination of this, but I have checked the HTML and relevants elements in devtools/accessibility tree and agree that it looks like these things you've removed are fixed.

I've not tested the screenreader tools either. But I know you have...

PR #4813, opened 2023-08-15
QR code validation

samuelhwilliams:
Validate a QR code's length at all possible stages:

1) When editing a template and inserting a fully-static QR code.
2) When sending a one-off letter and entering personalisation that allows a QR code to be created
3) When sending many letters via CSV upload

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -1934,6 +1934,42 @@ def test_send_one_off_letter_redirects_to_right_url(
     )
 
 
+def test_send_one_off_letter_qr_code_placeholder_too_big(
+    client_request,
+    platform_admin_user,
+    fake_uuid,
+    mock_get_service_letter_template_with_qr_placeholder,
+    mock_s3_upload,
+    mock_get_users_by_service,
+    mock_get_service_statistics,
+    mocker,
+):
+    mocker.patch("app.template_previews.get_page_count_for_letter", return_value=9)
+    with client_request.session_transaction() as session:
+        session["recipient"] = ""
+        session["placeholders"] = {
+            "address line 1": "foo",
+            "address line 2": "bar",
+            "address line 3": "",
+            "address line 4": "",
+            "address line 5": "",
+            "address line 6": "",
+            "address line 7": "SW1 1AA",
+        }
+
+    client_request.login(platform_admin_user)
+    page = client_request.post(
+        "main.send_one_off_step",
+        service_id=SERVICE_ONE_ID,
+        template_id=fake_uuid,
+        step_index=7,
+        _data={"placeholder_value": "content which makes the QR code too big " * 25},
+        _expected_status=200,
+    )
+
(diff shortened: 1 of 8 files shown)


CrystalPea (commented):
Looks good codewise.

Content-wise, isn't it a bit weird that the error message is very similar but different for sending one off, and for template editing?

"Cannot create a usable QR code - the text you entered makes the link too long" vs "Cannot create a usable QR code - the link you entered is too long" 🤔

samuelhwilliams:
There's been a fair chunk of back and forth on the QR code content in the doc here: https://docs.google.com/document/d/1EpiH8s-Ff_VAQwYafgrIo64mDUY0_BVIrHlarlZGB-M/edit

Hopefully that provides some context :)

In each case the error means a slightly different thing, or is shown in a slightly different situation, and so we do explain it differently. It's one of those cases where we think being uniform wouldn't be as helpful.

tests/app/main/views/test_templates.py
```diff
@@ -2312,6 +2312,31 @@ def test_should_not_create_too_big_template(
...
+        _data={
+            "name": "new name",
+            "subject": "subject",
+            "template_content": "qr: " + ("content" * 100),
+            "template_type": "letter",
+            "service": SERVICE_ONE_ID,
+        },
+        _expected_status=200,
+    )
+    assert "Cannot create a usable QR code - the link you entered is too long" in page.text
```
quis, line 2337: I prefer if these assertions were more specific. It means that if the content changes it’s much easier to see the difference, versus just saying that the expected content wasn’t somewhere in hundreds of lines of text.

```suggestion
    assert normalize_spaces(page.select_one(".error-message").text) == "Cannot create a usable QR code - the link you entered is too long" in page.text
```

tests/app/main/views/test_templates.py
```diff
@@ -2334,6 +2359,31 @@ def test_should_not_update_too_big_template(
...
```
quis, line 2384: Same here
```suggestion
    assert normalize_spaces(page.select_one(".error-message").text) == "Cannot create a usable QR code - the link you entered is too long"
```

tests/app/main/views/test_send.py
```diff
@@ -1934,6 +1934,42 @@ def test_send_one_off_letter_redirects_to_right_url(
...
+    page = client_request.post(
+        "main.send_one_off_step",
+        service_id=SERVICE_ONE_ID,
+        template_id=fake_uuid,
+        step_index=7,
+        _data={"placeholder_value": "content which makes the QR code too big " * 25},
+        _expected_status=200,
+    )
+
+    assert "Cannot create a usable QR code - the text you entered makes the link too long" in page.text
```
quis, line 1970: Same:
```suggestion
    assert normalize_spaces(page.select_one(".govuk-error-message").text) == "Cannot create a usable QR code - the text you entered makes the link too long"
```

quis (approved):
Good to go with a couple of non-essential changes to make the tests friendlier to our future selves

PR #4815, opened 2023-08-22
Bump utils to 70.0.2

quis:
Depends on:

- [x] https://github.com/alphagov/notifications-api/pull/3856

***

70.0.2
---
* Update sanitising of SMS content to include more obscure whitespace

70.0.0
---
* InvalidPhoneError messages have been updated. The new error messages are more user friendly.

69.0.0
---
* Remove old syntax for QR codes in letters (only `QR: http://example.com` will work now)

68.0.1
---
* Fix a bug with some HTML getting injected into QR codes

68.0.0
---
* Update return value of `BaseLetterTemplate.has_qr_code_with_too_much_data` from `bool` to `Optional[QrCodeTooLong]`.

67.0.0
---
* Add `has_qr_code_with_too_much_data` property to letter templates.
* Update RecipientCSV to detect and throw errors when rows generate QR codes with too much data in them. Anything using RecipientCSV to process letters will need to check for and report on the row-level property `qr_code_too_long`.

***

Complete changes: https://github.com/alphagov/notifications-utils/compare/66.1.0...70.0.2


quis:
New error messages when uploading a spreadsheet:

Screenshot 2023-08-23 at 10 11 48

PR #4818, opened 2023-08-30
Make font size smaller when numbers are very big

quis:
We have a problem where big numbers overflow the width of their columns. This happens when the numbers are bigger than we anticipated (for example when we designed the page we never imagined anyone would spend tens of millions of £ on text messages in a single year).

This commit adds some logic to reduce the font size for all the totals if any of them get so big that they overflow the width of the column. This gives us enough space to display the biggest numbers we currently store in the platform.

This happens in 3 places:

## 1. At the bottom of the service dashboard

Before | After
---|---
Screenshot 2023-08-30 at 11 35 00 | Screenshot 2023-08-30 at 11 35 11

## 2. The usage page for a service

Before | After
---|---
Screenshot 2023-08-30 at 11 33 52 | Screenshot 2023-08-30 at 11 34 07

## 3. The usage page for an organisation

Before | After
---|---
Screenshot 2023-08-30 at 11 31 51 | Screenshot 2023-08-30 at 11 32 00

***

Because we use tabular numbers for these elements we don’t need to worry about the width of 1,111,111 versus 9,999,999.

A function which takes 2 arguments, `smaller` and `smallest` isn’t a great piece of API design, but if we want to refactor that it should be as a separate PR. This commit just works within the confines of the existing argument names.

Similarly we can’t wrap this logic up in the `big_number` component because it doesn’t know about:
- the width of the column in which it’s shown
- whether other numbers on the page are too big

### DIFF
--- a/app/templates/views/dashboard/_usage.html
+++ b/app/templates/views/dashboard/_usage.html
@@ -1,9 +1,12 @@
 {% from "components/big-number.html" import big_number %}
 
+{% set is_high_volume = sms_cost >= 1_000_000 or letter_cost >= 1_000_000 %}
+{% set big_number_kwargs = {"smaller": not is_high_volume, "smallest": is_high_volume} %}
+
 <div class='govuk-grid-row ajax-block-container'>
   <div class='govuk-grid-column-one-third'>
     <div class="keyline-block">
-      {{ big_number("Unlimited", 'free email allowance', smaller=True) }}
+      {{ big_number("Unlimited", 'free email allowance', **big_number_kwargs) }}
     </div>
   </div>
   <div class='govuk-grid-column-one-third'>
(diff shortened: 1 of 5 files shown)


app/templates/views/dashboard/_usage.html
```diff
@@ -1,9 +1,12 @@
 {% from "components/big-number.html" import big_number %}
 
+{% set is_high_volume = sms_cost >= 1_000_000 or letter_cost >= 1_000_000 %}
+{% set big_number_kwargs = {"smaller": not is_high_volume, "smallest": is_high_volume} %}
```
samuelhwilliams, line 4:
```suggestion
{% set has_high_spender = sms_cost >= 1_000_000 or letter_cost >= 1_000_000 %}
{% set big_number_kwargs = {"smaller": not has_high_spend, "smallest": has_high_spend} %}
```

Maybe `spend` instead of `volume`? We talk about volumes elsewhere in terms of raw notifications sent so would be nice to separate these two concepts a bit. Definitely optional though.

Was tempted to go full tongue-in-cheek with `hey_big_spender` but 🤷

quis:
```suggestion
{% set is_big_spender = sms_cost >= 1_000_000 or letter_cost >= 1_000_000 %}
{% set big_number_kwargs = {"smaller": not is_big_spender, "smallest": is_big_spender} %}
```

samuelhwilliams (commented):
This works well 👍  Nice fix

samuelhwilliams (approved):

PR #4819, opened 2023-08-30
Make size of ‘Unlimted emails’ vary with count

quis:
We change the font size of the count of sent emails based on the service’s volumes, to stop the big numbers overflowing their container.

When we did this we missed that the allowance (unlimted) for emails also needs to change so the font sizes match.

Before | After
---|---
image | image


tombye (approved):
Looks good to me 👍🏻

PR #4820, opened 2023-08-31
Make link red in error message

samuelhwilliams:
We link to 'who can use notify' from an error message on email validation, but it's currently appearing blue. It should be red to look more consistent.

## Before
image

## After
image


app/main/validators.py
```diff
@@ -49,9 +49,11 @@ def __call__(self, form, field):
 
         message = """
             Enter a public sector email address or
-            <a class="govuk-link govuk-link--no-visited-state" href="{}">find out who can use Notify</a>
+            <a class="govuk-link govuk-link--no-visited-state notify-link--error" href="{}">
+            find out who can use Notify
+            </a>
         """.format(
-            url_for("main.guidance_who_can_use_notify")
+            url_for("main.guidance_who_can_use_notify").replace("\n", " ")
```
quis, line 56: Why is this being called on the output of `url_for`? Did you mean to call it on the string as a whole? Even then I’m not sure why it’s needed

quis:
I would kinda prefer we had a CSS rule to make any link inside an error message red. What do you think?

app/main/validators.py
samuelhwilliams: this shouldn't be here 🙃

samuelhwilliams:
> I would kinda prefer we had a CSS rule to make any link inside an error message red. What do you think?

My instinct is that that isn't how the design system/BEM stuff is supposed to work, but I would love a steer from @tombye on that. It does sound more consistent and simpler at first glance but I'm not an expert on the potential consequences and if it could cascade in an unexpected way, which I think the BEM thing is meant to address

Happy for this to wait until we get a steer. If that's a fine way to proceed it does feel more likely to give us a reliably style. Probably on something like `.govuk-error-message .govuk-link`?

samuelhwilliams:
Hey @tombye - would be interested in your thoughts on this as/when you have some time :)

tombye:
I'm a bit unsure about this tbh, because the links in the error summary link to errors in the page whereas the link we're converting here is external so it seems a bit of a shame to lose the colour as an indicator of that. Is it worth asking the design system team for their thoughts, or at least if they considered this when designing the error messages?

Btw, if we did want to have a blanket rule for styling all links in error messages red, I can't see why this wouldn't be a good use of the cascade. Design system themselves do this with error summary.

samuelhwilliams:
Going to close this off for now given we don't seem to have consensus.

PR #4821, opened 2023-08-31
Disable custom data retention for trial mode services

samuelhwilliams:
image

### DIFF
--- a/tests/app/main/views/service_settings/test_service_settings.py
+++ b/tests/app/main/views/service_settings/test_service_settings.py
@@ -1468,8 +1468,9 @@ class TestServiceDataRetention:
         ),
     )
     def test_shown_guidance_if_declared_over_1mil_notifications_per_year(
-        self, client_request, platform_admin_user, mocker, notification_volumes, show_guidance
+        self, client_request, platform_admin_user, mocker, service_one, notification_volumes, show_guidance
     ):
+        service_one["restricted"] = False
         for channel, volume in notification_volumes.items():
             mocker.patch(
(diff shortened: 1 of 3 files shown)


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -1592,6 +1667,99 @@ def test_shown_guidance_if_predicted_to_send_over_1mil_notifications_this_year(
...
+            _data={"days_of_retention": 10},
+            _expected_status=200,
+        )
+        assert mock_set.call_args_list == []
+        assert (
+            "You cannot change the data retention period yourself. "
+            "This is because your service is likely to send over 1 million messages this year."
+        ) in page.text
+
+    def test_shown_guidance_if_service_is_in_trial_mode(self, client_request, platform_admin_user, mocker, service_one):
```
CrystalPea, line 1714: I thought we try to avoid asserting on content when possible. It could work better to check that input fields for changing the retention are not present.

CrystalPea (commented):
Nice one.

Only one thing: quite a few tests for us not allowing users to change retention assert on content. I think the test would be stronger and less fragile if we asserted on absence of input fields connected to retention changes instead. Let me know what you think.

tests/app/main/views/service_settings/test_service_settings.py
CrystalPea: Ok, fair do.

CrystalPea (approved):

PR #4824, opened 2023-09-05
update normalised_service_name when changing service name

leohemsted:
notably, we still read from `email_from` when, for example, displaying email previews. we update both fields.

1. ~~add new column as nullable~~ (https://github.com/alphagov/notifications-api/pull/3865)
2. **write new values to both new and old column**
3. migrate old data. add unique and not null constraints.
4. delete references to old column in code
5. drop old column


samuelhwilliams (approved):
Looks good and works as expected locally. I think we'll need to clear the service caches before merging?

leohemsted:
so we're not _reading_ normalised_service_name anywhere yet - when someone updates a name it'll clear the old cache and replace it.

i think we _need_ to clear the cache between merging this and the next admin PR

i actually also think we _need_ to clear the cache between migrating the data in the next api PR, and then the code coming in that actually uses that column 🤔. think that PR needs to split into two

PR #4825, opened 2023-09-06
Update index.html

saimaghafoor:
This PR groups the links on the Using Notify page.

It will make it easier for users to scan the page to look for information instead of being faced with a long list of links.


samuelhwilliams:
This is what it looks like:

image

samuelhwilliams:
Previously these were all listed alphabetically. The groups make a lot more sense and easier to find the right section to look at, but within each section do we want to keep them alphabetical? I can't tell if they've been intentionally ordered in some other way. I think they're mostly alphabetical but there are a few that are definitely not 🤔

saimaghafoor:
Hey @samuelhwilliams, thanks for the preview, looks good! They've been ordered by popularity. The side nav will remain as is in alphabetical order

samuelhwilliams (approved):

PR #4831, opened 2023-09-15
Enable error summaries for service settings pages

BlessedDev:
This PR enables error summaries for the admin app service settings pages as specified in [this](https://trello.com/c/8JBMz4Nu/459-add-error-summaries-to-service-settings-page-forms) ticket.

After a discussion with @karlchillmaid and @tombye, `/services/<service_id>/service-settings/request-to-go-live/estimate-usage` is not being dealt with as a rethink of the pages content is being considered. It will be handled in a separate PR.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1755,11 +1755,14 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
 
     def validate(self):
         if self.contact_details_type.data == "url":
-            self.url.validators = [DataRequired(), URL(message="Must be a valid URL")]
+            self.url.validators = [
+                NotifyDataRequired(thing="a URL in the correct format"),
+                URL(message="Enter a URL in the correct format"),
+            ]
 
         elif self.contact_details_type.data == "email_address":
             self.email_address.validators = [
-                DataRequired(),
+                NotifyDataRequired(thing="an email address"),
                 Length(min=5, max=255, thing="email address"),
                 ValidEmail(),
             ]
@@ -1771,15 +1774,15 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
                 try:
                     normalised_number = normalise_phone_number(num.data)
                 except InvalidPhoneError as e:
-                    raise ValidationError("Must be a valid phone number") from e
+                    raise ValidationError("Enter a phone number in the correct format") from e
 
                 if normalised_number in {"999", "112"}:
-                    raise ValidationError("Must not be an emergency number")
+                    raise ValidationError("Phone number cannot be an emergency number")
 
                 return True
 
             self.phone_number.validators = [
-                DataRequired(),
+                NotifyDataRequired(thing="a phone number"),
                 Length(min=3, max=20, thing="phone number"),
                 valid_non_emergency_phone_number,
             ]
@@ -1788,7 +1791,7 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
 
 
 class ServiceReplyToEmailForm(StripWhitespaceForm):
-    email_address = make_email_address_field(label="Reply-to email address", gov_user=False)
+    email_address = make_email_address_field(label="Reply-to email address", thing="an email address", gov_user=False)
     is_default = GovukCheckboxField("Make this email address the default")
 
 
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2690,7 +2690,10 @@ class AcceptAgreementForm(StripWhitespaceForm):
             on_behalf_of_email=org.agreement_signed_on_behalf_of_email_address,
         )
 
-    version = GovukTextInputField("Which version of the agreement do you want to accept?")
+    version = GovukTextInputField(
+        "Which version of the agreement do you want to accept?",
+        validators=[NotifyDataRequired(thing="a version number")],
+    )
 
     who = GovukRadiosField(
         "Who are you accepting the agreement for?",
@@ -2715,9 +2718,16 @@ class AcceptAgreementForm(StripWhitespaceForm):
     )
 
     def __validate_if_nominating(self, field):
+
+        error_messages = {
+            "on_behalf_of_name": "Enter the name of the person accepting the agreement",
+            "on_behalf_of_email": "Enter the email address of the person accepting the agreement",
+        }
+
         if self.who.data == "someone-else":
             if not field.data:
-                raise ValidationError("Cannot be empty")
+                error_message = error_messages[field.name]
+                raise ValidationError(error_message)
         else:
             field.data = ""
 
@@ -2728,7 +2738,7 @@ class AcceptAgreementForm(StripWhitespaceForm):
         try:
             float(field.data)
         except (TypeError, ValueError) as e:
-            raise ValidationError("Must be a number") from e
+            raise ValidationError("Enter a version number in digits, like 3.1") from e
 
 
 class BroadcastAreaForm(StripWhitespaceForm):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1158,8 +1158,8 @@ class RenameServiceForm(StripWhitespaceForm):
     name = GovukTextInputField(
         "Service name",
         validators=[
-            DataRequired(message="Cannot be empty"),
-            MustContainAlphanumericCharacters(),
+            NotifyDataRequired(thing="a service name"),
+            MustContainAlphanumericCharacters(thing="Service name"),
             Length(max=255, thing="service name"),
         ],
     )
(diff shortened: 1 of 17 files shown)


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -4969,9 +4975,9 @@ def test_send_files_by_email_contact_details_displays_error_message_when_no_radi
 @pytest.mark.parametrize(
     "contact_details_type, invalid_value, error",
     [
-        ("url", "invalid.com/", "Must be a valid URL"),
-        ("email_address", "me@co", "Enter a valid email address"),
-        ("phone_number", "abcde", "Must be a valid phone number"),
+        ("url", "invalid.com/", "Enter a URL in the correct"),
```
karlchillmaid, line 4978:
```suggestion
        ("url", "invalid.com/", "Enter a URL in the correct format"),
```

tests/app/main/views/test_providers.py
```diff
@@ -385,7 +385,7 @@ def test_edit_sms_provider_ratio_submit(
     [
         ({"sms_provider_1": 90, "sms_provider_2": 20}, "The total must add up to 100%"),
         ({"sms_provider_1": 101, "sms_provider_2": 20}, "Must be between 0 and 100"),
-        ({"sms_provider_1": 99.9, "sms_provider_2": 0.1}, "Percentage must be a whole number"),
+        ({"sms_provider_1": 99.9, "sms_provider_2": 0.1}, "Enter percentage in digits"),
```
karlchillmaid, line 388:
```suggestion
        ({"sms_provider_1": 99.9, "sms_provider_2": 0.1}, "Enter a percentage in digits"),
```

karlchillmaid: Sam and I discussed this interaction and whether we could improve it.

For example, changing the design to up/down buttons that remove the need for users to enter 2 percentages that balance.

karlchillmaid (changes requested):
I’ve left a couple of small changes - hopefully they’re just things that got missed when updating the tests, but worth checking back against the errors themselves.

Apart from that, as far as I can tell without seeing everything in context, the content looks good. We discussed adding screenshots for every error, but it’s a lot of effort.

tests/app/main/views/test_providers.py
BlessedDev: @karlchillmaid  I have corrected the errors you highlighted

karlchillmaid (approved):
Can’t see any errors in the content.

samuelhwilliams:
I think we can also update `service_edit_email_reply_to` to include error summaries 👍

BlessedDev:
> This ticket talks about adding error summaries to "service settings pages". I've noticed quite a few pages under service settings that don't have error summaries enabled still after this PR. Have they been missed here, or are they part of some other ticket? If not the latter then I think we should try to cover off all of the service settings pages here.
> 
> Some missing pages:
> 
> * `service_edit_email_reply_to`
> * `email_branding_options`
> * `service_edit_sms_sender`
> * `service_edit_letter_contact`
> * `letter_branding_options`

I followed the ticket's specifications and they weren't covered in the ticket. I am not sure they are covered in other tickets. I do see your point about covering as much as possible rather than racking up work for later. I will address those pages as part of this PR.

app/main/forms.py
```diff
@@ -361,7 +361,7 @@ def pre_validate(self, form):
                 raise StopValidation(f"Enter {self.things} in digits")
 
             if not isinstance(self.data, int):
-                raise StopValidation(f"{sentence_case(self.things)} must be a whole number")
+                raise StopValidation(f"Enter {self.things} in digits")
```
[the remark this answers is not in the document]
BlessedDev: Agreed

samuelhwilliams:
The "Data retention" (platform admin) service settings edit page don't have an error summary on them either, I've just noticed.

image

Also the backlink on those pages is broken, if you want to fix that now go ahead - otherwise I'll try to remember to do it separately :)

BlessedDev:
> The "Data retention" (platform admin) service settings edit page don't have an error summary on them either, I've just noticed.
> 
> image
> Also the backlink on those pages is broken, if you want to fix that now go ahead - otherwise I'll try to remember to do it separately :)
> 
> Everything else here looks as expected so will be happy to 👍 ✅ this afterwards

Good catch. I will add a fix for both issues.

tombye (commented):
The code looks good to me except the comment I added and one thing I noticed that didn't seem right in how the validation for reply-to email addresses.

I used an existing reply-to email and saw this error: 'Enter an email address in the correct format, like name@example.gov.uk' when I was expecting one about using an existing address. I know it was in the right format because it was an existing address.

Other than that, it all looks great and is a ton of work so thanks for doing it all!

app/main/forms.py
```diff
@@ -1746,11 +1740,14 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
 
     def validate(self):
         if self.contact_details_type.data == "url":
-            self.url.validators = [DataRequired(), URL(message="Must be a valid URL")]
+            self.url.validators = [
```
tombye, line 1743: I meant to comment on this as part of my review (sorry to do it now instead). It seems odd to message with 'a URL in the correct format' for an 'input is required' error. I checked and this does match the spreadsheet so we should confirm with @karlchillmaid what the expected content is but it seems out of sync with how we do that type of error elsewhere here.

BlessedDev:
@samuelhwilliams I have pushed a fix for the broken back page link.

app/main/forms.py
```diff
@@ -2383,7 +2383,7 @@ class AdminServiceAddDataRetentionForm(StripWhitespaceForm):
     days_of_retention = GovukIntegerField(
         label="Days of retention",
         things="days of retention",
-        validators=[validators.NumberRange(min=3, max=90, message="Must be between 3 and 90")],
+        validators=[validators.NumberRange(min=3, max=90, message="The number of days must be between 3 and 90")],
```
samuelhwilliams, line 2386: We can also update the `AdminServiceEditDataRetentionForm` form below which is used on these pages too (or, well, editing an existing data retention as opposed to just creating one) :)

app/main/forms.py
```diff
@@ -1746,11 +1740,14 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
 
     def validate(self):
         if self.contact_details_type.data == "url":
-            self.url.validators = [DataRequired(), URL(message="Must be a valid URL")]
+            self.url.validators = [
+                NotifyDataRequired(thing="a URL in the correct format"),
```
karlchillmaid, line 1744:
```suggestion
                NotifyDataRequired(thing="a URL"),
```

app/main/forms.py
```diff
@@ -1746,11 +1740,14 @@ class ServiceContactDetailsForm(StripWhitespaceForm):
...
+                NotifyUrlValidator(),
```
karlchillmaid, line 1745: We need this validation error if the URL is not valid:

```suggestion
                NotifyUrlValidator(thing="a URL in the correct format"),
```

samuelhwilliams: That fragment is the default if none is provided, so it's best left blank here 👍

app/main/forms.py
karlchillmaid: @BlessedDev I’ve paired with Tom to correct the mistake I made in the spreadsheet for these 2 errors.

There’s another change described in the spreadsheet, which I’ll add here for clarity:

1. "Rename the form field `Link to a website`
2. Add hint text to the form field`For example, https://www.example.gov.uk`

app/main/forms.py
BlessedDev: @samuelhwilliams is right that it is the default error message and should doesn't need to be explicitly stated.

karlchillmaid: Sorry. Tom and I thought that this was the validation for a badly formatted URL.

app/main/forms.py
BlessedDev: > @BlessedDev I’ve paired with Tom to correct the mistake I made in the spreadsheet for these 2 errors.
> 
> There’s another change described in the spreadsheet, which I’ll add here for clarity:
> 
> 1. "Rename the form field `Link to a website`
> 2. Add hint text to the form field `For example, https://www.example.gov.uk`

These are great changes and I have implemented them. It is definitely much more user friendly.

Screenshot 2023-10-09 at 11 43 42

samuelhwilliams:
This page still needs error summaries enabled

image

samuelhwilliams (approved):

PR #4832, opened 2023-09-18
Update error handling for reply-to email address

BlessedDev:
This PR updates our error handling for when a user tires to add a reply-to email address that is already in use.
Instead of redirecting the user to service-settings/email-reply-to where a message is flashed, the error will now be displayed inline and as part of the error summary displayed at eh top of the page.

Before:
reply-to-address-already-in-use-before

After:
reply-to-address-already-in-use-after

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -451,34 +451,43 @@ def service_add_email_reply_to(service_id):
     form = ServiceReplyToEmailForm()
     first_email_address = current_service.count_email_reply_to_addresses == 0
     is_default = first_email_address if first_email_address else form.is_default.data
+
     if form.validate_on_submit():
         if current_user.platform_admin:
-            service_api_client.add_reply_to_email_address(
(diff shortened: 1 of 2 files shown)


samuelhwilliams (commented):
Good update here. There's still one case where we can throw a 500, which is when the user is a platform admin. Can we cover off that case as well please?

BlessedDev:
> Good update here. There's still one case where we can throw a 500, which is when the user is a platform admin. Can we cover off that case as well please?

I considered covering raising a separate ticket for it. However I guess since I am already in the neighbourhood I might as well address the issue in this PR.

samuelhwilliams:
I think that makes sense rather than adding overhead of another ticket :) thanks!

BlessedDev:
> I think that makes sense rather than adding overhead of another ticket :) thanks!

I have updated the logic so that the error handling is consistent for both platform admins and non platform admins.

samuelhwilliams (approved):
👏

PR #4834, opened 2023-09-20
Move Trial mode to Using Notify menu

karlchillmaid:
This PR updates our navigation to re-home the `Trial mode` page in the `Using Notify` section.

When we restructured our guidance at the start of the year we decided to put the Trial mode page in the Pricing section. We did this because trial mode is bound up in the transition from trying Notify for free to accepting responsibility for paying for messages. 

Our interpretation of support tickets and the card sort supported this to some extent, but it was by no means conclusive. There was lots of discussion at the time about whether this was the right place for this page - as you can make a case to put this `Trial mode` in any of the 3 main sections - so we said we’d keep it under review.

Lately Notify’s content designers have been more involved in user support. Seeing the context in which users ask about removing trial mode restrictions and requesting to go live, we think that the page might fit better in the Using Notify section.


samuelhwilliams:
image

image

There are a few extra things we still need to do here:

* Looks like we still need to update the left-hand nav for the page. This is done via the `navigation_links` part of the view: https://github.com/alphagov/notifications-admin/blob/2f0578a51f96b5d19f8639eae987acd24e244317/app/main/views/pricing.py#L46. We can replace `pricing_nav` with `using_notify_nav`, but we need to import that function as well (at the top of the file).
* We probably should update the URL from `/pricing/trial-mode` to `/using-notify/trial-mode`. This will also mean we need to add a redirect from the former to the latter so that we don't break any links.

Happy to pair on getting any of this sorted, but if you want to give it a crack yourself first feel free 🙌

samuelhwilliams (approved):
nice 👍

PR #4838, opened 2023-09-22
enable error summary for /verify

BlessedDev:
This PR enables error summary for `/verify`

### DIFF
--- a/app/templates/views/two-factor-sms.html
+++ b/app/templates/views/two-factor-sms.html
@@ -10,7 +10,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Check your phone</h1>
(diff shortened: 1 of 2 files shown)


app/templates/views/two-factor-sms.html
```diff
@@ -10,7 +10,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Check your phone</h1>
+    <h1 class="govuk-heading-xl">Check your phone</h1>
```
quis, line 13: We don’t use the `xl` size anywhere. It’s fine to keep this as `govuk-heading-l` now because enabling the error summary will fix the spacing
```suggestion
    <h1 class="govuk-heading-l">Check your phone</h1>
```

quis (approved):

PR #4846, opened 2023-09-28
Template version history tweaks

samuelhwilliams:
## commit 1: Add links from template history to each version

image

## commit 2: Paginate template history

We know some services have huge numbers of template versions, and this
page effectively breaks for those users.

Let's put pagination on this to reduce the HTML size for those services,
and keep the page a bit more responsive.

Note that this is a quick win rather than a full fix, as the API doesn't
support pagination on this endpoint and so still sends the entire
template history across. We paginate on the admin side only.

We should add pagination to the API to fix this properly, but it's more
complex since we're working across two apps and have to tweak our
template version caching as well.

### Pictures (page size reduced to 3 for screenshot purposes)

image

image

### DIFF
--- a/app/templates/views/templates/template_history.html
+++ b/app/templates/views/templates/template_history.html
@@ -1,17 +1,22 @@
 {% extends "withnav_template.html" %}
 {% from "components/page-footer.html" import page_footer %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+
+{% set page_title = "Version " + template._template.version|string + " : " + template.name %}
 
 {% block service_page_title %}
-  {{ template.name }}
+  {{ page_title }}
+{% endblock %}
+
+{% block backLink %}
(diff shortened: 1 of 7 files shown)


app/templates/views/templates/template_history.html
```diff
@@ -1,17 +1,22 @@
...
+{% block backLink %}
+  {{ govukBackLink({ "href": url_for('main.view_template_versions', service_id=current_service.id, template_id=template.id) }) }}
+{% endblock %}
 
-  <h1 class="heading-large">{{ template.name }}</h1>
+{% block maincolumn_content %}
+  <h1 class="heading-large">{{ page_title }}</h1>
 
   <div class="govuk-grid-row">
-    {% with show_title=False %}
```
whpearson, line 14: A quick query. where/how was show_title being used?

samuelhwilliams: It wasn't being used any more from my checks

whpearson (approved):
Tested locally all looks good.

PR #4848, opened 2023-09-29
Remove email_from

leohemsted:
1. ~~add new column as nullable~~ (https://github.com/alphagov/notifications-api/pull/3865)
2. ~~write new values to both new and old column~~ (https://github.com/alphagov/notifications-admin/pull/4824)
3. ~~migrate old data. add unique and not null constraints.~~ (https://github.com/alphagov/notifications-api/pull/3881)
4. ~~use new column~~ (https://github.com/alphagov/notifications-api/pull/3868)
5. **delete references to old column in code**
6. drop old column

we won't need to clear cache around this - the cache will have extra values that will jsut be ignored when the json is loaded into the `Service` model

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -109,7 +109,6 @@ def service_name_change(service_id):
             normalised_service_name = email_safe(form.name.data)
             current_service.update(
                 name=form.name.data,
-                email_from=normalised_service_name,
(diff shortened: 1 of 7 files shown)


whpearson (approved):
Changes look good, tests run fine.

CrystalPea (approved):

PR #4849, opened 2023-10-02
Extend maximum time for scheduled jobs to 7 days

quis:
4 days is sometimes not enough if there is a weekend with a bank holiday before and after it.

7 days is a sensible new limit because it:
- accounts for these longer weekends or holiday periods like Christmas
- without introducing any ambiguity about which ‘Thursday’ is meant (because only one of each weekday will ever be shown at once)

Before | After
---|---
image | image

***

Depends on:
- [ ] https://github.com/alphagov/notifications-api/pull/3911

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -105,11 +105,12 @@ from app.utils.user_permissions import (
 
 
 def get_time_value_and_label(future_time):
+    timestamp = future_time.astimezone(pytz.timezone("Europe/London"))
     return (
         future_time.replace(tzinfo=None).isoformat(),
         "{} at {}".format(
-            get_human_day(future_time.astimezone(pytz.timezone("Europe/London"))),
-            get_human_time(future_time.astimezone(pytz.timezone("Europe/London"))),
+            get_human_day(timestamp),
+            get_human_time(timestamp),
         ),
     )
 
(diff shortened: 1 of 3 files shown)


samuelhwilliams (approved):
Posted this approval comment on the wrong PR originally 🙃 But both sides look good.

I think this could go in (checked out locally and looks good).

image

But if we're providing these extra dates, I start to feel like including the date+month would be useful. Some people definitely will be date oriented (I want these messages to send on the 1st of the month) rather than day oriented, and I think in either case it helps to provide extra immediate clarity that these are all in the future.

So probably I'm suggesting/pondering:

* Today
* Tomorrow
* Sunday 8th October
* Monday 9th October
* Tuesday 10th October
* Wednesday 11th October
* Thursday 12th October

samuelhwilliams:
Pushed an update to add date/month:

image

image

app/main/forms.py
```diff
@@ -120,11 +121,14 @@ def get_human_time(time):
 
 def get_human_day(time):
     #  Add 1 hour to get ‘midnight today’ instead of ‘midnight tomorrow’
```
CrystalPea, line 123: this comment is in a wrong place now.

app/main/forms.py
```diff
@@ -120,11 +121,14 @@ def get_human_time(time):
...
-    time = (time - timedelta(hours=1)).strftime("%A")
-    if time == datetime.utcnow().strftime("%A"):
+    date_format = "%A %-d %B"
```
CrystalPea, line 124: would be useful to do inline comment like:

```
date_format = "%A %-d %B"  # for example "Monday 4 January"
```

CrystalPea (approved):
Looks good, only needs some tweaks around comments.

PR #4853, opened 2023-10-06
Fix spacing under new date picker

quis:
As part of the work<sup>1</sup> to build the new date and time picker one of our old spacing classes, `page-footer`, was replaced with one of the new spacing classes from GOV.UK Frontend, `govuk-!-margin-bottom-6`.

`page-footer` and `govuk-!-margin-bottom-6` both add 30px of bottom margin.

Both are inside a `<div>` with the class `bottom-gutter-3-2` which adds `45px` of bottom margin.

Before the new date picker the bottom margin of `page-footer` combines the bottom margin of `bottom-gutter-3-2` for a total of `45px` of (the larger of the 2 values).

With the new date picker the bottom margin of `page-footer` adds on to the bottom margin of `bottom-gutter-3-2` for a total of `75px` of (the 2 values summed together).

Not 100% sure why this has changed. On the page without the date picker (for example when you’re about to send a letter job) the appearance is the same with and without the `page-footer` class. So it’s something to do with the new date picker ‘activating’ the bottom margin on `page-footer` 🤔

Before before | Before | After
---|---|---
image | image | image
image | image | image

1. https://github.com/alphagov/notifications-admin/commit/afdf04148ba30726cdf4db61b882d628032d0747#diff-f10085db71986399c42e16e82dd64862f2a262b03e08491c304d758fc427f337L37

### DIFF
--- a/app/templates/views/check/ok.html
+++ b/app/templates/views/check/ok.html
@@ -34,7 +34,7 @@
 
   {{ template|string }}
   <div class="bottom-gutter-3-2">
-    <form method="post" enctype="multipart/form-data" action="{{url_for('main.start_job', service_id=current_service.id, upload_id=upload_id)}}" class="govuk-!-margin-bottom-6">
+    <form method="post" enctype="multipart/form-data" action="{{url_for('main.start_job', service_id=current_service.id, upload_id=upload_id)}}">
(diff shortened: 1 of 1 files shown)


tombye (approved):
Can't honestly say why this happened (sorry) but the fix looks good to me 👍🏻

PR #4857, opened 2023-10-12
Fix attributes missed from day format update

tombye:
Fixes this issue where the radios aren't focused when you move to the view showing the hours of a day:

|Currently|With this fix|
|---|---|
|radio-select-bug|radio-select-bug-after|

The format of the days we show changed to include the month and day of the month in [this pull request](https://github.com/alphagov/notifications-admin/pull/4849). It updated the human-readable days but didn't change the code that baked the day into these parts of the HTML:
- the names of radios, ie. `times-for-sunday-15-october`
- the id of fieldsets for a day, ie. `radio-select__times-for-sunday-15-october`

It's not surprising these were missed as there weren't any tests checking them. There should have been though, because not updating these parts of the code meant:
- it couldn't find the radios to shift focus to when the times view shows
- fieldsets wrapping the radios for a day had invalid id attributes

This fixes the code mentioned and adds tests to check those parts of the JS.

### DIFF
--- a/app/assets/javascripts/radioSelect.js
+++ b/app/assets/javascripts/radioSelect.js
@@ -64,10 +64,14 @@
         };
       }
 
+      function _getKeyFromDayLabel (label) {
+        return label.toLowerCase().replace(/\s/g, '-');
+      }
+
       const timesByDay = {};
 
       const days = this.$component.data('days').split(',').map(day => {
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
Oh no. Great spot and good idea to ask tests on this now 🙏

PR #4860, opened 2023-10-17
Restrict characters for SMS sender IDs

samuelhwilliams:
Our text message providers have let us know that having apostrophes in SMS sender IDs can, in some cases, cause issues at the interconnect between mobile network operators.

We need to stop allowing apostrophes in sender IDs. The new acceptable character set is:

Alphanumeric
ASCII spaces ( )
ASCII dash (-)
ASCII underscore (_)
ASCII period/full-stop (.)
ASCII ampersand (&)

This patch updates validation only, so that services cannot create/update sender IDs to contain these characters in the future.

We will separately ego through existing sender IDs with these invalid characters, bring them in line with the new rules, and notify service(s) as needed.


app/main/validators.py
```diff
@@ -157,11 +157,21 @@ def __call__(self, form, field):
...
+    - Ampersands
+    - Periods
+    - Plain dashes
+    - Plain underscores
+    """
 
-    def __init__(self, message="Text message sender can only include letters and numbers"):
+    regex = re.compile(r"^[a-zA-Z0-9 &.\-_]+$")
+
+    def __init__(self, message="Text message sender can only include: letters, numbers, spaces, and & . - _"):
```
samuelhwilliams, line 174: @karlchillmaid could I get your input on this error message please (no rush). The old error message was in fact lying, but either way it would be nice to update this now to describe the accepted characters properly :)

CrystalPea (approved):
Good work 👍🏼 .

Do we want to include the information about apostrophes in particular causing issues in our code? We could do it by adding a test case that includes an apostrophe to `test_incorrect_sms_sender_input` with a comment?

For example:
```
(r"Sender's name", "Text message sender can only include letters and numbers"),  # singling this out, as our suppliers warned us that apostrophes cause interoperability issues with MNOs
```

That could help if a user makes a request for apostrophe in the future and we consider adding it back.

samuelhwilliams:
There is this already:

```
pytest.param("'UC'", False, None, marks=pytest.mark.xfail), # Straight single quotes are not allowed
```

But I can expand the comment a bit 👍 

But a bit of git digging should still quickly turn up both the commit message, and this back PR, for the full context.

app/main/validators.py
samuelhwilliams: I will merge for this now, and we can revisit the error message if we need to.

PR #4866, opened 2023-10-20
Show recipients for scheduled jobs

quis:
This is useful so that:
- you can go back and check you uploaded the right thing
- someone else can see who a job is going to
- someone else who spots a mistake can correct it by downloading and re-uploading the original file

preview

### DIFF
--- a/app/assets/javascripts/removeInPresenceOf.js
+++ b/app/assets/javascripts/removeInPresenceOf.js
@@ -1,0 +1,23 @@
+(function (Modules) {
+
+  'use strict';
+
+  Modules.RemoveInPresenceOf = function () {
+
+    let elementToRemove;
+    let observer = new MutationObserver(function () {
+      if (document.getElementById(elementToRemove.dataset.targetElementId)) {
+        elementToRemove.parentNode.removeChild(elementToRemove);
+        observer.disconnect();
+      }
+    });
+
+    this.start = function ($elementToRemove) {
+
+      elementToRemove = $elementToRemove[0];
+      observer.observe(document.getElementById('main-content'), { childList: true, subtree: true });
+
+    };
+
+  };
+})(window.GOVUK.NotifyModules);
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -15,6 +15,7 @@ from flask import (
 )
 from markupsafe import Markup
 from notifications_python_client.errors import HTTPError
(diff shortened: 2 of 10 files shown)


quis:
I’ve redesigned this to not use a sticky footer because the sticky footer overlaps the horizontal scrollbar on the scrollable table.

Here’s how it looks now:

image

image

app/assets/javascripts/removeInPresenceOf.js
```diff
@@ -0,0 +1,23 @@
...
+        observer.disconnect();
+      }
+    });
+
+    this.start = function ($elementToRemove) {
+
+      elementToRemove = $elementToRemove[0];
+      observer.observe(document.getElementById('main-content'), { childList: true, subtree: true });
+
+    };
```
samuelhwilliams, line 20: from a quick functional test this seems to work nicely 👍  I'm not super critical of the actual JS though .. so if you want a valid second opinion maybe tag in Tom.

samuelhwilliams (approved):
nice 👍

app/main/views/jobs.py
```diff
@@ -112,6 +124,23 @@ def view_job_csv(service_id, job_id):
...
 
+@main.route("/services/<uuid:service_id>/jobs/<uuid:job_id>/original.csv")
+@user_has_permissions("view_activity")
+def view_job_original_file_csv(service_id, job_id):
+    job = Job.from_id(job_id, service_id=service_id)
+
+    if not job.scheduled:
+        abort(404)
+
+    original_file_contents = s3download(current_service.id, job.id)
```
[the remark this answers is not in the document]
quis: This should be caught by line 132–133. `job.scheduled` returns `False` once the job begins sending.

I think if the file was missing (deleted before the 7 scheduled days) then that’s an exception we’d want to know about.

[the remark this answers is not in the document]
samuelhwilliams: ah grand, didn't realise scheduled went to false 👍

app/assets/javascripts/removeInPresenceOf.js
quis: @tombye would you like to cast your eye over the JS here?

tombye: All looks good to me. I can see this could easily be a pattern we could use elsewhere in future so also nice to have this as an example of a clean way of reacting to our AJAX updates 👍🏻

PR #4868, opened 2023-10-25
Fix viewing letter template version pages

samuelhwilliams:
We weren't passing `page` through to template-preview, meaning that each page preview was just showing page 1. Passing this through renders an accurate preview of each page in the template.

Also fixes the 'view paginated template history' views

## Before
before

## After
after

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -313,7 +313,7 @@ def view_template_version(service_id, template_id, version):
 @user_has_permissions(allow_org_user=True)
 def view_template_version_preview(service_id, template_id, version, filetype):
     template = current_service.get_template(template_id, version=version)
-    return TemplatePreview.from_utils_template(template, filetype)
+    return TemplatePreview.from_utils_template(template, filetype, page=request.args.get("page"))
 
 
 def _add_template_by_type(template_type, template_folder_id):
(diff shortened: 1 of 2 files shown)


quis:
Is there a reason for not making `page` a required argument?

samuelhwilliams:
I _think_ we should be able to safely do that but I'm not 100% sure yet that nowhere calls that endpoint without a `page`, so this feels safer in the short-term until we can be certain.

We pass the preview URL into utils, and it looks like that always appends a page parameter (in a fairly naive way 🙃) so I think it should be alright...

leohemsted (approved):

quis:
I think this is why it’s optional in utils: https://github.com/alphagov/notifications-utils/pull/1041

Should be safe to make it non-optional in `TemplatePreview.from_utils_template`, but happy to prioritise fixing the problem first by merging this.

samuelhwilliams:
Just checked [last 7 days of Kibana logs](https://kibana.logit.io/s/9423a789-282c-4113-908d-0be3b1bc9d1d/goto/6148f6b8f69227c21cb25bc118471e73) and there's no calls to this view without a `page` param, so we're probably also fine from that perspective to make it required. But yeah let's get this in now.

PR #4871, opened 2023-10-26
Make `page` a required argument

samuelhwilliams:
Refactor this method signature slightly to make `page` required, and make `filetype` required via kwarg to make it a bit more readable.

### DIFF
--- a/app/template_previews.py
+++ b/app/template_previews.py
@@ -1,5 +1,6 @@
 import base64
 from io import BytesIO
+from typing import Optional
 
 import requests
 from flask import abort, current_app, json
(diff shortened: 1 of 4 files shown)


leohemsted (commented):
what's the general motivation for this?

i'm unsure as to why `from_database_object` needs a different signature to `from_utils_template`, `from_notification`, etc.  (other than the first arg obvs). from_example_template, from_valid, from_invalid are a bit snowflakey too.

Do we have concerns that `TemplatePreview.from_database_object(my_template, filetype='pdf')` is ambiguous 🤔 ?

I guess if filetype is `png`, then page _needs_ to be provided right. But if filetype is `pdf` then page _shouldn't be provided? I think? That's kinda tricky to unstick while `png` and 
`pdf` are all handled in the same endpoints and function calls throughout admin/template preview i guess.

i like having the args being moved into always-a-kwarg though

samuelhwilliams:
This will go in after Pea's refactor anyway so the png/pdf thing should disappear and I'll refactor this again, so it can kinda just hold for now...

samuelhwilliams:
But the general motivation is that we just had a bug in the `png` flow because `page` was missed off and if `page` was a required arg that probably wouldn't have happened.

quis:
Closing this because the code has since been refactored

PR #4872, opened 2023-10-26
Don’t reimport `Markup` from Flask

quis:
This is not supported in the newest versions of Flask. It should be directly imported from `markupsafe` instead.

Was deprecated in https://github.com/pallets/flask/pull/4996

Already did this work in utils in https://github.com/alphagov/notifications-utils/pull/874


whpearson (approved):
Looks good

PR #4874, opened 2023-10-27
Bump core dependencies to latest versions

quis:
Fixes #4869 (security update)

***

These changes work with the previous and new versions of the dependencies:
- Don’t re-import `Markup` from Flask
- Register `before_request` and `after_request` globally
- Remove blinker as explicit dependency

Then we update our core frameworks to the latest versions so they all play nicely with each other, including:
- Flask
- Flask-WTF
- wtforms
- Flask-Login (no release for the newest compatible code here, so we have to go by a commit SHA instead)
- Werkzeug (the package with the vulnerability)

Then these changes are to fix problems in our code caused by the new versions of the dependencies:
- Update method signature of `Form.validate`
- Update return value of `iter_choices`
- Don’t expected encoded URL querystring argument value

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -854,7 +854,7 @@ class OnOffField(GovukRadiosField):
             # This overrides WTForms default behaviour which is to check
             # self.coerce(value) == self.data
             # where self.coerce returns a string for a boolean input
-            yield (value, label, (self.data in {value, self.coerce(value)}))
+            yield (value, label, (self.data in {value, self.coerce(value)}), {})
 
 
 class OrganisationTypeField(GovukRadiosField):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1139,7 +1139,7 @@ class TwoFactorForm(StripWhitespaceForm):
 
     sms_code = SMSCode("Text message code")
 
-    def validate(self):
+    def validate(self, *args, **kwargs):
 
         if not self.sms_code.validate(self):
             return False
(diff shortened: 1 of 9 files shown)


app/main/forms.py
```diff
@@ -1139,7 +1139,7 @@ def __init__(self, validate_code_func, *args, **kwargs):
 
     sms_code = SMSCode("Text message code")
 
-    def validate(self):
+    def validate(self, *args, **kwargs):
```
leohemsted, line 1142: this function doesn't call `super().validate()` at all, and I think it should 🤔 

but also, we should make sure all our calls are to `super().validate(*args, **kwargs)` so we aren't discarding things if we start using `extra_validators` or some other new param in the future

tests/app/main/views/test_two_factor.py
```diff
@@ -326,7 +326,7 @@ def test_two_factor_sms_should_activate_pending_user(
 
 @pytest.mark.parametrize(
     "extra_args, expected_encoded_next_arg",
-    (({}, ""), ({"next": "https://example.com"}, "?next=https%3A%2F%2Fexample.com")),
+    (({}, ""), ({"next": "https://example.com"}, "?next=https://example.com")),
```
leohemsted, line 329: it's valid (https://stackoverflow.com/a/68151221)

but can we add a test where the `next` url has a querystring itself, and specifically an `&` i guess `https://example.com?foo=bar&baz=waz? i want to make sure that we dont strip querystrings from any login redirects

app/main/forms.py
quis: Fixed in 2fc293c

tests/app/main/views/test_two_factor.py
quis: Added an extra test in fb08afe

leohemsted (approved):

PR #4878, opened 2023-11-01
give zendesk tickets different subjects

leohemsted:
we often struggle to distinguish tickets as they come in, and appropriately identify them.

We can do a low-impact change here to make our lives simpler - by setting the subject based on what flow the user took through the triage process.

If they're not signed in, they get GENERAL
If they're signed in, they get PROBLEM or QUESTION depending on which radio they press

these new subjects will be seen by the user when we reply to their tickets

image


CrystalPea (approved):
Nice one!

PR #4879, opened 2023-11-02
Enable error summaries for template page forms

BlessedDev:
This PR enables error summaries for template page forms.
https://trello.com/c/JD46QBcO/457-add-error-summaries-to-templates-page-forms


karlchillmaid (approved):
I think all the content is consistent with the designs in the spreadsheet.

app/main/forms.py
```diff
@@ -1461,30 +1461,32 @@ def validate_address(self, field):
...
-                raise ValidationError("Last line of the address must be a UK postcode or another country")
+                raise ValidationError("The last line of the address must be a UK postcode or the name of a country")
             if address.international:
                 raise ValidationError("You do not have permission to send letters to other countries")
             raise ValidationError("Last line of the address must be a real UK postcode")
 
         if address.has_invalid_characters:
             raise ValidationError(
-                "Address lines must not start with any of the following characters: @ ( ) = [ ] ” \\ / , < > ~"
+                'Address lines cannot start with any of the following characters: @ ( ) = [ ] " " \\ / , < > ~'
```
karlchillmaid, line 1477: Should the quote marks be ‘smart’ quotes?

> “ ”

quis: These are the exact characters the validation is checking:
https://github.com/alphagov/notifications-utils/blob/3e6c6e018163c189c8aaf753cd423aab35903809/notifications_utils/postal_address.py#L30

So it’s a straight double quote, not a smart quote. Either way we shouldn’t have the quote repeated:
```suggestion
                'Address lines cannot start with any of the following characters: @ ( ) = [ ] " \\ / , < > ~'
```

Ideally this error message would reference the definition in code so it updates automatically but that doesn’t have to be for this PR.

quis: Made a PR to work on this here: https://github.com/alphagov/notifications-utils/pull/1071

quis: Fixed in [257434f](https://github.com/alphagov/notifications-admin/pull/4879/commits/257434fd3253e21cab4fd209d6a4108aa1e1a30f)

quis:
I’ve rebased this to resolve the merge conflicts I introduced by merging https://github.com/alphagov/notifications-admin/pull/4882

samuelhwilliams:
I've spotted a few places where we've got some issues around error summaries on template pages.

1) When sending a one-off notification, the `/one-off/step-<int:index>` page doesn't have error summaries
image


2) When creating a new folder, the 'Folder name' field doesn't generates an error message 'Cannot be empty'.
image


@BlessedDev I think we should fix these before merging :)

PR #4880, opened 2023-11-02
Enable error summaries for support pages

BlessedDev:
This PR enables error summaries for support page forms.
https://trello.com/c/hh1xh8il/458-add-error-summaries-to-support-page-forms

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1704,8 +1704,10 @@ class SupportRedirect(StripWhitespaceForm):
 
 class FeedbackOrProblem(StripWhitespaceForm):
     name = GovukTextInputField("Name (optional)")
-    email_address = make_email_address_field(label="Email address", gov_user=False, required=True)
-    feedback = TextAreaField("Your message", validators=[DataRequired(message="Cannot be empty")])
+    email_address = make_email_address_field(
+        label="Email address", gov_user=False, required=True, thing="your email address"
+    )
+    feedback = TextAreaField("Your message", validators=[NotifyDataRequired(thing="your message")])
 
 
 class Triage(StripWhitespaceForm):
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1715,7 +1715,7 @@ class Triage(StripWhitespaceForm):
             ("yes", "Yes"),
             ("no", "No"),
         ],
-        thing="yes or no",
+        thing="‘yes’ if you want us to contact you when we’re doing user research",
     )
 
 
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1703,11 +1703,11 @@ class SupportRedirect(StripWhitespaceForm):
 
 
 class FeedbackOrProblem(StripWhitespaceForm):
+    feedback = TextAreaField("Your message", validators=[NotifyDataRequired(thing="your message")])
     name = GovukTextInputField("Name (optional)")
     email_address = make_email_address_field(
         label="Email address", gov_user=False, required=True, thing="your email address"
     )
-    feedback = TextAreaField("Your message", validators=[NotifyDataRequired(thing="your message")])
 
 
 class Triage(StripWhitespaceForm):
@@ -1717,7 +1717,7 @@ class Triage(StripWhitespaceForm):
             ("yes", "Yes"),
             ("no", "No"),
         ],
-        thing="‘yes’ if you want us to contact you when we’re doing user research",
+        thing="‘yes’ if it is an emergency",
     )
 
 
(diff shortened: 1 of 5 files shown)


app/main/forms.py
```diff
@@ -1683,7 +1685,7 @@ class Triage(StripWhitespaceForm):
             ("yes", "Yes"),
             ("no", "No"),
         ],
-        thing="yes or no",
+        thing="‘yes’ if you want us to contact you when we’re doing user research",
```
quis, line 1688: This doesn’t match the label here: https://github.com/alphagov/notifications-admin/blob/73269d7b4b9e13b9ed016e430d9da8ec304ba17b/app/main/forms.py#L1681

app/main/forms.py
```diff
@@ -1672,8 +1672,10 @@ class SupportRedirect(StripWhitespaceForm):
 
 class FeedbackOrProblem(StripWhitespaceForm):
     name = GovukTextInputField("Name (optional)")
-    email_address = make_email_address_field(label="Email address", gov_user=False, required=True)
-    feedback = TextAreaField("Your message", validators=[DataRequired(message="Cannot be empty")])
+    email_address = make_email_address_field(
+        label="Email address", gov_user=False, required=True, thing="your email address"
+    )
+    feedback = TextAreaField("Your message", validators=[NotifyDataRequired(thing="your message")])
```
quis, line 1678: Can we make the order of the items in the summary (email address, message) match the order of the fields on the page (message, email address):

image

BlessedDev: Yes, I will fix the order in which they appear.

app/main/forms.py
BlessedDev: Good catch. I have updated the error message.

quis (approved):

app/main/forms.py
```diff
@@ -1671,9 +1671,16 @@ class SupportRedirect(StripWhitespaceForm):
 
 
 class FeedbackOrProblem(StripWhitespaceForm):
+    custom_field_order = (
+        "feedback",
+        "name",
+        "email_address",
+    )
```
samuelhwilliams, line 1678: Just to provide some extra context, we don't _need_ to use this here, we can just reorder the fields below to be in the 'correct' order.

`custom_field_order` is a break-glass solution for when we've got forms that declare fields across levels of inheritance.

It would be nice to remove this and just re-order the below fields as needed, but I'm not going to block you if you decide not to do it :)

BlessedDev: I take your point. `custom_field_order` is over kill so I have implemented the simpler solution

samuelhwilliams:
[suggestion]: The commit message is still `Add custom_field_order to FeedbackOrProblem Form`, which doesn't describe what the commit is doing, so maybe update this. :)

app/main/forms.py
```diff
@@ -1683,7 +1685,7 @@ class Triage(StripWhitespaceForm):
             ("yes", "Yes"),
             ("no", "No"),
         ],
-        thing="yes or no",
+        thing="‘yes’ if you want us to contact you if it is an emergency",
```
karlchillmaid, line 1688: I cannot find this one in the spreadsheet.

It looks like it’s based on a pattern we use for other errors.

If someone can talk me through the scenario and the trigger for this error, I might be able to improve the content slightly.

quis: Scenario:
1. you are signed in to Notify
2. it is out of hours (for example in the evening)
3. you go to raise a support ticket
4. you select ‘problem’ rather than ‘question’
5. you get an extra page where we ask whether or not it’s an emergency
6. you fill in the rest of your support message as usual

If you answer ‘yes’ to 5. then we will treat the ticket as a P1 incident, set off Pagerduty, etc

BlessedDev: > I cannot find this one in the spreadsheet.
> 
> It looks like it’s based on a pattern we use for other errors, but I’m not sure about having 2 ‘if’s in the error.
> 
> If someone can talk me through the scenario and the trigger for this error, I might be able to improve the content slightly.

@karlchillmaid Do you still want to update the error message?

karlchillmaid: @BlessedDev Yes. Sorry, I haven’t had time to look at this yet. I’ll pick it up tomorrow.

tombye (approved):
I'll leave the content and less frontendy dev reviews to the others here but the frontend changes look good to me 👍🏻

app/main/forms.py
karlchillmaid: @quis so the validation is more:

> Select ‘yes’ if it is an emergency

And not so much about whether or not the user wants us to contact them?

quis: @karlchillmaid that’s correct

karlchillmaid: @quis nice. easy fix then.

```suggestion
        thing="‘yes’ if it is an emergency",
```

PR #4884, opened 2023-11-03
Fix a bug where our custom unauthorised_handler lost query params

CrystalPea:
from url we want to redirect to after user signs in.

We were using request.path, which doesn't include the query params. Now we instead get the path and the params using urlparse and urlunparse.

Flask login use similar approach to get next_url, too: https://github.com/maxcountryman/flask-login/blob/main/src/flask_login/utils.py#L66

### DIFF
--- a/tests/app/main/views/test_sign_in.py
+++ b/tests/app/main/views/test_sign_in.py
@@ -338,18 +338,7 @@ def test_when_signing_in_as_invited_user_you_cannot_accept_an_invite_for_another
 
 def test_sign_in_again_handler_passes_next_url_with_params(client_request, fake_uuid):
     client_request.logout()
-    client_request.get(
-        "main.service_verify_reply_to_address",
-        service_id=SERVICE_ONE_ID,
-        notification_id=fake_uuid,
-        is_default=True,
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_sign_in.py
```diff
@@ -334,3 +334,22 @@ def test_when_signing_in_as_invited_user_you_cannot_accept_an_invite_for_another
...
+        _expected_redirect=url_for(
+            ".sign_in",
+            next=url_for(
+                ".service_verify_reply_to_address",
+                service_id=SERVICE_ONE_ID,
+                notification_id=fake_uuid,
+                is_default=True,
+            ),
+        ),
+    )
```
samuelhwilliams, line 355:
```suggestion
    url = url_for(
        "main.service_verify_reply_to_address", service_id=SERVICE_ONE_ID, notification_id=fake_uuid, is_default=True
    )
    client_request.get_url(url, _expected_redirect=url_for(".sign_in", next=url))
```

Might be tempted to reduce this like so, so that it's clearer the two `url_for` calls are identical at a glance.

samuelhwilliams (approved):

PR #4885, opened 2023-11-07
Remove code supporting the PaaS Redis

idavidmcdonald:
https://trello.com/c/9XD1L1cR/430-delete-old-paas-redis-and-references-to-it

Best reviewed commit by commit


manifest.yml.j2
```diff
@@ -51,12 +50,7 @@ applications:
       ROUTE_SECRET_KEY_1: '{{ ROUTE_SECRET_KEY_1 }}'
       ROUTE_SECRET_KEY_2: '{{ ROUTE_SECRET_KEY_2 }}'
       METRICS_BASIC_AUTH_TOKEN: {{ METRICS_BASIC_AUTH_TOKEN }}
-      {% if REDIS_URL is defined %}
       REDIS_URL: '{{ REDIS_URL }}'
-      {% endif %}
-      {% if REDIS_ENABLED is defined %}
```
klssmith, line 57: Do we want to keep the `REDIS_ENABLED` option in the manifest? That way we can turn it on and off with an environment variable since we are keeping the value in the creds repo (though I can't imagine us wanting to turn it off in production)

idavidmcdonald: So I wasn't sure but what helped me decide was

- this is only for the PaaS environments, ie preview, staging, production and those will be going away in just a couple of months time. If this was ECS and being kept around for a long time then I might choose otherwise.
- the chance of us needing to disable redis in any of those PaaS environments in the next few months is really slim so I erred towards deleting the code (it is easy to add back in if we want)
- also allows us to remove the values from the creds repo and tidy up a few lines of code extra there too

corlettb (approved):

PR #4888, opened 2023-11-08
fix copy letter template page

leohemsted:
we removed the ability to call "add template" for letters in fe7e9a32

however, this inadvertently broke the copy template page, which calls through to add template. change the guard to refer to the endpoint itself to allow copying.

Notably in the commit mentioned above we removed some complexity around validating QR code lengths. This can still be removed since the previous template you're copying will have already passed those checks when edited.

we also need to add the TemplateNameMixin back into the LetterTemplateForm since we need to copy across the name when creating a new copied template. We don't use this form when creating a new template, and when editing an existing template we pre-populate the form with the prev contents of the template, so the name simply gets passed through.

We'll need to be careful about these flows when we get to welsh content to ensure copying still works cleanly.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1503,7 +1503,7 @@ class EmailTemplateForm(BaseTemplateForm, TemplateNameMixin):
     subject = TextAreaField("Subject", validators=[DataRequired(message="Cannot be empty")])
 
 
-class LetterTemplateForm(BaseTemplateForm):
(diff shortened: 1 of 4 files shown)


samuelhwilliams (approved):

samuelhwilliams:
@quis We need to have a think about the copy letter journey and how we manage this for bilingual templates.

Leo also wondered if, and I've just checked and confirmed, that copying letter templates doesn't copy attachments at the moment.

PR #4890, opened 2023-11-09
Save original PDF for letter attachments

klssmith:
We had code to save the sanitised letter attachment and the original letter attachment (before we have changed it) but we were saving the sanitised version in both cases. This fixes the code to ensure we are saving the original file as well as the sanitised version.

Also removes a few unused test fixtures.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1147,6 +1147,7 @@ def _process_letter_attachment_form(service_id, template, form, upload_id):
         template_id=template.id,
         upload_id=upload_id,
         original_filename=original_filename,
+        original_file=pdf_file_bytes,
         sanitise_response=response,
     )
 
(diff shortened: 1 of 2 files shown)


app/main/views/templates.py
```diff
@@ -1176,7 +1177,7 @@ def _save_letter_attachment(*, service_id, template_id, upload_id, original_file
 
     # we've changed fonts and cmyk, so retain original in case we need to investigate errors
     backup_original_letter_to_s3(
-        file_contents,
+        original_file,
         upload_id=upload_id,
     )
```
samuelhwilliams, line 1182: [optional suggestion]: This should work, although I wonder if we should upload this backup before we try to sanitise the letter so that if sanitising fails we've got something we can investigate?

samuelhwilliams (approved):

app/main/views/templates.py
klssmith: If sanitising fails, we do [upload the original file](https://github.com/alphagov/notifications-admin/blob/f5cf26f724fe8121e473ad5334b0e862bfe7ea6e/app/main/views/templates.py#L1112) to a different bucket so I think that is covered

PR #4891, opened 2023-11-10
Edit bilingual letter templates

samuelhwilliams:
Add a new endpoint allowing letter templates that are set with
`letter_languages=welsh_then_english` to edit the Welsh subject and
content (body).

This page is not hooked up anywhere yet as Welsh pages of a letter aren't yet rendered on the 'view template' page, so we'll need to follow up with that later. However you can view the page by URL hacking, and it updates the letter in the DB appropriately.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1531,6 +1531,30 @@ class LetterTemplateForm(BaseTemplateForm, TemplateNameMixin):
     )
 
 
+class WelshLetterTemplateForm(BaseTemplateForm, TemplateNameMixin):
+    subject = TextAreaField("Main heading in Welsh", validators=[DataRequired(message="Cannot be empty")])
+    template_content = TextAreaField(
+        "Body in Welsh", validators=[DataRequired(message="Cannot be empty"), NoCommasInPlaceHolders()]
+    )
+
+    def __init__(self, *args, subject, content, letter_welsh_subject, letter_welsh_content, **kwargs):
+        # Populate subject and template_content form fields using the Welsh template values; we can discard the English
+        # data for this form.
+        super().__init__(*args, subject=letter_welsh_subject, content=letter_welsh_content, **kwargs)
+
+    @property
+    def new_template_data(self):
+        data = super().new_template_data
+
+        if "subject" in data:
+            data["letter_welsh_subject"] = data.pop("subject")
+
+        if "content" in data:
+            data["letter_welsh_content"] = data.pop("content")
+
+        return data
+
+
 class LetterTemplatePostageForm(StripWhitespaceForm):
     postage = GovukRadiosField(
         "Choose the postage for this letter template",
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1253,7 +1253,7 @@ def letter_template_change_language(template_id, service_id):
     if template.template_type != "letter" or not isinstance(template, TemplatedLetterImageTemplate):
         abort(404)
 
-    current_languages = template._template.get("letter_languages")
+    current_languages = template._template["letter_languages"]
     form = LetterTemplateLanguagesForm(data=dict(languages=current_languages))
     if form.validate_on_submit():
         languages = form.languages.data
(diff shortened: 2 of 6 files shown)


app/main/forms.py
```diff
@@ -1511,6 +1511,30 @@ class LetterTemplateForm(BaseTemplateForm, TemplateNameMixin):
...
+        "Body in Welsh", validators=[DataRequired(message="Cannot be empty"), NoCommasInPlaceHolders()]
+    )
+
+    def __init__(self, *args, subject, content, letter_welsh_subject, letter_welsh_content, **kwargs):
+        # Populate subject and template_content form fields using the Welsh template values; we can discard the English
+        # data for this form.
+        super().__init__(*args, subject=letter_welsh_subject, content=letter_welsh_content, **kwargs)
+
+    @property
+    def new_template_data(self):
```
CrystalPea, line 1526: I'm not fully understanding what is happening here - it's a property, but we don't seem to be using it anywhere, except within the property itself?

tests/app/main/views/test_templates.py
```diff
@@ -525,6 +525,23 @@ def test_editing_letter_template_should_have_hidden_name_field(
     assert name_input["type"] == "hidden"
 
 
+def test_editing_welsh_content_for_letter_template(
```
CrystalPea, line 528: From name of the test, I thought it would actually edit the content.

Could we rename it so it's clear we are testing page viewing? For example:

`test_GET_edit_service_template_letter_type_welsh_subject_and_content()`

tests/app/main/views/test_templates.py
```diff
@@ -2825,6 +2842,164 @@ def test_should_redirect_when_saving_a_template_email(
...
+        _expected_redirect=url_for(
+            "main.view_template",
+            service_id=SERVICE_ONE_ID,
+            template_id=fake_uuid,
+        ),
+    )
+    mock_update_service_template.assert_called_with(
+        service_id=SERVICE_ONE_ID,
+        template_id=fake_uuid,
+        name=name,
```
CrystalPea, line 2970: are we actually changing template name in that view then? 😮

CrystalPea (commented):
Good work 👏🏼 . I left a few comments, some are suggestions, and some are questions.

app/main/views/templates.py
```diff
@@ -76,12 +76,16 @@
...
-    "letter": LetterTemplateForm,
-    "broadcast": BroadcastTemplateForm,
-}
+
+def get_template_form(template_type: Literal["email", "sms", "letter", "broadcast"]):
+    if template_type == "email":
+        return EmailTemplateForm
+    elif template_type == "sms":
+        return SMSTemplateForm
+    elif template_type == "broadcast":
```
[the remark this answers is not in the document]
samuelhwilliams: Basically yes :) I think the rule at the moment is broadly "if you encounter broadcast stuff, and it's easy to delete, do it. If it's a chunk of work then leave it for later".

Here I think detangling that would be enough work that it would detract from the main focus of the PR.

tests/app/main/views/test_templates.py
samuelhwilliams: No, but it's present on the page as a hidden field because the "edit letter" template is the same as the "create letter" template. See this pr: https://github.com/alphagov/notifications-admin/pull/4888

I'm not undoing any of that .. yet

app/main/forms.py
```diff
@@ -1511,6 +1511,30 @@ class LetterTemplateForm(BaseTemplateForm, TemplateNameMixin):
     )
 
 
+class WelshLetterTemplateForm(BaseTemplateForm, TemplateNameMixin):
```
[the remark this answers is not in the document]
samuelhwilliams: See other thread, https://github.com/alphagov/notifications-admin/pull/4891#discussion_r1396081836

tests/app/main/views/test_templates.py
CrystalPea: @samuelhwilliams I made the proposed test name shorter - does that work better?

CrystalPea (approved):

PR #4892, opened 2023-11-13
Update for department name change

samuelhwilliams:
DIT was replaced by DBT:
https://www.gov.uk/government/organisations/department-for-international-trade

The insignia is unchanged.

---

image

image


idavidmcdonald:
What is the implication of this please?

For emails already sent with the old department name, will they be affected at all?
For services that are using the old branding, will their branding suddenly change for all new emails? If so, are we OK with changing their branding without consultation from them like that?

quis (approved):

samuelhwilliams:
This won't change any existing brandings - it really just updates the platform admin form that we use to create government logos.

idavidmcdonald:
Thanks, Sam :) All good then

PR #4901, opened 2023-11-15
Fix link between error summary and file upload

tombye:
https://trello.com/c/fCjM227k/568-links-to-form-fields-in-the-branding-image-upload-error-summary-dont-work

When we generate the links in the error summary, we set them to the id of the field with that error so clicking them jumps you to it.

Replacing the upload field with a button broke this link because the field linked to was now hidden. This copies the field id to the button so any error links will still work.


samuelhwilliams (approved):
Tested and works with/without JS 👍  nice fix, well spotted

PR #4902, opened 2023-11-15
Explain how text message ‘fragments’ affect pricing and the free allowance

karlchillmaid:
Users need to understand the difference between the number of text messages they send and the number of actual text message ‘fragments’ they use.

In the past we’ve always avoided using the term ‘fragments’ as its meaning may not be clear.

This pull request makes a few small changes to the `Text message pricing` page to introduce and explain fragments.

Updating our pricing information probably won’t solve the problem though. We also need to think about updating the UI content on the:
* Usage page
* Edit template page
* Template preview page

We might also want to think about whether our offline touchpoints (invoices, emails etc) are consistent with what we say on the Notify website.

Once we’ve discussed and agreed how we want to change content across Notify (and beyond), we can collate all those changes in this PR.

### DIFF
--- a/app/templates/views/guidance/pricing/text-message-pricing.html
+++ b/app/templates/views/guidance/pricing/text-message-pricing.html
@@ -25,7 +25,7 @@ Text message pricing
   <p class="govuk-body">When a service has used its annual allowance, it costs {{ sms_rate }} pence (plus VAT) for each text message you send.</p>
   <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="#long-text-messages">Long text messages</a> cost more, and will use more of your allowance.</p>
 
-  <p class="govuk-body">You may also pay more per message if you:</p>
+  <p class="govuk-body">You may also use more of your allowance, or pay more per message, if you:</p>
   <ul class="govuk-list govuk-list--bullet">
     <li>use certain <a class="govuk-link govuk-link--no-visited-state" href="#symbols">signs and symbols</a></li>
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -23,10 +23,10 @@ <h1 class="heading-large">Text message pricing</h1>
 
   <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
   <p class="govuk-body">When a service has used its annual allowance, it costs {{ sms_rate }} pence (plus VAT) for each text message you send.</p>
+  <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="#long-text-messages">Long text messages</a> cost more, and will use more of your allowance.</p>
 
-  <p class="govuk-body">You may use more free messages, or pay more for each message, if you:</p>
+  <p class="govuk-body">You may also pay more per message if you:</p>
```
karlchillmaid, line 28:
```suggestion
  <p class="govuk-body">You may also use more of your allowance, or pay more per message, if you:</p>
```

CrystalPea (approved):
Hey, current changes look good. 

Did we get anywhere with additional changes mentioned in pull request's description?

karlchillmaid:
@CrystalPea There’s a story in the backlog to work on this stuff – I’ve been speaking to Product and I hope we can pick it up soon.

quis:
Closing this because it’s been open for over a year. Can always re-open when we pick up that story

PR #4903, opened 2023-11-16
Change validation error for gov logo text

tombye:
https://trello.com/c/QzsiMA8c/569-fix-error-message-for-gov-identity-name

A recent accessibility audit raised an issue with the error 'Cannot be empty' being too general.

This updates it to be in line with new content created as part of the validation audit.

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2350,7 +2350,7 @@ class BrandingRequestForm(StripWhitespaceForm):
 class GovernmentIdentityLogoForm(StripWhitespaceForm):
     logo_text = GovukTextInputField(
         "Enter the text that will appear in your logo",
-        validators=[DataRequired("Enter the text that will appear in your logo")],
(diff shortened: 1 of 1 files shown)


app/main/forms.py
```diff
@@ -2350,7 +2350,7 @@ class BrandingRequestForm(StripWhitespaceForm):
 class GovernmentIdentityLogoForm(StripWhitespaceForm):
     logo_text = GovukTextInputField(
         "Enter the text that will appear in your logo",
-        validators=[DataRequired("Cannot be empty")],
+        validators=[DataRequired("Enter the text that will appear in your logo")],
```
quis, line 2353: Think the pattern now is to use the `NotifyDataRequired` class instead:
```suggestion
        validators=[NotifyDataRequired(thing="the text that will appear in your logo")],
```

karlchillmaid (approved):
Content looks good

quis (approved):

PR #4904, opened 2023-11-16
Bump Node to 16.16.0 LTS

samuelhwilliams:
This is the current 16 LTS version. It also, with https://github.com/alphagov/notifications-aws/pull/1791, brings both versions back in sync and adds comments to help keep them that way.

With them both in sync we should have faster and more reliable test builds.


docker/Dockerfile
```diff
@@ -72,7 +72,10 @@ CMD [ "web" ]
 ##### Test Image ##############################################################
 FROM production as test
 
-ARG NODE_VERSION=16.14.0
+# If bumping this node version, also bump:
+#   `.nvmrc` in notifications-admin
+#   `concourse/docker/unit-tests-image-node16.Dockerfile` in notifications-aws
+ARG NODE_VERSION=16.16.0
```
leohemsted, line 78: i think we should also add `ARG NVM_VERSION=0.39.1` to keep this in line with the aws repo (or `0.39.4` and then update the AWS repo, or indeed `0.39.5` and update both) just to make sure the version numbers are listed in the same place so they can be kept in lockstep

CrystalPea (approved):

docker/Dockerfile
samuelhwilliams: A fine shout good sir 👍  Will do

leohemsted (approved):

PR #4905, opened 2023-11-16
Let services set a custom email sender name

leohemsted:
A new setting in service settings

the pre-amble content is not final yet - cc @karlchillmaid - but main flow/form validation is all good I think

-------------
### Service settings index:
image

### Form page
| Happy Path | Unhappy Path |
|-|-|
| image  | image |


quis:
Two changes I’d suggest here:
- make ‘Email sender name’ the second item in the table, underneath ‘Send emails’
- move the _nhs@notifications…_ example to hint text for the ‘Same as service name’ option

leohemsted:
@quis first point is good - squashed in to 019b22b7 

second point i'm a lil uncertain, if you currently have a custom sender name set up, I think its good to show what that looks like (particularly to help explain to people that the local part is different but is but created based on the value you put in).

if i put it down to the hint, it might not be clear what your full sending email address currently is, if you have a custom name. Or what it will be set to, if you move to a custom name

app/templates/views/service-settings/custom-email-sender-name.html
```diff
@@ -0,0 +1,55 @@
...
+    </ul>
+
+    <p class="govuk-body">Do not use an acronym or initialism unless your users are already familiar with it.</p>
+
+    {% call form_wrapper() %}
+        {{ form.use_custom_email_sender_name(param_extensions={
+            "items": [
+                {},
+                {
+                    "hint": {"text": "Enter a sender name"},
```
samuelhwilliams, line 44: This hint should be on the conditional field so that it's only displayed when the input is visible (it doesn't make sense otherwise).

Currently we see:
image

But we should see:

samuelhwilliams (commented):
This looks good, only a couple of comments 👍  Needs content update/sign-off still of course.

app/templates/views/service-settings/custom-email-sender-name.html
```diff
@@ -0,0 +1,55 @@
+{% extends "withnav_template.html" %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/form.html" import form_wrapper %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/inset-text/macro.html" import govukInsetText %}
+
+{% set heading = 'Email address and sender name' %}
```
karlchillmaid, line 8:
```suggestion
{% set heading = 'Sender name and email address' %}
```

app/templates/views/service-settings/custom-email-sender-name.html
```diff
@@ -0,0 +1,55 @@
...
+{% endblock %}
+
+{% block backLink %}
+    {{ govukBackLink({ "href": url_for('main.service_settings', service_id=current_service.id) }) }}
+{% endblock %}
+
+{% block maincolumn_content %}
+    {{ page_header(heading) }}
+    <p class="govuk-body">This is the name your emails will come from.</p>
+
```
karlchillmaid, line 21: Remove this line.

```suggestion
```

app/main/forms.py
```diff
@@ -1917,6 +1917,50 @@ def __init__(self, channel, *args, **kwargs):
         super().__init__(name, *args, **kwargs)
 
 
+class ServiceEmailSenderForm(StripWhitespaceForm):
+    use_custom_email_sender_name = OnOffField(
+        "Where should emails come from for this service",
```
karlchillmaid, line 1922:
```suggestion
        "Choose your sender name and email address",
```

app/main/forms.py
```diff
@@ -1917,6 +1917,50 @@ def __init__(self, channel, *args, **kwargs):
...
+        choices_for_error_message="same or custom",
+        choices=[
+            (False, "Same as service name"),
+            (True, "A custom sender name"),
```
karlchillmaid, line 1926:
```suggestion
            (False, "Use your service name"),
            (True, "Something else"),
```

app/templates/views/service-settings/custom-email-sender-name.html
karlchillmaid:
```suggestion
                    "hint": {"text": "We’ll use this to create a new email address for your service"},
```

leohemsted:
image

Updated as per suggestions. keen to merge now to let us see it and poke it a bit

samuelhwilliams (approved):

PR #4906, opened 2023-11-17
Be consistent about how we display the email sender name

quis:


app/templates/views/service-settings/custom-email-sender-name.html
```diff
@@ -16,35 +15,52 @@
...
-                            param_extensions={
-                                "label": {"classes": "govuk-visually-hidden"},
-                                "hint": {"text": "Enter a sender name"}
-                            }
-                        )
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-five-sixths">
+            {{ page_header(heading) }}
+
+            <p class="govuk-body">Your sender name should tell the recipient what your message is about, as well as who it’s from.</p>
```
karlchillmaid, line 22:
```suggestion
            <p class="govuk-body">Your sender name should tell the recipient what your email is about, as well as who it’s from.</p>
```

app/main/forms.py
```diff
@@ -1918,6 +1918,13 @@ def __init__(self, channel, *args, **kwargs):
 
 
 class ServiceEmailSenderForm(StripWhitespaceForm):
+    BAD_EMAIL_LOCAL_PARTS = {
+        "noreply",
+        "no.reply",
+        "info",
+        "support",
```
samuelhwilliams, line 1925: `alert` could be one to add, too.

samuelhwilliams (approved):
The empty state for the custom input looks a bit odd to me, but if you're happy 👍 

image

app/main/forms.py
karlchillmaid: Would we want to pre-populate it with something like ‘Custom sender name’?

That would generate a full email address.

We’d still need something like this, though, in case the user leaves the field empty.

The other alternative is to have no preview if the field is empty.

quis: I’ve done this, think it works ok. I don’t like no preview because then everything jumps around when you start typing.

image

karlchillmaid: Yeah, I figured no preview would look a bit janky/jarring as soon as you enter something

PR #4907, opened 2023-11-20
Launch email sender name feature

quis:
This pull request removes the platform admin flag and makes it so that anyone can change their email sender name.

***

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/3956
- [x] https://github.com/alphagov/notifications-api/pull/3957

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -126,10 +126,8 @@ def service_name_change(service_id):
     )
 
 
-# TODO: change permissions to manage_service
-# @user_has_permissions("manage_service")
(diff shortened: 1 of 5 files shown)


leohemsted (approved):
just pending my last api PR

doesn't need to be this PR, but we might want to think about EmailPreviewTemplate - it still just says the service name when you're previewing an email you're about to send, it would be nice to reflect that too

quis:
@leohemsted  added the preview in [729c253](https://github.com/alphagov/notifications-admin/pull/4907/commits/729c253bc2853546798c12e643b63a8a0c47972e) since it wasn’t big change

image

PR #4908, opened 2023-11-21
Fix images in IE11 & mobile viewports

tombye:
Spotted a few visual issues with the radios-with-images component when doing other work. This adds some fixes for the IE11 issue and a proposed redsign for smaller screens.

## Images stretching to their full size in IE11

All screenshots are from IE11 but have also been tested in our other supported browsers.

### Current version

choose_logo_type_depart_or_custom_ie11_current

choose_custom_logo_type_ie11_current

### With fix

choose_logo_type_depart_or_custom_ie11_after

choose_custom_logo_type_ie11_after

### Before

radios_with_images_now

### After

choose_logo_type_depart_or_custom_mobile_after

## Background to the fixes / redesign

IE11 didn't respect the use of `max-width: 100%` and instead went by the widths set on each of the images. This changes to `width: 100%` to make sure they always fill the space in their container (which I think was always the intent of `max-width` anyway) but in a way that IE11 respects over the width on the element.

We also didn't do much work on this component on smaller screens and the current implementation makes the images small enough that some users might not be able to see the text in them properly. This stacks them vertically which gives
enough space for their text to be legible.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2389,7 +2389,7 @@ class EmailBrandingChooseBanner(OrderableFieldsForm):
             "image": {
                 "url": asset_fingerprinter.get_url("images/branding/org_banner.png"),
                 "alt_text": "An example of an email with a logo on a blue banner.",
-                "dimensions": {"width": 404, "height": 454},
+                "dimensions": {"width": 606, "height": 404},
             },
         },
         "org": {
@@ -2397,7 +2397,7 @@ class EmailBrandingChooseBanner(OrderableFieldsForm):
             "image": {
                 "url": asset_fingerprinter.get_url("images/branding/org.png"),
                 "alt_text": "An example of an email with a logo on a clear background.",
-                "dimensions": {"width": 404, "height": 454},
+                "dimensions": {"width": 606, "height": 404},
             },
         },
     }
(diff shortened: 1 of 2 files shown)


app/main/forms.py
```diff
@@ -2389,15 +2389,15 @@ class EmailBrandingChooseBanner(OrderableFieldsForm):
...
+                "dimensions": {"width": 606, "height": 404},
             },
         },
         "org": {
             "label": "No",
             "image": {
                 "url": asset_fingerprinter.get_url("images/branding/org.png"),
                 "alt_text": "An example of an email with a logo on a clear background.",
-                "dimensions": {"width": 404, "height": 454},
+                "dimensions": {"width": 606, "height": 404},
```
samuelhwilliams, line 2400: Good spot on the png sizes being wrong 👍 🤔

samuelhwilliams: It feels fairly safe to think we won't change the images again. I wonder if a test or something pinning the images to these dimensions would help in the unlikely case we do change them though?

samuelhwilliams (approved):
Tested in IE and Firefox, looks good 👍

app/main/forms.py
tombye: I see what you're saying but, in our case, the sizes being wrong didn't have any real consequences because the browser followed the CSS and scaled them to fit the space rather than respecting their dimensions. It could be a problem if the images were smaller than the largest space they have to fill (half the main content width - 20px) because they would appear pixelated... Do you think this is enough of a risk to test for?

I might just be picky but curious to know.

samuelhwilliams: I don't think we super need to :)

PR #4912, opened 2023-11-22
Add a page where users can request to join a service 

quis:
Reopens #4659

***

Depends on:
- https://github.com/alphagov/notifications-api/pull/3977

***

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2832,18 +2832,24 @@ class JoinServiceForm(StripWhitespaceForm):
         super().__init__(*args, **kwargs)
 
         self.users.choices = [(user.id, user.name) for user in users]
-        self.users.param_extensions = {
-            "items": [
-                {
-                    "hint": {
-                        "text": f"Last used Notify {format_date_human(user.logged_in_at)}",
-                    }
-                }
-                for user in users
-            ]
-        }
+        self.users.param_extensions["items"] = [
+            {"hint": {"text": f"Last used Notify {format_date_human(user.logged_in_at)}"}} for user in users
+        ]
 
     users = GovukCheckboxesField(
         "Who do you want to ask?",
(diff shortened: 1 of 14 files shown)


BlessedDev (approved):
Nice One

app/main/forms.py
```diff
@@ -2824,3 +2825,31 @@ def validate(self, *args, **kwargs):
...
+                    # This removes the `govuk-fieldset__legend--s` class, thereby
+                    # making the form label font regular weight, not bold
+                    "classes": "",
+                },
+            }
+        },
+    )
+    reason = GovukTextareaField(
+        "Explain why you need access",
+        param_extensions={"hint": {"text": "Optional"}},
```
samuelhwilliams, line 2854: It's not much of a pattern, but on another form we do this inline, ie `Explain why you need access (optional)`. Might be nice to keep that pattern (or see what Karl wants).

samuelhwilliams (approved):

app/main/views/join_service.py
```diff
@@ -0,0 +1,51 @@
...
+@user_is_gov_user
+def join_service(service_to_join_id):
+    service = Service.from_id(service_to_join_id)
+
+    if not service.organisation.can_ask_to_join_a_service:
+        abort(403)
+
+    form = JoinServiceForm(
+        users=service.active_users_with_permission("manage_service"),
+    )
```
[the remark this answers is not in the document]
quis: Think it’s worth guarding against, if only because it exposes the list of team members – changed in 52b6f79

app/main/forms.py
quis: Changed in [1419d59](https://github.com/alphagov/notifications-admin/pull/4912/commits/1419d591442a00393721d8480d9031dfae09c3f5)

PR #4914, opened 2023-11-23
Add sign in method guidance

karlchillmaid:


app/main/views/sub_navigation_dictionaries.py
```diff
@@ -94,6 +94,10 @@ def using_notify_nav():
                     "name": "Bulk sending",
                     "link": "main.guidance_bulk_sending",
                 },
+                {
+                    "name": "Data retention period",
+                    "link": "main.guidance_data_retention_period",
+                },
```
karlchillmaid, line 100:
```suggestion
                },
```

app/navigation.py
```diff
@@ -61,19 +61,21 @@ class HeaderNavigation(Navigation):
...
             "guidance_attach_pages",
             "guidance_bulk_sending",
-            "guidance_message_status",
+            "guidance_data_retention_period",
             "guidance_delivery_times",
             "guidance_email_branding",
             "guidance_formatting",
             "guidance_letter_branding",
             "guidance_links_and_URLs",
+            "guidance_message_status",
```
karlchillmaid, line 70:
```suggestion
            "guidance_message_status",
```

tests/route-list.json
```diff
@@ -401,6 +401,7 @@
     "/using-notify/guidance",
     "/using-notify/guidance/branding-and-customisation",
     "/using-notify/guidance/bulk-sending",
+    "/using-notify/guidance/data_retention_period",
```
karlchillmaid, line 404:
```suggestion
    "/using-notify/guidance/data-retention-period",
```

tests/route-list.json
```diff
@@ -415,6 +416,7 @@
     "/using-notify/guidance/schedule-emails-and-text-messages",
     "/using-notify/guidance/schedule-messages",
     "/using-notify/guidance/send-files-by-email",
+    "/using-notify/guidance/sign_in_method",
```
karlchillmaid, line 419:
```suggestion
    "/using-notify/guidance/sign-in-method",
```

tests/route-list.json
```diff
@@ -401,6 +401,7 @@
     "/using-notify/guidance",
     "/using-notify/guidance/branding-and-customisation",
     "/using-notify/guidance/bulk-sending",
+    "/using-notify/guidance/data-retention-period",
```
samuelhwilliams, line 404: These don't match the defined URLs. See:

```
@main.route("/using-notify/data-retention-period")
def guidance_data_retention_period():
    return render_template(
        "views/guidance/using-notify/data-retention-period.html",
        navigation_links=using_notify_nav(),
    )
```

The route is on `/using-notify/data-retention-period`, but this line has `/guidance` in the middle.

I'm not sure which is correct, but they need to match :)

karlchillmaid: @samuelhwilliams They shouldn’t include `/guidance`. That was the old URL structure. I assume there are still a few of these in there because they weren’t removed when the URLs were updated and redirects put in place. 

For example, line 422 is the old version:

```"/using-notify/guidance/upload-a-letter",```

While line 439 is the newer (correct) version:

```"/using-notify/upload-a-letter",```

If that’s right, can we remove all the old `/guidance` ones?

Comment (samuelhwilliams) [reply tests/route-list.json]: I think let's not worry about removing things in this PR - we can clean up separately later if needed.

Just need to update these new URLs to not have the `guidance` bit 👍

Comment (karlchillmaid) [reply tests/route-list.json]: Thanks @samuelhwilliams – I think I fixed those new URLs already.

File: app/templates/views/guidance/using-notify/data-retention-period.html [hunk 6]
```diff
@@ -0,0 +1,32 @@
...
+  </ul>
+
+  <p class="govuk-body">By default, this data is only held for 7 days.</p>
+
+  <p class="govuk-body">Once your service is live, you can choose the number of days you want Notify to keep details of the messages you send.</p>
+
+  <p class="govuk-body">To change your data retention period:</p>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
```
samuelhwilliams, line 28: In order to use this, we need to `{% macro service_link(service, endpoint, link_text) -%}` at the top of the file, below extends

samuelhwilliams: Same for the other page as well :)

app/templates/views/guidance/using-notify/data-retention-period.html
```diff
@@ -0,0 +1,32 @@
+{% extends "content_template.html" %}
+
```
karlchillmaid, line 2:
```suggestion
{% extends "content_template.html" %}
{% macro service_link(service, endpoint, link_text) -%}
```

app/templates/views/guidance/using-notify/sign-in-method.html
```diff
@@ -0,0 +1,40 @@
+{% extends "content_template.html" %}
+
```
karlchillmaid, line 2:
```suggestion
{% extends "content_template.html" %}
{% macro service_link(service, endpoint, link_text) -%}

```

app/templates/views/guidance/using-notify/data-retention-period.html
karlchillmaid: @samuelhwilliams Done

app/templates/views/guidance/using-notify/sign-in-method.html
```diff
@@ -0,0 +1,41 @@
...
+  <p class="govuk-body">You must keep to the [terms of use for signing in to Notify].</p>
+
+  <h2 class="heading-medium">How to change the sign-in method for your team</h2>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
+    <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Sign-in method</b> row.</li>
+  </ol>
+
+{% endblock %}
```
karlchillmaid, line 41:
```suggestion

{% endmacro %}
{% endblock %}
```

app/templates/views/guidance/using-notify/data-retention-period.html
```diff
@@ -0,0 +1,33 @@
...
+  <p class="govuk-body">Once your service is live, you can choose the number of days you want Notify to keep details of the messages you send.</p>
+
+  <p class="govuk-body">To change your data retention period:</p>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
+    <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Data retention period</b> row.</li>
+  </ol>
+
+{% endblock %}
```
karlchillmaid, line 33:
```suggestion

{% endmacro %}
{% endblock %}
```

app/templates/views/guidance/using-notify/sign-in-method.html
```diff
@@ -0,0 +1,42 @@
...
+
+  <h2 class="heading-medium">How to change the sign-in method for your team</h2>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
+    <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Sign-in method</b> row.</li>
+  </ol>
+
+{% endmacro %}
+{% endblock %}
```
karlchillmaid, line 42:
```suggestion
{% endblock %}
{% endmacro %}
```

app/templates/views/guidance/using-notify/data-retention-period.html
```diff
@@ -0,0 +1,34 @@
...
+
+  <p class="govuk-body">To change your data retention period:</p>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
+    <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Data retention period</b> row.</li>
+  </ol>
+
+{% endmacro %}
+{% endblock %}
```
karlchillmaid, line 34:
```suggestion
{% endblock %}
{% endmacro %}
```

samuelhwilliams:
image

image

app/templates/views/guidance/using-notify/sign-in-method.html
```diff
@@ -0,0 +1,41 @@
...
+  <p class="govuk-body">If signing in with a text message is a problem for your team, you can also sign in using a link that’s sent in an email.</p>
+
+  <p class="govuk-body">For security, you should only do this if you use:</p>
+
+  <ul class="govuk-list govuk-list--bullet">
+    <li>a work-managed device</li>
+    <li>a work email address protected by multi-factor authentication (MFA)</li>
+  </ul>
+
+  <p class="govuk-body">You must keep to the [terms of use for signing in to Notify].</p>
```
samuelhwilliams, line 32: `terms of use for signing in to Notify].`

This looks like a placeholder?

karlchillmaid: Good catch. Updated terms aren’t ready yet – so I’ll remove this line for now.

app/templates/views/guidance/using-notify/data-retention-period.html
```diff
@@ -0,0 +1,33 @@
...
+
+  <p class="govuk-body">By default, this data is only held for 7 days.</p>
+
+  <p class="govuk-body">Once your service is live, you can choose the number of days you want Notify to keep details of the messages you send.</p>
+
+  <p class="govuk-body">To change your data retention period:</p>
+
+  <ol class="govuk-list govuk-list--number">
+    <li>Go to the {{ service_link(current_service, 'main.service_settings', 'Settings') }} page.</li>
+    <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">Data retention period</b> row.</li>
```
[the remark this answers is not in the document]
karlchillmaid: It’s a very good point.

We write the guidance as numbered steps for anyone who’s not signed in.

Once the user signs in we make the page they need to go a link. I think we stuck to only making the page a link to avoid filling the steps with links to each and every clickable thing.

This raises an interesting point – that instead of numbered steps we could just provide a link to the action for signed in users.

I don’t want to explore that in this PR, because it affects a lot of other pages. I’d like time to weigh up the benefits of sticking with a numbered list for signed-in users, because there’s a chance it teaches users the structure of Notify, so they can repeat actions and explore with confidence, without always needing to turn to guidance or support. One for separate discussion, though.

samuelhwilliams (approved):

PR #4917, opened 2023-11-24
Add 'Edit' buttons for each language on bilingual letters

samuelhwilliams:
This is a big one, best done commit-by-commit.

First of all, we lift and shift the LetterImagePreview class from utils. This is only used in admin, and we want direct access to the template so that we can overlay UI elements in arbitrary places (on the first page of the English language template). (Related utils PR: https://github.com/alphagov/notifications-utils/pull/1076)

Then we do a refactor to make sure that the template can access all of the page counts from the template-preview API call, rather than just the total page count for the template.

Then we move an attribute from the base letter template to the image preview template only, since it's not applicable for precompiled letters.

Finally, we move the 'edit template' UI elements from the main template to the new `TemplatedLetterImageTemplate` template. This is mixing responsibility slightly, but it's the cleanest way I could think of applying the design style that we're after.

## On the page
image


app/template_previews.py
```diff
@@ -111,7 +111,7 @@ def get_page_count_for_letter(db_template, values=None):
         headers={"Authorization": f"Token {current_app.config['TEMPLATE_PREVIEW_API_KEY']}"},
     )
 
-    page_count = json.loads(response.content.decode("utf-8"))["count"]
+    page_count = json.loads(response.content.decode("utf-8"))
```
CrystalPea, line 114: [suggestion] should we maybe rename that variable, and function name, to reflect that it returns `page_counts_dict` now?

app/utils/templates.py
```diff
@@ -131,28 +135,64 @@ def __init__(self, *args, **kwargs):
...
-        self._page_count = get_page_count_for_letter(self._template)
+        self._all_page_counts = get_page_count_for_letter(self._template, self.values)
+        redis_client.set(cache_key, json.dumps(self._all_page_counts), ex=cache.DEFAULT_TTL)
 
-        redis_client.set(cache_key, self._page_count, ex=cache.DEFAULT_TTL)
+        return self._all_page_counts
 
-        return self._page_count
+    @property
+    def page_count(self):
```
CrystalPea, line 166: [proposition] what do you think about naming it `total_page_count`? Then it implies it's a page count for all the things 😁 

I am conscious that we'd have to change the name in quite a few places though.

tests/app/utils/test_templates.py
```diff
@@ -598,3 +609,79 @@ class LetterImageTemplateWithAttachment(TemplatedLetterImageTemplate):
         '<img alt="" loading="lazy" src="http://example.com/endpoint.png?page=7"/>',
         '<img alt="" loading="lazy" src="http://example.com/endpoint.png?page=8"/>',
     ]
+
+
+class TestTemplatedLetterImageTemplate:
```
CrystalPea, line 615: Nice one creating a class for those tests - I would love more of our tests to be split into classes.

tests/app/utils/test_templates.py
```diff
@@ -598,3 +609,79 @@ class LetterImageTemplateWithAttachment(TemplatedLetterImageTemplate):
...
+    def test_first_english_page(self, mocker, fake_uuid, mocker_kwargs, expected_value):
+        do_mock_get_page_count_for_letter(mocker, **mocker_kwargs)
+        t = TemplatedLetterImageTemplate(template_json(service_id=SERVICE_ONE_ID, id_=fake_uuid, type_="letter"))
+
+        assert t.first_english_page == expected_value
+
+    @pytest.mark.parametrize(
+        "mocker_kwargs, expected_value",
+        (
+            (dict(count=1), None),
```
CrystalPea, line 663: Is there an advantage to notating dictionaries like this instead of standard `{"count": 1}` ?

I'm asking because if there isn't, isn't it best to maintain one style for the ease of reading?

tests/conftest.py
```diff
@@ -3745,16 +3745,33 @@ def mock_get_returned_letter_summary_with_no_returned_letters(mocker):
     )
 
 
+def do_mock_get_page_count_for_letter(mocker, count, welsh_page_count=0, attachment_page_count=0):
```
CrystalPea, line 3748: is "do_mock" some convention I am not aware of?

If it's not, I think it would work better to call this one `mock_get_page_count_for_letter` and the function below it `mock_get_page_count_for_letter_fixture` - it would be more descriptive that way.

app/utils/templates.py
```diff
@@ -209,6 +202,10 @@ def attachment(self):
         if attachment := self.get_raw("letter_attachment"):
             return LetterAttachment(attachment)
 
+    @property
+    def jinja_render_data(self):
+        return super().jinja_render_data | {"first_page_of_attachment": self.first_attachment_page}
```
CrystalPea: [comment] Ooh, I didn't know that pipe can be a union operator :o

app/templates/templates/letter_image_template.jinja2
```diff
@@ -1,10 +1,38 @@
+{% if template.include_letter_edit_ui_overlay %}
+<div class="template-container template-container--with-attach-pages-button">
+  {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) and not template.too_many_pages %}
+    <a href="{{ url_for(".set_sender", service_id=current_service.id, template_id=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-get-ready-to-send">
+      Get ready to send<span class="govuk-visually-hidden"> a letter using this template</span>
+    </a>
+  {% endif %}
+  {% if current_user.has_permissions('manage_templates') %}
```
CrystalPea, line 8: I am aware you just copied this bit over, but I wonder - shouldn't the permission for changing branding be "manage_service" permission?

app/templates/templates/letter_image_template.jinja2
```diff
@@ -1,10 +1,38 @@
...
+      <a href="{{ url_for(".edit_service_template", service_id=current_service.id, template_id=template.id, language='welsh') }}" class="govuk-button govuk-button--secondary edit-template-link-letter-body">Edit<span class="govuk-visually-hidden"> Welsh pages of the letter template</span></a>
+    {% else %}
+      <a href="{{ url_for(".edit_service_template", service_id=current_service.id, template_id=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-body">Edit<span class="govuk-visually-hidden"> letter template</span></a>
+    {% endif %}
+    {% if current_service.count_letter_contact_details %}
+      <a href="{{ url_for(".set_template_sender", service_id=current_service.id, template_id=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-contact">Edit<span class="govuk-visually-hidden"> letter contact block</span></a>
+    {% else %}
+      <a href="{{ url_for(".service_add_letter_contact", service_id=current_service.id, from_template=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-contact">Edit<span class="govuk-visually-hidden"> letter contact block</span></a>
```
CrystalPea, line 21: Interesting how we have two different links here, but with the same content. 

I wonder if it's something to highlight to Karl? As in, wouldn't we want Edit/Add depending on circumstances?

app/templates/templates/letter_image_template.jinja2
```diff
@@ -1,10 +1,38 @@
...
+{% endif %}
+
 {% for page_number in page_numbers %}
   <div class="letter page--{{ loop.cycle('odd', 'even') }}{% if loop.first %} page--first{% endif %}{% if loop.last %} page--last{% endif %}">
     {% if loop.first and show_postage %}
       <p class="letter-postage {{ postage_class_value }}">
         Postage: {{ postage_description }}
       </p>
     {% endif %}
+    {% if template.include_letter_edit_ui_overlay and template.welsh_page_count and page_number == first_page_of_english %}
```
CrystalPea, line 33: Nice one 👏🏼 . Wow it required so much groundwork to get here 😳

app/templates/views/templates/_letter_template.html
```diff
@@ -8,26 +8,7 @@
...
-    <a href="{{ url_for(".edit_service_template", service_id=current_service.id, template_id=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-body">Edit<span class="govuk-visually-hidden"> letter template</span></a>
-    {% if current_service.count_letter_contact_details %}
-      <a href="{{ url_for(".set_template_sender", service_id=current_service.id, template_id=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-contact">Edit<span class="govuk-visually-hidden"> letter contact block</span></a>
-    {% else %}
-      <a href="{{ url_for(".service_add_letter_contact", service_id=current_service.id, from_template=template.id) }}" class="govuk-button govuk-button--secondary edit-template-link-letter-contact">Edit<span class="govuk-visually-hidden"> letter contact block</span></a>
-    {% endif %}
-  {% endif %}
-  {{ template|string }}
-</div>
+{{ template|string }}
```
CrystalPea, line 18: [question] does the pipe here convert template object to string?

CrystalPea (commented):
Nice one 👏🏼 .

I left several comments, as it's a lot of changes, but nothing major ☺️

tests/app/utils/test_templates.py
```diff
@@ -102,3 +113,488 @@ def test_get_page_count_for_letter_does_not_cache_for_personalised_letters(
...
+        "html.parser",
+    )
+    assert [(img["src"], img["loading"]) for img in page.select("img")] == [
+        ("http://example.com/endpoint.png?page=1", "eager"),
+        ("http://example.com/endpoint.png?page=2", "lazy"),
+        ("http://example.com/endpoint.png?page=3", "lazy"),
+    ]
+
+
+def test_letter_image_template_marks_first_page_of_attachment():
```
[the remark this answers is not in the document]
samuelhwilliams: thanks for checking!

app/templates/templates/letter_image_template.jinja2
samuelhwilliams: yep 😅

app/templates/templates/letter_image_template.jinja2
samuelhwilliams: Yes, it may well be. I think given this is pre-existing we can keep it to a separate PR though :)

@karlchillmaid fyi

app/templates/templates/letter_image_template.jinja2
samuelhwilliams: Fixed in 40d6752bd

tests/conftest.py
samuelhwilliams: Often we use the `mock_` prefix in fixtures (things that can be used in tests via dependency injection).

This is not a fixture, so I wanted to differentiate it. It does the mocking on-demand, as a normal function.

It feels a bit simpler than the [indirect parameterization](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments) that supports this "natively".

tests/app/utils/test_templates.py
samuelhwilliams: me too :)

app/utils/templates.py
samuelhwilliams: I kept it as `page_count` because this overrides the `page_count` attribute in the base class.

I think we can revisit the `total_page_count` as/when we look at template-preview's return values and rationalise that bit :)

app/templates/views/templates/_letter_template.html
CrystalPea: Thanks! The pipe has so many uses 😆

app/templates/templates/letter_image_template.jinja2
```diff
@@ -0,0 +1,40 @@
+{% for page_number in page_numbers %}
```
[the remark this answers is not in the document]
samuelhwilliams: done :)

tests/app/utils/test_templates.py
CrystalPea: We have a firebreak project right there ;P

app/template_previews.py
```diff
@@ -95,7 +95,11 @@ def get_png_for_letter_attachment_page(cls, attachment_id, page=None):
         return response.content, response.status_code, cls.get_allowed_headers(response.headers)
 
 
-def get_page_count_for_letter(db_template, values=None):
+def get_page_counts_for_letter(db_template, values=None):
+    """
+    Expected return value format (mimics the template-preview endpoint:
```
CrystalPea, line 100: :chefs-kiss: great idea with the docstring ^_^

CrystalPea (approved):
:shipit:

PR #4920, opened 2023-11-27
Prevent sender name from being an email address

quis:
People are trying to enter an email address here, but the sender email address is something we generate. It looks weird and phishy to have the email look like it’s coming from one address but actually comes from another. So let’s stop that from happening.

***

image

### DIFF
--- a/tests/app/main/views/service_settings/test_service_settings.py
+++ b/tests/app/main/views/service_settings/test_service_settings.py
@@ -6569,6 +6569,8 @@ class TestServiceEmailSenderChange:
             ("info", "Sender name needs to be more specific"),
             ("Support", "Sender name needs to be more specific"),
             ("ALERT", "Sender name needs to be more specific"),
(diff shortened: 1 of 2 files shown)


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -6569,6 +6569,8 @@ def test_service_email_sender_change_page_shows_your_current_email_sender_name(
             ("info", "Sender name needs to be more specific"),
             ("Support", "Sender name needs to be more specific"),
             ("ALERT", "Sender name needs to be more specific"),
+            ("test@example.com", "Sender name cannot be an email address"),
+            ("Foo.BAR@example.gov.uk", "Sender name cannot be an email address"),
```
leohemsted, line 6573: optional: do we want to add that `SETI @ Home` example as a name that succeeds?

PR #4922, opened 2023-11-28
Remove Emergency Alerts

quis:
Emergency Alerts has been forked [into it’s own codebase](https://github.com/alphagov/emergency-alerts-admin) so we don’t need to maintain it here any more.

This pull request completely removes Emergency Alerts from this repo. While the API will still return emergency-alert-related data the admin app will now ignore it.

I’ve tried to split this into logical commits. The tests should pass after each and every commit.

***

Removing all this code cuts 20% off the run time of the integration tests.

***

Existing services with the `broadcast` permission will look like any other Notify service:

image

Message templates for emergency alerts will be hidden:

image

Emergency-alerts-related settings won’t be shown:

image


leohemsted (approved):
there's a line `!app/broadcast_areas/source_files/**` we can remove out of the gitignore if we want

PR #4929, opened 2023-12-01
Give iframes titles

tombye:
From this card:

https://trello.com/c/3vTS0QFY/588-give-all-iframes-titles

## Rational

Iframes need explaining (why there’s another document in the page) and those without an accessible name, though their title attribute, will be given one by screen readers which will be inferred from the rest of the page so could be confusing.

More detail: [Be Sure to Provide Titles for Iframes | Accessibility Tips](https://dequeuniversity.com/tips/provide-iframe-titles).

I think the content I have set the iframe titles to, and that of the documents they contain, is right but will need 👀 from @karlchillmaid, who I discussed this with.


tombye:
@karlchillmaid I've made some changes to let us set the iframe title and the title of the page in it on a page-by-page basis. I found a few variants we didn't discuss too so I think we'll need to catch up so I can talk you through the changes.

tombye:
Have now discussed the batch of changes since @karlchillmaid's comment with him directly. I since made a change to the iframe for the example of a government identity logo, based on his advice. This was the only change requested to the content.

There have now been quite a few changes since @samuelhwilliams review so the code changes will now need another look.

BlessedDev (approved):

PR #4931, opened 2023-12-05
Update default Welsh content for letters

samuelhwilliams:
And also toggle the English letter content to be language-specific for Welsh templates. Remove the language specification when removing Welsh.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -727,6 +727,9 @@ def edit_service_template(service_id, template_id, language=None):
         form=form,
         template=template,
         heading_action="Edit",
+        language=language,
+        letter_languages=template.get_raw("letter_languages"),
+        language_options=LetterLanguageOptions,
         back_link=url_for("main.view_template", service_id=current_service.id, template_id=template.id),
     )
 
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1248,6 +1248,27 @@ def _get_page_numbers(page_count):
     return [*range(1, page_count + 1)]
 
 
+def _change_template_language(service_id, template, language: LetterLanguageOptions):
+    update_kwargs: dict[str, Optional[str]] = {}
+
+    if language == LetterLanguageOptions.english:
+        if template.subject == "English heading":
+            update_kwargs["subject"] = "Main heading"
+        if template.content == "English body text":
+            update_kwargs["content"] = "Body"
+        update_kwargs["letter_welsh_subject"] = None
+        update_kwargs["letter_welsh_content"] = None
+    else:
+        if template.subject == "Main heading":
+            update_kwargs["subject"] = "English heading"
+        if template.content == "Body":
+            update_kwargs["content"] = "English body text"
+        update_kwargs["letter_welsh_subject"] = "Welsh heading"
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -1012,8 +1011,67 @@ def test_POST_letter_template_change_to_welsh_and_english_sets_subject_and_conte
...
+    client_request,
+    service_one,
+    mocker,
+    fake_uuid,
+    active_user_with_permissions,
+    subject,
+    content,
+    extra_kwargs,
+):
+    def _get(service_id, template_id, version=None, postage="second"):
```
CrystalPea, line 1038: How come we use here the syntax that we usually keep in the fixtures in conftest?

Also if we use it in this broader context, would we maybe want to rename `_get` to `_get_template` so it says what it's getting?

app/main/views/templates.py
```diff
@@ -1248,6 +1251,27 @@ def _get_page_numbers(page_count):
...
 
 
+def _change_template_language(service_id, template, language: LetterLanguageOptions):
+    update_kwargs: dict[str, Optional[str]] = {}
+
+    if language == LetterLanguageOptions.english:
+        if template.subject == "English heading":
+            update_kwargs["subject"] = "Main heading"
+        if template.content == "English body text":
+            update_kwargs["content"] = "Body"
```
karlchillmaid, line 1261: I think this should be `Body text` now

app/main/views/templates.py
```diff
@@ -1248,6 +1251,27 @@ def _get_page_numbers(page_count):
...
+        update_kwargs["letter_welsh_subject"] = None
+        update_kwargs["letter_welsh_content"] = None
+    else:
+        if template.subject == "Main heading":
```
karlchillmaid, line 1265:
```suggestion
        if template.subject == "Heading":
```

app/main/views/templates.py
```diff
@@ -1248,6 +1251,27 @@ def _get_page_numbers(page_count):
...
+            update_kwargs["subject"] = "English heading"
+        if template.content == "Body":
```
karlchillmaid, line 1267:
```suggestion
        if template.content == "Body text":
```

app/main/views/templates.py
```diff
@@ -1248,6 +1251,27 @@ def _get_page_numbers(page_count):
     return [*range(1, page_count + 1)]
 
 
+def _change_template_language(service_id, template, language: LetterLanguageOptions):
+    update_kwargs: dict[str, Optional[str]] = {}
+
+    if language == LetterLanguageOptions.english:
+        if template.subject == "English heading":
+            update_kwargs["subject"] = "Main heading"
```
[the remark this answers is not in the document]
samuelhwilliams: Tricky unfortunately. This will require a change in the API to do going forward, but any letters that already exist with default content will keep 'Main heading' 🤔 

I think it's OK to just ignore historical letters in their initial state, so I can update the API to set it to 'Heading' by default. PR coming...

samuelhwilliams:
All updated @karlchillmaid - the new base content will only affect new letter templates going forward though :)

samuelhwilliams:
I believe the content is correct now. If I'm wrong we can fix up in a follow-up PR :)

PR #4933, opened 2023-12-07
Adding validators for numeric and generic sender IDs

whpearson:
Numeric sender IDs should be valid uk mobile phone numbers or shortcodes.

Generic sender IDs (like Info, Verify and Alert) are being used by spam, so should be avoided.

Further work would be needed if we were going to have a long list of generic wordsm which might be coming soon. But I'm not sure it is worth it now.

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -165,6 +165,39 @@ class DoesNotStartWithDoubleZero:
             raise ValidationError(self.message)
 
 
+class IsNotAGenericSenderID:
+    generic_sender_ids = ["info", "verify", "alert"]
+
+    def __init__(
+        self,
+        message="Text message sender ID cannot be Alert, Info or Verify as those are prohibited due to "
+        "usage by spam",
+    ):
+        self.message = message
+
+    def __call__(self, form, field):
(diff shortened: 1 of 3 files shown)


app/main/validators.py
```diff
@@ -165,6 +165,39 @@ def __call__(self, form, field):
...
+    def __init__(
+        self,
+        message="Text message sender ID cannot be Alert, Info or Verify as those are prohibited due to "
+        "usage by spam",
+    ):
+        self.message = message
+
+    def __call__(self, form, field):
+        if field.data and re.match(self.regex, field.data):
+            raise ValidationError(self.message)
```
CrystalPea, line 180: [suggestion]

if we used a list, for example:
```
generic_sender_ids = ["info", "verify", "alert"]
```

It would be easy to add to that list, or even abstract it into a file as it gets larger.

Then in the `__call__` we could do the if statement like:
```
if field.date.lower() in self.generic_sender_ids:
```

CrystalPea (commented):
Nice work - I added a suggestion for making the generic sender names validator more extendable. Let me know what you think.

PR #4934, opened 2023-12-08
Handle more verbose exception message in Python 3.11

quis:
This internal error message was changed in python/cpython@0cb765b

So that our tests keep passing we need to check for both the old or the new error message.

This matches the change made in https://github.com/alphagov/notifications-utils/pull/1078/commits/7c323fe0f17870c1c6b57ae07d4b489b5411486d (this still needs testing in both places because the behaviour of the 2 objects is subtly different).

### DIFF
--- a/tests/app/models/test_base_model.py
+++ b/tests/app/models/test_base_model.py
@@ -1,3 +1,5 @@
+import sys
+
 import pytest
 
 from app.models import JSONModel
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Is this test just failing locally for you because you're on 3.11?

I can see why we'd change utils, because that could be pulled into various apps that may be on different python versions. But admin is just on 3.9 at the moment. Think it makes sense to do this change when we migrate to 3.11 rather than before?

quis:
I just figured I’d get ahead of it while I was doing the work on utils and save someone else digging back into it when we do decide to upgrade

leohemsted (approved):

PR #4935, opened 2023-12-08
Redirect to English preview after editing English content

quis:
If you edit some content or add an attachment to a letter it’s good if you can see the change you have made immediately, without having to scroll.

We do this for attachments, this makes it work the same way for bilingual letters. If you edit the English content in a bilingual letter we will jump you down the page to the preview of the English content when redirecting.


samuelhwilliams (approved):
This works - nice usability tweak. It's a bit weird how the behaviour/placement between `first-page-of-attachment` and `first-page-of-english-in-bilingual-letter` is different though.

samuelhwilliams:
Nice 🎉

PR #4938, opened 2023-12-12
Allow multiple distinct services to be mocked

samuelhwilliams:
`mock_get_service` uses a naive implementation which returns a fixed set of service JSON, with a service_id depending on what is passed into the `get_service` call.

This makes it difficult to test code that touches multiple services with different attributes.

This updates `mock_get_service` to use a dictionary underneath that can store custom service JSON blobs. If a request is made for a service that isn't in the explicit dict, we return a default service still.

PR #4939, opened 2023-12-12
New 'copy template' journey

samuelhwilliams:
- [x] **Sits on top of #4938 and https://github.com/alphagov/notifications-api/pull/3975 - to be reviewed/merged first**

---

The current 'copy template' journey is a bit more of a 'pre-filled
create new template' journey. The user is shown a list of all templates
they can see, and when they select one, we take them to the 'add new
template' page with the relevant form fields pre-filled.

This means that all data from the source template needs to be shown on
the page in order to get copied into the new template. As templates get
more complex (eg letter templates with welsh and english pages, and
possibly attachments), this falls outside of our existing design
pattern.

Instead we can simply show confirmation of the template they've
selected, by showing the template preview, and give them a chance to
provide a new name for the template copy (but pre-filled with a sensible
copy name).

This means that we can make sure all of the data is copied correctly
behind-the-scenes. When a user confirms that they're copying the correct
template, we still do a POST to our API to create a new template, but
populated with all of the data from the source data automatically.

To minimise the changeset, I am leaving letter attachments aside for now
still. That will be handled in a future PR.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -583,11 +583,6 @@ def delete_template_folder(service_id, template_folder_id):
 )
 @user_has_permissions("manage_templates")
 def add_service_template(service_id, template_type, template_folder_id=None):
-    # letter templates are created for the user straight from the choose_template page, but this
-    # code path is still accessed if the user has come via the copy an existing template flow
-    if template_type == "letter" and request.endpoint == "main.add_service_template":
-        abort(404)
-
     if template_type not in current_service.available_template_types:
         return redirect(
             url_for(
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -36,6 +36,7 @@ from app.constants import QR_CODE_TOO_LONG, LetterLanguageOptions
 from app.formatters import character_count, message_count
 from app.main import main, no_cookie
 from app.main.forms import (
+    CopyTemplateForm,
     EmailTemplateForm,
     LetterTemplateForm,
     LetterTemplateLanguagesForm,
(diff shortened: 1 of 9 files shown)


samuelhwilliams:
@karlchillmaid would I be able to get your content eyes on this new 'copy template' page please? 🙏

karlchillmaid:
> @karlchillmaid would I be able to get your content eyes on this new 'copy template' page please? 🙏
> 
> I wonder if the page(s) that currently say 'Copy an existing template' should say something more like 'Select the template to copy', then the "confirmation" page could say "Copy an existing template"?
> 
> Possibly the 'template name' field makes more sense as something like 'name of the template copy', but I can't think of good words there...

I’ve only had a quick look, but how about:

1. `Copy an existing template` becomes `Choose the template you want to copy` or `Choose a template to copy` or just `Choose a template` (although this last one lacks a little context).
2. As you say, let’s try `Copy an existing template` for the confirmation page. This is consistent with the first button the user clicks to trigger this flow. If it doesn’t look right we can try alternatives.
3. The button on the confirmation page can be `Copy this template`

app/main/views/templates.py
```diff
@@ -399,42 +400,66 @@ def choose_template_to_copy(
...
+        template.from_name = current_service.email_sender_name
+    template.name = _get_template_copy_name(template._template, current_service.all_templates)
 
-    template_folder = template_folder_api_client.get_template_folder(from_service, template["folder"])
+    template_folder = template_folder_api_client.get_template_folder(from_service_id, template.get_raw("folder"))
     if not current_user.has_template_folder_permission(template_folder, service=current_service):
         abort(403)
 
-    if request.method == "POST":
-        return add_service_template(service_id, template["template_type"])
```
leohemsted, line 413: as a minor note the `add_service_template` function can be updated - the template_type=='letter' 404 check now doesn't need to check the endpoint, and can just be

```
    if template_type == "letter":
        abort(404)
```

since we don't ever hit that page for letters now

samuelhwilliams: ah nice catch 👍  cleaned up

PR #4940, opened 2023-12-15
Fix error when platform admin users try to use the make service live journey when a service has no organisation

quis:
Code elsewhere was assuming that the service has an organisation by the time it gets to this point.

For platform admin users it’s possible to get into the new ‘make service live’ journey without the service having an organisation.

This pull request first makes sure that the user gets a `403` instead of `500` at any point in the make service live journey. Then it adds a custom error page which is shown at the start of the journey.

***

image


klssmith:
Would it be worth adding a check to see if the agreement has been signed too? To match what the checks we do if you put a service live from the service settings table

klssmith (approved):

PR #4943, opened 2023-12-16
Update letter prices and prevent regression

samuelhwilliams:


leohemsted (approved):

app/formatters.py
```diff
@@ -454,3 +454,12 @@ def format_auth_type(auth_type, with_indefinite_article=False):
 def sentence_case(sentence):
     first_word, rest_of_sentence = (sentence + " ").split(" ", 1)
     return f"{first_word.title()} {rest_of_sentence}"[:-1]
+
+
+def format_currency(pennies: int, long: bool) -> str:
```
quis, line 459: We already have this, is it worth refactoring into one function or renaming to make the difference clear?
https://github.com/alphagov/notifications-admin/blob/2667dcacad4801d2cb52e7dc2a73af93bbdc7fe7/app/formatters.py#L274-L278

samuelhwilliams: yeah let's - although I'll follow up with that in a separate PR shortly

PR #4944, opened 2023-12-16
Copy templates to target folder

samuelhwilliams:
If a user is in a folder when they start the 'copy template' journey, we
should make sure to create the new template inside that folder (as
opposed to just at the root).


klssmith (approved):
Tested locally and all worked well

PR #4947, opened 2023-12-19
Make it explicit that link text only works in emails

karlchillmaid:
We’ve had a couple of support tickets in the last few months where users were trying to add Markdown for link text to text message templates.

This PR makes a few small changes to (hopefully) make it clearer that this is not possible.

### DIFF
--- a/app/templates/views/guidance/using-notify/links-and-URLs.html
+++ b/app/templates/views/guidance/using-notify/links-and-URLs.html
@@ -48,16 +48,18 @@
 
   <p class="govuk-body">Never use ‘click here’, ‘click link’, ‘this link’ or ‘more’, as these do not make sense when read out of context.</p>
 
-  <h3 class="heading-small">How to format link text</h3>
+  <h3 class="heading-small">How to add link text to an email</h3>
   <p class="govuk-body">GOV.UK Notify uses Markdown to format link text.</p>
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/using-notify/links-and-URLs.html
```diff
@@ -48,16 +48,18 @@ <h2 class="heading-medium" id="link-text-in-emails">Link text in emails</h2>
...
 
-  <h3 class="heading-small">How to format link text</h3>
+  <h3 class="heading-small">How to add link text to an email</h3>
   <p class="govuk-body">GOV.UK Notify uses Markdown to format link text.</p>
   <p class="govuk-body">To convert text into a link, use square brackets around the link text and round brackets around the URL. Make sure there are no spaces between the brackets or the link will not work. For example:</p>
 
   <pre class="formatting-example"><code class="lang-md">[Apply now](https://www.gov.uk/example)</code></pre>
 
-  <p class="govuk-body">To see formatting instructions while you’re editing a template:</p>
+  <p class="govuk-body">You cannot add link text to text message or letter templates.</p>
```
karlchillmaid, line 57:
```suggestion
  <p class="govuk-body">You cannot use Markdown to add link text to a text message or letter template.</p>
```

CrystalPea (approved):

PR #4950, opened 2023-12-20
Add api_key_id to notifications sent report that users can download

CrystalPea:
We already provide sender name and email, and job id, makes sense to tell users which api key notifications were sent with. Added after a user requested it.

Needs to go in after: https://github.com/alphagov/notifications-api/pull/3985

### DIFF
--- a/app/utils/csv.py
+++ b/app/utils/csv.py
@@ -59,7 +59,18 @@ def generate_notifications_csv(**kwargs):
         original_column_headers = original_upload.column_headers
         fieldnames = ["Row number"] + original_column_headers + ["Template", "Type", "Job", "Status", "Time"]
     else:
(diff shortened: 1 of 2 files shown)


leohemsted (dismissed):
I don't think we can use this - the API key ID is not something that users have access to (the API key they store is `{name}-{service_id}-{secret}`. We should probably just return the API key name here.

CrystalPea:
Good catch @leohemsted 🙌🏼

CrystalPea:
@leohemsted Ok, done now, it needed a small change in the API, too: https://github.com/alphagov/notifications-api/pull/3985

idavidmcdonald (approved):

leohemsted (approved):

PR #4952, opened 2023-12-21
Set up prototype for ‘join a service’ user research

quis:
This branch should not be merged, but instead deployed to the prototype environment, following the instructions here: https://github.com/alphagov/notifications-manuals/wiki/Deploy-a-prototype

To test this you need to set up a fake organisation in the relevant environment, and assign the `example.gov.uk` domain to it.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -153,6 +153,7 @@ class Test(Development):
 
 
 class Preview(Config):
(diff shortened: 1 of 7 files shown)


quis:
Closing because we have finished this round of research

PR #4953, opened 2023-12-22
Update guidance for scheduling jobs

samuelhwilliams:
We updated Notify to be able to schedule jobs a week in advance, but forgot to update the guidance page.

image

### DIFF
--- a/app/templates/views/guidance/using-notify/schedule-messages.html
+++ b/app/templates/views/guidance/using-notify/schedule-messages.html
@@ -12,7 +12,7 @@
 {% block content_column_content %}
   <h1 class="heading-large">Schedule emails and text messages</h1>
 
-  <p class="govuk-body">You can schedule messages up to 4 days in advance.</p>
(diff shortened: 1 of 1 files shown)


klssmith (approved):

idavidmcdonald (approved):
Technically fine. Not sure whether it should be 'up to 1 week' rather than 'up to a week'.

Let's merge this as is, as it is an improvement but then maybe @karlchillmaid / @saimaghafoor when back from holidays can follow up with a further change if needed.

samuelhwilliams:
Gonna change it to `one week` as I vaguely remember Karl that 'one' should be spelled out and 2+ put as numbers. I can't find that slack message now I'm searching for it, but I think more than 50% confident that was a thing.

So will just make it that. Karl can still review it if that's not right when the holiday's over...

idavidmcdonald:
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#numbers

PR #4954, opened 2023-12-27
Replace hardcoded 7 day retention statement with service specific

idavidmcdonald:
The API integration page was hardcoded to 7 days, which may not be correct as services can have retentions between 3 and 90 days.

## Before
image

If the service has consistent retention, ie email, letter and sms have the same retention periods, it is easy to update this number to make it accurate.

If the service does not have consistent retention, which is rare, but possible (only platform admins can put a service in this state) then I've opted for a quick solution of just removing the line from the page. Happy for this to be adapated as designers see fit (we could for example show the retention per channel again if we wanted to, or instead refer generically to the retention settings for their service).

I've tested this with both live and trial mode services (trial mode services don't have any rows in the service_data_retention table).

## Service with custom retention of 80 days for all channels
image

## Trial mode service (which therefore has the default of 7 days for all channels)
image

## Service with inconsistent custom retention across channels
image

@karlchillmaid, @saimaghafoor, @quis - one to be aware of and review from a content/design perspective. Feel free to change the approach I've gone for too (but hopefully this is a positive improvement and further changes can build on top of it).

I also spotted some other places where 7 days is hardcoded.

https://github.com/alphagov/notifications-admin/blob/main/app/templates/views/guidance/features/security.html#L26
https://github.com/alphagov/notifications-admin/blob/main/app/templates/views/guidance/using-notify/delivery-times.html#L17

These will both want to be updated by a content designer please.

### DIFF
--- a/app/templates/views/api/index.html
+++ b/app/templates/views/api/index.html
@@ -39,9 +39,11 @@
         <p class="api-notifications-item__meta">
           When you send messages via the API they’ll appear here.
         </p>
+        {% if current_service.get_consistent_data_retention_period() %}
         <p class="api-notifications-item__meta">
-          Notify deletes messages after 7 days.
+          Notify deletes messages after {{ current_service.get_consistent_data_retention_period() }} days.
         </p>
+        {% endif %}
       </div>
     {% endif %}
     {% for notification in api_notifications.notifications %}
@@ -92,9 +94,11 @@
             Only showing the first 50 messages.
           </p>
         {% endif %}
+        {% if current_service.get_consistent_data_retention_period() %}
         <p class="api-notifications-item__meta">
-          Notify deletes messages after 7 days.
+          Notify deletes messages after {{ current_service.get_consistent_data_retention_period() }} days.
         </p>
+        {% endif %}
       </div>
     {% endif %}
   </div>
--- a/tests/app/main/views/test_api_integration.py
+++ b/tests/app/main/views/test_api_integration.py
@@ -4,12 +4,18 @@ from unittest.mock import call
 import pytest
 from flask import url_for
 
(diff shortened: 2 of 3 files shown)


app/templates/views/api/index.html
```diff
@@ -39,9 +39,11 @@ <h2 class="heading-small">
         <p class="api-notifications-item__meta">
           When you send messages via the API they’ll appear here.
         </p>
+        {% if service.get_consistent_data_retention_period() %}
         <p class="api-notifications-item__meta">
-          Notify deletes messages after 7 days.
+          Notify deletes messages after {{ service.get_consistent_data_retention_period() }} days.
```
quis, line 44: You don’t need to pass in `service` here, `current_service` is available as a global in all Jinja templates
```suggestion
          Notify deletes messages after {{ current_service.get_consistent_data_retention_period() }} days.
```

app/templates/views/api/index.html
```diff
@@ -92,9 +94,11 @@ <h3>
             Only showing the first 50 messages.
           </p>
         {% endif %}
+        {% if service.get_consistent_data_retention_period() %}
         <p class="api-notifications-item__meta">
-          Notify deletes messages after 7 days.
+          Notify deletes messages after {{ service.get_consistent_data_retention_period() }} days.
         </p>
+        {% endif %}
```
quis, line 101: I’m happy with this only being shown if the retention period is consistent

app/templates/views/api/index.html
idavidmcdonald: Good spot. Have changed and rebased this in.

quis (approved):
Nice, very thorough tests

PR #4956, opened 2023-12-28
dont run accessibility statement tests if git errors

leohemsted:
we check the accessibility statement for local changes by using a subprocess to run `git diff` against the origin/main branch.

We check the status code of that git diff to determine if there's been any changes.

However, if we run the tests on a copy of the admin source code that doesn't have an active git checkout, then the command will error and will return statuscode 1.

So, to determine if there were changes to the accessibility statement that we should check the update date of, we should check the diff output (in stdout) as well to make sure that has content.

-----

You can test this by renaming your `.git` folder to something else in your local checkout and running `pytest tests/app/test_accessibility_statement.py`

```
❯ mv .git .git-backup
❯ git diff --unified=0 --exit-code origin/main -- app/templates/views/accessibility_statement.html
error: Could not access 'origin/main'
```

### DIFF
--- a/tests/app/test_accessibility_statement.py
+++ b/tests/app/test_accessibility_statement.py
@@ -8,13 +8,17 @@ def test_last_review_date():
 
     # test local changes against main for a full diff of what will be merged
     statement_diff = subprocess.run(
-        [f"git diff --unified=0 --exit-code origin/main -- {statement_file_path}"], stdout=subprocess.PIPE, shell=True
+        [f"git diff --unified=0 --exit-code origin/main -- {statement_file_path}"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
So this appears to do what it says...

I've tested it by pulling it down locally, making a change to the accessibility statement without updating the date and confirming that the test still fails (which is the most important bit). I've also tested that no change to the accessibility statement also passes the test.

Importantly though, why do we need this?

Is it because in our docker image for the admin app, although we have the admin app files, it is not a git repository? ie there is no `.git` directory? If so, how have you confirmed this?

And is the solution to make our test more flexible to handle this case? Or should we just make sure that the `.git` directory is in our admin image? Or is that a bad thing to do?

leohemsted:
closed in favour of #4957

PR #4960, opened 2024-01-02
Get letter rates from API

quis:


idavidmcdonald:
Just a quick driveby comment. Very good to store them in Redis to protect our DB.

What is the plan for when we do update the rates? They will probably be cached in redis for 7 days be default? So I'm guessing a developer needs to go clear the cache manually for them to take affect? 

Sorry if I'm missing context/detail from just a very quick skim read!

quis:
Yeah, part of the process of updating the rates will be to go into platform admin and clear the cache.

Alternative strategy would be to reduce the TTL to something like 60 mins which we do for the count of services/organisations: https://github.com/alphagov/notifications-admin/blob/da3798c02f942c32521ebf24fd10364423727384/app/notify_client/status_api_client.py#L8C71-L8C75

I think the former is better so we can synchronise with when we send comms out.

tests/app/main/views/test_platform_admin.py
```diff
@@ -741,6 +749,13 @@ def test_clear_cache_shows_form(
             ],
             "Removed 26 objects across 13 key formats for service, organisation",
         ),
+        (
+            ["letter_rates"],
+            [
+                call("letter-rates"),
+            ],
+            "Removed 2 objects across 1 key formats for letter_rates",
```
idavidmcdonald, line 757: Doesn't really make any difference, but we only expect to see one key object here so would suggest 'Remove 1 object' rather than 'Removed 2 objects'.

tests/app/main/views/test_pricing.py
```diff
@@ -1,4 +1,4 @@
-def test_guidance_pricing_letters(client_request):
+def test_guidance_pricing_letters(client_request, mock_get_letter_rates):
     page = client_request.get(".guidance_pricing_letters")
 
     first_row = page.select_one("main table tbody tr")
```
idavidmcdonald, line 4: Do we want to remove the 'update these values if letter prices change' comment? They should only need updating if the mock is updated, right? In which case it should be self explanatory

app/models/letter_rates.py
```diff
@@ -0,0 +1,46 @@
...
+        "europe": "International",
+    }
+
+    @property
+    def rates(self):
+        return tuple(rate.rate_in_pennies for rate in self)
+
+    @property
+    def sheet_counts(self):
+        return sorted(set(rate.sheet_count for rate in self))
```
idavidmcdonald, line 36: Do we need to convert to a `set` before passing to `sorted`? `sorted` transforms an iterable it into a list so not immediately clear to me if `set` is needed? But I might be missing something

idavidmcdonald (approved):
3 optional code comments but generally things look good.

In terms of the manual cache purge, I slightly lean towards reducing the TTL and letting it expire naturally rather than relying on a manual purge. This is because it is an additional thing that needs to be remembered to do. Especially because if someone updates the letter rates in our DB (like https://github.com/alphagov/notifications-api/pull/3931/files) they will then need to remember to come and purge the cache on the start date the rates take affect. In that PR's case there, you would need to remember to purge it about 1 week later. Now that may be organised as part of comms, but remembering to do a small action 1 week later might be easy to forget.

You mention the benefit of timing the purging alongside comms going out. Do comms usually go out on the day of the new rate change? I thought not?

But more generally, is the long cache duration worth it? Do we get enough benefit versus a short cache duration?

app/models/letter_rates.py
quis: `set()` will turn `[1, 1, 2, 2, 3, 3]` into `{1, 2, 3}`

tests/app/main/views/test_platform_admin.py
quis: This is an artefact of how the mock is set up (the number of objects will always be double the number of key formats)

quis:
@idavidmcdonald I think you’re right about the short TTL, have changed to 1 hour in 755f5db59bbface2af50dd3d3f44d4a290b52257

idavidmcdonald (approved):

PR #4962, opened 2024-01-03
Duplicate and copy attachments when copying templates

samuelhwilliams:
When a user copies a letter template, we will also copy across any letter attachment for the template.

We make a full duplicative copy of the template so that, if copying templates across service boundaries, the copied template references an attachment within its own service namespace in the S3 bucket.

We still make a full copy if the template is copied within the same service, so that in theory if we clean up letter attachments its easier to reason about - an attachment is only ever used by one template (albeit possibly multiple template versions).


CrystalPea (approved):
LGTM 👍🏼

PR #4965, opened 2024-01-05
Fix API unknown errors

samuelhwilliams:
`requests.Response` defines `__bool__` as `self.ok`, which returns True if the status code is < 400.

Therefore, all 4xx and 5xx responses are considered `False`. Which means that when we try to log errors, and check `if error.response`, it always evaluates to `False`, and so we never log the response URL - just `unknown`.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -3,6 +3,7 @@ import pathlib
 from time import monotonic
 
 import jinja2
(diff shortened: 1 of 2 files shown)


leohemsted (approved):
ahhhh tricksy requests

PR #4966, opened 2024-01-05
Highlight the relevant platform admin navigation item

klssmith:
This highlights the relevant link in the left hand platform admin navigation when viewing a platform admin page. We do this for the other types of left hand navigation, so this just adds the same approach to the platform admin pages.

### Before
Screenshot 2024-01-05 at 16 54 17

### After
Screenshot 2024-01-05 at 16 54 02


quis (approved):
Nice

PR #4967, opened 2024-01-09
Only call template preview for letter templates

quis:
If someone hacks the URL to try to get a preview of an email template then they will get a 500 back from template preview.

Lets catch this early and return a 404 instead because we have no way of showing a PNG or PDF for an email or text message template.

### DIFF
--- a/app/template_previews.py
+++ b/app/template_previews.py
@@ -64,21 +64,6 @@ class TemplatePreview:
         return response.content, response.status_code, cls.get_allowed_headers(response.headers)
 
     @classmethod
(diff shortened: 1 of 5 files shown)


app/template_previews.py
```diff
@@ -19,6 +19,10 @@ def get_allowed_headers(headers):
     def get_preview_for_templated_letter(cls, db_template, filetype, values=None, page=None):
```
leohemsted, line 19: need to add `branding_filename` here as an optional and pull it in to the data dict if supplied

PR #4969, opened 2024-01-09
Fix missing `branding_filename` argument

quis:
…and add a test that doesn’t rely on a mock to pass

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -293,6 +293,7 @@ def letter_branding_preview_image(filename):
             "here, content here’, making it look like readable English."
         ),
         "template_type": "letter",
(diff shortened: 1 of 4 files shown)


klssmith:
This gave an error of `KeyError: 'is_precompiled_letter'` when I tried to preview changing the letter branding and the `letter_branding_preview_image` route was called.

quis:
@klssmith good catch, fixed in [f05ce11](https://github.com/alphagov/notifications-admin/pull/4969/commits/f05ce11e6ba6f98fc51bb39d58ae59c296806731)

klssmith (approved):
Tested, and all worked well

PR #4978, opened 2024-01-17
Fix letter branding preview when no service present

quis:
Because we serve letter branding images without cookies there’s no`current_service` available in the request context.

This was causing an error when trying to preview a letter with no branding, because the fallback is to look at the branding for the current service.

This commit adds a conditional for this case, and makes the test more end-to-end to ensure this edge case is accounted for.

### DIFF
--- a/tests/app/main/views/test_templates.py
+++ b/tests/app/main/views/test_templates.py
@@ -1937,23 +1937,31 @@ def test_letter_branding_preview_image(
     original_filename,
     new_filename,
 ):
-    mocked_preview = mocker.patch(
-        "app.main.views.templates.TemplatePreview.get_preview_for_templated_letter", return_value="foo"
-    )
+    class MockedResponse:
+        content = "foo"
+        status_code = 200
+        headers = {}
+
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -1937,23 +1937,31 @@ def test_letter_branding_preview_image(
...
         filename=original_filename,
     )
 
     mocked_preview.assert_called_with(
-        {
-            "subject": "An example letter",
-            "content": ANY,
-            "template_type": "letter",
-            "is_precompiled_letter": False,
+        "http://localhost:9999/preview.png",
```
CrystalPea, line 1952: I don't understand how the assertion worked both with the old version of args and the new version? Not that much changed in the code to change the call so much?

CrystalPea: Ohh, yes, of course, I missed that the mock changed, thank you 🙏🏼

CrystalPea (approved):

PR #4979, opened 2024-01-18
Fix back link to join service journey appearing on ‘Add service’ page

quis:
Apparently `block` tags in Jinja are always evaluated even if they are inside a conditional which evaluates to `False`.

Moving the `if` inside the `block` makes things work as intended.

This also `403`s the page that was being linked to for good measure.

### DIFF
--- a/tests/app/main/views/test_add_service.py
+++ b/tests/app/main/views/test_add_service.py
@@ -6,7 +6,7 @@ from notifications_python_client.errors import HTTPError
 
 from app.utils.user import is_gov_user
 from tests import organisation_json, service_json
-from tests.conftest import SERVICE_ONE_ID, SERVICE_TWO_ID, normalize_spaces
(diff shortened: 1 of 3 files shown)


tests/app/main/views/test_add_service.py
```diff
@@ -395,12 +404,36 @@ def test_email_auth_user_creates_service_with_email_auth_permission(
...
 
 
+@pytest.mark.parametrize(
+    "organisation",
+    (
+        None,
+        organisation_json(ORGANISATION_ID),
+    ),
+)
+def test_join_or_add_service_page_redirects_without_permission(
```
CrystalPea, line 414: this test doesn't actually test that the we redirect, but that we throw 403, so maybe name could reflect that? 

```suggestion
def test_join_or_add_service_page_403s_without_permission(
```

CrystalPea (approved):
One test needs name change I thing, otherwise it all looks fine 👍🏼

CrystalPea (approved):

PR #4983, opened 2024-01-24
Include onwards request headers in requests to template-preview, api

risicle:
https://trello.com/c/n4FmWuff/585-annotate-logs-with-cloudfront-http-request-ids

For api requests, we were already including trace id and span id manually, but this formalises that using `NotifyRequest.get_onwards_request_headers()` (and as a result handles parent span id properly too).

For template-preview, I did a little refactoring to pull `get_page_counts_for_letter` and `sanitise_letter` into the singleton pseudo-client class you already have. This allowed me to factor the auth header generation into a shared method and at the same time include the onwards request headers.

Requests the the antivirus api are handled by utils' `AntivirusClient` which was instrumented in https://github.com/alphagov/notifications-utils/pull/1093 and the utils bump should bring that to this app.

That said, I've actually got another utils tracing improvement waiting in https://github.com/alphagov/notifications-utils/pull/1094 that I'd love to include in this PR if someone would review that...

### DIFF
--- a/app/notify_client/__init__.py
+++ b/app/notify_client/__init__.py
@@ -30,18 +30,14 @@ class NotifyAdminAPIClient(BaseAPIClient):
             "X-Custom-Forwarder": self.route_secret,
             "User-agent": f"NOTIFY-API-PYTHON-CLIENT/{__version__}",
         }
-        return self._add_request_id_header(headers)
-
-    @staticmethod
-    def _add_request_id_header(headers):
-        if not has_request_context():
-            return headers
-
-        headers["X-B3-TraceId"] = request.request_id
-        headers["X-B3-SpanId"] = request.span_id
-
-        if g.user_id:
-            headers["X-Notify-User-Id"] = g.user_id
+        if has_request_context():
+            if hasattr(request, "get_onwards_request_headers"):
+                headers = {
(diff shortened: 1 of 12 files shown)


app/notify_client/__init__.py
```diff
@@ -30,18 +30,13 @@ def generate_headers(self, api_token):
...
-
-        if g.user_id:
-            headers["X-Notify-User-Id"] = g.user_id
+        if has_request_context() and hasattr(request, "get_onwards_request_headers"):
+            headers = {
+                **request.get_onwards_request_headers(),
+                **headers,
+            }
+            if g.user_id:
+                headers["X-Notify-User-Id"] = g.user_id
```
leohemsted, line 39: what does the `hasattr(request, "get_onwards_request_headers")` check do in practice - when might it not have this?

i only ask as this g.user_id stuff is now stuck inside that hasattr check, so won't be attached to the headers in that case where previously it would.

risicle: It only might not have it if for some reason `NotifyRequest` isn't being used. But yeah, I should probably pull it out of that clause because it doesn't really make any sense.

leohemsted (approved):

c4seyla (approved):

PR #4989, opened 2024-02-07
config: set `NOTIFY_REQUEST_LOG_LEVEL` to `INFO`

risicle:
To match other apps. This means we can see the same request logs on both paas and ECS, allowing us to e.g. set up common alert conditions.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -17,6 +17,7 @@ class Config:
 
     # Logging
     DEBUG = False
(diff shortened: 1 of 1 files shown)


klssmith (approved):

risicle:
Oops forgot about this

klssmith (approved):

PR #4993, opened 2024-02-14
Add tudor crown PNG file

whpearson:
This is just to be used in the email template. The crown you see in the website header is an SVG, in
the page HTML.

It is going out in admin so that it can be referenced in utils.


leohemsted (approved):
checked, it's available locally - lets double check it's in s3 bucket correctly after we deploy

PR #4995, opened 2024-02-19
Bumping utils for crown in email template preview

whpearson:
**74.12.1**
email template: use the new crown and associated styling
**74.12.0**
Remove celery.* structured logging for now. It can return when we have more certainty over what values it will log or can rule out that it is ever given sensitive values.
**74.11.0**

- Add option to ignore list of web paths from request logging. Defaults to /_status and /metrics.
- Add new fields to web request log messages (user_agent, host, path)

**74.9.1**
logging: set celery.worker.strategy logging to WARNING to prevent sensitive information being logged
**74.9.0**
logging: also attach handlers etc. to celery.worker and celery.redirected
**74.8.1**
Always hide barcodes when printing a letter - not only when adding a NOTIFY tag

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -27,7 +27,7 @@ fido2==1.1.0
 # PaaS
 awscli-cwlogs>=1.4,<1.5
 itsdangerous==2.1.2
-notifications-utils @ git+https://github.com/alphagov/notifications-utils.git@74.8.0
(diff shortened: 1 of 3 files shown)


quis (approved):

quis:
Run locally:

image

PR #5001, opened 2024-02-29
Email branding iframe

BlessedDev:
The iframe used to preview email branding currently fails WCAG reflow criteria.
This PR which is based on the requirements from this [ticket](https://trello.com/c/BhMRHbFK/590-preview-iframe-should-resize-with-page), addresses this issue by reducing the content of the example email displayed to the user.

 `main.email_template` now checks for an `email_branding_preview` flag that is set
 so that the content of `email-branding-preview-example-email.md` which is
 more compliant with WCAG reflow guidelines is served within the iframe.

To add more context, the iframe is set to no scrolling and the comparatively verbose text
 in `example-email.md` is not fully displayed within the iframe at any resolution.
 `example-email.md` is being preserved for it's content to be served by /_email for other purposes.

**Before**
before 1
before 2

**After**
after1

after 2

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -70,6 +70,7 @@ def design_content():
 def email_template():
     branding_style = request.args.get("branding_style")
     subject = request.args.get("title", default="Preview of email branding")
+    email_branding_preview = request.args.get("email_branding_preview", False)
 
     if not branding_style or branding_style in {"govuk", FieldWithNoneOption.NONE_OPTION_VALUE}:
         branding = EmailBranding.govuk_branding()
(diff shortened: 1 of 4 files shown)


tombye (commented):
The code changes make sense but I'm unsure why we're creating another markdown file for example email content. Is `email_template` used anywhere else that uses `example-email.md`?

tombye (approved):
Chatting with @BlessedDev on Slack, it seems there is more than one use of the `/_email` endpoint so these changes make sense.

Note: I've added a section to the docs on the email template referencing the endpoint, to make it clearer that it can also be used in isolation.

PR #5004, opened 2024-03-01
Fix bug where GP org name exists: catch the error and show user an actionable message

CrystalPea:


tombye:
I vote for the second option, because the actions we ask them to do are both outside of this page so linking to a field in the page doesn't make sense. We should also link to both actions in the text:
- our support page
- the page to change your service name

...to follow the validation pattern of linking to the thing you need to fix.

BlessedDev (approved):
Tested this locally, functionality works and improves the UX.

BlessedDev (approved):

PR #5008, opened 2024-03-11
Tell users about the free allowance cap

karlchillmaid:
This PR adds some content to the ‘Text message pricing’ page to explain that we may remove the free allowance from high-volume services.

The content is not specific about thresholds because these can change year-on-year – so using old numbers is a bit misleading.

### DIFF
--- a/app/templates/views/guidance/pricing/text-message-pricing.html
+++ b/app/templates/views/guidance/pricing/text-message-pricing.html
@@ -57,6 +57,8 @@ Text message pricing
  will no longer get a free allowance.</p>
   </div>
 
+    <p class="govuk-body">We cover the cost of the free allowance to help support those teams who need it the most. Services that send a very high volume of text messages in a single year will not get a free allowance. We’ll contact you if we need to stop your allowance.</p>
+
   <h2 class="heading-medium" id="long-text-messages">Long text messages</h2>
   <p class="govuk-body">If a text message is longer than 160 characters (including spaces), it counts as more than one message.</p>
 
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -57,6 +57,8 @@ <h2 class="heading-medium" id="free-text-message-allowance">Free text message al
  will no longer get a free allowance.</p>
   </div>
 
+    <p class="govuk-body">We cover the cost of the free allowance to help support those teams who need it the most. Services that send a very high volume of text messages in a single year may not be eligible for the free allowance. We’ll contact you in advance if we need to stop your allowance.</p>
```
karlchillmaid, line 60:
```suggestion
    <p class="govuk-body">GOV.UK Notify covers the cost of the free allowance. We do this to help support those teams who need it most. If your service sends a very high volume of text messages in a single year, we may have to stop your free allowance.</p>
```

app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -57,6 +57,8 @@ <h2 class="heading-medium" id="free-text-message-allowance">Free text message al
  will no longer get a free allowance.</p>
   </div>
 
+    <p class="govuk-body">GOV.UK Notify covers the cost of the free allowance. We do this to help support those teams who need it most. If your service sends a very high volume of text messages in a single year, we may have to stop your free allowance.</p>
+
```
karlchillmaid, line 61:
```suggestion
    <p class="govuk-body">GOV.UK Notify covers the cost of the free allowance. We do this to help support those teams who need it most. If your service sends a very high volume of text messages in a single year, we may not renew your free allowance the following year.</p>

```

quis (approved):

PR #5010, opened 2024-03-12
Accessibility fixes for data tables

tombye:
Fixes an accessibility issue with the tables we use to show placeholder data where we show headings for the rows in the data visually but hide them in the HTML.

https://trello.com/c/46giIml8/589-fix-tables-of-notification-data-in-screen-readers

image

### How the tables work

The tables have fixed row headings visually. This is achieved by having a clone of the table that sits on top of it but is hidden from assistive tech, like screen readers, and only shows its row headings:

image

`div.fullscreen-scrollable-table` is the real one. `div.fullscreen-fixed-table` is the fake.

When we say that the row headings are hidden, we mean the first cells for each row of `div.fullscreen-scrollable-table` which currently have `display: none` in their CSS.

### The Apple 'scroll bounce'

For most cases, showing the `div.fullscreen-scrollable-table` row headings doesn't matter because they're hidden by those of the `div.fullscreen-fixed-table` table. Except on Apple computers/devices because of their 'scroll bounce'.

When you scroll past the end of a scroll area on Apple computers/laptops, it makes the content you scrolled overshoot and then 'bounce back', like it's on an elastic band:

data_table_bouncy_scroll

When this happens if you show the row headings, it means you see the row headings from both tables. These changes instead make the text and border colour transparent to hide them visually. Worth saying that I tested with `visibility: hidden` and found this caused some screen readers to ignore them so we couldn't do it.

### Issue with how real and fake tables are laid out

I found some small inconsistencies with how the 2 tables were rendered because their CSS is set so they have `position` set to different values (`static` for the real table, `absolute` for the fake). These changes include some JS that fixes that for the first column by setting the real table's column width to that of the fake, because the fake is what users see.

### What's already tested and what to review

Reviewers should focus on whether these changes make sense and whether the code is sensible.

I've tested these changes with [the browsers and assistive technologies we support](https://github.com/alphagov/notifications-manuals/wiki/Support-for-browsers,-email-clients-and-assistive-technologies).

If you're comfortable using a screen reader, please feel free to test with one but don't feel you have to.

Likewise, if you want to check in any of the browsers we support but you wouldn't have installed, I can get you access to our testing tool for this but don't feel you have to.


tests/javascripts/fullscreenTable.test.js
```diff
@@ -295,6 +294,42 @@ describe('FullscreenTable', () => {
...
+
+      screenMock.reset();
+
+    });
+
+    // table dimensions are set when the module starts and on page resizes
+    // we can't fake the columns being different when it starts so our test needs to run after a resize
+    test('when the page has resized', () => {
+
+      let tableFrameNumberColumnHeader = container.querySelector('.table-field-heading-first');
```
leohemsted, line 310: minor: i'd possibly consider making this a `document.querySelector('.fullscreen-scrollable-table .table-field-heading-first'` so it's easier to compare and contrast with the other selector below

leohemsted: could consider renaming this variable so it's apparent what's interesting about it and why we care about it too

`let invisibleScrollableTopLeftCell = ...`

tests/javascripts/fullscreenTable.test.js
```diff
@@ -295,6 +294,42 @@ describe('FullscreenTable', () => {
...
+      let numberColumnFrameNumberColumnHeader;
+
+      // set main content column width (used by module as gauge for table width)
+      screenMock.window.setWidthTo(1024);
+      document.querySelector('main').setAttribute('style', 'width: 712px');
+
+      // start module
+      window.GOVUK.notifyModules.start();
+
+      numberColumnFrameNumberColumnHeader = document.querySelector('.fullscreen-fixed-table .table-field-heading-first');
```
leohemsted, line 320: think this could be named something to draw attention to it being:

* the cell on the top left
* visible and non-scrollable

for example
`fixedTopLeftHeaderCell = ...`

leohemsted (commented):
thanks for taking the time to show this to me - it looks good and makes sense. 

just a couple of comments about those test variable names

tests/javascripts/fullscreenTable.test.js
tombye: All sounds good to me. I made the changes but worth saying they had to involve moving some code around as that node isn't available in the DOM with that selector before `window.GOVUK.notifyModules.start` is called so I had to move the assignment to after that.

PR #5016, opened 2024-03-19
Update optional-content.html

saimaghafoor:
Added content to advise users they cannot add placeholders for personalisation within optional content. Users frequently ask this question in zendesk.

### DIFF
--- a/app/templates/views/guidance/using-notify/optional-content.html
+++ b/app/templates/views/guidance/using-notify/optional-content.html
@@ -25,7 +25,7 @@
   <pre class="formatting-example">((under18??Please get your application signed by a parent or guardian.))</code></pre>
 
 <div class="govuk-inset-text"> 
-    <p class="govuk-body">You cannot add placeholders for personalisation within optional content. </p>
+    <p class="govuk-body">You cannot add a placeholder for personalisation within optional content. </p>
  </div>
  
 <p class="govuk-body">For each person you send the message to, specify ‘yes’ or ‘no’ to show or hide the optional content.</p>
--- a/app/templates/views/guidance/using-notify/optional-content.html
+++ b/app/templates/views/guidance/using-notify/optional-content.html
@@ -24,7 +24,11 @@
 
   <pre class="formatting-example">((under18??Please get your application signed by a parent or guardian.))</code></pre>
 
-  <p class="govuk-body">For each person you send the message to, specify ‘yes’ or ‘no’ to show or hide the optional content.</p>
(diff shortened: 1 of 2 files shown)


app/templates/views/guidance/using-notify/optional-content.html
```diff
@@ -24,7 +24,11 @@ <h1 class="heading-large">Optional content</h1>
 
   <pre class="formatting-example">((under18??Please get your application signed by a parent or guardian.))</code></pre>
 
-  <p class="govuk-body">For each person you send the message to, specify ‘yes’ or ‘no’ to show or hide the optional content.</p>
+<div class="govuk-inset-text"> 
+    <p class="govuk-body">You cannot add placeholders for personalisation within optional content. </p>
```
karlchillmaid, line 28: I think it might it might look a little cleaner if we say `a placeholder` singular. For example:

> You cannot add a placeholder for personalisation within optional content.

I also thought about alternatives to ‘within’. For example:

```You cannot add a placeholder for personalised content inside your optional content.```

```You cannot add a placeholder for personalised content to your optional content.```

Of those, I think `to` is the best.

Lasly, it might be nothing, but I wonder if users will notice that we’re using placeholder here to refer to personalisation, whereas we use it to refer to optional content further up this page.

I tried to think about how we might address this, but it ends up being a little wordy. For example:

```You cannot add a placeholder for personalised content to your optional content placeholder.```

I’m only sharing these for consideration. I’ll leave it to you to decide the final version of this content. :-)

saimaghafoor: Thanks for the spot on plural placeholders and other feedback. I've changed it to singular now :)

saimaghafoor:
@idavidmcdonald If you're happy we can merge and deploy :)

idavidmcdonald:
Here is a screenshot of the change you've made, just to make sure you are happy with it

image

I think there is one other place to make the equivalent change. That would be on the page when you edit a template and the formatting guidance is shown at the bottom of the page:

image

saimaghafoor:
@idavidmcdonald Great spot. Added content to the edit templates page.  Can be merged and deployed if we're happy with it.

idavidmcdonald:
Screenshot of your new addition, Saima:

image

saimaghafoor:
Looks good to me. Happy for it to be merged and deployed.

idavidmcdonald (approved):
All fine from a tech perspective. Let me know if you I am good to merge for you, @saimaghafoor

PR #5019, opened 2024-03-26
Get the SMS rate from the API

quis:
Copies the approach we used for letters from
- https://github.com/alphagov/notifications-admin/pull/4960

Gets the rates from:
- https://github.com/alphagov/notifications-api/pull/4049

Caches them in Redis for 1 hour so we don’t hit the API on high-traffic public pages (for example the homepage)

Uses the current rate as test data so the existing tests pass unmodified.

Adds a conditional to only show incoming pricing message when old pricing in effect.


app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -17,16 +17,21 @@ <h1 class="heading-large">Text message pricing</h1>
...
+          sms_rate.valid_from
+        ]|max|format_date_normal
       }
     ) }}
 
   <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
   <p class="govuk-body">When a service has used its annual allowance, it costs {{ sms_rate }} pence (plus VAT) for each text message you send.</p>
 
-  <div class="govuk-inset-text">  
+  {% if sms_rate != 2.27 %}
```
CrystalPea, line 30: We will need to remember to delete that all if statement next week, right?

quis: Yup, I will put up a PR to delete it once this is merged

quis:
Thanks @CrystalPea. I have:
- fixed the pounds/pence formatting in 1339fa214c8a5b7dde61e73bf1aaa92d20a2d612
- added the SMS rates to the cache-clearing form in b0ca67b19c87a838b63cb568d734322ef9bc87ea

CrystalPea (approved):
Looks good, ship it! I tested cache clearing, too 😁 

Screenshot 2024-03-27 at 15 25 03

karlchillmaid (approved):
Content changes look correct.

I like the addition of logic to display the right content based on the current SMS rate.

The tests look consistent with the code changes.

PR #5022, opened 2024-03-27
Make data tables focusable only when they scroll

tombye:
Adds logic to see if the data tables need to scroll before making them focusable. To fix this issue:

https://trello.com/c/X63MGfzJ/668-address-placeholder-data-tables-are-focusable-even-if-they-dont-scroll

image

It's confusing for anyone using the keyboard to navigate to have these tables focusable when they don't need to scroll (which was the point of them being given focus in the first place).

Scrolling is possible via the keyboard using arrow keys but the thing you're scrolling needs to be focusable for this to work.

Note: also includes a fix for the test that checks the component is set up properly for receiving focus. It didn't check tabindex=0 was set before.

## Notes for reviewers

I've tested this in all [the browsers we support](https://github.com/alphagov/notifications-manuals/wiki/Support-for-browsers,-email-clients-and-assistive-technologies) so, other than seeing if the code is ok, it's worth trying these tables with data that makes them scroll (usually >=4 columns) and with data that doesn't.

### DIFF
--- a/app/assets/javascripts/fullscreenTable.js
+++ b/app/assets/javascripts/fullscreenTable.js
@@ -9,6 +9,7 @@
       this.$table = this.$component.find('table');
       this.nativeHeight = this.$component.innerHeight() + 20; // 20px to allow room for scrollbar
       this.topOffset = this.$component.offset().top;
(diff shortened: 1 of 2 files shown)


kr8n3r (approved):
works as designed

PR #5026, opened 2024-03-28
Stop trial mode GPs sending SMS

leohemsted:
I haven't had time to test this manually but unit tests work well

* Set sms_message_limit=0 for GPs when the service is added
* skip the tour for GPs as this would now error
* Set sms_message_limit=0 for a trial mode service if its organisation is changed to GP

we already handle by default the sms limit being set to 250k when a service goes live

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,8 @@ htmlcov/
 coverage.xml
 test_results.xml
 *,cover
+.testmon*
+
 
 # Translations
 *.mo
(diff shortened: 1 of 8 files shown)


quis:
Tested locally by hacking my default organisation type.

Confirm that I skip the tour and have a rate limit of 0:
image

However I was still able to send myself a text message (which also arrived on my phone):

image

quis:
Update: I think this is because I wasn’t running Redis locally. But now I can’t get Redis to start up so can’t confirm

quis:
Update, got Redis running locally and now sending one-off messages is also blocked:

image

quis (approved):

PR #5027, opened 2024-04-02
Fix divide by 0 error with estimated usage

quis:
If a user is part way through the first few days of the financial year then we consider this – as a percentage – to be 0%. This is incorrect, at the end of April 2nd you are 0.55% through the financial year.

This causes problems because elsewhere we assume the percentage will always be >0%:
https://github.com/alphagov/notifications-admin/blob/ca0d5dade5cf36e688945a3eed9d39d6658d2685/app/utils/services.py#L20

This means that – for the first few days of the year only – anyone trying to change their data retention will get a `ZeroDivisionError` exception because we can’t calculate how much we expect them to send in the coming year.

This commit increases the resolution of the calculation to about 5 minutes, from about 3 days. It also sets the minimum bound of the return value to about 5 minutes, so that it never returns 0%.


CrystalPea (approved):

samuelhwilliams:
😨

PR #5031, opened 2024-04-08
Remove OpenStreetMap from safe image sources

quis:
We were loading map tiles from OpenStreetMap for Emergency Alerts. Now that Emergency Alerts is no longer part of Notify we don’t need to load any images from OpenStreetMap’s servers, so we can remove their domain from our content security policy.


kr8n3r (approved):

risicle:
:cry:

PR #5032, opened 2024-04-08
Preview letter branding with `img`, not `iframe`

quis:
The iframe we use to show letter brandings only has value as a holder for the image of the letter PDF so should just be that image.

Presenting the branding example as a document (in an iframe) is confusing for screen reader users when it has no actual content other than the image.

The auditors testing the pages for setting email and letter branding raised issues for this in the audit on Oct 9th.

***

https://trello.com/c/Y7cgOc3u/497-iframe-for-letter-branding-should-just-be-an-image


kr8n3r (approved):
looks good, I see no issues with python code, but a dev might wade if if needed

Before
Screenshot 2024-04-08 at 14 21 20

after
Screenshot 2024-04-08 at 14 20 18

quis:
🚨 deploy blocked until frontend pipeline is fixed 🚨

PR #5033, opened 2024-04-08
Adding setting for Sending SMS to UK landlines

whpearson:
Work to use it will be in another card


whpearson:
Correspondin api changes here https://github.com/alphagov/notifications-api/pull/4058

leohemsted (approved):

PR #5035, opened 2024-04-10
Remove direct dependency on govuk-bank-holidays

quis:
We only use this indirectly through the module in utils: https://github.com/alphagov/notifications-utils/blob/6791aebe7c68b8ad7de7b1dd5b02791974f8ae07/notifications_utils/bank_holidays.py

Not pinning the dependency will make it easier to keep the list of bank holidays up to date, because we will only need to bump the version in utils.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,7 +2,6 @@
 # with package version changes made in requirements-app.txt
 
 ago==0.0.95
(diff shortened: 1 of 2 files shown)


tombye (approved):
Makes sense 👍🏻

PR #5036, opened 2024-04-11
Only ask for recipients once in one-off journey

quis:
If someone uses a placeholder which is the same as the key we use for the recipient <sup>1</sup>, we ask for the value twice, but only store it once.

This is even worse if the user has a the same placeholder with different space/casing in the same template<sup>2</sup>. In this case we would ask for the value once for each different variant of spacing/case, but still only store the first one entered.

This commit normalises all the placeholders after merging them with the special recipient keys.

This means that every instance of that placeholder will be populated in the template straight away, and the user won’t be asked for it again.

***

1. For example, `email address`, `phone number` or `address line 1`, `address line 2`, etc
2. For example `email address` and `EMAILADDRESS`

***

https://www.pivotaltracker.com/story/show/181220935

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -15,7 +15,7 @@ from flask import (
 from flask_login import current_user
 from notifications_python_client.errors import HTTPError
 from notifications_utils import SMS_CHAR_COUNT_LIMIT
-from notifications_utils.insensitive_dict import InsensitiveDict
+from notifications_utils.insensitive_dict import InsensitiveDict, InsensitiveSet
 from notifications_utils.postal_address import PostalAddress, address_lines_1_to_7_keys
 from notifications_utils.recipients import RecipientCSV, first_column_headings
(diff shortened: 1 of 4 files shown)


app/main/views/send.py
```diff
@@ -782,10 +782,14 @@ def start_job(service_id, upload_id):
 
 def fields_to_fill_in(template, prefill_current_user=False):
     if "letter" == template.template_type:
-        return letter_address_columns + list(template.placeholders)
+        return tuple(InsensitiveDict.from_keys(letter_address_columns + list(template.placeholders)).values())
```
whpearson, line 785: Might be nice to extract tuple(InsensitiveDict.from_keys(X).values()) as a function perhaps as toInsensitiveTuple()?

quis: @whpearson good idea – I extracted it here: https://github.com/alphagov/notifications-utils/pull/1111

whpearson (approved):
Changes look good

PR #5038, opened 2024-04-12
Make ‘sign-in method’ consistent

karlchillmaid:
This PR updates our form fields to use the term `sign-in method` consistently.


kr8n3r (approved):
I couldn't find any other

PR #5042, opened 2024-04-15
Update form_wrapper macro action attribute

BlessedDev:
This PR fixes the issue raised in [this](https://trello.com/c/bnjIRzze/695-fix-accessibility-issue-with-post-redirect-get-form-submission-pattern) ticket.
Updating the `form_wrapper` macro deploys the change across all the form pages on Notify.
Setting the `action` attribute to `#` seems to be the simplest and optimal solution

### DIFF
--- a/app/templates/components/form.html
+++ b/app/templates/components/form.html
@@ -1,6 +1,6 @@
 {% macro form_wrapper(
   method="post",
-  action=None,
+  action="#",
   autocomplete=False,
   class=None,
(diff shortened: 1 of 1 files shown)


tombye:
I'm closing this because we've since decided to try the alternative approach of setting the `action` attribute of the form to the URL of the page, without the `#`, to force the `#` out of future requests and stop it appearing in the browser history.

PR #5043, opened 2024-04-16
Clarify NHS Notify timeline

kr8n3r:
As requested and approved by NHS Notify stakeholders and copy agreed.

Screenshot 2024-04-16 at 09 47 35


karlchillmaid (approved):
Content looks good.

PR #5045, opened 2024-04-16
Show organisation membership on user page for platform admins

idavidmcdonald:
As requested in:
https://gds.slack.com/archives/C037E9EL8RE/p1713261630095319

Follows the same pattern as the services shown below. We can move this section higher or lower on the page if we wish.

image


quis (approved):
LGTM

PR #5050, opened 2024-04-19
Add permissions decorator to all letter branding endpoints

klssmith:
Some letter branding endpoints were not checking if a user was logged in, which resulted in a `500` status code when trying to visit these pages and logged out.


whpearson (approved):
Changes make sense

PR #5052, opened 2024-04-24
Resolve Dependabot alerts

quis:
Closes #5037

Closes #5040

Closes #5041


whpearson (approved):
Bootstrapped locally and things still work

PR #5054, opened 2024-04-25
Update route to serve letter template brand image

kr8n3r:
When a letter branding template is saved and then served for the preview before user is looking to change it, the url was `/templates/letter-preview-image/?branding\_style=uuid` which does not exist.

This PR changes the path to `/templates/letter-preview-image?branding\_style=uuid` which serves the preview

Fix from https://github.com/alphagov/notifications-admin/pull/5032

Before
Screenshot 2024-04-25 at 08 58 41

After
Screenshot 2024-04-25 at 10 00 19


app/main/views/templates.py
```diff
@@ -272,7 +272,7 @@ def view_letter_template_preview(service_id, template_id, filetype):
 
 
 @no_cookie.route("/templates/letter-preview-image")
-@no_cookie.route("/templates/letter-preview-image/<filename>")
+@no_cookie.route("/templates/letter-preview-image<filename>")
```
leohemsted, line 275: this isn't quite right - if we're passing in a filename of `foo.png` to `url_for('no_cookie.letter_branding_preview_image', filename='foo.png')` we'll end up with `/templates/letter-preview-imagefoo.png`, and that's definitely not right.

-----

i think instead we need to change where the macro calls this. That macro's passing in `filename=filename`, which isn't defined in the macro.

https://github.com/alphagov/notifications-admin/blob/71f646a7119bbe8d87a1d993a0b4e704c152a1bb/app/templates/components/branding-preview.html#L19-L23

When `url_for` tries to generate a URL, it tries to match any kwargs passed in with parameters in the URL path. So it'll match the `filename=None` with this second filename route. Any kwargs that aren't matched go as query strings (like branding_style). 

So rather than changing the second route, our approach should be "we need to ensure url_for coerces to the generic path with no filename". hope that makes sense

quis: Yeah, this is what I figured out too.

`filename=filename` where `filename` is undefined results in `filename=""` in Jinja.  This matches the second route, so the generated URL is `/templates/letter-preview-image/` (which doesn’t exist)

If we remove `filename=filename` that is equivalent to `filename=None`, which matches the first route, and the generated URL will be `/templates/letter-preview-image` (which does exist).

kr8n3r:
template.py has `filename = letter_branding_client.get_letter_branding(branding_style)["filename"]` which I assume resolves to `?branding\_style=<uuid>`

app/main/views/templates.py
quis: It only resolves to the wrong thing when filename is undefined. So I think the fix looks like cc64c9272221a0cd7ed3160dc04abcc0fc8727cb

In retrospect all those instances of `filename=""` in the tests were a bit of a code smell.

kr8n3r: yeah we found that pushing updated commit

kr8n3r: updated commit e20039eb026228e436b369fc57f869e4a9d7273 to resolve this

quis (approved):
Nice work. Definitely some niche behaviour of `url_for` that we didn’t anticipate here

PR #5055, opened 2024-04-25
Tidy up top level domains not stored in our database that sign up to use Notify

idavidmcdonald:
Review commit by commit

Worth reading https://gds.slack.com/archives/C06S65BV0K1/p1714055580394099 first

### DIFF
--- a/app/utils/email_domains.txt
+++ b/app/utils/email_domains.txt
@@ -6,6 +6,5 @@ police.uk
 cjsm.net
 ac.uk
 sch.uk
-onevoicewales.wales
 suttonmail.org
 hscni.net
--- a/app/utils/email_domains.txt
+++ b/app/utils/email_domains.txt
@@ -6,7 +6,6 @@ police.uk
 cjsm.net
 ac.uk
 sch.uk
-suttonmail.org
 hscni.net
 llyw.cymru
 nhs.wales
--- a/app/utils/email_domains.txt
+++ b/app/utils/email_domains.txt
@@ -8,3 +8,6 @@ ac.uk
 sch.uk
 suttonmail.org
 hscni.net
+llyw.cymru
+nhs.wales
+judiciary.uk


app/utils/email_domains.txt
```diff
@@ -6,6 +6,8 @@ police.uk
 cjsm.net
 ac.uk
 sch.uk
-onevoicewales.wales
 suttonmail.org
```
leohemsted, line 9: this one also stands out - five users from four trial mode services, the most recent of whom logged in three years ago

idavidmcdonald: suttonmail?

So yeah, it does. If you do a google search for the domain (https://www.google.com/search?q=suttonmail.org), then you will see that there are a few schools that have email addresses ending in suttonmail on their websites.

The MX record for suttonmail.org points at lgflmail.org/

Googling that gives me https://lgfl.net/ which looks like an email provider just for schools. They aren't an official public sector org administering who can sign up (like the other domains we have).

So maybe we remove them off this list because we can't trust their processes as much. And anyone who works at a school that uses the domain will need to fill in a support ticket to get access to Notify?

Given the super small numbers you mention, seems like we can probably take that risk of removing their domain from here and having to deal with any support tickets if they come in?

app/utils/email_domains.txt
```diff
@@ -6,6 +6,5 @@ police.uk
 cjsm.net
 ac.uk
```
CrystalPea, line 7: Do we still support universities? I seem to remember we stopped, so maybe that one needs to be removed, too?

idavidmcdonald: So we don't support universities anymore... but I think we still support colleges which also have ac.uk addresses.

I think we should keep ac.uk here because I see this list being a list of 'domains for the public sector' rather than a list of 'which domains are in the public sector AND are eligible to use Notify'.

This way, someone can still sign up as a college and we can approve their request to go live. Or they can sign up as a university and we would then reject their request to go live...

Open to changing that... but I think removing ac.uk will add more burden to us (we will need to add domains for every college that wants to use Notify) and I'm not sure how much burden that will be so suggest keeping the status quo for the moment...

app/utils/email_domains.txt
idavidmcdonald: I've added a new commit that removes suttonmail.org

quis (approved):

app/utils/email_domains.txt
leohemsted: good sleuthing!

yeah i vote remove em 🚮 . we can re-approach it if they come back.

PR #5056, opened 2024-04-25
Make button positions based on container size

tombye:
The size of the image for the preview of a letter scales relative to the viewport. This makes the top position of the 'edit' buttons do the same thing, using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries).

view of the letter preview page, with buttons positioned in different place ontop of an image odf the letter. There are arrows marking the width (with 'A') and height (with 'B') of the image, plus one ('C') for the position of a button relative to the top of the image.

# How it works

1. the image keeps its aspect ratio at whatever size so we can always apply it to the width (A) to get the height (B)
2. the image width is set to `100%` so it will always match that of it's container, which scales with the page
3. if we set the image container as the [container context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries#using_container_queries), we can access the image's dimensions in [container query units](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries#container_query_length_units)
4. finally, if we convert the top position of a button (C) to container query units, it will scale with the container (and so, the image).

# Browsers that don't know container queries

We still have some browsers in our support matrix that don't understand container queries:
- [which browsers understand them](https://caniuse.com/css-container-queries)
- [our browser support matrix](https://github.com/alphagov/notifications-manuals/wiki/Support-for-browsers,-email-clients-and-assistive-technologies#browsers)

For them, we just provide a static value, which will get used because CSS engines ignore any code they don't understand.

### DIFF
--- a/app/assets/stylesheets/views/template.scss
+++ b/app/assets/stylesheets/views/template.scss
@@ -5,7 +5,7 @@
 // - the height, calculated from the width, which is made of:
 //   - the width of the image, in container query width units: 100cqw
 //   - the height of the image as a fraction of the width: (1008 / 712)
-@mixin scalable_top($top) {
+@mixin scalable-top($top) {
   $max_height_of_image: 1008;
   $max_width_of_image: 712;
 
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
👏 👏

kr8n3r (approved):

PR #5058, opened 2024-04-26
Enforce types on URLs with `email`, `letter` (and sometimes `sms`) in

quis:
Using a URL convertor means we:
- automatically serve `404`s for things which aren’t valid template types, or things in the wrong case
- get useful exceptions if we try to build URLs using invalid values at development time 

This commit adds relevant URL convertors, and adds test coverage to catch the same thing happening in the future.


tests/app/main/test_permissions.py
```diff
@@ -300,42 +300,46 @@ def test_routes_have_permissions_decorators():
...
+            )
 
 
-def test_routes_require_uuids(client_request):
+def test_routes_require_types(client_request):
     for rule in current_app.url_map.iter_rules():
         for param in re.findall("<([^>]*)>", rule.rule):
             if "_id" in param and not param.startswith("uuid:"):
-                pytest.fail(("Should be <uuid:{}> in {}").format(param, rule.rule))
+                pytest.fail("Should be <uuid:{param}> in {rule.rule}")
```
leohemsted, line 341:
```suggestion
                pytest.fail(f"Should be <uuid:{param}> in {rule.rule}")
```

tests/app/main/test_permissions.py
```diff
@@ -300,42 +300,46 @@ def test_routes_have_permissions_decorators():
...
+            if ("template_type" in param or "notification_type" in param) and not (
+                param.startswith("template_type:") or param.startswith("branding_type:")
```
leohemsted, line 343: confused about template/notification in the contains check vs template/branding in the startswith check. is that right?

i also think this might be quite an error prone/nuanced check.  maybe it makes sense to split this into a couple bits - one where we split the param into a param name and an optional type, and then separately we can assert on known combinations.

leohemsted: we're missing a bunch of places in service_settings/branding.py that i would have expected to fail (that have `<notification_type>` or `<branding_type>` so there's definitely something going on.

leohemsted: actually, can we just enforce that all params need an enforced type, and then just have a catch-all string type for those random ones that dont fit into uuids or enum types?

currently, looks like the following are the only untyped URL params we have in admin

```
<branding_type>
<filename>
<filetype>
<language>
<logo>
<notification_type>
<permission>
<return_to>
<token>
<variant>
```

some of these are definitely ripe for having a defined type (for example, `<filetype>` is just png or pdf). Some of them are more open like `<logo>` but maybe `<str:logo>` would be a good enough aide memoire for us to think about what we're writing when we create new routes?

leohemsted (commented):
i like this move in general but think we could go further and cover the whole param space. also, there's a couple endpoints that have been missed in branding.py. not sure how the tests passed?

eg 
https://github.com/alphagov/notifications-admin/blob/enforce-template-type-param/app/main/views/service_settings/branding.py#L498

tests/app/main/test_permissions.py
quis: I’ve added a few more commits so that:
- every URL parameter must specify a type (even if it’s only `string:`)
- certain params which match a given pattern (for example `_id`) must still be a specific type (for example `uuid:`)

app/main/views/service_settings/index.py
```diff
@@ -1046,26 +1046,26 @@ def set_per_minute_rate_limit(service_id):
...
-@main.route("/services/<uuid:service_id>/service-settings/set-<notification_type>-branding", methods=["GET", "POST"])
+@main.route(
+    "/services/<uuid:service_id>/service-settings/set-<branding_type:branding_type>-branding",
+    methods=["GET", "POST"],
+)
 @user_is_platform_admin
-def service_set_branding(service_id, notification_type):
-    notification_type = notification_type.lower()
+def service_set_branding(service_id, branding_type):
+    branding_type = branding_type.lower()
```
quis, line 1056:
```suggestion
```

leohemsted (approved):

leohemsted (approved):

PR #5059, opened 2024-04-26
Adding validator for potentially malicious sender IDS

whpearson:
Following on from an incident, it was identified that we could block senderids that were for companies, such as Evri etc.

This work is the initial effort in this regard, we will look at expanding this list of companies and perhaps keeping it in the database in later work.

The work is case insensitive for the sender IDs involved in the incident as well as some banks.

The error looks like 

Screenshot 2024-04-26 at 16 22 18

@karlchillmaid when you have a moment could you review the error message?

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -165,7 +165,7 @@ class IsNotAPotentiallyMaliciousSenderID:
         if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
             current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
             raise ValidationError(
-                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing"
+                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing scams"
             )
 
 
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -165,7 +165,7 @@ class IsNotAPotentiallyMaliciousSenderID:
         if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
             current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
             raise ValidationError(
-                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing scams"
+                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients from phishing scams"
             )
 
 
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -146,6 +146,29 @@ class IsNotAGenericSenderID:
             raise ValidationError(self.message)
 
 
+class IsNotAPotentiallyMaliciousSenderID:
+    potentially_malicious_sender_ids = [
+        "amazon",
+        "evri",
+        "lloydsbank",
+        "coinbase",
+        "fromnab",
+        "nab",
+        "hsbc",
+        "natwest",
+        "tsb",
+        "barclays",
+        "nationwide",
+    ]
+
+    def __call__(self, form, field):
+        if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
+            current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
+            raise ValidationError(
+                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing"
+            )
+
+
 class IsAUKMobileNumberOrShortCode:
     number_regex = re.compile(r"^[0-9\.]+$")
     mobile_regex = re.compile(r"^07[0-9]{9}$")
--- a/tests/app/main/forms/test_service_sms_senders_form.py
+++ b/tests/app/main/forms/test_service_sms_senders_form.py
@@ -34,7 +34,7 @@ from app.main.forms import ServiceSmsSenderForm
         (
             "Evri",
             True,
-            "Text message sender ID cannot be ‘Evri’ - this is to protect recipients from phishing",
+            "Text message sender ID cannot be ‘Evri’ - this is to protect recipients from phishing scams",
         ),
         pytest.param("'UC'", False, None, marks=pytest.mark.xfail),  # Apostrophes can cause SMS delivery issues
     ],
--- a/tests/app/main/forms/test_service_sms_senders_form.py
+++ b/tests/app/main/forms/test_service_sms_senders_form.py
@@ -1,3 +1,5 @@
+import logging
+
 import pytest
 
 from app.main.forms import ServiceSmsSenderForm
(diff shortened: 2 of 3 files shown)


app/main/validators.py
```diff
@@ -146,6 +146,21 @@ def __call__(self, form, field):
             raise ValidationError(self.message)
 
 
+class IsNotAPotentiallyMaliciousSenderID:
+    potentially_malicious_sender_ids = ["amazon", "evri", "lloydsbank", "coinbase", "fromnab", "nab", "hsbc", "natwest"]
```
CrystalPea, line 150: Do we want to also include other major banks? Like "tsb"?

CrystalPea (commented):
This is good to go when Karl reviews the content - code looks good.

There aren't very many senderIds in the list so far - is this on purpose? Would we want to ask MMG about the list they have, and base ours on theirs?

app/main/validators.py
risicle: I would recommend against having this list in a public repo. Makes it very easy to circumvent in one shot.

risicle:
Could this also emit a log message (at warning level?) when it's triggered?

app/main/validators.py
whpearson: That is the next step. This is just to make it a bit less easy to target people.

whpearson:
> This is good to go when Karl reviews the content - code looks good.
> 
> There aren't very many senderIds in the list so far - is this on purpose? Would we want to ask MMG about the list they have, and base ours on theirs?

That is the next bit of work in this stream.

tests/app/main/forms/test_service_sms_senders_form.py
```diff
@@ -29,6 +29,12 @@
         ("UK-GOV", False, None),  # Simple dashes are allowed
         ("UK.GOV", False, None),  # Full stops are allowed
         ("UK&GOV", False, None),  # Ampersands are allowed
+        (
+            "Evri",
+            True,
+            "Text message sender ID cannot be one used by a well known company as those are prohibited due to usage "
+            "by spam",
```
karlchillmaid, line 36: Is it possible to play back the sender ID they tried to use?

I know we have some validation errors that play back the banned characters a user entered.

If so, how about:

```Text message sender ID cannot be ‘EVRI’```

Or, if we want to provide more explanation:

```Text message sender ID cannot be ‘EVRI’ - this is to protect recipients from spam```

File: app/main/validators.py [hunk 3]
```diff
@@ -146,6 +146,29 @@ def __call__(self, form, field):
             raise ValidationError(self.message)
 
 
+class IsNotAPotentiallyMaliciousSenderID:
+    potentially_malicious_sender_ids = [
```
karlchillmaid, line 150: Do we want to add more couriers to this list?

RoyalMail
PostOffice
DPD
DPD Local
FedEx
Yodel
InPost
DHL
DHL Parcel
DXDelivery
UPS

Banks and financial companies sometimes use multiple sender IDs. For example:
HSBC UK
HSBCUK
HSBCBank
PayPal
PayPal UK
Aviva
Churchill

Will we exclude anything that includes ‘HSBC’ or do we need to add each ID?

What about shops, utilities and other companies? For example:
VirginMedia
John Lewis
Tesco

karlchillmaid: @idavidmcdonald am I right in thinking we’re considering blocking some IDs based on the organisation?

For example, only NHS organisations can use NHS in the sender ID?

Do we need to rethink the default of GOVUK that everyone gets when they first join Notify? Letting non gov senders use that sender ID (and email branding) is potentially risky. Perhaps the default should be something super-generic that obviously *needs* to be changed as part of the onboarding process?

app/main/validators.py
```diff
@@ -146,6 +146,29 @@ def __call__(self, form, field):
...
+        "tsb",
+        "barclays",
+        "nationwide",
+    ]
+
+    def __call__(self, form, field):
+        if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
+            current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
+            raise ValidationError(
+                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing"
```
karlchillmaid, line 168:
```suggestion
                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing scams"
```

tests/app/main/forms/test_service_sms_senders_form.py
```diff
@@ -29,6 +31,11 @@
         ("UK-GOV", False, None),  # Simple dashes are allowed
         ("UK.GOV", False, None),  # Full stops are allowed
         ("UK&GOV", False, None),  # Ampersands are allowed
+        (
+            "Evri",
+            True,
+            "Text message sender ID cannot be ‘Evri’ - this is to protect recipients from phishing",
```
karlchillmaid, line 37:
```suggestion
            "Text message sender ID cannot be ‘Evri’ - this is to protect recipients from phishing scams",
```

app/main/validators.py
idavidmcdonald: So for the moment, the plan is to just put in place a small list like this which will not catch much. Then going to write up a separate ticket which will be to flesh out all the names we want to block (potentially sourcing the list from BT).

And then as a different piece of work we need to figure out the problem of blocking some IDs based on the organisation... but definitely haven't worried about that yet (but we will need to shortly).

So consider this very MVP, to specifically block the names that were used in the phishing incident.

app/main/validators.py
idavidmcdonald: Is it really that much of an issue to have the list in a public repo?

If someone wants to send a malicious text message with the name 'Barclays', they can either look at our list in the public repo... or they can attempt to add the sender name which will take them about 2 seconds to attempt. 

If someone wants to know what sender names have we missed, then having this list in the public repo makes it easier. They can go ah, they have forgotten that niche building society! But I'm not sure how big an issue that is... as it is probably much harder to find a customer of that niche building society to phish. 

I'm also unsure if we have a quick and easy way of moving the list somewhere private too.

1. We can put it in the database - that means new database migration, new API route, admin calls the API, making looking into some form of caching so it doesn't have to call the API everytime. Seems a reasonable chunk of work. Not difficult, but a few steps.
2. We could put it in an environment variable. But these get passed from AWS SSM Parameter Store and the docs are vague but POTENTIALLY the the limit of the number of characters for the value is about 1000 - if each sender name is 10 characters long that is only 100 sender names we can block. This point might be worth double checking though as would be fine to be told I was wrong as this might be a reasonably quick way of doing it.
3. Do something like the digital marketplace bad words repo (if anyone remembers that) - new private repo with a list of bad sender names that gets pulled in as a package at build time. 

At the moment I lean towards keeping it public as it is a bit more work to make it private and I'm not sure really how much more protection we will get... Thoughts?

app/main/validators.py
karlchillmaid: Thanks @idavidmcdonald.

I didn’t spot that some of this had already been discussed in resolved comments.

Thanks for re-explaining!

app/main/validators.py
leohemsted, line 168: there's some rogue string concatenation going on here too btw

app/main/validators.py
```diff
@@ -146,6 +146,29 @@ def __call__(self, form, field):
...
+        "tsb",
+        "barclays",
+        "nationwide",
+    ]
+
+    def __call__(self, form, field):
+        if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
+            current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
+            raise ValidationError(
+                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients " f"from phishing scams"
```
leohemsted, line 168:
```suggestion
                f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients from phishing scams"
```

CrystalPea (approved):

PR #5064, opened 2024-04-30
Fix set-letter-branding preview javascript

kr8n3r:
As we've changed the way we serve letter branding previews from iframe to image, we need to update
the admin brand preview on `set-letter-branding` platform admin route to serve them in an image as well.

No change to email and sms brand previews

Fixes: https://trello.com/c/egKyrIg1/717-fix-js-for-platform-admin-set-letter-branding-page

As we previously did not make any aria live roles on the iframe, I'm not sure if we want them now either?

**Before**
image

**After**
Screenshot 2024-04-30 at 11 49 29

### DIFF
--- a/app/assets/javascripts/previewPane.js
+++ b/app/assets/javascripts/previewPane.js
@@ -13,7 +13,13 @@
   const $paneWrapper = $('<div class="govuk-grid-column-full"></div>');
   const $form = $('form');
   const previewType = $form.data('previewType');
-  const $previewPane = $(`<iframe src="/_${previewType}?${buildQueryString(['branding_style', branding_style])}" class="branding-preview" scrolling="no"></iframe>`);
+  const letterBrandingPreviewRootPath = `templates/${previewType}-preview-image`;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/previewPane.js
```diff
@@ -12,8 +12,14 @@
 
   const $paneWrapper = $('<div class="govuk-grid-column-full"></div>');
   const $form = $('form');
+  const letterBrandingPreviewRootPath = '/templates/letter-preview-image?';
```
tombye, line 15: Having the `?` here read a bit odd when looking at the template strings that use it are alongside those for the iframe. Just because this has it hidden in a variable but the other doesn't.

Could we do one of:
- move the `?` out of here into the template strings
- make a `emailBrandingPreviewRootPath` variable made of `_${previewType}?`

...to make it more similar?

tombye (changes requested):
This is a nice fix, particularly the use of `imagePreviewPane` and `iframePreviewPane` to differentiate. I did notice something and added a comment to help, what I found was, a readability thing.

app/assets/javascripts/previewPane.js
kr8n3r: appreciate that,  have made the changes

tombye (approved):
Thanks for making that change. It reads a lot better now.

PR #5065, opened 2024-05-03
Use hidden attribute instead of pure CSS to hide things in cookies page

tombye:
Using this instead means we make explicit what's hidden when the page loads, in the HTML. It also means users won't see things they shouldn't if the CSS doesn't load, which is often raised as a requirement in accessibility accessments.

Having state represented in the HTML also makes it easier for our Python tests to know what's happening, because they often make assertions based on what's in the HTML of a returned page.


kr8n3r (approved):
🙌

PR #5068, opened 2024-05-07
Fix problem with skiplinks adding hashes to url

tombye:
Changes to all our forms, to prevent them sending requests to URLs with a fragment identifier (`#`) in.

This is to fix an accessibility issue where the `#` for the main content area (shared by all pages) forces the browser to jump straight there when the page loads

https://trello.com/c/bnjIRzze/695-fix-accessibility-issue-with-post-redirect-get-form-submission-pattern

## Rationale

There are 3 ways this could have been handled:
1. setting `action` to a blank `#` (initially proposed in https://github.com/alphagov/notifications-admin/pull/5042)
2. setting `action` to the URL of the current page, without any `#`s
3. using the [Content-Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header to set the next URL to not include the `#`

Way 1. is a bit mucky, because of it changing the browser history, and 3. involves messing with how we do redirects, which seemed a bit complex and I wasn't confident I understood the impact well enough. So I chose 2.

## Notes on implementation

By default, setting `action` to `{{ request.url }}`, it ends up as the whole URL, including the protocol and domain. Like:

`<form action="https://www.notify.works/email-branding/create">`

instead of:

`<form action="/email-branding/create">`

This works, and I think wouldn't introduce any problems with the app running on different domains, but is a bit yuck.

Instead, I added a `extract_path_from_url` template filter to remove the protocol and hostname. I'm not sure if that introduces too much complexity or performance cost so put it in a commit (the last one) for easy removal if so.


kr8n3r (approved):
tested locally, strips out url hashes as promised.

Confused about `format_url_as_absolute_path` name. If i understand correctly request.url returns absolute url and this filter extracts a relative path, right?
if we just want the [path](https://docs.python.org/3/library/urllib.parse.html#url-parsing), could we do
```
from urllib.parse import urlparse

return urlparse(url).path
```

tombye:
> tested locally, strips out url hashes as promised.
> 
> Confused about `format_url_as_absolute_path` name. If i understand correctly request.url returns absolute url and this filter extracts a relative path, right? if we just want the [path](https://docs.python.org/3/library/urllib.parse.html#url-parsing), could we do
> 
> ```
> from urllib.parse import urlparse
> 
> return urlparse(url).path
> ```

I took the term 'absolute path' from a spec for URIs, which is probably old enough now to be less relevant, or at least readable now. The filtering I'm doing is intended not to just produce the path of URLs but also their query params. I took the route to that I thought was safer: removing the protocol and hostname instead of composing the URL myself.

We had a chat about this outside of this PR and Jani was happy with the filtering approach. I also for consensus on the name of the template filter and we came up with `extract_path_from_url` as the closest in meaning to what it does so I changed the last commit.

kr8n3r (approved):
I'm happy with this :)

PR #5069, opened 2024-05-07
Make content changes to join service journey

quis:
These are the straightforward changes from this doc: https://docs.google.com/document/d/15jG6UW7aoCQ15E8uTVfpMG_lEYtCbJueEg-ww2O-6Nc/edit?pli=1

We’re changing the wording here based on things we observed in user research.


app/templates/views/invite-user.html
```diff
@@ -5,7 +5,7 @@
 {% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
 
 {% block service_page_title %}
-  Invite a team member
+  {{ 'Let someone join your service' if user_to_invite else 'Invite a team member' }}
```
tombye, line 8: Can this be the same as the h1?

tombye (changes requested):
Only 1 thing I spotted, otherwise all good once that's looked at 👍🏻

app/templates/views/invite-user.html
quis: I think the reason this didn’t match the `<h1>` before is so we don’t put people’s names (PII) in the page title, because page titles go into Google Analytics, browser history, and maybe some other places

tombye: Another reason to either remove GA or exercise better control over what we send it. Happy for this not to block these changes going out though.

tombye (approved):
Only request for change now discussed and resolved so looks good to me 👍🏻

PR #5075, opened 2024-05-09
Remove environment knowledge from application.

corlettb:
What
----

Remove environment knowledge from application. Environment configuration will be passed in during deployments.

Replace environment name checks with feature flags.

Why
----

We are updating our notify deployment model to allow for many more environments. As such environment settings will be kept in as few places as possible (eventually just one place).

The main benefit of this is to be able to add new environments without having to modify source code


app/config.py
```diff
@@ -58,23 +60,31 @@ class Config:
...
+    S3_BUCKET_MOU = os.environ.get("S3_BUCKET_MOU", "local-mou")
+    S3_BUCKET_TRANSIENT_UPLOADED_LETTERS = os.environ.get(
+        "S3_BUCKET_TRANSIENT_UPLOADED_LETTERS", "local-transient-uploaded-letters"
+    )
+    S3_BUCKET_PRECOMPILED_ORIGINALS_BACKUP_LETTERS = os.environ.get(
+        "S3_BUCKET_PRECOMPILED_ORIGINALS_BACKUP_LETTERS", "local-precompiled-originals-backup-letters"
+    )
+    S3_BUCKET_LETTER_ATTACHMENTS = os.environ.get("S3_BUCKET_LETTER_ATTACHMENTS", "local-letter-attachments")
+    LOGO_CDN_DOMAIN = os.environ.get("LOGO_CDN_DOMAIN")
+    ROUTE_SECRET_KEY_1 = os.environ.get("ROUTE_SECRET_KEY_1", "")
```
klssmith, line 78: `ROUTE_SECRET_KEY_1`, `ROUTE_SECRET_KEY_2` and `CHECK_PROXY_HEADER` aren't needed since [this change](https://github.com/alphagov/notifications-admin/pull/5063), so can be removed

corlettb: Updated and rebased.

app/config.py
```diff
@@ -58,23 +60,28 @@ class Config:
...
+    S3_BUCKET_LOGO_UPLOAD = os.environ.get("S3_BUCKET_LOGO_UPLOAD", "public-logos-local")
+    S3_BUCKET_MOU = os.environ.get("S3_BUCKET_MOU", "local-mou")
+    S3_BUCKET_TRANSIENT_UPLOADED_LETTERS = os.environ.get(
+        "S3_BUCKET_TRANSIENT_UPLOADED_LETTERS", "local-transient-uploaded-letters"
+    )
+    S3_BUCKET_PRECOMPILED_ORIGINALS_BACKUP_LETTERS = os.environ.get(
+        "S3_BUCKET_PRECOMPILED_ORIGINALS_BACKUP_LETTERS", "local-precompiled-originals-backup-letters"
+    )
+    S3_BUCKET_LETTER_ATTACHMENTS = os.environ.get("S3_BUCKET_LETTER_ATTACHMENTS", "local-letter-attachments")
+    LOGO_CDN_DOMAIN = os.environ.get("LOGO_CDN_DOMAIN")
```
risicle, line 77: No sensible default? `static-logos.notify.tools` like in `Development`? Otherwise this is a mandatory env var we're adding.

risicle:
Actually I think my comment does need to be resolved but the rest gets my approval.

app/config.py
corlettb: Added default.

risicle (approved):

PR #5076, opened 2024-05-10
Save selection when going back to first page in email or letter branding journey

klssmith:
https://trello.com/c/3hU3e6ls/709-going-back-to-first-page-in-email-or-letter-branding-journeys-doesnt-save-your-selection

This change ensures that if you click "back" to return to the first page of the email or letter branding journey, your choice is still selected.

To do this, we ensure that the `branding_choice` query parameter is always passed when leaving or returning to the first page in the journey so that it can be used to pre-populate the form.

### DIFF
--- a/app/main/views/service_settings/branding.py
+++ b/app/main/views/service_settings/branding.py
@@ -82,6 +82,7 @@ def email_branding_options(service_id):
                     ".branding_option_preview",
                     service_id=current_service.id,
                     branding_option=branding_choice,
+                    branding_choice=branding_choice,
                     branding_type="email",
                 )
             )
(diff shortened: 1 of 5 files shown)


app/main/views/service_settings/branding.py
```diff
@@ -71,17 +72,21 @@ def email_branding_options(service_id):
...
+                url_for(".email_branding_govuk", service_id=current_service.id, branding_choice=branding_choice)
+            )
 
         elif branding_choice in current_service.email_branding_pool.ids:
             return redirect(
                 url_for(
                     ".branding_option_preview",
                     service_id=current_service.id,
                     branding_option=branding_choice,
+                    branding_choice=branding_choice,
```
klssmith, line 89: `branding_option` will be removed in a second PR to avoid breaking things for people using the branding flow as this is deployed.

kr8n3r (approved):
tested locally, option on the first page is saved as required

PR #5079, opened 2024-05-15
Remove unused branding query param

klssmith:
In order to rename the branding query parameter `branding_option` to `branding_choice`, we first added `branding_choice` which had the same value of `branding_option` and wrote the code to check both https://github.com/alphagov/notifications-admin/pull/5076. This was done so that the existing journey didn't break if someone was in the middle of using it when the change was deployed.

Now that the change has been deployed for some days, `branding_option` should not be being used so can be removed.

### DIFF
--- a/app/main/views/service_settings/branding.py
+++ b/app/main/views/service_settings/branding.py
@@ -85,7 +85,6 @@ def email_branding_options(service_id):
                 url_for(
                     ".branding_option_preview",
                     service_id=current_service.id,
-                    branding_option=branding_choice,
                     branding_choice=branding_choice,
                     branding_type="email",
                 )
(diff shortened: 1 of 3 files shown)


idavidmcdonald (approved):
Seems good.

I did wonder if we should also change the method names from `branding_option_preview` to `branding_choice_preview` but assume it is fine/preferable to leave as is?

klssmith:
I think trying to change `option` to `choice` everywhere would be a bigger task because it would involve renaming views and html files etc. so I'll probably leave that. It was necessary to change the name in the query parameters because it means we can then use the helper methods which were created - they expect the name to be `branding_choice`

PR #5081, opened 2024-05-16
Add more allowed domain names to the list of who can use Slack

karlchillmaid:


app/templates/views/support/index.html
```diff
@@ -24,7 +24,7 @@
     {{ page_footer('Continue') }}
   {% endcall %}
 
-  <p class="govuk-body">If you have a .gov.uk, .gov.scot or .gov.wales email address, you can also <a class="govuk-link govuk-link--no-visited-state" href="https://ukgovernmentdigital.slack.com/messages/C0E1ADVPC">contact us on Slack</a>.</p>
+  <p class="govuk-body">If you have a .gov.uk, .gov.scot, .gov.wales, .police.uk or .mod.uk email address, you can also <a class="govuk-link govuk-link--no-visited-state" href="https://ukgovernmentdigital.slack.com/messages/C0E1ADVPC">contact us on Slack</a>.</p>
```
karlchillmaid, line 27: The list is looking a bit long now. I’m reluctant to turn it into a bullet point list as I think that makes it too prominent.

We could do this instead, if we think .gov.uk .gov.scot and .gov.wales users will understand what we mean.

```suggestion
  <p class="govuk-body">If you have a .gov, .police.uk or .mod.uk email address, you can also <a class="govuk-link govuk-link--no-visited-state" href="https://ukgovernmentdigital.slack.com/messages/C0E1ADVPC">contact us on Slack</a>.</p>
```

idavidmcdonald (approved):
I think I lean towards your original suggestion but any would be fine. Your original suggestion is more explicit which I think is useful when talking about precise things like domains

app/templates/views/support/index.html
CrystalPea: I'm not sure about this, .gov is an actual domain, and it's for US government 😓

karlchillmaid:
Good catch @CrystalPea!

I think we’ll go with the more explicit (longer) list.

PR #5086, opened 2024-05-17
removed sign in issues from emergencies on support page

rparke:
Removed the list element "nobody on your team can sign in" as an emergency from the support page


idavidmcdonald (approved):
Technically, looks all good. You might want to double check with Saima from a content perspective that we are still OK to have just a 2 item list given it doesn't have many items. I think it is probably OK but no harm asking her to double check

rparke:
Okay, I'll check in with Saima on Monday before merging.

PR #5087, opened 2024-05-20
Raise a zendesk ticket if a forbidden senderID is set

whpearson:
In order to be alerted to people potentially trying to send a phishing email or other spam email, send a zendesk ticket.

This only triggers with small list of SenderIDS there is a story in the backlog to extend this.

The zendesk email looks like 
Screenshot 2024-05-20 at 10 58 12

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -1,7 +1,9 @@
 import re
 from abc import ABC, abstractmethod
 
-from flask import current_app
+from flask import current_app, render_template
+from flask_login import current_user
+from notifications_utils.clients.zendesk.zendesk_client import NotifySupportTicket, NotifyTicketType
 from notifications_utils.field import Field
 from notifications_utils.formatters import formatted_list
 from notifications_utils.recipients import InvalidEmailError, validate_email_address
@@ -11,7 +13,7 @@ from wtforms import ValidationError
 from wtforms.validators import URL, DataRequired, InputRequired, StopValidation
 from wtforms.validators import Length as WTFormsLength
 
-from app import antivirus_client
+from app import antivirus_client, current_service, zendesk_client
 from app.formatters import sentence_case
 from app.main._commonly_used_passwords import commonly_used_passwords
 from app.models.spreadsheet import Spreadsheet
@@ -146,6 +148,25 @@ class IsNotAGenericSenderID:
             raise ValidationError(self.message)
 
 
+def create_phishing_senderid_zendesk_ticket(senderID=None):
+    ticket_message = render_template(
+        "support-tickets/phishing-senderid.txt",
+        senderID=senderID,
+    )
+    ticket = NotifySupportTicket(
+        subject=f"Possible Phishing sender ID - {current_service.name}",
+        message=ticket_message,
+        ticket_type=NotifySupportTicket.TYPE_INCIDENT,
+        notify_ticket_type=NotifyTicketType.NON_TECHNICAL,
+        user_name=current_user.name,
+        user_email=current_user.email_address,
+        org_id=current_service.organisation_id,
+        org_type=current_service.organisation_type,
+        service_id=current_service.id,
+    )
+    zendesk_client.send_ticket_to_zendesk(ticket)
+
+
 class IsNotAPotentiallyMaliciousSenderID:
     potentially_malicious_sender_ids = [
         "amazon",
@@ -163,6 +184,7 @@ class IsNotAPotentiallyMaliciousSenderID:
 
     def __call__(self, form, field):
         if field.data and field.data.lower() in self.potentially_malicious_sender_ids:
+            create_phishing_senderid_zendesk_ticket(senderID=field.data)
             current_app.logger.warning("User tried to set sender id to potentially malicious one: %s", field.data)
             raise ValidationError(
                 f"Text message sender ID cannot be ‘{field.data}’ - this is to protect recipients from phishing scams"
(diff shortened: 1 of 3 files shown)


whpearson:
Oh I'll add a test here too.

app/templates/support-tickets/phishing-senderid.txt
```diff
@@ -0,0 +1,12 @@
+Organisation: {% if current_service.organisation -%}
+  {{ current_service.organisation.name }}
+{%- else -%}
+  Can’t tell (domain is {{ current_user.email_domain }})
+{%- endif %}
+Service: {{ current_service.name }}
+{{ url_for('main.service_dashboard', service_id=current_service.id, _external=True) }}
+
+---
+The user asked for a sender ID that is potentially used for phishing - {{ senderID }}. It has been blocked.
```
idavidmcdonald, line 10: In your template above, we don't actually know which user asked for the sender ID (we only know the service). It would be useful to add this please.

app/templates/support-tickets/phishing-senderid.txt
```diff
@@ -0,0 +1,12 @@
...
+
+Check whether this is just testing, if not raise a security incident.
```
idavidmcdonald, line 12: Optional but encouraged: I'd love to flesh this out a bit more about what we should do (although I appreciate we haven't decided exactly what we should do but very open to you making some more detailed suggestions). You could also create a runbook entry in our team wiki and link to that from here (we have that pattern for some of our other zendesk tickets)

whpearson:
I've updated the template. Example zendesk ticket is here

https://govuk.zendesk.com/agent/tickets/5832195

app/templates/support-tickets/phishing-senderid.txt
whpearson: Done :)

app/main/validators.py
```diff
@@ -146,6 +148,25 @@ def __call__(self, form, field):
...
+    )
+    ticket = NotifySupportTicket(
+        subject=f"Possible Phishing sender ID - {current_service.name}",
+        message=ticket_message,
+        ticket_type=NotifySupportTicket.TYPE_INCIDENT,
+        notify_ticket_type=NotifyTicketType.NON_TECHNICAL,
+        user_name=current_user.name,
+        user_email=current_user.email_address,
+        org_id=current_service.organisation_id,
+        org_type=current_service.organisation_type,
```
idavidmcdonald, line 164: I think we could probably drop the org type and ID. I don't think these would be too helpful for investigating. I think the main things are the user and the service.

app/main/validators.py
```diff
@@ -146,6 +148,25 @@ def __call__(self, form, field):
...
+def create_phishing_senderid_zendesk_ticket(senderID=None):
+    ticket_message = render_template(
+        "support-tickets/phishing-senderid.txt",
+        senderID=senderID,
+    )
+    ticket = NotifySupportTicket(
+        subject=f"Possible Phishing sender ID - {current_service.name}",
+        message=ticket_message,
+        ticket_type=NotifySupportTicket.TYPE_INCIDENT,
+        notify_ticket_type=NotifyTicketType.NON_TECHNICAL,
```
idavidmcdonald, line 160: Optional: I think we should probably make this a technical ticket because I think devs might be best placed to investigate. They can more easily do a query of all notifications in the DB sent by that service and see what sender names have been used for a service, for example. Although it doesn't need to be a technical person to start looking at the ticket... but I suspect might end up needing a dev if we investigate further at all...

I probably lean 75% towards making it technical. What do you think?

app/templates/support-tickets/phishing-senderid.txt
```diff
@@ -0,0 +1,19 @@
...
+---
+The user - {{current_user.email_address}} asked for a sender ID that is potentially used for phishing - {{ senderID }}. It has been blocked.
+
+Check whether this is just testing, if not raise a security incident:
+
+Security Incident runbook: https://github.com/alphagov/notifications-manuals/wiki/What-to-do-during-an-incident#security-incidents-and-data-breaches'
+
+To determine the extent of the problem:
+ - look at sender ids both current and archived : https://github.com/alphagov/notifications-manuals/wiki/Support-Runbook#look-at-senderid-history
+ - look at emails: {{url_for("main.view_notifications", service_id=current_service.id, message_type='email', status='delivered', _external=True)}}
```
idavidmcdonald, line 18: I don't think there is much need to look at emails, I wouldn't have thought given bad sender IDs we are worried about are only for text messages. I reckon you could get rid of this line.

idavidmcdonald (commented):
Looking pretty good. Some minor comments about what goes in the zendesk ticket, but code wise all goods already!

app/templates/support-tickets/phishing-senderid.txt
whpearson: I thought we should look for malicious activity generally, for reporting to NCSC and looking at scope of the incident. I  was unsure whether looking at letters was worth while, being slower and maybe not ncsc's remit. But thinking about it more I'm not sure I'd want to miss malicious letters because there wasn't a link. What do you think?

app/main/validators.py
whpearson: It was meant to be technical :+1:

app/main/validators.py
idavidmcdonald: You'll also want to remove the variables you are passing through into the template here (you have removed them from the template already)

app/templates/support-tickets/phishing-senderid.txt
idavidmcdonald: Maybe let's leave email and text messaging here. I reckon we would end up having a good look around their service anyway so think we would have a glance at letter naturally if we see anything else suspicious looking

idavidmcdonald (approved):
Nice work!

PR #5088, opened 2024-05-20
Hide the edit links for archived letters

klssmith:
We were showing all the edit links on archived letter templates. This meant that people could go through the flow of sending a letter, only to get a `500` error code when clicking the Send button (when api does some checks and determines that the template has been archived).

It's easy to accidentally try and send an archived letter, because if it has been used and archived recently, it will be linked to from the service dashboard under the recently used templates section. This change hides all the edit buttons for archived letters.


whpearson (approved):
Change looks good. Makes sense

PR #5092, opened 2024-05-21
Update unsubscribe-links.html

saimaghafoor:
New content  added for 'one-click unsubscribe link in the header of an email'.  Signpost users to API documentation for more information.

Removed reference to google and yahoo

Story: https://trello.com/c/WoTzSW56

image
image

### DIFF
--- a/app/templates/views/guidance/using-notify/unsubscribe-links.html
+++ b/app/templates/views/guidance/using-notify/unsubscribe-links.html
@@ -15,9 +15,9 @@
   <p class="govuk-body">If you do not, recipients can report your emails as spam. If this happens we may stop you from using GOV.UK Notify.</p>
   <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>
 
-  <p class="govuk-body">There are 2 ways to let recipients unsubscribe from your emails:</p>
+  <p class="govuk-body">There are 2 ways to let recipients opt out</p>
   <ol class="govuk-list govuk-list--number">
-    <li>Add a link to the body of your email</li>
+    <li>Add an unsubscribe link to the body of your email.</li>
     <li>Add a one-click unsubscribe button to the email header.</li>
   </ol>
   <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
--- a/app/templates/views/guidance/using-notify/unsubscribe-links.html
+++ b/app/templates/views/guidance/using-notify/unsubscribe-links.html
@@ -20,7 +20,7 @@
   <p class="govuk-body">There are 2 ways to let recipients opt out:</p>
   <ol class="govuk-list govuk-list--number">
     <li>Add an unsubscribe link to the body of your email.</li>
-    <li>Add a one-click unsubscribe button to the email header.</li>
+    <li>Add one-click unsubscribe to the email header.</li>
   </ol>
 
   <h2 class="heading-medium">Add an unsubscribe link to the email body</h2>
@@ -35,8 +35,8 @@
   <pre class="formatting-example"><code class="lang-md">[Unsubscribe](mailto:example@gov.uk?subject=unsubscribe)</code></pre>
   <p class="govuk-body">The email address should be a shared inbox managed by your team, not your own email address.</p>
 
-  <h2 class="heading-medium">Add a one-click unsubscribe button to the email header</h2>
+  <h2 class="heading-medium">Add one-click unsubscribe to the email header</h2>
   <p class="govuk-body">This is currently an API-only feature. You’ll need a developer on your team to set it up for you.</p>
-  <p class="govuk-body">To add a one-click unsubscribe button, follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
+  <p class="govuk-body">To add one-click unsubscribe, follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
     href="https://www.notifications.service.gov.uk/using-notify/api-documentation">API documentation</a>.</p>
 {% endblock %}
--- a/app/templates/views/guidance/using-notify/unsubscribe-links.html
+++ b/app/templates/views/guidance/using-notify/unsubscribe-links.html
@@ -17,7 +17,7 @@
    <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
     href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#transactional-messages">transactional and subscription messages</a>.</p>
 
-  <p class="govuk-body">There are 2 ways to let recipients opt out</p>
+  <p class="govuk-body">There are 2 ways to let recipients opt out:</p>
   <ol class="govuk-list govuk-list--number">
     <li>Add an unsubscribe link to the body of your email.</li>
     <li>Add a one-click unsubscribe button to the email header.</li>
--- a/app/templates/views/guidance/using-notify/unsubscribe-links.html
+++ b/app/templates/views/guidance/using-notify/unsubscribe-links.html
@@ -11,9 +11,10 @@
 
   <h1 class="heading-large">Unsubscribe links</h1>
 
-  <p class="govuk-body">You must let recipients opt out of subscription emails.</p>
+  <p class="govuk-body">You must let people opt out of receiving subscription emails.</p>
+  <p class="govuk-body">If you do not, recipients can report your emails as spam. If this happens we may stop you from using GOV.UK Notify.</p>
   <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,21 @@
...
 
-  <p class="govuk-body">Subscription emails must include an unsubscribe link in the body of the message.</p>
-  <p class="govuk-body">If they do not:</p>
-  <ul class="govuk-list govuk-list--bullet">
-    <li>recipients could report your emails as spam</li>
-    <li>email providers may refuse to deliver your messages</li>
-  </ul>
+  <p class="govuk-body">You must let recipients opt out of subscription emails.</p>
   <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>
+  <p class="govuk-body">If recipients report your emails as spam we can stop you from using GOV.UK Notify.</p>
```
karlchillmaid, line 16:
```suggestion
  <p class="govuk-body">You must let recipients opt out of subscription emails.</p>
  <p class="govuk-body">If you do not, they can report your emails as spam. If this happens we may stop you from using GOV.UK Notify.</p>
  <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>

```

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,21 @@
...
+  <p class="govuk-body">There are 2 ways to let recipients unsubscribe from your emails:</p>
+  <ol class="govuk-list govuk-list--number">
+    <li>Add a link to the body of your email</li>
+    <li>Add a one-click unsubscribe button to the email header.</li>
+  </ol>
```
karlchillmaid, line 21:
```suggestion
  <p class="govuk-body">There are 2 ways to let recipients opt out</p>
  <ol class="govuk-list govuk-list--number">
    <li>Add an unsubscribe link to the body of your email</li>
    <li>Add a one-click unsubscribe button to the email header.</li>
  </ol>
```

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,21 @@
...
   <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
```
karlchillmaid, line 22: This line feels like it belongs with ‘Transactional emails do not need to include an unsubscribe link.’ – but it also feels less important than the 2 bullet points.

If you can ask a dev to attach screenshots to the PR, seeing how the page will look might help with the final decision.

saimaghafoor:
@BlessedDev could we have a preview of the content please if possible?

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,21 @@
 
   <h1 class="heading-large">Unsubscribe links</h1>
 
-  <p class="govuk-body">Subscription emails must include an unsubscribe link in the body of the message.</p>
-  <p class="govuk-body">If they do not:</p>
-  <ul class="govuk-list govuk-list--bullet">
-    <li>recipients could report your emails as spam</li>
-    <li>email providers may refuse to deliver your messages</li>
-  </ul>
+  <p class="govuk-body">You must let recipients opt out of subscription emails.</p>
```
karlchillmaid, line 14:
```suggestion
  <p class="govuk-body">You must let people opt out of receiving subscription emails.</p>
```

Just putting this in as something to consider. It feels right to qualify what we mean by ‘opt out’, but we probably don’t want recipients and receiving together in the same sentence.

app/templates/views/guidance/using-notify/unsubscribe-links.html
karlchillmaid: The more I re-read this, the more I think ‘move it above the bullets’... 🤔

saimaghafoor: Yes i think it sits better above the bullets when we're referencing subscription/transactional emails

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,22 @@
...
-  </ul>
+  <p class="govuk-body">You must let people opt out of receiving subscription emails.</p>
+  <p class="govuk-body">If you do not, recipients can report your emails as spam. If this happens we may stop you from using GOV.UK Notify.</p>
   <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>
-  <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
+   <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
     href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#transactional-messages">transactional and subscription messages</a>.</p>
 
-  <h2 class="heading-medium">Add your unsubscribe links by 1 April 2024</h2>
+  <p class="govuk-body">There are 2 ways to let recipients opt out</p>
```
karlchillmaid, line 20:
```suggestion
  <p class="govuk-body">There are 2 ways to let recipients opt out:</p>
```

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -41,4 +35,8 @@ <h2 class="heading-medium">How to add an unsubscribe link to an email</h2>
   <pre class="formatting-example"><code class="lang-md">[Unsubscribe](mailto:example@gov.uk?subject=unsubscribe)</code></pre>
   <p class="govuk-body">The email address should be a shared inbox managed by your team, not your own email address.</p>
 
+  <h2 class="heading-medium">Add a one-click unsubscribe button to the email header</h2>
```
BlessedDev, line 38: Technically it could be presented as a button or a link depending on the email client, so it should be along the lines of `Add a one-click unsubscribe link to appear in the email header` in order to reduce confusion.

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -41,4 +35,8 @@ <h2 class="heading-medium">How to add an unsubscribe link to an email</h2>
...
+  <p class="govuk-body">This is currently an API-only feature. You’ll need a developer on your team to set it up for you.</p>
+  <p class="govuk-body">To add a one-click unsubscribe button, follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
```
BlessedDev, line 40: I would suggest replacing `button` with `link`

app/templates/views/guidance/using-notify/unsubscribe-links.html
```diff
@@ -11,28 +11,22 @@
...
   <p class="govuk-body">Transactional emails do not need to include an unsubscribe link.</p>
-  <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
+   <p class="govuk-body">The GOV.​UK Service Manual explains the difference between <a class="govuk-link govuk-link--no-visited-state"
     href="https://www.gov.uk/service-manual/design/sending-emails-and-text-messages#transactional-messages">transactional and subscription messages</a>.</p>
 
-  <h2 class="heading-medium">Add your unsubscribe links by 1 April 2024</h2>
+  <p class="govuk-body">There are 2 ways to let recipients opt out:</p>
+  <ol class="govuk-list govuk-list--number">
+    <li>Add an unsubscribe link to the body of your email.</li>
+    <li>Add a one-click unsubscribe button to the email header.</li>
```
BlessedDev, line 23: Technically it could be presented as a button or a link depending on the email client, so it should be along the lines of `Add a one-click unsubscribe link to appear in the email header` in order to reduce confusion.

BlessedDev (changes requested):
Well done on the content, it is very succinct. There is a lot of to be worked out regarding unsubscribe links and I think we should say as little as possible while giving as much helpful information as possible, and this PR lays a solid foundation and strikes the right balance. 
I have asked for a minor change to use the term `link` instead of `button`. This is because the behaviour of email clients is not uniform and the link may be presented as  a button or a link or something else entirely.

app/templates/views/guidance/using-notify/unsubscribe-links.html
saimaghafoor: Thanks Gozie. This is interesting. If it'll appear as either a button or link then perhaps we can refer to it as just 'one-click unsubscribe' that appears in the email header. 

I've checked Google and they refer to it as one-click unsubscribe

saimaghafoor:
Tweaked content to remove references to a button. Used 'one-click unsubscribe' (which Google also use)

BlessedDev (approved):

PR #5095, opened 2024-05-22
Add announcement of 1 July price increase

karlchillmaid:
We expect Royal Mail to increase their prices on 1 July.

We don’t know any details yet, but we need to announce this to our users as early as possible.

This PR adds an announcement to the top of the page. We need to deploy this before we send an email about the price change to our users.

### DIFF
--- a/app/templates/views/guidance/pricing/letter-pricing.html
+++ b/app/templates/views/guidance/pricing/letter-pricing.html
@@ -21,7 +21,7 @@
     ) }}
 
   <div class="govuk-inset-text">
-    Letter prices will go up on 1 July 2024. This is because Royal Mail is increasing its rates.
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/pricing/letter-pricing.html
```diff
@@ -20,6 +20,10 @@ <h1 class="heading-large">Letter pricing</h1>
       }
     ) }}
 
+  <div class="govuk-inset-text">
+    Letter prices will go up on 1 July 2024.
```
karlchillmaid, line 24:
```suggestion
    Letter prices will go up on 1 July 2024. This is because Royal Mail is increasing its rates.
```

kr8n3r (approved):

PR #5099, opened 2024-05-30
Remove `process_type` from API client

quis:
This is cleaning up a property we have removed. Follow up to
- https://github.com/alphagov/notifications-admin/pull/4567
- https://github.com/alphagov/notifications-api/pull/3711


idavidmcdonald (approved):
I think this shouldn't have any unforeseen impact.

These look like the final uses of `process_type` in the admin repo too which is nice.

I think we may be able to remove `process_type` from some places in our API code too (without having to drop the database column, which is a bit scarier). I'll let you decide if you investigate doing that too or not.

PR #5102, opened 2024-05-31
Make callout to callback documentation consistent

idavidmcdonald:
The callout for delivery status callbacks and inbound text message callbacks was different, for no obvious reason. I've made these consistent, picking what I thought is the better and clearer message of the 2 ways we do it.

Will need review of a content designer please.

## Before

image
image

## After
image


karlchillmaid (approved):
Content looks 👍

PR #5103, opened 2024-05-31
Update code to use new Zendesk form

klssmith:
Every ticket type is now a "task" and should have the `notify_task_type`
drop down menu populated instead of the `ticket_categories` dropdown,
which has been removed.

The exception to this is the tickets that get opened using the support
form - these haven't changed.

Needs https://github.com/alphagov/notifications-utils/pull/1121 to be merged first


whpearson (approved):
Changes look good for once the utils is merged and bumped

PR #5104, opened 2024-06-04
Remove redundant section from accessibility statement

tombye:
I removed the last issue from this seciton in https://github.com/alphagov/notifications-admin/pull/5096 but didn't delete the section. This does that.

Example of what it looked like before can be seen [in this version of the statement before that PR](https://github.com/alphagov/notifications-admin/blob/927672d6780faf37bcaaeaa6957b3369ba76f5b5/app/templates/views/accessibility_statement.html#L101)

### DIFF
--- a/app/templates/views/accessibility_statement.html
+++ b/app/templates/views/accessibility_statement.html
@@ -16,7 +16,7 @@
       "Next review due": "25 June 2024"
     }
     ) }}
-    {# Last non-functional changes: 21 Feb 2024 #}
+    {# Last non-functional changes: 4 June 2024 #}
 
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):

kr8n3r (approved):

tombye:
Sorry @kr8n3r but I originally pushed a change that included bumping the 'last updated' date. That was wrong, because this change is just a fix rather than a real update to the content. Not great as it would make using 'last updated' as a log of actual changes to our accessibility status harder.

I rewrote the commit to instead log this as a functional change, apologies for the extra work to re-approve.

PR #5106, opened 2024-06-07
Remove unreachable code

quis:
> This PR still requires some frontend attention. Will work with Chris to update the templates.

https://github.com/alphagov/notifications-admin/commit/c858869a523349c6eeecc8a157948b3badea2deb (8 years ago)


klssmith (approved):
Better late than never 😄

PR #5107, opened 2024-06-07
Add a warning to support tickets from non-authenticated users

klssmith:
This adds a note to the start of support tickets created when a user is logged out of Notify. When someone is logged out, they can enter any email address so this helps make it clear that the user hasn't been verified.

---
Tickets created when not logged in look like this:
Screenshot 2024-06-07 at 15 30 00

### DIFF
--- a/app/templates/support-tickets/support-ticket.txt
+++ b/app/templates/support-tickets/support-ticket.txt
@@ -1,3 +1,6 @@
+{% if not current_user.is_authenticated -%}
+**This ticket was opened using the public support form and not by a logged in user**
+{% endif %}
 {{ content }}
 {% if current_service -%}
 Service: "{{ current_service.name }}"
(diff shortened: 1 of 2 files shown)


app/templates/support-tickets/support-ticket.txt
```diff
@@ -1,3 +1,6 @@
+{% if not current_user.is_authenticated -%}
+**This ticket was opened using the public support form and not by a logged in user**
+{% endif %}
```
leohemsted, line 3: just as a note, this'll be visible to the user in email correspondance, right?

it's probably okay in this case, though we'd need to be careful with the exact content in case people get concerned. i guess if we don't want this, the alternative would be to modify the `zendesk_api_client` to have the ability to add an internal note to a ticket, then we could create the ticket then immediately add a note.

klssmith: Yeah, I'll close this for now because I forgot that it will be visible to everyone. We can then rethink how to do this

PR #5109, opened 2024-06-10
Bump utils

leohemsted:
Brings in breaking changes to the recipient validation structure

Also brings in a smaller change to the redis client that nonetheless required a lot of unit test changes.

### DIFF
--- a/tests/app/notify_client/test_service_api_client.py
+++ b/tests/app/notify_client/test_service_api_client.py
@@ -6,13 +6,16 @@ import pytest
 from app import invite_api_client, service_api_client, user_api_client
 from app.notify_client.service_api_client import ServiceAPIClient
 from tests.conftest import SERVICE_ONE_ID
+from tests.utils import RedisClientMock
 
 FAKE_TEMPLATE_ID = uuid4()
 
 
 def test_client_posts_archived_true_when_deleting_template(mocker):
     mocker.patch("app.notify_client.current_user", id="1")
-    mock_redis_delete_by_pattern = mocker.patch("app.extensions.RedisClient.delete_by_pattern")
+    mock_redis_delete_by_pattern = mocker.patch(
+        "app.extensions.RedisClient.delete_by_pattern", new_callable=RedisClientMock
+    )
     expected_data = {"archived": True, "created_by": "1"}
     expected_url = f"/service/{SERVICE_ONE_ID}/template/{FAKE_TEMPLATE_ID}"
 
@@ -25,7 +28,7 @@ def test_client_posts_archived_true_when_deleting_template(mocker):
 
     client.delete_service_template(SERVICE_ONE_ID, FAKE_TEMPLATE_ID)
     mock_post.assert_called_once_with(expected_url, data=expected_data)
-    assert call(f"service-{SERVICE_ONE_ID}-template-{FAKE_TEMPLATE_ID}*") in mock_redis_delete_by_pattern.call_args_list
+    mock_redis_delete_by_pattern.assert_called_with_args(f"service-{SERVICE_ONE_ID}-template-{FAKE_TEMPLATE_ID}*")
 
 
 def test_client_gets_service(mocker):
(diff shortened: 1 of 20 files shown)


tests/utils.py
```diff
@@ -37,3 +38,16 @@ def check_render_template_forms(calls: list[mock.call]):
...
+    if not any(call[0][: len(args)] == args for call in mocked.call_args_list):
+        err_msg = "\n".join(
+            (
+                f"call{args} not found in Mock {mocked}",
+                "  call_args_list: [",
+                *[f"    {c}" for c in mocked.call_args_list],
+                "  ]",
+            )
+        )
+        pytest.fail(err_msg)
```
leohemsted, line 53: sample err msg:
```
>           pytest.fail(err_msg)
E           Failed: call(call('organisations'),) not found in Mock <MagicMock name='delete' id='4756761040'>
E             call_args_list: [
E               call('domains', raise_exception=True)
E               call('organisations', raise_exception=True)
E               call('organisations', raise_exception=True)
E               call('domains', raise_exception=True)
E             ]
```

idavidmcdonald (dismissed):
Thanks Leo.

So I'm happy to approve this.

However, I'm not entirely happy with the weakening of the test assertions for the redis deletion... I would be tempted instead to change the redis client to have `raise_exception=True` as it's default for the `delete` method (because I believe any failed deleted has the potential to end us with stale data). This would then enable us to keep our test assertions a little bit stricter with

```
assert call("organisations") in redis_delete_mock.call_args_list
assert call(f'organisation-{organisation_one["id"]}-name') in redis_delete_mock.call_args_list
assert call("domains") in redis_delete_mock.call_args_list
```

instead of the slightly weaker assertions of what you propose

```
assert_mock_has_any_call_with_first_n_args(redis_delete_mock, f'organisation-{organisation_one["id"]}-name')
assert_mock_has_any_call_with_first_n_args(redis_delete_mock, "domains")
assert_mock_has_any_call_with_first_n_args(redis_delete_mock, "organisations")
```

But that is extra work and I don't think should be a blocker to your change going in... but is maybe the direction I'd like to see our redis deletion code go. I did ponder changing the default for `raise_exception` in my original PR but thought I'd try and do a smaller thing to start with... but maybe I should have been more aggressive.

You don't have to reply to my musings if you don't want to. Feel free just to merge if you want to crack on.

tests/app/notify_client/test_service_api_client.py
```diff
@@ -511,33 +514,18 @@ def test_client_updates_guest_list(mocker):
     )
 
 
-def test_client_doesnt_delete_service_template_cache_when_none_exist(
```
idavidmcdonald, line 514: What led you to delete this test out of curiousity please?

tests/app/notify_client/test_service_api_client.py
```diff
@@ -463,28 +466,28 @@ def test_deletes_caches_when_modifying_templates(
...
-    mock_redis_delete = mocker.patch("app.extensions.RedisClient.delete")
-    mock_redis_delete_by_pattern = mocker.patch("app.extensions.RedisClient.delete_by_pattern")
+    mock_redis_delete = mocker.patch("app.extensions.RedisClient.delete", new_callable=RedisClientMock)
 
     mocker.patch("notifications_python_client.base.BaseAPIClient.request", return_value={"data": ""})
 
     service_api_client.archive_service(SERVICE_ONE_ID, ["my-user-id1", "my-user-id2"])
 
-    assert call("user-my-user-id1", "user-my-user-id2") in mock_redis_delete.call_args_list
-    assert call(f"service-{SERVICE_ONE_ID}-template*") in mock_redis_delete_by_pattern.call_args_list
```
idavidmcdonald, line 487: What led you to drop the assertion about the service template being deleted from the cache?

leohemsted:
@idavidmcdonald thanks for the comments - i agree with you and as is the way, after taking another look at it thought of a much better way to write these asserts with a custom mock rather than just a random function, and think they look much nicer now with two obvious functions. (I'm open to renaming one or either of these functions if you think of better names, but think they're good enough for now)

idavidmcdonald (approved):
Nice, I prefer this approach too.

2 small comments but hopefully don't require any changes as I assume they have good reason for them.

tests/app/notify_client/test_service_api_client.py
leohemsted: Short: We _DO_ delete the service template cache when no templates exist.

Long:

when we created this test, we were using a map to expand a list of templates to delete as kwargs - if there were no templates, we'd call `redis_client.delete(*[])`, which is an error as at least one argument must be provided. Hence adding a test for that specific case.

https://github.com/alphagov/notifications-admin/blob/7e15d0c928525e11ad655d9edb3f9bb00876dd82/app/notify_client/service_api_client.py#L9-L15


However, this was superceded by us starting to use `delete_by_pattern`, which doesn't behave differently if there's nothing to delete, so there's no need for us to have a separate unit test case. (eventually this function was removed and it was all moved into a decorator).

https://github.com/alphagov/notifications-admin/blob/01a3df6edce08e3cf0cca42b60dcf95127155497/app/notify_client/service_api_client.py#L9-L10

idavidmcdonald: Thanks

tests/app/notify_client/test_service_api_client.py
leohemsted: good point 🦅👀, think this was missed out by accident. squashed back in

idavidmcdonald (approved):

PR #5112, opened 2024-06-12
Conditionally show validation messages in an error-summary for "attach pages to a letter" journey

kr8n3r:
## What

Add logic to show for letter attachment file upload error messages in the error summary only for the initial file upload validation if form empty, file not a PDF or file bigger than maximum allowed size.

All content validation error messages remain being shown as they are currently

https://trello.com/c/IWs0jRtA/773-make-attach-a-page-to-a-letter-form-meet-validation-requirements

## Why

We’ve been updating file upload validation to error summary in 
https://trello.com/c/exLvftli/772-make-upload-contact-list-and-upload-addresses-phone-numbers-form-meet-validation-requirements
https://trello.com/c/Uu9YoBkj/710-make-forms-for-uploading-files-meet-validation-requirements

which deal with upload only. We’ve decide to tackle content validation separately in https://trello.com/c/N9RipR7p/590-evaluate-and-update-our-complex-validation-error

## How to review

## Visual changes

**New Error summary for upload validation only**

Screenshot 2024-06-12 at 12 03 21

**Current error banner for content validation only**

Screenshot 2024-06-12 at 12 03 52

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1177,7 +1177,7 @@ def _process_letter_attachment_form(service_id, template, form, upload_id):
         current_app.logger.info("Invalid PDF uploaded for service_id: %s", service_id)
         raise LetterAttachmentFormError(
             title="There’s a problem with your file",
-            detail="Notify cannot read this PDF.<br>Save a new copy of your file and try again.",
+            detail="Notify cannot read this PDF - save a new copy and try again",
         ) from None
 
     file_location = get_transient_letter_file_location(service_id, upload_id)
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -1058,6 +1058,7 @@ def letter_template_attach_pages(service_id, template_id):
     error = {}
     letter_attachment_image_url = None
     attachment_page_count = 0
+    use_error_summary = False
     if form.validate_on_submit():
         upload_id = uuid.uuid4()
         try:
@@ -1073,6 +1074,7 @@ def letter_template_attach_pages(service_id, template_id):
 
     if form.file.errors:
         error = get_error_from_upload_form(form.file.errors[0])
+        use_error_summary = True
 
     if not template.attachment:
         return (
@@ -1083,6 +1085,8 @@ def letter_template_attach_pages(service_id, template_id):
                 error=error,
                 letter_attachment_image_url=letter_attachment_image_url,
                 page_numbers=_get_page_numbers(attachment_page_count),
+                error_summary_enabled=use_error_summary,
+                use_error_summary=use_error_summary,
             ),
             400 if error else 200,
         )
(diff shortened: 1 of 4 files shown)


app/main/views/templates.py
```diff
@@ -1058,6 +1058,7 @@ def letter_template_attach_pages(service_id, template_id):
     error = {}
     letter_attachment_image_url = None
     attachment_page_count = 0
+    new_validation_pattern=False
```
CrystalPea, line 1061: that variable name could be more descriptive, maybe:

`simple_validation_message` ?

CrystalPea (commented):
It looks good, just the variable name is not very descriptive at present.

CrystalPea:
Also one thing is missing from the PR message I think - that is the WHY. Why do we want to change the validation pattern for those initial errors on file upload page?

app/main/views/templates.py
kr8n3r: updated variable name

kr8n3r:
> Also one thing is missing from the PR message I think - that is the WHY. Why do we want to change the validation pattern for those initial errors on file upload page?

updated

app/main/views/templates.py
```diff
@@ -1083,6 +1085,9 @@ def letter_template_attach_pages(service_id, template_id):
                 error=error,
                 letter_attachment_image_url=letter_attachment_image_url,
                 page_numbers=_get_page_numbers(attachment_page_count),
+                error_summary_enabled=use_error_summary,
```
CrystalPea, line 1088: I don't see us using this `error_summary_enabled` variable anywhere?

quis: It’s used in a few places, most prominently here: https://github.com/alphagov/notifications-admin/blob/9aaabeb7335b4a12004f659a3ff12fc74dbb7747/app/templates/withnav_template.html#L35-L37

CrystalPea: Ahh fair 🙌🏼 . I thought since it's introduced in this PR, the use would be added in the same PR, but it's pre-existing, makes sense 😳 .

Then do we need to pass in both `use_error_summary` and `error_summary_enabled`, since they both point to the same thing? Wouldn't `error_summary_enabled` be sufficient?

kr8n3r: https://github.com/alphagov/notifications-admin/blob/main/app/main/views/send.py and https://github.com/alphagov/notifications-admin/blob/main/app/main/views/uploads.py

kr8n3r: > Then do we need to pass in both `use_error_summary` and `error_summary_enabled`, since they both point to the same thing? Wouldn't `error_summary_enabled` be sufficient?

this flow is special as outlined in https://github.com/alphagov/notifications-admin/pull/5112/commits/ad6488ac8ad3c21823aed2cbed833c412b6929e5 - differs from other flows in as such it renders the same page. As we want to different the error message treatment, we've introduced `use_error_summary` additional param.

CrystalPea (approved):

CrystalPea (approved):

PR #5113, opened 2024-06-12
Switch from internal list of protected sender IDs to api

whpearson:
This makes admin query api/protected-sender-ids/check?sender_id=$NAME to find out whether an id is in the protected list, rather than in the hardcoded list in admin.

This requires mocking out the client in a bunch of places.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -111,6 +111,7 @@ from app.notify_client.performance_dashboard_api_client import (
 )
 from app.notify_client.platform_admin_api_client import admin_api_client
 from app.notify_client.platform_stats_api_client import platform_stats_api_client
+from app.notify_client.protected_sender_id_api_client import protected_sender_id_api_client
 from app.notify_client.provider_client import provider_client
 from app.notify_client.service_api_client import service_api_client
 from app.notify_client.sms_rate_client import sms_rate_api_client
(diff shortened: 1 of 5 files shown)


whpearson:
Relevant API PR https://github.com/alphagov/notifications-api/pull/4109

idavidmcdonald (commented):
Nice one, Will. Looking pretty good. Two hopefully minor and uncontroversial comments but if those are resolved then happy to approve.

One other comment, which may be a bit too late, is that I'm not so sure on the language of 'protected' sender id. Protected to me is something that we want want to use and keep safe for ourselves. That to me would be something like a sender name 'NHS'. I'd consider these more banned or not allowed sender names like 'Evri'. But naming things is not a binary choice and you and Mervi seemed happy with it, having already approved the API equivalent PR. So fine to stick as is with the naming convention but thought I'd mention anyway!

whpearson:
> Nice one, Will. Looking pretty good. Two hopefully minor and uncontroversial comments but if those are resolved then happy to approve.
> 
> One other comment, which may be a bit too late, is that I'm not so sure on the language of 'protected' sender id. Protected to me is something that we want want to use and keep safe for ourselves. That to me would be something like a sender name 'NHS'. I'd consider these more banned or not allowed sender names like 'Evri'. But naming things is not a binary choice and you and Mervi seemed happy with it, having already approved the API equivalent PR. So fine to stick as is with the naming convention but thought I'd mention anyway!

I had thought the table might grow into being a list of protected IDs as well, basically consuming all the information from the csv. So I wanted a generic name.

idavidmcdonald (approved):
Nice work! And thanks for the context with your naming choice too! 🚢

PR #5115, opened 2024-06-14
Use request cache delete decorator function where possible

idavidmcdonald:
We recently put in some work to make our request cache delete function more faultproof so we aren't at risk of stale data if redis goes down:
https://github.com/alphagov/notifications-utils/pull/1120

This commit finds a few places in the admin repo that we aren't using this request cache delete function (and so aren't getting the protections of the above pull request).

For the letter jobs client and `archive_organisation` function these are fairly straightforward.

For the `update_organisation` function, it used to delete the cache keys only if it needed to based on the type of update for the organisation. Whilst this is more efficient, I think it is more important to get the guarantees of no stale data than the efficiency. Calls out to redis are fairly quick and this route is not very often used so if this does slow down the route slightly, the impact will be small. I think this trade off is worth making.

Note, we still do have a few redis_client.delete calls in our admin code that it would still be great to remove. However, if you search the code for `redis_client.delete` you will see these are all passing multiple strings into `redis_client.delete` which is not currently supported by `cache.delete`. We will need to think about how we migrate these in the future.


leohemsted:
some of the ones we're passing multiple strings in could be replaced by deleting by pattern 🤔 

https://github.com/alphagov/notifications-admin/blob/38a7c35ebe078282be1ec1d4c97b3f737dc5f750/app/notify_client/template_folder_api_client.py#L41-L42

this could be replaced with `@cache.delete_by_pattern("service-{service_id}-template*")` for example (actually we could remove the other decorators too at that point)

the archive_service and update_organisation ones might be a bit tougher though

leohemsted (approved):

idavidmcdonald:
Agree on Leo's suggestion but ran out of time before I leave. The above is a sensible thing to do if anyone wants to do it please...

the archive_service and update_organisation ones I'm not too sure what the best approach is though...

PR #5121, opened 2024-06-19
Remove Google Analytics

kr8n3r:
## Why
See https://trello.com/c/lZupW4Kj/816-remove-analytics-code

## What

Removal of code that aids Google Analytics implementation such as
- analytics initialisation and event tracking
- cookie banner component, styles and Javascript functions for settings consent cookie values
- cookie banner content change and removal of cookie settings form
- error tracking data attributes
- removal of redundant analytics javascript tests
- update of python tests

Addition of 
 - a simple function that checks if GA and/or `cookies_policy`` cookies are present and if so removes them from user's browser as they're no longer needed and it's good practice to do so.
- test for said function
-  updates test cookie helpers to add `getCookie` helper and remove `deleteCookie` helper.

### Cookies page content update
Screenshot 2024-06-19 at 15 13 00


tombye (approved):
Looks good to me 👍🏻 

Really nice work, catching lots of places we might not have checked (CSP, print styles) and makes a good record in case we need to add it back in.

PR #5123, opened 2024-06-25
Remove notice about upcoming price increase

klssmith:
⚠️ **This should not be merged until 1 July** ⚠️

On 1 July 2024, the new prices will be displayed on the page (since the data comes from the api) and we can remove the message about an upcoming increase.


karlchillmaid (approved):
Approved. Merge and deploy on 1 July.

PR #5132, opened 2024-07-03
fix(forms): enforce character length validation in template name

joybytes:
## PR Summary
- form validation for template name set to 255 characters.
- we were enforcing this validation on the database but not at the form level

### How was this tested?
-  all templates now have validation for that field
- top error message has a link that takes to the error field

**before**:

FIELD REQUIRED VALIDATION

Screenshot 2024-07-08 at 10 37 51

**after**:
LENGTH VALIDATION

Screenshot 2024-07-08 at 10 35 54

FIELD REQUIRED VALIDATION

Screenshot 2024-07-08 at 10 36 06

### Related ticket:
[Add max length validation to the template name field in the admin app](https://trello.com/c/7v4l6uri/795-add-max-length-validation-to-the-template-name-field-in-the-admin-app)

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1348,8 +1348,8 @@ class TemplateNameMixin:
     name = GovukTextInputField(
         "Template name",
         validators=[
-            NotifyDataRequired(thing="a name for this template"),
-            Length(max=255, thing="a name for this template"),
+            NotifyDataRequired(thing="Template name"),
+            Length(max=255, thing="Template name"),
         ],
     )
 
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1345,7 +1345,10 @@ class ConfirmPasswordForm(StripWhitespaceForm):
 
 
 class TemplateNameMixin:
-    name = GovukTextInputField("Template name", validators=[NotifyDataRequired(thing="a name for this template")])
+    name = GovukTextInputField("Template name", validators=[
+        NotifyDataRequired(thing="a name for this template"),
+        Length(max=255, thing="a name for this template"),
+    ])
 
 
 class RenameTemplateForm(StripWhitespaceForm, TemplateNameMixin):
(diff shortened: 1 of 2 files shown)


klssmith:
Tested locally and this works well 🎉

@karlchillmaid do you have thoughts on the error message content? For the service name length error we have `Service name cannot be longer than...`, so I'm not sure if we want to make both consistent

BlessedDev (approved):

BlessedDev:
> Tested locally and this works well 🎉
> 
> @karlchillmaid do you have thoughts on the error message content? For the service name length error we have `Service name cannot be longer than...`, so I'm not sure if we want to make both consistent

This error message content seems to better, should we consider making the Service error message content to be consistent with this?

app/main/forms.py
```diff
@@ -1345,7 +1345,13 @@ def validate_password(self, field):
 
 
 class TemplateNameMixin:
-    name = GovukTextInputField("Template name", validators=[NotifyDataRequired(thing="a name for this template")])
+    name = GovukTextInputField(
+        "Template name",
+        validators=[
+            NotifyDataRequired(thing="a name for this template"),
+            Length(max=255, thing="a name for this template"),
```
karlchillmaid, line 1352:
```suggestion
            Length(max=255, thing="Template name"),
```

karlchillmaid:
We should use `Template name` to be consistent with the existing ‘thing is too long’ validation errors. For example ‘Service name’, ‘Organisation name’, ‘Password’, ‘Email address’...

app/main/forms.py
joybytes: https://github.com/alphagov/notifications-admin/pull/5132/commits/7ad06aa1b8d59bc5ea2457b7634da353c5cd3245

BlessedDev (approved):

PR #5135, opened 2024-07-05
Remove issues from accessibility statement

tombye:
The reason for each removal is in the commit that removes them.

The decision to remove them was made in collaboration with @karlchillmaid.

### DIFF
--- a/app/templates/views/accessibility_statement.html
+++ b/app/templates/views/accessibility_statement.html
@@ -49,7 +49,6 @@
   <ul class="govuk-list govuk-list--bullet">
     <li>if a team adds template folders, screen readers may announce the page heading in a confusing way</li>
     <li>screen reader users may find it hard to expand or collapse groups of checkboxes</li>
-    <li>one page links to a PDF document that is not fully accessible</li>
     <li>the Notify status page has several accessibility issues</li>
   </ul>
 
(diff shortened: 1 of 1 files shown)


app/templates/views/accessibility_statement.html
```diff
@@ -114,13 +109,6 @@ <h3 class="heading-small" id="pdfs-documents">
     PDFs and other documents
   </h3>
 
-  <p class="govuk-body">
-    One page links to a PDF document that is not fully accessible. This is not a fail under any WCAG 2.1 success criterion.</p>
-
-  <p class="govuk-body">
-    We provide a written description of the information in the PDF on an HTML page as an accessible alternative.
-  </p>
-
```
karlchillmaid, line 123: We should keep this content somewhere easily accessible, in case we get any support tickets about the PDF.

karlchillmaid (approved):
Removals and date bump look good.

app/templates/views/accessibility_statement.html
tombye: Where do you think it should live, the support manual?

karlchillmaid: Yeah, let’s put it there for now. 🙂

tombye: Done. I created a section called 'Accessibility' and put it in there.

kr8n3r (approved):

PR #5131, opened 2024-07-03
Improve the ‘security code not received’ guidance 

karlchillmaid:
When users do not receive their security code text message, they contact support.

To help users troubleshoot the possible causes of this issue before they have to contact us, we’re going to publish some of the information we usually send through support on our website.

PR #5138, opened 2024-07-08
Improve the email sign-in link guidance

karlchillmaid:
If a user does not receive their email link to sign in to Notify, they may contact support.

To help users troubleshoot the possible causes of this issue before they have to contact us, we’re going to publish some of the information from our support macro on the website.

We’re doing this work to be consistent with the improvements we made for the SMS security code (https://github.com/alphagov/notifications-admin/pull/5131).


kr8n3r (approved):
looks good
Screenshot 2024-07-09 at 12 39 38

PR #5136, opened 2024-07-08
Email unsubscribe request dashboard

rparke:

PR #5139, opened 2024-07-09
Add unusbscribe reports summary page

BlessedDev:
This PR adds the main frame work of the unsubscribe report pages to list and batch unsubscribe requests as described in [this](https://trello.com/c/MhGtZtKW/52-add-pages-to-list-email-unsubscribe-requests-and-batch-them-up) card.

Batching of unsubscribe requests is initiated when a report is downloaded. Creating and downloading reports will be addressed in an later [card](https://trello.com/c/BeZGTW0E/53-let-users-download-email-unsubscribe-requests).

A summary is generated for un-batched subscribe requests, with a `is_a_batched_report` value of `False`, which will be used to create a batched report to be persisted in the database when the user initiates download.

The work done here is dependent on this [PR](https://github.com/alphagov/notifications-api/pull/4128) which builds the endpoints it relies on.

When there are no unsubscribe requests, this is what the page looks like:
empty unsubscribe requests

Available unsubscribe requests are displayed as follows in reverse order of chronology, showing the latest report:

unsubscribe reports

An unbatched report has the "Mark as Completed" checkbox greyed out, a hint message encouraging download of the report, and no update button:

un-batched unsubscribe report

A batched report has the "Mark as Completed" checkbox available, a hint message encouraging processing of the report, and an update button. Validation is to be dealt with in a later [card](https://trello.com/c/5hJS5Oia/54-let-users-mark-email-unsubscribe-requests-as-completed)

batched unsubscribe report


app/notify_client/service_api_client.py
```diff
@@ -503,6 +503,10 @@ def get_notification_count(self, service_id, notification_type):
         count = redis_client.get(daily_limit_cache_key(service_id, notification_type=notification_type)) or 0
         return int(count)
 
+    @cache.set("service-{service_id}-unsubscribe-request-reports-summary")
+    def get_unsubscribe_reports_summary(self, service_id):
```
quis, line 507: Looks like this method isn’t used anywhere in this PR. Can we stick to implementing it in this PR instead:
https://github.com/alphagov/notifications-admin/pull/5136/files#diff-28c72005108ae72f73bb3a3d0ea065dc08d2633b7f9c723cef584a30a48c3f6fR6

app/main/views/unsubscribe_requests.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@main.route("/services/<uuid:service_id>/unsubscribe-request-report/<uuid:report_id>")
+@user_has_permissions("view_activity")
+def unsubscribe_request_report(service_id, report_id):
+    report_data = None
+    if reports_summary_data := _get_unsubscribe_request_reports_summary():
+        for report_summary in reports_summary_data:
+            if report_summary["report_id"] == report_id:
+                report_data = report_summary
```
quis, line 33: This kind of logic would be much better in a model class, but happy to save that for a follow up PR if this works

app/templates/views/unsubscribe-request-report.html
```diff
@@ -0,0 +1,37 @@
...
+    {{ page_header('{} until {}'.format(earliest_timestamp|format_date_normal, latest_timestamp|format_date_normal)) }}
+
+ <p class="govuk-body">
+     {{ count|format_thousands}} new requests to unsubscribe
+ </p>
+
+    <p class="govuk-body">You must: </p>
+      <ol class="govuk-list govuk-list--number">
+        <li><span class="bottom-gutter"><a download href="" class="govuk-link heading-small">Download the report </a>
+          &nbsp(available for 90 days) </span> </li>
```
quis, line 27: If you really want a non-breaking space here it needs to be encoded like this with no other whitespace around it:
```suggestion
        <li><span class="bottom-gutter"><a download href="" class="govuk-link heading-small">Download the report</a>&nbsp;(available for 90 days) </span> </li>
```
Otherwise:
```suggestion
        <li><span class="bottom-gutter"><a download href="" class="govuk-link heading-small">Download the report </a>
          (available for 90 days) </span> </li>
```

app/templates/views/unsubscribe-request-reports-summary.html
```diff
@@ -0,0 +1,52 @@
...
+        field_headings=['Report'],
+        field_headings_visible=False
+    ) %}
+    {% call row_heading() %}
+        {%  if item.processed_by_service_at and item.is_a_batched_report %}
+            {%  set processed_status = "Completed" -%}
+        {% elif  item.is_a_batched_report and not item.processed_by_service_at %}
+            {%  set processed_status = "Downloaded" -%}
+        {%  else %}
+            {%  set processed_status = "Not Downloaded" -%}
```
quis, line 33:
```suggestion
            {%  set processed_status = "Not downloaded" -%}
```

app/main/forms.py
```diff
@@ -2882,3 +2882,31 @@ def __init__(self, *args, organisation, **kwargs):
             ("main.choose_service_to_join", "Join an existing service"),
         ),
     )
+
+
+class ProcessUnsubscribeRequestForm(StripWhitespaceForm):
+    report_has_been_processed = GovukCheckboxField("Mark as Completed")
```
quis, line 2888:
```suggestion
    report_has_been_processed = GovukCheckboxField("Mark as completed")
```

app/templates/views/unsubscribe-request-reports-summary.html
```diff
@@ -0,0 +1,52 @@
...
+        </a>
+       <p class="file-list-hint">
+
+         {{ item.count|format_thousands}} {{ item.count | message_count_label('unsubscribe request', suffix='') }}
+           <span class="banner-dashboard-meta">
+        {{ processed_status }}
+      </span>
+
+       </p>
+    {% endcall %}
```
quis, line 47: This should be a separate table column, which you can achieve like so:
```suggestion
    </p>
    {% endcall %}
    {% call field(align='right') %}
      {{ processed_status }}
    {% endcall %}
```

app/notify_client/service_api_client.py
quis: My mistake, was confused by the naming here.

I think this PR should use:
- `service/{service_id}/unsubscribe-request/reports` as the endpoint
- `service-{service_id}-unsubscribe-request-reports` as the cache key

@rparke’s PRs should use:
- `/<uuid:service_id>/unsubscribe-request/statistics`<sup>1</sup> as the endpoint
- `service-{service_id}-unsubscribe-request-statistics` as the cache key

***

1. `statistics` instead of `summary` because:
    - you’ve both used `summary` to mean different things so it’s obviously ambiguous
    - it’s the word we use for a similar endpoint with returned letters: https://github.com/alphagov/notifications-admin/blob/3bf4695fc69c4d54b3b546e4a03539d8f3a263ca/app/notify_client/service_api_client.py#L490-L492

BlessedDev: Good points. I will change the endpoints to  `/services/<uuid:service_id>/unsubscribe-requests/reports` for the reports and `/services/<uuid:service_id>/unsubscribe-requests/reports/<uuid:report_id>` for individual reports

app/templates/views/unsubscribe-request-report.html
BlessedDev: I think it looks better with a non breaking space. I will go with your suggestion to remove the whitespace.

app/templates/views/unsubscribe-request-reports-summary.html
BlessedDev: You are right. I shall update it.

app/notify_client/service_api_client.py
BlessedDev: Regarding the cache key. I think it should remain as `"service-{service_id}-unsubscribe-request-reports-summary"` to be consistent with the pattern for returned letters https://github.com/alphagov/notifications-admin/blob/3bf4695fc69c4d54b3b546e4a03539d8f3a263ca/app/notify_client/service_api_client.py#L494
Also it retrieves summaries of the reports. Another `service_api_client` method will be added in subsequent PRs to retrieve the actual reports for the download endpoint, so `service-{service_id}-unsubscribe-request-reports` should be save for it.

quis: OK, fine to use `service-{service_id}-unsubscribe-request-reports-summary` and `service/{service_id}/unsubscribe-request-reports/summary` in this PR for consistency then.

For an individual report it should probably be `service/{service_id}/unsubscribe-request-reports/{batch_id}` and won’t need caching.

app/templates/views/unsubscribe-request-reports-summary.html
```diff
@@ -0,0 +1,50 @@
...
+
+    {{ page_header('Email unsubscribe requests') }}
+
+<div class="dashboard-table">
+    {% call(item, row_number) list_table(
+        data,
+        caption="Unsubscribe Request Reports",
+        caption_visible=False,
+        empty_message='If you have unsubscribed requests they will be listed here',
+        field_headings=['Report'],
```
quis, line 24: Because the table has 2 columns now it should also have 2 field headings
```suggestion
        field_headings=['Report', 'Status'],
```

app/main/views/unsubscribe_requests.py
```diff
@@ -0,0 +1,49 @@
...
+                latest_timestamp=report_data["latest_timestamp"],
+                processed_by_service_at=report_data["processed_by_service_at"],
+                batch_id=report_data["batch_id"],
+                is_a_batched_report=report_data["is_a_batched_report"],
+                service_id=service_id,
+                form=form,
+            )
+
+    else:
+        return redirect(url_for("main.unsubscribe_request_reports_summary", service_id=service_id))
```
quis, line 49: If there is no data for the service this will redirect where a `404` would be more appropriate.

If the `batch_id` doesn’t exist then the code won’t reach this point and the function will return `None` – a `404` would be a lot better.

```suggestion
    abort(404)
```

app/main/views/unsubscribe_requests.py
```diff
@@ -0,0 +1,24 @@
...
+
+@main.route("/services/<uuid:service_id>/unsubscribe-requests/summary")
+@user_has_permissions("view_activity")
+def unsubscribe_request_reports_summary(service_id):
+    return render_template("views/unsubscribe-request-reports-summary.html")
+
+
+@main.route("/services/<uuid:service_id>/unsubscribe-requests/reports/<uuid:batch_id>")
+@user_has_permissions("view_activity")
+def unsubscribe_request_report(service_id, batch_id):
```
quis, line 17: Think this is what we need to make the page work for unbatched requests:
```suggestion
@main.route("/services/<uuid:service_id>/unsubscribe-requests/reports/<uuid:batch_id>")
@main.route("/services/<uuid:service_id>/unsubscribe-requests/reports/latest")
@user_has_permissions("view_activity")
def unsubscribe_request_report(service_id, batch_id=None):
```

BlessedDev: Yes, that will make it work.

quis (approved):
Good to go now I reckon

PR #5140, opened 2024-07-10
Add guidance for team members on support

karlchillmaid:
We just did some work to [warn our team when a Zendesk ticket has come from a user who’s not signed in to Notify](https://trello.com/c/bPywVnSY/825-reduce-chance-of-impersonation-on-our-support-form-for-unauthenticated-users).

Now we need to provide some practical guidance for team members on support to know how to check a user is who they say they are.

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -19,6 +19,13 @@ from app.utils import hide_from_search_engines
 
 bank_holidays = BankHolidays(use_cached_holidays=True)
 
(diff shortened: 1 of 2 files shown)


klssmith:
It turned out to be fairly easy to format the content just with new line symbols. Here's what it looks like in action:

Screenshot 2024-07-10 at 14 28 47

kr8n3r (approved):

PR #5143, opened 2024-07-12
Make branding journeys remember the radio button choices

klssmith:
https://trello.com/c/vgGPEQav/853-branding-journeys-dont-remember-your-choices

Best reviewed commit by commit. This change ensures we always remember the radio buttons selected when going through the flow to select branding. We had previously done this for the initial radio buttons you see, but the email branding journey has two more pages with radio buttons and we weren't remembering the selection made there.

To do this, we pass query params with the choices on the forms made through the requests, using the helper methods (`_email_branding_flow_query_params` and `_letter_branding_flow_query_params`) to do this.

* The choice made on the `ChooseEmailBrandingForm` is stored in the `branding_choice` query param
* The choice made on the `EmailBrandingChooseLogoForm` is stored in the `logo_type` query param
* The choice made on the `EmailBrandingChooseBanner` is stored in the `logo_type` query brand_type

### DIFF
--- a/app/main/views/service_settings/branding.py
+++ b/app/main/views/service_settings/branding.py
@@ -401,7 +401,7 @@ def email_branding_set_alt_text(service_id):
 
 
 def _email_branding_flow_query_params(request, **kwargs):
-    """Return a dictionary containing values for the new email branding flow.
+    """Return a dictionary containing values for the email branding flow.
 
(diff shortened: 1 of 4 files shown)


kr8n3r (approved):
changes look good and work as expected

PR #5144, opened 2024-07-12
Bump utils to 81.1.0

tombye:
Brings in changes from these pull requests:
- [Refactor of `InvalidPhoneError` class - breaking change](https://github.com/alphagov/notifications-utils/pull/1125)
- [Introduction of `PhoneNumber` class](https://github.com/alphagov/notifications-utils/pull/1126)

Admin doesn't use any of the bits of the `InvalidPhoneError` API changed by this bump so this doesn't require any further work.

Admin doesn't use `PhoneNumber` anywhere yet. [Leo hinted we will use it as the base of all phone number validation in future](https://github.com/alphagov/notifications-utils/blob/main/CHANGELOG.md#8110) but not yet.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -16,7 +16,7 @@ pyexcel-ods3==0.6.1
 notifications-python-client==8.0.1
 fido2==1.1.0
 
-notifications-utils @ git+https://github.com/alphagov/notifications-utils.git@80.0.1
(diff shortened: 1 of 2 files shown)


quis (approved):
Pro tip: if you run `make bump-utils` it will generate you a nice commit message automatically

PR #5146, opened 2024-07-16
Upgrade to use Design System (govuk-frontend) version 5

kr8n3r:
Adresses: https://trello.com/c/lJgxnzZW/869-upgrade-admin-to-use-govuk-frontend-version-5-design-system

Upgrade [govuk-frontend](https://github.com/alphagov/govuk-frontend/releases/tag/v5.4.1) and [govuk-frontend-jinja](https://github.com/LandRegistry/govuk-frontend-jinja/releases/tag/3.1.0) to latest available versions, which are major releases.

With version 5, legacy browsers, including IE11, no longer get Javascript enhancements, so we needed to do to the work to create two separate Javascript bundles as govuk-frontend requires to run in script type='module'

As we're upgrading from version 4.8.0, lots of other changes needed to be made as outlined in the govuk-frontend [CHANGELOG](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md) as well as upgrading our custom checkboxes, nested radios and radios-with-images with changes from upstream.

See individual commits for detailed changes


app/assets/javascripts/esm/all-esm.mjs
```diff
@@ -8,4 +8,10 @@ createAll(Header)
 createAll(Radios)
 createAll(ErrorSummary)
 createAll(SkipLink)
-createAll(Tabs)
\ No newline at end of file
+createAll(Tabs)
+
+// for fileUpload.js
+window.GOVUKFrontendButton = Button;
```
tombye, line 14: I'm not 100% sure how strongly I feel about this so feel free to argue against but it seems funny to assign to `window` here instead of exporting things in the right order and letting RollupJS assign them to a 'namespace' variable on `window`, as we do currently: https://github.com/alphagov/notifications-admin/blob/main/app/assets/javascripts/esm/all.mjs#L88.

app/templates/views/templates/_move_to.html
```diff
@@ -1,10 +1,12 @@
 {% from "components/page-footer.html" import page_footer %}
 
-<div id="sticky_template_forms">
+<div id="sticky_template_forms template-forms">
+  <h2 class="govuk-heading-m js-hidden">Template and folder actions</h2>
   <button name="operation" value="unknown" hidden></button>
   <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
   {% if form.move_to.choices and template_list.templates_to_show %}
     <div id="move_to_folder_radios" class="sticky-template-form" role="region" aria-label="Choose the folder to move selected items to">
+      <h3 class="govuk-heading-s js-hidden">Move to a folder</h3>
```
tombye, line 9: Do you think we could have these as visually hidden when our JS runs and replace the `aria-label` on the parent div with an `aria-describedby` linked to them? Seems like duplication of copy otherwise.

This does look so much better when JS is off. Makes a massive difference.

app/assets/stylesheets/components/message.scss
```diff
@@ -63,12 +63,13 @@ a {
     &-with-checkbox {
 
       .template-list-item-hint {
-        margin-top: govuk-spacing(-3);
+        margin-top: govuk-spacing(1);
```
tombye, line 66: The current `margin-top` is `0`. This is complex so I might be missing something but shouldn't this match that?

tombye (commented):
I know you're still working on some fixes for how template list items render when live search is used but thought it was still worth submitting this review, so you've access to it.

Additional: I've really enjoyed the commits in this PR. It's a lot of changes and some of them complicated but the commits explain everything so well.

app/assets/stylesheets/components/message.scss
```diff
@@ -56,9 +58,21 @@ a {
...
+      .template-list-item-hint {
+        margin-top: govuk-spacing(-3);
+        padding-left: ($govuk-checkboxes-label-padding-left-right + $govuk-touch-target-size);
+      }
+
+      input {
+        margin-right: $govuk-checkboxes-label-padding-left-right;
+      }
+      a {
+        width: calc(100% - (#{$govuk-touch-target-size} + #{$govuk-checkboxes-label-padding-left-right}));
```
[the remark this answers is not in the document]
kr8n3r: Screenshot 2024-07-29 at 11 45 41
without interpolation things break

app/assets/stylesheets/components/message.scss
kr8n3r: yeah, was still finding the adjustments for when search is performed. reset to 0 now

gulpfile.js
```diff
@@ -63,14 +63,13 @@ const copy = {
...
 
-const javascripts = () => {
-  // JS from third-party sources
-  // We assume none of it will need to pass through Babel
-  const vendored = src(paths.src + 'javascripts/esm/all.mjs')
+const javascripts = {
+  commonJsBundle: () => {
+    // JS from third-party sources
+    // We assume none of it will need to pass through Babel
+    const vendored = src(paths.src + 'javascripts/all.js',{allowEmpty: true})
```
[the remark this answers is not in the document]
kr8n3r: doing just 

```
commonJsBundle: () => {
    // JS from third-party sources
    // We assume none of it will need to pass through Babel
    const vendored = src(
      [
        paths.npm + 'hogan.js/dist/hogan-3.0.2.js',
        paths.npm + 'jquery/dist/jquery.min.js',
        paths.npm + 'query-command-supported/dist/queryCommandSupported.min.js',
        paths.npm + 'timeago/jquery.timeago.js',
        paths.npm + 'textarea-caret/index.js',
        paths.npm + 'cbor-js/cbor.js'
      ]
    )
   
```
results in a completely diff of compiled all.js with JS errors on execution. As it's been a while since I last worked with gulp, perhaps I'm missing something here to make it work

app/assets/javascripts/esm/all-esm.mjs
kr8n3r: while we can bundle Morphdom into the other bundle, we can't do the same with govuk-frontend. equally We can't initialise v5 with window.namespace if I'm understanding you suggestion correctly

app/templates/views/templates/_move_to.html
kr8n3r: hmm not sure how that would work and help. Likely it's obvious but can't see it

app/assets/stylesheets/components/message.scss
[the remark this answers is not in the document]
kr8n3r: added comment about sass version and treatment of `calc()`

app/assets/javascripts/esm/all-esm.mjs
tombye: We discussed this offline so I was able to see what Jani was talking about. Now I've seen this, I'm happy with this being marked with a comment explaining why we're doing it.

app/templates/views/templates/_move_to.html
tombye: Discussed offline and Jani has since made changes reflecting what I described so marking this as resolved.

tombye (approved):
All comments addressed and a quick test across our browser matrix looks good 👍🏻 

Thanks for the patience with this, a nice piece of work.

PR #5147, opened 2024-07-17
Fix wonky empy state for email unsubscribe request table

quis:
When we show a ‘this table is empty’ message we want it to span the full width of the table.

Previously we were doing this by setting `colspan` to an arbitrary high number (10), which is higher than the number of columns any of our tables have.

At some point this seems to have stopped having the desired effect, in Chrome at least.

This commit fixes the problem by setting `colspan` to exactly the number of columns the table has.

# Before

image

# After 

image

### DIFF
--- a/app/templates/components/table.html
+++ b/app/templates/components/table.html
@@ -37,7 +37,7 @@
     {% endfor %}
     {% if not items %}
       {% call row() %}
-        <td class="table-empty-message" colspan="10">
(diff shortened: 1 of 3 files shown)


kr8n3r:
how about not rendering a table at all if there is no data?

quis:
@kr8n3r sometimes we still want to show the header row even if there’s no data in the table body

kr8n3r (approved):

PR #5149, opened 2024-07-22
Add validation for letters which have a "no fixed abode" address 

klssmith:
This adds validation when sending a one-off letter and uploading a file
of letters. Template-preview validates the precompiled letters and gives
them the validation message of "no-fixed-abode-address" in the metadata,
so we also need to add a user-friendly message for this code for the
precompiled letters.


tests/app/utils/test_letters.py
```diff
@@ -143,56 +143,63 @@ def test_get_letter_validation_error_for_unknown_error():
         ),
         (
             "address-is-empty",
-            None,
+            [1],
```
klssmith, line 146: The test data for these parameterized tests was wrong, so this fixes it - the errors with addresses list the first page as invalid

quis (approved):

PR #5151, opened 2024-07-24
fix: if users are login via sms auth we do not want to show sms-specific message when they do not receive the email

joybytes:
## Summary
- when users login via text message AND have not verified their email for more than 90 days, we force them to verify their email address prior to login by sending them an email link
- In the case that they do not receive that email link, they can click on `Not received an email?` and it will take you to this page:
Screenshot 2024-07-24 at 17 26 42

**Issue to resolve**: Since the user started their journey via sms-verification, showcasing text regarding 'If you want to sign in with a text message code instead' can be quite confusing

**Expectation**: do not show that section if the user came from a text-message verification journey.

**After fix**:
Screenshot 2024-07-24 at 17 13 06

### Related task
[Fix  login journey for users who authenticate via sms and have not logged in a while](https://trello.com/c/KHF3gK7N/882-fix-login-journey-for-users-who-authenticate-via-sms-and-have-not-logged-in-a-while)

### DIFF
--- a/app/main/views/code_not_received.py
+++ b/app/main/views/code_not_received.py
@@ -5,7 +5,6 @@ from app.main import main
 from app.main.forms import TextNotReceivedForm
 from app.models.user import User
 from app.utils.login import (
-    email_needs_revalidating,
     redirect_to_sign_in,
 )
 
@@ -54,7 +53,7 @@ def check_and_resend_verification_code():
 def email_not_received():
     # user login from two-factor-sms refactor does not need to see the two-factor-sms details on page
     user = User.from_email_address(session["user_details"]["email"])
-    login_via_sms = email_needs_revalidating(user)
+    login_via_sms = user.sms_auth
 
     redirect_url = request.args.get("next")
     return render_template("views/email-not-received.html", redirect_url=redirect_url, login_via_sms=login_via_sms)
--- a/app/main/views/code_not_received.py
+++ b/app/main/views/code_not_received.py
@@ -4,7 +4,10 @@ from app import user_api_client
 from app.main import main
 from app.main.forms import TextNotReceivedForm
 from app.models.user import User
-from app.utils.login import redirect_to_sign_in
+from app.utils.login import (
+    email_needs_revalidating,
+    redirect_to_sign_in,
+)
 
 
 @main.route("/resend-email-verification")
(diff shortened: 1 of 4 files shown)


app/main/views/code_not_received.py
```diff
@@ -49,8 +52,12 @@ def check_and_resend_verification_code():
 @main.route("/email-not-received", methods=["GET"])
 @redirect_to_sign_in
 def email_not_received():
+    # user login from two-factor-sms refactor does not need to see the two-factor-sms details on page
+    user = User.from_email_address(session["user_details"]["email"])
+    login_via_sms = email_needs_revalidating(user)
+
     redirect_url = request.args.get("next")
-    return render_template("views/email-not-received.html", redirect_url=redirect_url)
+    return render_template("views/email-not-received.html", redirect_url=redirect_url, login_via_sms=login_via_sms)
```
kr8n3r, line 60: not sure the variable makes sense. maybe canLoginviaSms?

app/main/views/code_not_received.py
```diff
@@ -49,8 +52,12 @@ def check_and_resend_verification_code():
 @main.route("/email-not-received", methods=["GET"])
 @redirect_to_sign_in
 def email_not_received():
+    # user login from two-factor-sms refactor does not need to see the two-factor-sms details on page
+    user = User.from_email_address(session["user_details"]["email"])
+    login_via_sms = email_needs_revalidating(user)
```
klssmith, line 57: `email_needs_revalidating ` will return `True` if someone logs in with email auth but just hasn't logged in for over 90 days, so it can't be used to check someone's auth type. However, you can just do `user.sms_auth` instead.

joybytes: nice, it works thanks 
Screenshot 2024-07-29 at 08 45 04

joybytes: https://github.com/alphagov/notifications-admin/pull/5151/commits/f6850b29b2b15cf96d883b76f7714ac1a3a57582

app/main/views/code_not_received.py
joybytes: Yes, that would have been the expected behaviour because you probably have signed in via email less than 90 days ago.

Katie has proposed a new way of verifying the route of sign-in which by checking a field I did not know was available. `sms_auth`

klssmith (approved):

PR #5160, opened 2024-08-01
Make unsubscribe report titles friendlier

quis:
At the moment the title of our reports follow the same format no matter:
- the range they cover
- how recent they are

This commit makes the titles aware of their context (in other words what time it is now and what other reports exist).

This makes them better in 2 situations especially:
- there are multiple reports on the same day (where we now add the time to help differentiate)
- there are very recent reports (where we now say today or yesterday instead of the numeric date)

The alternative would be to always show the time and year for every report, but this would feel noisy and overly-specific in most cases.

Example from the test cases:

Before | After
---|---
22 June 2024 to 22 June 2024 | Today at 4:00pm
22 June 2024 to 22 June 2024 | Today at midday to today at 2:17pm
22 June 2024 to 22 June 2024 | Today until midday
15 June 2024 to 21 June 2024 | 15 June to yesterday
8 December 2023 to 14 January 2024 | 7 December 2023 to 13 January

### DIFF
--- a/app/formatters.py
+++ b/app/formatters.py
@@ -102,8 +102,8 @@ def format_date_human(date):
     return get_human_day(date)
 
 
-def format_datetime_human(date, date_prefix="on"):
-    return f"{get_human_day(date, date_prefix=date_prefix)} at {format_time(date)}"
+def format_datetime_human(date, date_prefix="on", separator="at"):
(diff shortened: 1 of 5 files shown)


BlessedDev (approved):
Looks good to me. 👍

PR #5162, opened 2024-08-05
Make ‘mark as completed’ checkbox available as soon as unsubscribe report is downloaded

quis:
At the moment you can’t mark a report as completed until you have downloaded it.

However when you click the ‘Download this report’ link you stay on the same page while the report downloads. This means you have downloaded the report but still can’t click the checkbox.

This commit changes the flow when clicking the download link to:
- batch up the requests into a report
- server-side redirect to the page for that report (checkbox now enabled)
- client-side redirect to the download (while the user stays on the same page)

So the effect – without any AJAX – is that the report is downloaded and the checkbox is enabled simultaneously.

### DIFF
--- a/app/main/views/unsubscribe_requests.py
+++ b/app/main/views/unsubscribe_requests.py
@@ -87,8 +87,9 @@ def create_unsubscribe_request_report(service_id):
     created_report_data = service_api_client.create_unsubscribe_request_report(service_id, unbatched_report_data)
     return redirect(
         url_for(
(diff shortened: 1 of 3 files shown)


CrystalPea:
@kr8n3r isn't it only a failure for periodic refresh? If I understand well this code would only refresh once? (I might be missing something)

rparke (approved):

kr8n3r:
yeah, it'll likely be fine

kr8n3r (approved):

PR #5157, opened 2024-07-30
Download unsubscribe request reports

BlessedDev:
This PR enables downloading an `unsubscribe-request-report` as described in [this](https://trello.com/c/BeZGTW0E/53-let-users-download-email-unsubscribe-requests) card.
The PR for the API endpoints can be found [here](https://github.com/alphagov/notifications-api/pull/4152).

PR #5165, opened 2024-08-06
Fix test_create_unsubscribe_request_report_creates_batched_report

BlessedDev:
This PR fixes the issues encountered while merging in https://github.com/alphagov/notifications-admin/pull/5157

### DIFF
--- a/tests/app/main/views/test_unsubscribe_requests.py
+++ b/tests/app/main/views/test_unsubscribe_requests.py
@@ -1,5 +1,6 @@
 import pytest
 from freezegun import freeze_time
+from notifications_utils.timezones import utc_string_to_aware_gmt_datetime
 
 from app import service_api_client
 from app.models.unsubscribe_requests_report import UnsubscribeRequestsReports
@@ -320,8 +321,8 @@ def test_create_unsubscribe_request_report_creates_batched_report(client_request
     summary_data = [
         {
             "count": 34,
-            "earliest_timestamp": "2024-06-22",
-            "latest_timestamp": "2024-07-01",
+            "earliest_timestamp": "Thu, 18 Jul 2024 15:32:28 GMT",
+            "latest_timestamp": "Sat, 20 Jul 2024 18:22:11 GMT",
             "processed_by_service_at": None,
             "batch_id": None,
             "is_a_batched_report": False,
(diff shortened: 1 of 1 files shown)


tests/app/main/views/test_unsubscribe_requests.py
```diff
@@ -339,8 +341,8 @@ def test_create_unsubscribe_request_report_creates_batched_report(client_request
         SERVICE_ONE_ID,
         {
             "count": 34,
-            "earliest_timestamp": "2024-06-22",
-            "latest_timestamp": "2024-07-01",
+            "earliest_timestamp": ANY,
+            "latest_timestamp": ANY,
```
quis, line 345: Using `ANY` here makes the tests pass but I’m not sure if it fixes the problem. The API will still be called with a `datetime` object not a string.

It also makes it easy to introduce bugs in the future, for example I could swap `earliest` and `latest` and the tests would still pass

tests/app/main/views/test_unsubscribe_requests.py
```diff
@@ -320,8 +322,8 @@ def test_create_unsubscribe_request_report_creates_batched_report(client_request
     summary_data = [
         {
             "count": 34,
-            "earliest_timestamp": "2024-06-22",
-            "latest_timestamp": "2024-07-01",
+            "earliest_timestamp": "2024-06-22 10:30:21",
+            "latest_timestamp": "2024-07-01 21:19:21",
```
quis, line 326: Can you check the datetime format here is what’s actually returned by the API?

BlessedDev: At this stage the API returns just the `report_id` and I have mocked the API call. The purpose of the test was to ensure that `service_api_client.create_unsubscribe_request_report` was called with the expected values.

quis: If the API is only returning the `id` then that’s all that should be in the mocked data here.

But I think it’s returning more – is it not the data in the `dict` here?
https://github.com/alphagov/notifications-api/blob/4cd15052a18819ecda8aca6734e93365b99c6cc8/app/service/rest.py#L1173-L1187

BlessedDev: Oh I see. This test was specifically for the `create_unsubscribe_request_report` endpoint, there is a separate test that covers the `download_unsubscribe_request_report` endpoint and that test covers what the API returns https://github.com/alphagov/notifications-admin/blob/97e7ab259b5cd38aa984ea257ef26e894b44b069/tests/app/main/views/test_unsubscribe_requests.py#L349.

quis: OK, I can see that is mocked elsewhere. Can you still check the date format here matches what the API returns. I feel like if we don’t get this matching we’ll end up with some BST/UTC/GMT weirdness

BlessedDev: I have updated both tests test_data to be exact replicas of the what the API returns.

quis (approved):
Let’s get this merged now to fix the build

PR #5171, opened 2024-08-07
Only allow the agreement to be accepted by people with a public sector email address

quis:


leohemsted (approved):
Code looks good pending product sign-off

PR #5172, opened 2024-08-08
Enable Design System error-summary validation pattern on tour pages

kr8n3r:
Enable error-summary Design System validation pattern on tour pages to match the rest of the admin.

Screenshot 2024-08-08 at 12 55 24

Fixes: https://trello.com/c/D5EEpWDV/689-tour-pages-validation-doesnt-have-meet-design-system-requirements

As field's auto focus interferes with the required error-summary autofocus logic was added to only autofucs when there's to form error.

Additionally, this also fixes the bug on one-off-send journey where autofocus should be set, but it isn't working

Before

Screenshot 2024-08-08 at 12 44 47
Screenshot 2024-08-08 at 12 46 36

After

Screenshot 2024-08-08 at 12 43 10
Screenshot 2024-08-08 at 12 43 17

### DIFF
--- a/app/templates/views/send-test.html
+++ b/app/templates/views/send-test.html
@@ -18,7 +18,7 @@
 
   {% call form_wrapper(
     class='js-stick-at-top-when-scrolling send-one-off-form' if template.template_type != 'sms' else 'send-one-off-form',
-    module="autofocus" if form.placeholder_value.label.text not in ['email address', 'phone number'] else None,
+    module="autofocus" if not form.errors else None,
     data_kwargs={'force-focus': True}
   ) %}
     <div class="govuk-grid-row">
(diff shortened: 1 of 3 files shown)


app/templates/views/send-test.html
```diff
@@ -18,8 +18,7 @@
 
   {% call form_wrapper(
     class='js-stick-at-top-when-scrolling send-one-off-form' if template.template_type != 'sms' else 'send-one-off-form',
-    module="autofocus" if form.placeholder_value.label.text not in ['email address', 'phone number'] else None,
-    data_kwargs={'force-focus': True}
+    module="autofocus" if not form.errors | length > 0 else None
```
tombye, line 21: I think this can just check `form.errors` as jinja treats empty lists as falsey.
```suggestion
    module="autofocus" if not form.errors else None
```

tombye (changes requested):
Only one thing I noticed and very minor + a bit subjective but otherwise this is nice work and good to go 👍🏻

tombye (approved):
All looks good now 👍🏻

PR #5174, opened 2024-08-09
Change ‘Your profile’ to ‘Your account’

karlchillmaid:
We use the terms ‘profile’ and ‘account’ inconsistently.

This PR replaces the less common ‘Your profile’ with ‘Your account’.

This is only a superficial change to the public-facing content. We still use ‘profile’ a lot in the code. And it looks like we use ‘account’ to mean ‘service’ in the code too. We’ll need to tackle this as part of our attempts to be clear and consistent about the meaning of ‘account’ and ‘service’.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -965,7 +965,7 @@ class PermissionsForm(StripWhitespaceForm):
                     {
                         "hint": {
                             "text": "Not available because this team member has not added a "
-                            "phone number to their profile"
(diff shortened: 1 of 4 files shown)


kr8n3r:
superseeded by https://github.com/alphagov/notifications-admin/pull/5395

tombye:
Karl confirmed, this is no longer needed as the changes are now on live, just as @kr8n3r said.

PR #5181, opened 2024-08-16
Sms to landline validation one shot and csv flows

rparke:
Adds support on admin for one shot and csv flows to send to UK landlines if the service has the sms_to_uk_landline permission enabled.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -82,7 +82,9 @@ from app.main.validators import (
     StringsNotAllowed,
     ValidEmail,
     ValidGovEmail,
+    ValidInternationalOrUKLandline,
     ValidInternationalPhoneNumber,
+    ValidUKLandlineNumber,
     ValidUKMobileNumber,
 )
 from app.models.branding import (
@@ -244,8 +246,18 @@ def uk_mobile_number(label="Mobile number"):
     return PhoneNumber(label, validators=[DataRequired(message="Cannot be empty"), ValidUKMobileNumber()])
 
 
+def uk_landline_number(label="Mobile number"):
+    return PhoneNumber(label, validators=[NotifyDataRequired(thing="a mobile number"), ValidUKLandlineNumber()])
+
+
 def international_phone_number(label="Mobile number"):
     return PhoneNumber(label, validators=[NotifyDataRequired(thing="a mobile number"), ValidInternationalPhoneNumber()])
+
+
+def international_or_uk_landline_number(label="Mobile number"):
+    return PhoneNumber(
+        label, validators=[NotifyDataRequired(thing="a mobile number"), ValidInternationalOrUKLandline()]
+    )
 
 
 def make_password_field(label="Password", thing="a password", validate_length=True):
@@ -2264,12 +2276,18 @@ def get_placeholder_form_instance(
     dict_to_populate_from,
     template_type,
     allow_international_phone_numbers=False,
+    sms_to_uk_landline=False,
 ):
     if InsensitiveDict.make_key(placeholder_name) == "emailaddress" and template_type == "email":
         field = make_email_address_field(label=placeholder_name, gov_user=False, thing="an email address")
     elif InsensitiveDict.make_key(placeholder_name) == "phonenumber" and template_type == "sms":
-        if allow_international_phone_numbers:
+        if sms_to_uk_landline & allow_international_phone_numbers:
+            field = international_or_uk_landline_number(label=placeholder_name)
+        elif sms_to_uk_landline:
+            field = uk_landline_number(label=placeholder_name)
+        elif allow_international_phone_numbers:
             field = international_phone_number(label=placeholder_name)
+
         else:
             field = uk_mobile_number(label=placeholder_name)
     else:
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -243,14 +243,20 @@ def valid_phone_number(label="Mobile number", international=False, sms_to_uk_lan
     if not (sms_to_uk_landline or international):
         return PhoneNumber(
             label,
-            validators=[DataRequired(message="Cannot be empty"), ValidPhoneNumber(is_international=international)],
+            validators=[
+                DataRequired(message="Cannot be empty"),
+                ValidPhoneNumber(allow_international_sms=international),
+            ],
         )
     else:
         return PhoneNumber(
             label,
             validators=[
                 NotifyDataRequired(thing="a mobile number"),
-                ValidPhoneNumber(sms_to_landline=sms_to_uk_landline, is_international=international),
+                ValidPhoneNumber(
+                    allow_sms_to_uk_landlines=sms_to_uk_landline,
+                    allow_international_sms=international,
+                ),
             ],
         )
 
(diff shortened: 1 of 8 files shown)


app/main/forms.py
```diff
@@ -2263,12 +2275,18 @@ def get_placeholder_form_instance(
     dict_to_populate_from,
     template_type,
     allow_international_phone_numbers=False,
+    sms_to_uk_landline=False,
 ):
     if InsensitiveDict.make_key(placeholder_name) == "emailaddress" and template_type == "email":
         field = make_email_address_field(label=placeholder_name, gov_user=False, thing="an email address")
     elif InsensitiveDict.make_key(placeholder_name) == "phonenumber" and template_type == "sms":
-        if allow_international_phone_numbers:
+        if sms_to_uk_landline & allow_international_phone_numbers:
```
rparke, line 2283: I don't like this pattern. It's probably fine for 4 different combinations of permissions (can do everything, can only send to uk mobiles, can send to uk and International mobiles, can send to uk mobiles and landlines), but it isn't manageable for even slightly larger sets of permissions

app/main/forms.py
```diff
@@ -2263,12 +2275,18 @@ def get_placeholder_form_instance(
...
+            field = international_or_uk_landline_number(label=placeholder_name)
+        elif sms_to_uk_landline:
+            field = uk_landline_number(label=placeholder_name)
+        elif allow_international_phone_numbers:
             field = international_phone_number(label=placeholder_name)
```
rparke, line 2288: I don't like this pattern. It's probably fine for 4 different combinations of permissions (can do everything, can only send to uk mobiles, can send to uk and International mobiles, can send to uk mobiles and landlines), but it isn't manageable for even slightly larger sets of permissions

rparke: Some sort of `phone_number_field_factory` method and corresponding class we create dynamically could work(?)

```
def phone_number_field_factory(label, sms_to_landline, ....):
   class ValidPhoneNumberPlaceholder(ValidPhoneNumber):
       sms_to_landline = sms_to_landline
    return PhoneNumber(
        label, validators=[NotifyDataRequired(thing="a mobile number"), ValidPhoneNumberPlaceholder()]
    )
```

app/main/forms.py
```diff
@@ -2263,12 +2275,18 @@ def get_placeholder_form_instance(
     dict_to_populate_from,
     template_type,
     allow_international_phone_numbers=False,
+    sms_to_uk_landline=False,
```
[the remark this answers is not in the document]
rparke: Ah right, didn't know whether it could be used with both a class and an instance. That's easier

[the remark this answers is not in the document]
rparke: I'll change to `allow_`

app/main/forms.py
rparke: Made this change. Didn't realise that it was an instance of the class vs the class we were passing as a validator, so yeh that's all been refactored into the init method now.

app/main/forms.py
```diff
@@ -242,22 +240,17 @@ class PhoneNumber(GovukTextInputFieldMixin, TelField):
...
-    )
+def valid_phone_number(label="Mobile number", international=False, sms_to_uk_landline=False):
+    if not (sms_to_uk_landline or international):
+        return PhoneNumber(label, validators=[DataRequired(message="Cannot be empty"), ValidPhoneNumber()])
+    else:
+        return PhoneNumber(
+            label,
+            validators=[
+                NotifyDataRequired(thing="a mobile number"),
+                ValidPhoneNumber(sms_to_landline=sms_to_uk_landline, is_international=international),
```
rparke, line 251: This is currently needed because we expect different error messages for different permissions.

app/main/validators.py
```diff
@@ -82,8 +82,10 @@ def __call__(self, form, field):
 
 
 class ValidPhoneNumber:
-    is_international = False
-    message = None
+    def __init__(self, is_international=False, sms_to_landline=False, message=None):
```
leohemsted, line 85: what do you think about changing to `allow_international` and `allow_sms_to_landline`?

i just want it to be really explicit that we aren't defining what the input is, but we're defining what the input is allowed to be once we've validated it. i think it made sense when this was just a class value defined once and never referred to outside of the class definition, but now it's being mentioned in form definitions i think it might be nice to be a bit clearer

(also in line with, for example, the arguments in `get_placeholder_form_instance`)

app/main/views/send.py
```diff
@@ -616,6 +616,7 @@ def _check_messages(service_id, template_id, upload_id, preview_row):
         ),
         remaining_messages=remaining_messages,
         allow_international_sms=current_service.has_permission("international_sms"),
+        allow_sms_to_uk_landline=current_service.has_permission("sms_to_uk_landlines"),
```
ipdigital24, line 619: could this be a const `sms_to_uk_landlines`? it's repeating.

app/main/validators.py
```diff
@@ -96,7 +98,11 @@ class ValidPhoneNumber:
     def __call__(self, form, field):
         try:
             if field.data:
-                validate_phone_number(field.data, international=self.is_international)
+                if self.sms_to_landline:
+                    PhoneNumber(field.data, allow_international=self.is_international)
+                else:
+                    validate_phone_number(field.data, international=self.is_international)
```
ipdigital24, line 104: any reason why an object is being created only on if part but not here?

rparke: Instantiating the object triggers validation. It'll raise an exception if the phone number isn't valid.

We can explicitly call the validation method, but this is redundant.

In some ways this isn't great, we should want to be explicit when we're using the class to check validation. The reasoning behind this design decision is that a PhoneNumber object, if it exists, should represent a phone number that is valid for that service.

Essentially, if you have a PhoneNumber object, you can send notifications to it.

app/main/views/send.py
rparke: Which bit do you think should be a constant? In what scope?

This can't be a global constant because its value changes depending on what service is sending notifications.

Some services can, some can't.

rparke: It is common to a couple of objects though so I'm open to any suggestions of making it more DRY

app/main/validators.py
rparke: We could be even more explicit, change it to something like: service_has_<permission name>.

It's probably going to make life easier for whoever next has to touch this code "why are we setting this? Oh that's why" should be a lot more obvious

app/main/validators.py
ipdigital24: if it is possible and not too much of hassle, following similar pattern on both sides of if else would make it easier to understand.

rparke: If I understand what you're asking, this would entail re-writing the entire `PhoneNumber` class, introducing a breaking change, and applying that fix to both admin and api.

This might be a good idea, but I don't think this PR is the place to do it. We already have a ticket for removing the old validation code (`validate_phone_number`). The branching logic above isn't going to exist in a few weeks time anyway. My preference is for all phone number validation to use the `PhoneNumber` class/object before we start making breaking changes to how this class is instantiated/which methods from it are called.

app/main/views/send.py
quis: Not using raw strings here is a good idea. There are several ways we could change the `current_service` object to accommodate this, and enforce that the value is a genuine permission. Such a refactor should definitely be a separate pull request though because there are 30+ examples of `current_service.has_permission()` in the codebase, as well as 10+ examples of `x in current_service.permissions`.

rparke: A permissions object would maybe be nice:

```
class Permission():
   def __init__(self, permission_string):
      if self._check_valid_permission(permission_string):
         self.permission = permission_string
      else:
         raise InvalidPermissionError
```

But I agree, that refactor is beyond the scope of this PR. For better or worse, raw strings **are** the pattern we're currently using throughout admin.

app/main/validators.py
ipdigital24: okay

leohemsted (approved):

leohemsted (approved):

PR #5182, opened 2024-08-16
fix: manual csv download without row limit

joybytes:
## PR Summary
- previously on notifications-api, an endpoint was created to count the notifications_sent by a service within its retention period
- here, this endpoint is used to check notifications_count does not exceed `max_notifications_for_download` which has been set to 250k
- if the number of notifications does exceed this value, we do not want to show the "Download your report" link anymore, instead a message for "Contact support to download this report" is presented which will direct the user to the "ask-question-give-feedback" page

**Default behaviour:**
Screenshot 2024-08-16 at 15 00 42

**Removed button with message for contacting support**
Screenshot 2024-08-19 at 16 40 04

## Related task 
[advise service users to use api instead of using admin page to download notifications if volume is very large](https://trello.com/c/VYDXkKxu/911-advise-service-users-to-use-api-instead-of-using-admin-page-to-download-notifications-if-volume-is-very-large)

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -182,9 +182,20 @@ def view_job_updates(service_id, job_id):
 @main.route("/services/<uuid:service_id>/notifications/<template_type:message_type>", methods=["GET", "POST"])
 @user_has_permissions()
 def view_notifications(service_id, message_type=None):
+    partials_data = _get_notifications_dashboard_partials_data(service_id, message_type)
+
+    notifications_count = notification_api_client.get_notifications_count_for_service(
+        service_id,
+        message_type,
+        partials_data["service_data_retention_days"],
+    )
+
+    max_notifications_for_download = 250000
+    can_download = notifications_count <= max_notifications_for_download
+
     return render_template(
         "views/notifications.html",
-        partials=_get_notifications_dashboard_partials_data(service_id, message_type),
+        partials=partials_data,
         message_type=message_type,
         status=request.args.get("status") or "sending,delivered,failed",
         page=request.args.get("page", 1),
@@ -210,6 +221,7 @@ def view_notifications(service_id, message_type=None):
             message_type=message_type,
             status=request.args.get("status"),
         ),
+        can_download=can_download,
     )
 
 
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -192,6 +192,15 @@ def view_notifications(service_id, message_type=None):
 
     max_notifications_for_download = 250000
     can_download = notifications_count <= max_notifications_for_download
+    download_link = None
+
+    if can_download:
+        download_link = url_for(
+            ".download_notifications_csv",
+            service_id=current_service.id,
+            message_type=message_type,
+            status=request.args.get("status"),
+        )
 
     return render_template(
         "views/notifications.html",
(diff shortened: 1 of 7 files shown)


joybytes:
Actions:
- remove csv button
- display get in touch with support with link

karlchillmaid (commented):
Content changes look good.

I haven’t reviewed the code.

app/main/views/jobs.py
```diff
@@ -182,9 +182,29 @@ def view_job_updates(service_id, job_id):
...
 def view_notifications(service_id, message_type=None):
+    partials_data = _get_notifications_dashboard_partials_data(service_id, message_type)
+
+    notifications_count = notification_api_client.get_notifications_count_for_service(
+        service_id,
+        message_type,
+        partials_data["service_data_retention_days"],
+    )
+
+    max_notifications_for_download = 250000
```
ipdigital24, line 193: could this be moved to `main/app/config.py`?

app/main/views/jobs.py
```diff
@@ -182,9 +182,29 @@ def view_job_updates(service_id, job_id):
 @main.route("/services/<uuid:service_id>/notifications/<template_type:message_type>", methods=["GET", "POST"])
 @user_has_permissions()
 def view_notifications(service_id, message_type=None):
+    partials_data = _get_notifications_dashboard_partials_data(service_id, message_type)
+
+    notifications_count = notification_api_client.get_notifications_count_for_service(
+        service_id,
+        message_type,
+        partials_data["service_data_retention_days"],
```
ipdigital24, line 190: any merit in moving this const to `main/app/config.py`?

app/main/views/jobs.py
joybytes: I looked into the `[app/config.py](https://github.com/alphagov/notifications-admin/blob/da7b91cd3d6da7f313d5b56546e4e9e134d672f5/app/config.py)` and it seems like there were mostly environment variables

I think the constants file would be the appropriate location for this https://github.com/alphagov/notifications-admin/pull/5182/commits/00dcadc72e0b475e5ac80b102131834b8fa7713f

app/main/views/jobs.py
joybytes: I don't think I can find the path for this file you are mentioning 
the only config.py I see in admin is https://github.com/alphagov/notifications-admin/blob/da7b91cd3d6da7f313d5b56546e4e9e134d672f5/app/config.py

ipdigital24: just rethinking based on your previous comment above, how about moving to constants.py like `MAX_NOTIFICATION_FOR_DOWNLOAD` and `PERMISSION_CAN_MAKE_SERVICES_LIVE`

ipdigital24: we leave it as is and discuss things like this on tech forum

leohemsted (approved):
i'm still concerned about the words in the link and how we manage that. my two concerns are:

1. i think that people won't know what to write in the support form
2. i think people who are used to downloading the report but now cant, vs people that are new to notify, will react in different ways. i think people who are new to notify might not know _what_ the link is for and what they can ask us to help them with

i also think what we as a team do when someone sends us a support ticket is still not clarified and if we do custom data extracts and if we ask them what columns they need or just replicate the original report is not clear etc.

but i think if the rest of us are happy with the content despite these issues, and with us taking the burden of helping people download the report, the code for this is good and we're ready to go.

PR #5183, opened 2024-08-19
Rewrite Collapsible checkboxes as an ES Module

kr8n3r:
Make collapsible checkboxes component work better with screenreader users. Addresses https://trello.com/c/izqbbBLc/891-collapsible-components-work-poorly-for-screen-reader-users

## What

Primarily, changes needed to be made to the functionality as screenreader users were experiencing unexpected behaviour and labelling. Described in detail in the report, linked to the Trello ticket.

Secondly, we're in the process of migrating our legacy code to ES modules and as proposed changes in functionality are such that, while they could have be done in the original module, the module is a good candidate for a rewrite and removal
of dependency on JQuery along the way.

Functional changes

- Toggle button is now inserted directly before the checkboxes container and it instead of the fieldset is being toggled.

This was it follows the pattern accordion uses, where toggled content follows immediately in the DOM.
With this there is no longer the  need to auto focus on the group of checkboxes.

- Toggle button indicates collapsed / expanded state and text now includes legend text

Showing the state visually as well as making it clearer to screen reader users what this section is.

- Removal of "done" button and sticky footer

"Done" button was there to collapse the content and it was wrapped in a "sticky" container to be in view for mouse users to be able to close the section quicker on longer lists.

Removing the "done" button entirely addresses the concern of "...describing the done button as expanded indicates that I have selected that button and expanded the content...this is no the expected method of expanding content and it was very confusing."

The second issue with having a "done" or "collapse this section button" is that once triggered, there isn't a consistent next tabbabble element to place focus on, as this component is used in two different places where next form elements
differ. Placing the focus back on to the toggle button would not make sense either.

The removal of making the button sticky is mostly down to above removal of functionality.
While we could insert a none tabbable element there for visual users, the script itself is not written in a way that can be used is ES modules as well as in legacy code, where it still used.

There have been attempts to use Intersection Observer to make the toggle button visible in view on long lists, it would not be "sticky" at the bottom nor would it be aware and introduce scroll to avoid overlaying a focussed checkbox.

## Appearance

**Template folder management view**

Screenshot 2024-08-29 at 12 15 49
Screenshot 2024-08-29 at 12 15 54
Screenshot 2024-08-29 at 12 16 01

**User access management view**

Screenshot 2024-08-29 at 12 14 42
Screenshot 2024-08-29 at 12 14 47

**No change to built static JS files with updated babel-core**
Screenshot 2024-08-19 at 12 11 34


kr8n3r:
@quis when you're back to speed after your holiday, if you have any thoughts on this, please feel free to comment

quis:
I’m ok with this, if we can use the same arrow style as the details component:
> [ ▶ Collapsed ]
>
> [ ▼ Expanded ]
> ☑️ 
> ☑️ 

But then I wonder if we shouldn’t just be using the details component anyway. Thoughts?

kr8n3r:
quick prototype using `details`
https://github.com/alphagov/notifications-admin/compare/collapsible-checkboxes-with-details?expand=1

Screenshot 2024-08-28 at 12 10 37
Screenshot 2024-08-28 at 12 10 47
Screenshot 2024-08-28 at 12 10 42

package.json
```diff
@@ -31,8 +31,9 @@
     "timeago": "1.6.5"
   },
   "devDependencies": {
-    "@babel/core": "7.4.0",
+    "@babel/core": "^7.25.2",
```
tombye, line 34: Can we lock this down to an exact version number please, to match the others? Same with all others added in the same way.

tests/javascripts/collapsibleCheckboxes.test.mjs
```diff
@@ -1,21 +1,14 @@
...
-beforeAll(() => {
-  // TODO: remove this when tests for sticky JS are written
-  require('../../app/assets/javascripts/stick-to-window-when-scrolling.js');
-
-  require('../../app/assets/javascripts/collapsibleCheckboxes.js');
-});
+import CollapsibleCheckboxes from '../../app/assets/javascripts/esm/collapsible-checkboxes.mjs'
+import * as helpers from './support/helpers.js'
 
 afterAll(() => {
```
tombye, line 4: Doesn't look like we need this `afterAll` now.

tombye (changes requested):
Spotted a few things, all commented, but this is some nice work overall and thanks for doing all the wrangling with jest and babel.

If we leave it as a button, not a `<details>`, I think we should make the arrow the same as the day/time picker:

image

## Things from cross browser testing

The button seems to start with spacing above and then move on the first click on Safari on MacOS and iOS. This is on MacOS:

moving_button_macos

I noticed the folder icon for the summary still appearing on IE11. Not a game changer but didn't think this was intentional:

image

## Suggestion for changing the order of the HTML

The other thing I'd suggest changing is to move the button out of the fieldset. It's currently a bit strange to hear the button after the legend:
- "[Group] Folders this team member can see"
- "Choose folders [button, expanded]"
- "[Unordered list]" (of checkboxes)

Moved outside, it's more like:
- "Choose folders [button, expanded]"
- "[Group] Folders this team member can see"
- "[Unordered list]" (of checkboxes)

...which I think makes more sense as the button label sets the expectation of folders and the group label following it meets that expectation.

kr8n3r:
> Spotted a few things, all commented, but this is some nice work overall and thanks for doing all the wrangling with jest and babel.
> 
> If we leave it as a button, not a `<details>`, I think we should make the arrow the same as the day/time picker:
> 
> image
> ## Things from cross browser testing
> The button seems to start with spacing above and then move on the first click on Safari on MacOS and iOS. This is on MacOS:
> 
> moving_button_macos [ moving_button_macos ](https://private-user-images.githubusercontent.com/87140/362359575-0d108c34-f798-465f-9242-ed1cc40a0b86.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ5Mjg3MzIsIm5iZiI6MTcyNDkyODQzMiwicGF0aCI6Ii84NzE0MC8zNjIzNTk1NzUtMGQxMDhjMzQtZjc5OC00NjVmLTkyNDItZWQxY2M0MGEwYjg2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODI5VDEwNDcxMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRkMmU4N2RiNTc1MzJmNjFmMTY4ZmU2ZTQyNTI1YjU3MzlmZGY2ZTlkNWVlMmQ5N2Q4OTE0NjZjNmYwMTBmMDkmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.I6YI532crv2geZ_h9w1zDZmanlfeOby8b7AGFspN5pQ) [ ](https://private-user-images.githubusercontent.com/87140/362359575-0d108c34-f798-465f-9242-ed1cc40a0b86.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ5Mjg3MzIsIm5iZiI6MTcyNDkyODQzMiwicGF0aCI6Ii84NzE0MC8zNjIzNTk1NzUtMGQxMDhjMzQtZjc5OC00NjVmLTkyNDItZWQxY2M0MGEwYjg2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODI5VDEwNDcxMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRkMmU4N2RiNTc1MzJmNjFmMTY4ZmU2ZTQyNTI1YjU3MzlmZGY2ZTlkNWVlMmQ5N2Q4OTE0NjZjNmYwMTBmMDkmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.I6YI532crv2geZ_h9w1zDZmanlfeOby8b7AGFspN5pQ)
> 
> I noticed the folder icon for the summary still appearing on IE11. Not a game changer but didn't think this was intentional:
> 
> image
> ## Suggestion for changing the order of the HTML
> The other thing I'd suggest changing is to move the button out of the fieldset. It's currently a bit strange to hear the button after the legend:
> 
> * "[Group] Folders this team member can see"
> * "Choose folders [button, expanded]"
> * "[Unordered list]" (of checkboxes)
> 
> Moved outside, it's more like:
> 
> * "Choose folders [button, expanded]"
> * "[Group] Folders this team member can see"
> * "[Unordered list]" (of checkboxes)
> 
> ...which I think makes more sense as the button label sets the expectation of folders and the group label following it meets that expectation.

- updated arrow design
- not showing folder icon in non-supported browsers
- fixed Safari spacing
- moved the button before the fieldset

tombye (approved):
All issues fixed and it now makes much more sense when used with a screen reader due to the changes to the order of the elements. Nice work, ready to go now 👍🏻

PR #5187, opened 2024-08-29
Utils Bump to 86.2.0 - adding allow landline flag

rparke:
Updated phonenumber validation code to use allow_landline flag
In order to migrate phonenumber validation code to use the PhoneNumber utils class
    we need to support passing the SMS_TO_UK_LANDLINES permissions as an attribute to
    instances of the class so that it can correctly validate depending on the permissions
    service has. This is also required to fully migrate all phonenumebe validation logic
    away to PhoneNumber objects.

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -103,7 +103,11 @@ class ValidPhoneNumber:
         try:
             if field.data:
                 if self.allow_sms_to_uk_landlines:
-                    PhoneNumber(field.data, allow_international=self.allow_international_sms)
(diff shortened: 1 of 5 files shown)


BlessedDev:
@rparke 
Looks good. `notifications-utils` is now at `86.2.0`

rparke:
> @rparke Looks good. `notifications-utils` is now at `86.2.0`

Thanks! I'll just bump the version

BlessedDev (approved):

PR #5189, opened 2024-08-30
Ensure Zendesk tickets about phishing senders are not sent to users 

klssmith:
Setting the user details when creating the creating a Zendesk ticket about potential phishing senders means that the ticket appears to come _from_ them and they also see the ticket. The ticket should be internal only, so this removes the user details from the ticket object. The user and user's service are still visible in the ticket note.

Also fixes the page heading:
**Before**
Screenshot 2024-08-30 at 08 48 17

**After**
Screenshot 2024-08-30 at 08 48 08

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -2,7 +2,6 @@ import re
 from abc import ABC, abstractmethod
 
 from flask import current_app, render_template
-from flask_login import current_user
 from notifications_utils.clients.zendesk.zendesk_client import NotifySupportTicket, NotifyTicketType
 from notifications_utils.field import Field
 from notifications_utils.formatters import formatted_list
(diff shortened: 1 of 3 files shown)


app/main/validators.py
```diff
@@ -204,28 +203,12 @@ def create_phishing_senderid_zendesk_ticket(senderID=None):
...
-        user_name=current_user.name,
-        user_email=current_user.email_address,
-        service_id=current_service.id,
         notify_task_type="notify_task_blocked_sender",
     )
     zendesk_client.send_ticket_to_zendesk(ticket)
 
 
 class IsNotAPotentiallyMaliciousSenderID:
-    potentially_malicious_sender_ids = [
```
klssmith, line 216: This was removed since it's no longer being used

quis (approved):

PR #5191, opened 2024-09-02
Rejig dates for completed and deleted

quis:
Makes it so the information about how long the report is available for:
- matches the [updated retention policy](https://github.com/alphagov/notifications-api/pull/4167#issuecomment-2324372565)
- is relative (for example _3 days_ not _18 October 2024_), like we do elsewhere 
- is consistently in the same place on the page 

Before | After
---|---
image | image
image | image
image | image

Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/4167

### DIFF
--- a/app/assets/stylesheets/components/checkboxes.scss
+++ b/app/assets/stylesheets/components/checkboxes.scss
@@ -163,14 +163,14 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
 
 .govuk-checkboxes__item--single-with-hint {
 
-  margin-top: -10px;
+  margin-top: -1 * govuk-spacing(2);
 
   .govuk-checkboxes__label::before {
-    margin-top: 10px;
+    top: govuk-spacing(2) + 2px;  // 2px is existing top position
   }
 
   .govuk-checkboxes__label::after {
-    margin-top: 10px;
+    top: govuk-spacing(2) + 13px;  // 13px is existing top position
   }
 
 }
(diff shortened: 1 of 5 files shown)


kr8n3r (commented):
both pseudo elements are absolutely positioned with a `top` adjustment, so just overwrite that, instead of adding margin.
if you want to change 10px to govuk-spacing(2), but not too fussed about it as we have other magic numbers

app/assets/stylesheets/components/checkboxes.scss
```diff
@@ -160,3 +160,17 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
     transform:translateY(1px) rotate(225deg) scale(1);
   }
 }
+
+.govuk-checkboxes__item--single-with-hint {
+
+  margin-top: -10px;
```
quis, line 166:
```suggestion
  margin-top: -1 * govuk-spacing(2);
```

app/assets/stylesheets/components/checkboxes.scss
```diff
@@ -160,3 +160,17 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
...
+
+  .govuk-checkboxes__label::before {
+    margin-top: 10px;
```
quis, line 169:
```suggestion
    top: govuk-spacing(2) + 2px;  // 2px is existing top position
```

app/assets/stylesheets/components/checkboxes.scss
```diff
@@ -160,3 +160,17 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
...
+  }
+
+  .govuk-checkboxes__label::after {
+    margin-top: 10px;
```
quis, line 173:
```suggestion
    top: govuk-spacing(2) + 13px;  // 13px is existing top position
```

kr8n3r (approved):

kr8n3r (approved):

PR #5192, opened 2024-09-03
Fix back behaviour when returning to the set-sender page after sending an email or text message

klssmith:
**Fix back links from set-sender and start of one-off journey**
If you are going through the flow to send a message and you have selected an SMS or email sender, the back link on the next page (`/one-off/step-0`) was taking you back to the template or list of templates and skipping the set-sender page. It now takes you back to the set-sender page.

The back link on the set-sender page was hardcoded to always go back to `.view_template`, but not all users have permission to view a template so it now takes you back to the template list if you don't.

**Remember "sender_id" on set-sender page if going "back"**
The "set-sender" page was setting `session["sender_id"]` to `None`, but if you are returning to the page having clicked "back" when sending, we want to be able to remember your choice and to pre-populate the form with it.

This change now means that we don't clear the `sender_id` from the session if you're going back to the set-sender page and instead highlight the option that was previously selected.

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -181,6 +181,18 @@ def get_example_csv(service_id, template_id):
     )
 
 
+def _should_show_set_sender_page(service_id, template) -> bool:
+    if template.template_type == "letter":
+        return False
+
+    sender_details = get_sender_details(service_id, template.template_type)
+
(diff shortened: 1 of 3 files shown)


kr8n3r (approved):
changes work as expected

PR #5194, opened 2024-09-04
Change port mapping to use the loopback address

klssmith:
Docker defaults to using `0.0.0.0`, the default route listening on all network interfaces, as opposed to the loopback address (`127.0.0.1`). This has security implications and we have been asked to change it.


tombye (approved):
Makes sense, and matches [the approach shown in Docker's own doc examples](https://docs.docker.com/reference/cli/docker/container/run/#publish) so 👍🏻

PR #5203, opened 2024-09-10
Add confirmation step to public-facing unsubscribe page

quis:
Before | After
---|---
[redirect] | image
image | image
image | image


tests/app/main/views/test_unsubscribe_requests.py
```diff
@@ -615,5 +615,5 @@ def _post(notification_id, token):
 
     mock_post.assert_called_once_with(f"/unsubscribe/{fake_uuid}/abc123", None)
 
-    assert normalize_spaces(page.select_one("h1").text) == "Unsubscribe"
-    assert normalize_spaces(page.select_one("p").text) == "Your request to unsubscribe could not be processed."
+    assert normalize_spaces(page.select_one("h1").text) == "There is a problem"
+    assert normalize_spaces(page.select_one("p").text) == "We could not process your request to unsubscribe."
```
saimaghafoor, line 628: Because a user would only get this error by copy/pasting the link incorrectly, we could keep this in-line with page not found errors , for example:

"Page not found"
"Please check the link and try again."

BlessedDev (approved):
I have tested this and it works as expected.

app/templates/views/unsubscribe-failed.html
```diff
@@ -7,10 +7,10 @@
...
   <h1>
-    Unsubscribe
+    There is a problem
   </h1>
   <p>
-    Your request to unsubscribe could not be processed.
+    We could not process your request to unsubscribe.
   </p>
   <p>
     Please check the link and try again.
```
saimaghafoor, line 16: Change to:

We could not unsubscribe you from these emails. 

If you pasted the web address from an email, check you copied the entire address.

(Sorry, no change suggested icon appears for me so doing it this way)

PR #5206, opened 2024-09-16
Update approve join request template when user exists

spatel033:
When a user successfully requests to join a service they should get an email request has been approved. More details on [this card](https://trello.com/c/FHhogwve/904-send-a-different-invite-email-if-the-user-has-requested-to-join-a-service).

### DIFF
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -46,6 +46,8 @@ def invite_user(service_id, user_id=None):
         folder_permissions=[f["id"] for f in current_service.all_template_folders],
     )
 
+    is_join_request = False if user_id is None else True
+
     if user_id:
         user_to_invite = User.from_id(user_id)
         if user_to_invite.belongs_to_service(current_service.id):
@@ -78,6 +80,7 @@ def invite_user(service_id, user_id=None):
             form.permissions,
             form.login_authentication.data,
             form.folder_permissions.data,
+            is_join_request,
         )
 
         flash(f"Invite sent to {invited_user.email_address}", "default_with_tick")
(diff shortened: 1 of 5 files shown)


app/main/views/manage_users.py
```diff
@@ -78,6 +80,7 @@ def invite_user(service_id, user_id=None):
             form.permissions,
             form.login_authentication.data,
             form.folder_permissions.data,
+            is_join_request,
```
quis, line 83:
```suggestion
            is_join_request=bool(user_id),
```

app/main/views/manage_users.py
```diff
@@ -46,6 +46,8 @@ def invite_user(service_id, user_id=None):
         folder_permissions=[f["id"] for f in current_service.all_template_folders],
     )
 
+    is_join_request = False if user_id is None else True
+
```
quis, line 50:
```suggestion
```

app/models/user.py
```diff
@@ -489,22 +489,11 @@ def __init__(self, _dict):
...
     @classmethod
     def create(
-        cls,
-        invite_from_id,
-        service_id,
-        email_address,
-        permissions,
-        auth_type,
-        folder_permissions,
+        cls, invite_from_id, service_id, email_address, permissions, auth_type, folder_permissions, is_join_request
```
quis, line 492: This has so many arguments now that it would be good to enforce keywords arguments (this avoid accidentally mixing them up)
```suggestion
        cls, *, invite_from_id, service_id, email_address, permissions, auth_type, folder_permissions, is_join_request
```

PR #5209, opened 2024-09-17
Send `user_created_at` field to Zendesk

klssmith:
We want to populate a new field on the Notify Zendesk form showing when
a Notify user account was created.

For users who are logged in when using the feedback form, we now pass
the UTC datetime that their Notify user account was created as an
argument to the support ticket. For non-logged in users we can just pass
`None` as the argument, which ensures the new date picker won't get
populated.

The `user_created_at` argument is optional.

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -141,6 +141,7 @@ def feedback(ticket_type):
             org_id=current_service.organisation_id if current_service else None,
             org_type=current_service.organisation_type if current_service else None,
             service_id=current_service.id if current_service else None,
+            user_created_at=get_user_created_at_for_ticket(current_user),
         )
         zendesk_ticket_id = zendesk_client.send_ticket_to_zendesk(ticket)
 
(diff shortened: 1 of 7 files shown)


app/main/views/feedback.py
```diff
@@ -248,3 +249,10 @@ def get_zendesk_ticket_type(ticket_type):
         return NotifySupportTicket.TYPE_INCIDENT
 
     return NotifySupportTicket.TYPE_QUESTION
+
+
+def get_user_created_at_for_ticket(user) -> datetime | None:
```
quis, line 254: Optional, but you could avoid this function entirely by writing the model as:

```python
class User:
    def created_at(self):
        return utc_string_to_aware_gmt_datetime(self._dict["created_at"])

class AnonymousUser:
    created_at = None
```

quis (approved):

app/main/views/feedback.py
quis: I’ll tackle this as part of https://github.com/alphagov/notifications-admin/pull/5210 – you can go ahead and merge this

quis (approved):

PR #5210, opened 2024-09-17
Rewrite `JSONModel` to use annotations 

quis:
Instead of an arbitrarily-named `set` floating around we can define fields directly on the class itself using [annotations](https://docs.python.org/3/howto/annotations.html) syntax.

This also gives us somewhere to specify type information. This means we can do:
- validation
- default values
- type coercion

The last one is particularly interesting. For example if we mark a field as a `datetime` we can convert it to a `datetime` object with the correct timezone, rather than having to do this repeatedly across the codebase in different ways.


app/models/user.py
```diff
@@ -45,24 +50,19 @@ def is_invited_user(self):
...
+    current_session_id: Any
+    failed_login_count: int
+    email_access_validated_at: datetime
+    logged_in_at: datetime
+    mobile_number: str
+    password_changed_at: datetime
+    receives_new_features_email: bool
+    state: str
+    take_part_in_research: bool
+    created_at: datetime
```
klssmith, line 65: `created_at` is an annotation on the `BaseUser`, so can be removed here

app/models/__init__.py
```diff
@@ -34,12 +38,21 @@ def __hash__(self):
 
 
 class JSONModel(SerialisedModel, SortingAndEqualityMixin):
+
+    ALLOWED_PROPERTIES = set()  # This is deprecated
```
klssmith, line 42: Is this left here since it's what people are likely to be searching for when adding new properties?

klssmith (commented):
I like the idea of using annotations here, and it worked well when I tried running this locally. I think it's tested indirectly, so this is optional, but you could add some tests specifically for `coerce_value_to_type` to the tests for the base model.

app/models/__init__.py
quis: Left here for now because this requires it to be defined: https://github.com/alphagov/notifications-utils/blob/d9172141300d97624634637218e24c32b7638cc7/notifications_utils/serialised_model.py#L23-L26

Eventually will move most of these changes upstream then we can get rid of it.

app/models/user.py
quis, line 65:
```suggestion
```

klssmith (approved):

PR #5212, opened 2024-09-18
Get notifications for CSV reports through separate endpoint

CrystalPea:
We had one endpoint doing too much, so it became complex and hard to follow. So now we will use two separate endpoints for getting notifications for CSV.

Needs to go in after https://github.com/alphagov/notifications-api/pull/4192

### DIFF
--- a/app/notify_client/notification_api_client.py
+++ b/app/notify_client/notification_api_client.py
@@ -8,28 +8,22 @@ class NotificationApiClient(NotifyAdminAPIClient):
         job_id=None,
         template_type=None,
         status=None,
-        paginate_by_older_than=None,
-        older_than=None,
         page=None,
         page_size=None,
         count_pages=None,
         limit_days=None,
         include_jobs=None,
         include_from_test_key=None,
-        format_for_csv=None,
         to=None,
         include_one_off=None,
     ):
         params = {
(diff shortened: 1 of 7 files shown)


rparke (approved):

app/notify_client/notification_api_client.py
```diff
@@ -49,6 +43,35 @@ def get_notifications_for_service(
...
+        service_id,
+        job_id=None,
+        template_type=None,
+        status=None,
+        older_than=None,
+        page=None,
+        page_size=None,
+        limit_days=None,
+    ):
+        params = {
```
leohemsted, line 57: think we just need to add `"format_for_csv": True` to these params 🤔

rparke: Is there a test we should add to spot code changes that would cause yesterdays issue? I couldn't quite see where it came from.

CrystalPea: @leohemsted I think you are right - I will add it and do some manual testing, and write a unit test for it.

@rparke I think a unit test will be sufficient 🙌🏼

rparke: Cool, if we can pare it down to a unit test that's perfect

PR #5217, opened 2024-09-23
Add validator to restrict sender ids like NHSNoReply

whpearson:
NCSC are trying to reduce the number of sender IDs in use to help people know they are coming from a legitimate place.

One of the sender ids with lots of variation is NHSNoReply

This PR stops new people  registering new sender ids with nhs and no and reply in any order, apart from NHSNoReply

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -193,6 +193,18 @@ class IsNotAGenericSenderID:
             raise ValidationError(self.message)
 
 
+class IsNotLikeNHSNoReply:
+
+    def __call__(self, form, field):
+        lower_cased_data = field.data.lower()
+        if (
+            field.data
+            and ("nhs" in lower_cased_data and "no" in lower_cased_data and "reply" in lower_cased_data)
+            and field.data != "NHSNoReply"
+        ):
+            raise ValidationError("Text message sender ID must match other NHS services - change it to ‘NHSNoReply’")
+
+
 def create_phishing_senderid_zendesk_ticket(senderID=None):
     ticket_message = render_template(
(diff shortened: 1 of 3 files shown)


quis (commented):
Code looks ok but content of the error message could use some work. When the difference between correct/incorrect input is this subtle I think it helps to put both in the message, for example:
> This text message sender must be written as ‘NHSNoReply’, not ‘NHS NO REPLY’

app/main/validators.py
```diff
@@ -193,6 +193,23 @@ def __call__(self, form, field):
...
+        message="Text message sender ID has to be NHSNoReply if being used for that purpose.",
+    ):
+        self.message = message
+
+    def __call__(self, form, field):
+        lower_cased_data = field.data.lower()
+        if (
+            field.data
+            and ("nhs" in lower_cased_data and "no" in lower_cased_data and "reply" in lower_cased_data)
+            and not field.data == "NHSNoReply"
```
joybytes, line 208:
```suggestion
            and field.data != "NHSNoReply"
```

joybytes (approved):
minor syntax suggestion

karlchillmaid:
Suggested content changes:

`Text message sender ID must be ‘NHSNoReply’ - this is to match other NHS services`

`Text message sender ID must match other NHS services - change it to ‘NHSNoReply’`

`Text message sender ID must be ‘NHSNoReply’ to match other NHS services`

`To match other NHS services the text message sender ID must be ‘NHSNoReply’`

`To match other NHS services the text message sender ID must be ‘NHSNoReply’, not ‘{invalid sender id}’`

PR #5218, opened 2024-09-23
Reorder buttons on letter template view

kr8n3r:
Fixes https://trello.com/c/7IYHVqM5/947-visual-order-of-buttons-on-letter-template-page-doesnt-match-the-html

Accessibility audit raised an issue with the focus order of buttons on the letter template.

Cited as "The tab order is unpredictable, as it
moves in the following order:
- Get ready to send
- Add logo
- Change postage
- Edit
- Change your contact details

It is not the expected tab order of moving from top to bottom, left to right", it affects keyboard-only users."

This PR moves the "Get ready to send" but to the suggested order of:

- Add logo
- Change postage
- Get ready to send
- Change your contact details
- Edit

There is extra complexity as not all buttons exist for all users, so a decision was made to break-up the existing `if user has "manage templates"` block to add the button.

Tests needed to also be updated to match the changed button order.

## Before

https://github.com/user-attachments/assets/6b9451e4-b40d-4d0e-83d6-26315a2ec6df

## After

https://github.com/user-attachments/assets/10372974-000c-44a8-9bbd-38be43f6b867

### DIFF
--- a/app/templates/partials/templates/letter_image_template.jinja2
+++ b/app/templates/partials/templates/letter_image_template.jinja2
@@ -1,10 +1,5 @@
 {% if template.include_letter_edit_ui_overlay %}
 <div class="template-container template-container--with-attach-pages-button">
-  {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) and not template.too_many_pages %}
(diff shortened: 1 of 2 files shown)


tombye (approved):
Looks good and works well. The decision makes sense too 👍🏻

PR #5222, opened 2024-09-24
Show skiplink on CSV upload page when focussed

klssmith:
The skiplink on the page to upload a CSV file wasn't appearing when you tab to it, which is confusing for keyboard users who need to see the focus indicator to know where they are in the page.

This makes it visible by giving it an opacity of `1`. The current positioning would place the skiplink on top of the upload button on mobile, so now only applies to bigger screens.

### On larger screens
Screenshot 2024-09-24 at 11 44 30

### On mobile
Screenshot 2024-09-24 at 11 44 16


tombye (approved):
Makes sense and works well. Thanks for making the extra changes to the link hrefs 👍🏻

PR #5225, opened 2024-09-25
Persist notification search

spatel033:
If user search for a notification on the emails/sms/letters dashboard pages, click through to one of the notification and go back, the search term doesn’t persist. More details on [this card](https://trello.com/c/rsHVv64U/903-live-search-on-notification-dashboards-doesnt-persist-search-terms).

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -202,6 +202,17 @@ def cache_search_query(search_term, service_id, search_query_hash):
     return search_query_hash, cached_search_term
 
 
+def redirect_to_main_view_notification(current_service, message_type, search_query):
+    return redirect(
+        url_for(
+            "main.view_notifications",
+            service_id=current_service.id,
+            message_type=message_type,
+            search_query=search_query or None,
+        )
+    )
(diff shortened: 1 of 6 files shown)


app/main/views/jobs.py
```diff
@@ -179,6 +180,39 @@ def view_job_updates(service_id, job_id):
     return jsonify(**get_job_partials(job))
 
 
+def make_cache_key(query_hash, service_id):
+    return f"service-{service_id}-notification-search-query-hash-{query_hash}"
+
+
+def cache_search_query(search_term, service_id, search_query_hash):
+    cached_search_term = ""
```
joybytes, line 188:
```suggestion
   if not search_term:
        return "", ""
```

joybytes (approved):
looks good!

app/main/views/jobs.py
spatel033: If there is no `search_term` from user then still require to check `search_query_hash`.

joybytes (approved):

PR #5232, opened 2024-09-27
Add a `Notification` model

quis:

***

Notifications are one of the last database objects in the admin app that we are still passing around as raw JSON, rather than wrapping them in a proper Python object.

This is a bit fiddly because the fields present are different depending which API endpoint the list of notifications is coming from. So I’ve made multiple models instead of trying to overload one.

I’ve also made a model for inbound text messages, so they can be sorted and combined with outgoing ones.

### DIFF
--- a/app/models/notification.py
+++ b/app/models/notification.py
@@ -27,7 +27,6 @@ class Notification(JSONModel):
     job_row_number: int
     service: Any
     template_version: int
-    personalisation: dict
     postage: str
     notification_type: str
     reply_to_text: str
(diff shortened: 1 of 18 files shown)


app/models/notification.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __init__(self, *args, **kwargs):
+        resp = self.client_method(*args, **kwargs)
+        self.items = resp["notifications"]
+        self.prev = resp.get("links", {}).get("prev", None)
+        self.next = resp.get("links", {}).get("next", None)
+
+
+class NotificationForCSV(Notification):
+    created_at: str  # API returns this field pre-formatted in Europe/London timezone
```
joybytes, line 117: so because it's pre-formatted it comes on a string format instead of datetime?

joybytes (approved):

PR #5235, opened 2024-10-01
Api docs sections landing page

CrystalPea:


app/templates/views/guidance/using-notify/api-documentation.html
```diff
@@ -10,6 +10,16 @@
 {% block content_column_content %}
 
   <h1 class="heading-large">Documentation</h1>
+
+    {% if current_user.platform_admin %}
+      <a href="{{ url_for(".guidance_api_documentation_section") }}"
+        class="govuk-link govuk-link--no-visited-state"
+        target="_blank" rel="noopener">
+        Send a link to docs section
+      </a>
```
saimaghafoor, line 19:
```suggestion
        target="_blank" rel="noopener">
       Send a link to the Notify API documentation
      </a>
```

app/templates/views/guidance/using-notify/api-documentation-section.html
```diff
@@ -0,0 +1,27 @@
+{% extends "content_template.html" %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/page-footer.html" import page_footer %}
+
+{# Used by the content_template.html layout, prefixes the "navigation" accessible name #}
+{% set navigation_label_prefix = 'Using Notify' %}
+{% set page_title = "Send a link to an API docs section" %}
```
saimaghafoor, line 7:
```suggestion
{% set page_title = "Send a link to the Notify API documentation" %}
```

app/main/forms.py
```diff
@@ -2280,6 +2281,50 @@ def validate(self, *args, **kwargs):
...
+        "Which version of the docs would you like to view?",
+        choices=[
+            ("python", "Python"),
+            ("ruby", "Ruby"),
+            ("java", "Java"),
+            ("node", "Node JS"),
+            ("net", ".Net"),
+            ("php", "PHP"),
+            ("rest-api", "Rest API"),
+            ("rest-api", "Any is fine"),
```
joybytes, line 2321: is this duplication intentional?

app/main/forms.py
```diff
@@ -2280,6 +2281,50 @@ def validate(self, *args, **kwargs):
...
+        self.url.validators.append(self.check_url)
+        return super().validate(*args, **kwargs)
+
+    def check_url(self, *args, **kwargs):
+        parsed_url = urlparse(self.url.data)
+
+        if parsed_url.hostname == "docs.notifications.service.gov.uk" and parsed_url.fragment:
+            return parsed_url
+        else:
+            raise ValidationError("Must be a valid https URL, pointing to a section within the GOV.UK Notify API docs.")
```
joybytes, line 2303: feel free to ignore
```suggestion
             raise ValidationError("The URL must point to a specific section in the GOV.UK Notify API docs, and start with 'https://docs.notifications.service.gov.uk'.")
```

app/main/views/index.py
```diff
@@ -159,6 +159,39 @@ def guidance_api_documentation():
     )
 
 
+@main.route("/using-notify/api-documentation/section", methods=["GET", "POST"])
+def guidance_api_documentation_section():
+
+    form = UrlForm()
+
+    if form.validate_on_submit():
+        section_tag = form.url.data.split("#")[-1]
```
joybytes, line 168: would it be useful to do a check in case of no section_tag?

app/main/views/index.py
```diff
@@ -159,6 +159,39 @@ def guidance_api_documentation():
     )
 
 
+@main.route("/using-notify/api-documentation/section", methods=["GET", "POST"])
```
klssmith, line 162: Should this either be platform admin only or have the `@hide_from_search_engines` decorator? There's nothing sensitive on the page, but it might be confusing to people if they come across it

klssmith:
You've probably already seen this but not all the headings are consistent, so that might need to be looked at before this could be merged. No major differences, just things like `#get-a-pdf-for-a-letter-notification` vs `#get-a-pdf-for-a-letter`

app/main/forms.py
CrystalPea: Yes, it's explained in the commit message 🙌🏼

app/main/views/index.py
CrystalPea: We do validate that there is a section tag on form validation, this if statement:

```
if parsed_url.hostname == "docs.notifications.service.gov.uk" and parsed_url.fragment:
            return parsed_url
```

app/main/views/index.py
CrystalPea: Good idea to hide it from search engines 🙌🏼

app/main/forms.py
CrystalPea: Nice one, I will add this to code (and tests)

CrystalPea:
@klssmith yeah, I thought they probably aren't - we have upcoming work on the team to get API docs more in line - maybe we could start with making the headers match 🙌🏼 . What do you think @saimaghafoor ? Could be a good goal for first pass over the docs?

app/main/views/index.py
joybytes: sorry, I missed that!

quis:
- agree that we should fix the mismatched headings ([comparison tool](https://alphagov.github.io/notifications-tech-docs-checker/reports/2024-07-11-at-1357.html))
- don’t think we need radio buttons here – the choices could just be links
- I don’t like adding additional, platform-admin-only elements (the link) to user-facing pages because it confuses people on support about what users can/cannot see

quis:
As an interim step could we automatically validate that the header exists in all the docs before generating the link?

app/main/forms.py
```diff
@@ -2280,6 +2281,50 @@ def validate(self, *args, **kwargs):
...
+
+
+class ChooseDocsForm(StripWhitespaceForm):
+
+    def __init__(self, section_tag):
+        super().__init__(section_tag=section_tag)
+
+    docs_version = GovukRadiosField(
+        "Which version of the docs would you like to view?",
```
karlchillmaid, line 2312:
```suggestion
        "Which version of the API documentation would you like to view?",
```

app/main/forms.py
```diff
@@ -2280,6 +2281,50 @@ def validate(self, *args, **kwargs):
...
+            ("python", "Python"),
+            ("ruby", "Ruby"),
+            ("java", "Java"),
+            ("node", "Node JS"),
+            ("net", ".Net"),
+            ("php", "PHP"),
+            ("rest-api", "Rest API"),
+            ("rest-api", "Any is fine"),
+        ],
+        thing="a language version of GOV.UK Notify's API docs",
```
karlchillmaid, line 2323: Apart from the REST API documentation, we refer to these as ‘client libraries’ elsewhere.

app/templates/views/guidance/using-notify/api-documentation-section-choose-docs.html
```diff
@@ -0,0 +1,23 @@
+{% extends "content_template.html" %}
+{% from "components/form.html" import form_wrapper %}
+{% from "components/page-footer.html" import page_footer %}
+
+{# Used by the content_template.html layout, prefixes the "navigation" accessible name #}
+{% set navigation_label_prefix = 'Using Notify' %}
+{% set page_title = "You have been sent a link to GOV.UK Notify API documentation" %}
```
karlchillmaid, line 7:
```suggestion
{% set page_title = "You have been sent a link to the GOV.UK Notify API documentation" %}
```

```suggestion
{% set page_title = "You have been sent a link to GOV.UK Notify’s API documentation" %}
```

```suggestion
{% set page_title = "You have been sent a link to the Notify API documentation" %}
```

```suggestion
{% set page_title = "You have been sent a link to Notify’s API documentation" %}
```

app/templates/views/guidance/using-notify/api-documentation-section.html
```diff
@@ -0,0 +1,27 @@
...
+
+
+{% block per_page_title %}
+{{page_title}}
+{% endblock %}
+
+{% block content_column_content %}
+
+<h1 class="heading-large">{{page_title}}</h1>
+<p class="govuk-body">Below, copy paste the full URL for docs section you would like to send to a service user, and click "Submit".</p>
```
karlchillmaid, line 17:
```suggestion
<p class="govuk-body">Enter the full URL for a section of the API documentation and select <b class="govuk-!-font-weight-bold">Submit</b>.</p>
```

app/templates/views/guidance/using-notify/api-documentation-section.html
```diff
@@ -0,0 +1,27 @@
...
+
+<p class="govuk-body">You will be taken to
+a landing page for that section. Send link to that page to your user, and they will be able to choose which language version they want to see.</p>
```
karlchillmaid, line 20:
```suggestion
a landing page for that section. Send the URL of that page to the user. When they open it, they’ll be able to choose a client library.</p>
```

app/main/forms.py
karlchillmaid:
```suggestion
        "Choose a client library",
```

app/templates/views/guidance/using-notify/api-documentation.html
karlchillmaid: If this is for platform admins, do we need to specify ‘Notify’?

Can we just say `the API documentation` instead?

app/templates/views/guidance/using-notify/api-documentation-section.html
karlchillmaid: If this is for platform admins, do we need to specify ‘Notify’?

Can we just say `the API documentation` instead?

PR #5238, opened 2024-10-04
Delete unused code

klssmith:
Almost all from the tests to remove fixtures they're not using, but also a few unused functions.


CrystalPea (approved):
Nice weeding session 👏🏼

PR #5240, opened 2024-10-04
feat: Add permission-based filters for platform admins on manage users page

joybytes:
## Summary
- Introduced permission-based filters on the manage users page, visible only to platform admins.
- Platform admins can filter team members by specific permissions, such as "Manage settings, team and usage" and "Manage API integration."
- Updated tests to verify the visibility of the filter section and the "Apply Filters" button for platform admins.
Added asserts to ensure the filter section is hidden for non-platform admins.
- Added Last Login field on each team member

## Task:
[Add a filter/sort by function and 'last logged in' for ‘manage service‘ users on the 'team members' page
](https://trello.com/c/BxoOjSiF/58-add-a-filter-sort-by-function-and-last-logged-in-for-manage-service-users-on-the-team-members-page)

### Manual Testing:
**Default behaviour**
- Less than 7 users on services  = No search box and filters showing
Screenshot 2024-10-04 at 15 27 13

**Search box present with no Filters for users who are not platform admins**
Screenshot 2024-10-04 at 15 29 02

**Filters present for Platform admins**
Screenshot 2024-10-04 at 15 31 40

**Last Login details for each member**
Screenshot 2024-10-08 at 14 30 11

### DIFF
--- a/app/main/views/manage_users.py
+++ b/app/main/views/manage_users.py
@@ -23,15 +23,26 @@ from app.utils.user import is_gov_user, user_has_permissions
 from app.utils.user_permissions import permission_options
 
 
-@main.route("/services/<uuid:service_id>/users")
+@main.route("/services/<uuid:service_id>/users", methods=["GET"])
 @user_has_permissions(allow_org_user=True)
 def manage_users(service_id):
+    form = SearchUsersForm(request.args)
+    users = current_service.team_members
+    permission_filters = request.args.getlist("permissions")
+
+    if permission_filters:
+        users = [
+            user
+            for user in users
+            if all(user.has_permission_for_service(service_id, perm) for perm in permission_filters)
+        ]
+
     return render_template(
         "views/manage-users.html",
-        users=current_service.team_members,
+        users=users,
         current_user=current_user,
-        show_search_box=(len(current_service.team_members) > 7),
-        form=SearchUsersForm(),
+        show_search_box=(len(users) > 7),
+        form=form,
         permissions=permission_options,
     )
 
--- a/app/templates/views/manage-users.html
+++ b/app/templates/views/manage-users.html
@@ -75,6 +75,16 @@
               {% endif %}
               </span>
             </h2>
+
+            <br>
+              {% if user.logged_in_at %}
+                  <p class="govuk-body">Last logged in
+                      <time class="timeago" datetime="{{ user.logged_in_at }}">
+                          {{ user.logged_in_at|format_delta }}
+                      </time>
+                  </p>
+              {% endif %}
+
             <ul class="tick-cross__list">
               {% for permission, label in permissions %}
                 {{ tick_cross(
--- a/app/templates/views/manage-users.html
+++ b/app/templates/views/manage-users.html
@@ -2,6 +2,8 @@
 {% from "components/tick-cross.html" import tick_cross %}
 {% from "components/live-search.html" import live_search %}
 {% from "govuk_frontend_jinja/components/button/macro.html" import govukButton %}
+{% from "govuk_frontend_jinja/components/checkboxes/macro.html" import govukCheckboxes %}
+{% from "govuk_frontend_jinja/components/details/macro.html" import govukDetails %}
 
 {% block service_page_title %}
   Team members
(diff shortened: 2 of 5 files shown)


app/main/views/manage_users.py
```diff
@@ -23,15 +23,39 @@
...
 
 
-@main.route("/services/<uuid:service_id>/users")
+@main.route("/services/<uuid:service_id>/users", methods=["GET"])
 @user_has_permissions(allow_org_user=True)
 def manage_users(service_id):
+    form = SearchUsersForm(request.args)
+    users = current_service.team_members
+
+    search_term = request.args.get("search", "").strip()
```
klssmith, line 32: `search_term` isn't populated, since the search functionality comes [from the JavaScript](https://github.com/alphagov/notifications-admin/blob/main/app/assets/javascripts/liveSearch.js). So the `if search_term:` section below isn't reached, but also isn't needed for searching

joybytes: thanks Katie, there was there because I separated some PRs and that's part of the next one.
thanks again

app/templates/views/manage-users.html
```diff
@@ -48,6 +75,16 @@ <h2 class="user-list-item-heading" title="{{ user.email_address }}">
               {% endif %}
               </span>
             </h2>
+
+            <br>
+              {% if user.logged_in_at %}
+                  <p class="govuk-body">Last logged in
```
leohemsted, line 81: nice! just checking we want to show this to all users, not just platform admins?

quis:
Some design feedback (and apologies for raising now but I was away all last week).

# Adding platform admin stuff to the team members page

I don’t love adding platform-admin-only things to user-facing pages because it means we don’t see the page as the user sees it.

# User need

Feels like our real user need here is being able to email all the ‘admins’ on a team. Think this need would be better met by a separate, platform-admin-only page with either (in order of sophistication):
1. A list of email addresses to copy/paste
2. A CSV file we can download then reupload to Notify (with some columns we can filter on in Google Sheets)
3. A page where you can pick the team members then write a message which gets sent to them through Notify

# User interface 

More minor, but…

When you apply the filters your selection is either:
- not visible (because they are collapsed)
- not on the page at all (if there are fewer than 7 team members which match the filter)

It’s also not clear if you tick both whether that’s strict (users with both permissions) or loose (users with either permission)

joybytes:
> Some design feedback (and apologies for raising now but I was away all last week).
> 
> # Adding platform admin stuff to the team members page
> I don’t love adding platform-admin-only things to user-facing pages because it means we don’t see the page as the user sees it.
> 
> # User need
> Feels like our real user need here is being able to email all the ‘admins’ on a team. Think this need would be better met by a separate, platform-admin-only page with either (in order of sophistication):
> 
> 1. A list of email addresses to copy/paste
> 2. A CSV file we can download then reupload to Notify (with some columns we can filter on in Google Sheets)
> 3. A page where you can pick the team members then write a message which gets sent to them through Notify
> 
> # User interface
> More minor, but…
> 
> When you apply the filters your selection is either:
> 
> * not visible (because they are collapsed)
> * not on the page at all (if there are fewer than 7 team members which match the filter)
> 
> It’s also not clear if you tick both whether that’s strict (users with both permissions) or loose (users with either permission)

Hi @quis,

User Need:
I think there might have been some confusion regarding the intent of this work. This PR is aimed exclusively at platform admins, not general users. There hasn’t been any research done around user-facing needs for this feature.

The main goal here is to assist non-technical support staff in quickly identifying user permissions. Currently, they have to manually go through each individual team member to check their permissions and recent activity, which is time-consuming during support interactions. We're aiming to streamline that process with this feature.

Permissions & UI:
The filter logic is designed to be strict—if you're selecting users with both permissions, it will only show those who meet both criteria. Anyone missing one or the other won't appear in the filtered list, which is standard filter behavior.

As for the UI, the filters are applied using the same conditional logic as the search box. The search box only appears when there are more than 7 team members on the page, so I followed the existing behavior for consistency.

Platform-Admin-Only Pages:
You make a good point about not liking platform-admin-only functionality on user-facing pages because it affects how we see the page compared to the end-user. However, this is the page we were targeting with this implementation. Is there another page or location where you feel this feature would be more appropriate?

joybytes:
Moving this to draft as there is a different strategy to showcase this information being discussed.

PR #5241, opened 2024-10-07
Update date and style links in accessibility statement

kr8n3r:
Add missing `govuk-link` class to links in "Non-accessible content" list and update date to today.

Before

www notify works_accessibility-statement

After
Screenshot 2024-10-07 at 14 16 36

### DIFF
--- a/app/templates/views/accessibility_statement.html
+++ b/app/templates/views/accessibility_statement.html
@@ -12,8 +12,8 @@
   {{ content_metadata(
     data={
       "Published": "23 September 2020",
-      "Last updated": "4 October 2024",
-      "Next review due": "4 December 2024"
+      "Last updated": "7 October 2024",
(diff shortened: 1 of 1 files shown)


quis (approved):
Good to merge, but can we follow up and see why the missing class wasn’t caught by this test: https://github.com/alphagov/notifications-admin/blob/708409d25000e5ff88e7cf8f1455ba2cbadc8bc7/tests/conftest.py#L3101

kr8n3r:
@quis maybe because that method is part of client_request, but the a11y page test doesn't include it?

quis:
@kr8n3r we should add it to the list of pages here then: https://github.com/alphagov/notifications-admin/blob/7a889e2c4de1ec1becba8682a783bdc1d7add8d5/tests/app/main/views/test_index.py#L90-L115

PR #5243, opened 2024-10-09
Adjust button positioning on letter template preview page

kr8n3r:


tombye (changes requested):
Such a relief to see this done and I 😍 the changes that make more button text visible on smaller screens.

I just have one comment, which is that the buttons underneath the letter image don't have a margin between them and this now just looks odd. Can we add it please?

image

kr8n3r:
> Such a relief to see this done and I 😍 the changes that make more button text visible on smaller screens.
> 
> I just have one comment, which is that the buttons underneath the letter image don't have a margin between them and this now just looks odd. Can we add it please?
> 
 done

tombye (approved):
Thanks for adding the fix. Now good to go 👍🏻

PR #5244, opened 2024-10-09
Use returned service ID inside unsubscribe model

quis:
We won’t ever want to have one service batching up unsubscribe requests for another.

As of https://github.com/alphagov/notifications-api/pull/4191 we are serialising the service ID so the method on the model doesn’t need to take it as an argument any more.

***

I’ve also done a bunch of refactoring of the test data to use a common fixture and be less verbose (see separate commits).


joybytes (approved):
lgtm!

PR #5252, opened 2024-10-15
Update the guidance for short URLs

karlchillmaid:
Our guidance incorrectly says that anyone with a `gov.uk` domain can request a short URL from GDS.

We’ve had a couple of support tickets asking about this – so we need to correct the guidance to avoid misleading anyone.


klssmith (approved):
Change looks good:

---

Screenshot 2024-10-16 at 07 59 30

PR #5253, opened 2024-10-17
Update service navigation to show full service and organisation name

kr8n3r:
## What

Update the service navigation design to 
- display full name of the organisation (if allowed to see)
- display full name of the service

regardless of the viewport size or text zoom level. 
For this we mimicked the gov.uk breadcrumb pattern, where a long item drops onto the next line with a chevron preceding it. 

Fixes https://trello.com/c/UkFnJkCH/997-service-navigation-has-visual-issues-on-smaller-screens

## Why
Organisation and service names were being truncated if they were too long of their container, which fails 1.4.10 Reflow

## How to review

- see individual commits for detailed logic. Once reviewed I'll combine CSS changes into one commit and add all the messages into it.
- test locally with a users that can/cannot see the org link

## Visual examples

**Short names**
desktop-long-names

**Long names**
desktop-short-names

**400% zoom**
Screenshot 2024-10-21 at 13 11 02

**IE11**
Screenshot 2024-10-21 at 12 37 07

### DIFF
--- a/app/assets/stylesheets/components/navigation.scss
+++ b/app/assets/stylesheets/components/navigation.scss
@@ -8,14 +8,7 @@
 
   &-service-name,
   &-organisation-link {
-    display: inline-block;
-    overflow: hidden;
-    // aligning to the baseline with overflow: hidden adds to the parent's height
-    // aligning to the top doesn't
-    // see: https://stackoverflow.com/questions/23529369/why-does-x-overflowhidden-cause-extra-space-below#answer-51088033
-    vertical-align: top;
-    white-space: nowrap;
-    text-overflow: ellipsis;
+    float: left;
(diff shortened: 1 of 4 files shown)


tombye (changes requested):
This is excellent work. As a user, it's such a relief to lose the ellipsis hiding content and be able to see the org and service name.

I have one suggestion, which is to float the items in the service/org nav to remove the very slight extra spacing added after the org name (or 'Organisations' for org nav) because of it being display-block:

image

I suggest float because design system do that, I'd guess for the same reasons

tombye (approved):
Thanks for addressing my comment, and for rewriting the commits for quicker reading. Now looks ready to go 👍🏻

PR #5256, opened 2024-10-21
Fix sticky footer overlapping some focused elements

kr8n3r:
Fixes: https://trello.com/c/wPRphm25/998-the-sticky-footer-sometimes-overlaps-focused-elements

## What
Add a condition that checks for the type of focused element and if it's either a checkbox or radio then take the parent element (which will be `govuk-checkboxes__item` or `govuk-radios__item`) and adjust the scroll for its height.

This fixes an issue where checkboxes on templates page were not brought into view when focused if the sticky footer happens to overlap them.

While that's due to them having a minimal height for visual composition, it also helps to now fully show the hint text as that is usually lower than the height of the checkbox itself.

## Changes

https://github.com/user-attachments/assets/872c7739-7a3e-492a-a914-acb077b9a85d

https://github.com/user-attachments/assets/4a0e1df2-2bae-4eee-9be7-37caf2ec9d87

https://github.com/user-attachments/assets/45706fb7-e960-4f95-a7de-89697f88bb86

https://github.com/user-attachments/assets/a8be0e1b-e9ba-42c2-9287-e3458503c056

#

### DIFF
--- a/app/assets/javascripts/stick-to-window-when-scrolling.js
+++ b/app/assets/javascripts/stick-to-window-when-scrolling.js
@@ -68,6 +68,7 @@
   };
   ScrollArea.prototype.scrollToRevealElement = function ($el) {
     var nodeName = $el.get(0).nodeName.toLowerCase();
+    var nodeType =  $el.get(0).getAttribute('type');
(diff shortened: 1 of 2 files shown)


tombye (approved):
Nice piece of work and works well when I tested. Good to go 👍🏻

PR #5257, opened 2024-10-21
Service join request approve and set permissions 

spatel033:
For service-join-request when approver clicks on the link from email, it takes user on Yes/No options. This PR is for when user selects Yes then user is asked to set permissions for requester. If user selects No then is shows refused message. More details on [this card](https://trello.com/c/IdQPhhuy/1000-update-the-servicejoinrequest-when-request-is-approved-rejected).

Approver will be asked to set permissions for the requester:
image

https://github.com/user-attachments/assets/044b8f38-8cd2-475e-bef0-26d080c11679

Approver will be asked to inform the requester on refuse:
image

https://github.com/user-attachments/assets/c2620194-ee5d-4304-aabf-ead60e4af116

**Steps to test locally:**
This card is only for notify-admin changes. But you will require specific notify-api changes.
Check out this branch
Check out notify-api : https://github.com/alphagov/notifications-api/tree/feat/update-service-join-request (optional: and merge main into this branch)

*Test organisation and test users*
You will need Organisation with “Users can ask to join service” enabled
You will need two users:
Approver: who has manage-users access for any live service
Requester: who is in the org without live service access

*From Requester account:*
- Request to join a live service
- In `service_join_requests` table one record will be added or you can add manually if you want skip above steps
- Approver URL: http://notify.localhost:6012/services/<service_join_requests.service_id>/join-request/<service_join_requests.id>/approve
- Update ids in above URL
- Update the request status from approved/rejected to pending manually in db to skip above steps for other scenarios


app/templates/views/join-service-request-set-permissions.html
```diff
@@ -0,0 +1,29 @@
...
+{% block backLink %}
+  {{ govukBackLink({ "href": url_for('main.service_join_request_approve', service_id=current_service.id, request_id=request_id) }) }}
+{% endblock %}
+
+{% block maincolumn_content %}
+
+  {{ page_header('Choose permissions for {}'.format(requester.name)) }}
+
+  {% call form_wrapper() %}
+    <p class="govuk-body">
```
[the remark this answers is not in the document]
spatel033: For the `notify-api` app, `main` branch needs to be merged into feature branch, updated the instructions. Check out notify-api : https://github.com/alphagov/notifications-api/tree/feat/update-service-join-request and merge main into this branch.

app/main/views/manage_users.py
```diff
@@ -128,16 +126,77 @@ def service_join_request_approve(service_id, request_id):
...
+    requested_service = Service.from_id(service_join_request.service_id)
+
+    if form.validate_on_submit():
+        if form.join_service_approve_request.data == SERVICE_JOIN_REQUEST_APPROVED:
+            return redirect(
+                url_for(
+                    "main.service_join_request_set_permissions",
+                    service_id=service_id,
+                    request_id=request_id,
+                    requester_id=requested_by["id"],
```
klssmith, line 149: `requester_id` isn't used by the next `service_join_request_set_permissions` route so can be removed.

app/main/views/manage_users.py
```diff
@@ -128,16 +126,77 @@ def service_join_request_approve(service_id, request_id):
...
+
+@main.route("/services/<uuid:service_id>/join-request/<uuid:request_id>/approve", methods=["GET", "POST"])
+@user_has_permissions("manage_service")
+def service_join_request_approve(service_id, request_id):
+    is_redirect = validate_service_join_request(service_id, request_id)
+    if is_redirect:
+        return is_redirect
+
+    form = JoinServiceRequestApproveForm()
+    service_join_request = ServiceJoinRequest.from_id(request_id)
```
klssmith, line 138: I realise it wasn't added in this PR, but could we rename the `service_join_request_id` argument that the `ServiceJoinRequest` model has to `id`? (It will need an api change too, but the api part isn't merged yet so it's simpler to do it now before it is.)

The classes that inherit from JSONModel expect `id` to be there so bits of functionality don't work if it's not.

joybytes: good point Katie!

app/templates/views/join-service-request-set-permissions.html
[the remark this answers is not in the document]
spatel033: Could you try pulling latest notify-api (https://github.com/alphagov/notifications-api/tree/feat/update-service-join-request). It was behind main branch.

kr8n3r:
I'm happy with html part. Python review needs to be done by someone else

app/notify_client/service_api_client.py
```diff
@@ -538,8 +538,14 @@ def parse_custom_email_sender_name_http_error(cls, http_error):
 
         return None
 
+    @cache.set("service_join_request-{request_id}")
```
klssmith, line 541: Really minor, but if these are `service-join-request-{request_id}` (instead of the underscores) it will match the pattern for we have for other service cache keys. You could also add them to [this list](https://github.com/alphagov/notifications-admin/blob/a3fe11c22d1f631dce8caa0e043104a8b9bec058/app/main/views/platform_admin.py#L579) so that we have a way of clearing the keys through Platform Admin

klssmith:
If someone is added to a service, then we need to make sure we clear their key out of the Redis since that includes the services they are a member of - without this, they can't see the new service they have joined.

app/main/views/manage_users.py
```diff
@@ -128,16 +132,76 @@ def service_join_request_approve(service_id, request_id):
...
+                    service_id=service_id,
+                    request_id=request_id,
+                )
+            )
+        if form.join_service_approve_request.data == SERVICE_JOIN_REQUEST_REJECTED:
+            service_join_request.update(
+                status=SERVICE_JOIN_REQUEST_REJECTED,
+                status_changed_by_id=current_user.id,
+            )
+            return render_template(
```
leohemsted, line 162: rather than rendering a template, we should follow the Post/Redirect/Get pattern here (https://en.wikipedia.org/wiki/Post/Redirect/Get) - the same as we do when the request was approved.

If you're making stateful changes, it's good to redirect so that if users refresh it doesn't try and perform the action twice

we'll need a new endpoint and a new URL `/services/<service_ID>/join-request/<request_ID>/refused` that we can just redirect to here.

app/main/views/manage_users.py
```diff
@@ -128,16 +132,76 @@ def service_join_request_approve(service_id, request_id):
...
         form=form,
         requester=requested_by,
         requested_service=requested_service,
         error_summary_enabled=True,
+        service_id=service_id,
+        request_id=request_id,
+    )
+
+
+@main.route("/services/<uuid:service_id>/join-request/<uuid:request_id>/set-permissions", methods=["GET", "POST"])
```
leohemsted, line 180: karl's suggestion was to change this from set to choose

`/services/<service_ID>/join-request/<request_ID>/choose-permissions`

app/main/views/manage_users.py
```diff
@@ -128,16 +132,76 @@ def service_join_request_approve(service_id, request_id):
...
-    # if form.validate_on_submit():
-    # (form.join_service_approve_request.data)
-    # Once permissions/reject template created, redirect from here
+
+@main.route("/services/<uuid:service_id>/join-request/<uuid:request_id>/approve", methods=["GET", "POST"])
+@user_has_permissions("manage_service")
+def service_join_request_approve(service_id, request_id):
+    is_redirect = validate_service_join_request(service_id, request_id)
+    if is_redirect:
+        return is_redirect
```
leohemsted, line 141: this is a bit misleading as this sometimes raises a 403, sometimes returns a rendered template, and sometimes returns None

we should probably change the `is_redirect` variable to name it to something to make it clear that it's a rendered template with an error message

app/main/views/manage_users.py
joybytes: redirected calls to new endpoint
https://github.com/alphagov/notifications-admin/pull/5257/commits/b7f539c1656960f978559907be8e1bb165001626

app/main/views/manage_users.py
joybytes: Updated variables and endpoint to be in line with choose-permissions

whpearson (approved):
Tested locally. Code looks good.

PR #5258, opened 2024-10-22
Make add reply-to address flow work with screen readers

tombye:
Fixes the add reply-to address flow so updates are announced to screen readers.

https://trello.com/c/2JfcApXt/999-page-for-checking-a-reply-to-email-address-has-problems-for-screen-reader-users

All changes paired on by @klssmith and me.

### Changing the H1 and title

We changed the title and H1 of the page that checks and confirms/rejects the email address to 'Checking reply-to address' so it isn't the same as the previous page and explains the page better.

### Move the banners before the H1

We moved the banners you see if the email address is confirmed or rejected to be above the H1, to match existing error summary behaviour.

### Focus the banners

We added some JS to focus either banner when it appears to match the behaviour of error summaries and ensure the banner text is announced to screen readers.

## Before

https://github.com/user-attachments/assets/075eb9e2-f9e2-4905-b058-ab5f0ac3c290

## After

https://github.com/user-attachments/assets/c9e4e91e-fd0d-441e-b755-3acd5f27c2ec

Both examples use the JAWS screen reader. Neither video has audio so use the text descriptions at the bottom to see what JAWS is announcing.

## Notes for reviewers

Worth reviewing commit by commit.

If you don't feel comfortable testing with screen readers, just review the code because this work was paired on throughout so 2 pairs of 👀s have seen it working already.

If you have access to a assistivelabs, it's worth testing with JAWS or NVDA. If you're testing with Voiceover, bear in mind you might need to turn 'Automatically speak the web page' off and turn 'Speak web page summary' so it announces the title when the page loads. In Safari too, Chrome doesn't seem to respect that setting.

Testing is loads easier if you hack it by:
- setting the `REPLY_TO_EMAIL_ADDRESS_VALIDATION_TIMEOUT` to 5 in `app/config.py`
- changing the conditional logic in `app/templates/views/service-settings/email-reply-to/_verify-updates.html` so you see the banner you want, because the flow on local will always fail

### DIFF
--- a/app/assets/javascripts/esm/all-esm.mjs
+++ b/app/assets/javascripts/esm/all-esm.mjs
@@ -2,6 +2,7 @@
 import { createAll, Header, Button, Radios, ErrorSummary, SkipLink, Tabs } from 'govuk-frontend';
 
 import CollapsibleCheckboxes from './collapsible-checkboxes.mjs';
+import FocusBanner from './focus-banner.mjs';
 
 // Modules from 3rd party vendors
 import morphdom from 'morphdom';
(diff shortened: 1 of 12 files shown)


kr8n3r (approved):
👍  quality commit messages made reviewing this super easy

PR #5265, opened 2024-10-24
Don’t coerce `row_number` in CSV download

quis:
The `row_number` that the API returns for a CSV download can be either an empty string or a number.

When we deserialise the JSON from the API we check for `None` values before trying to cast to another type, but we don’t check for empty string.

Therefore marking a field as `int` causes an exception because empty string cannot be coerced to an integer.

This commit change the type to `Any`, which means no coercion will be attempted.

This also moves the field to the `NotificationForCSV` model, since it’s only returned when getting notifications to download and not from the morre general `GET /notifications` endpoint.

***

Comes from: https://github.com/alphagov/notifications-api/blob/8d787fcfca5e6e68720bace6204465b7e963e3c9/app/models.py#L1590

### DIFF
--- a/app/models/notification.py
+++ b/app/models/notification.py
@@ -23,7 +23,6 @@ class Notification(JSONModel):
     created_at: datetime
     created_by: Any
     updated_at: datetime
-    row_number: int
     job_row_number: int
     service: Any
     template_version: int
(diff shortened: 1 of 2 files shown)


risicle (approved):
Seems sensible

PR #5266, opened 2024-10-24
Put email and sms details in summary list

tombye:
Change the metadata on the email template page from being wrapped in a table to [a summary list](https://design-system.service.gov.uk/components/summary-list/).

https://trello.com/c/DFMje2N6/975-fix-the-table-on-the-template-preview-page-wcag-22

The accessibility audit described the current table as being confusing, partly because it doesn't have a caption[^1]. We've either converted most other pages with lists of key:value pairs from tables to summary lists, on the recommendation of previous accessibility audits. So we decided to do the same here.

## Details of the changes

The HTML for this table is currently imported from notifications-utils, along with its rendering class and tests. But the admin app is the only user of this code so these changes move both into the admin app and make the changes there. If this pull request is merged, we will remove the then-redundant code from utils.

### Desktop

email_template_desktop_before

### Mobile

email_template_mobile_before

### Desktop

email_template_desktop_after

### Mobile

email_template_mobile_after

## Notes for reviewers

This needs 2 types of review:
1. frontend code review, to compare with the existing (visual) UI
2. application code (python and Jinja) review, to ensure the class, jinja and tests still do what they did in their utils version
3. some 👀 from interaction design, to check the changes made to the mobile UI make sense

[^1]: See page 43 of [the audit report](https://drive.google.com/file/d/1qR8t3mcEka6sUVGqfhnjEpyhqqQhGNFi/view?usp=sharing)

### DIFF
--- a/app/assets/stylesheets/components/email-message.scss
+++ b/app/assets/stylesheets/components/email-message.scss
@@ -1,4 +1,5 @@
-$email-message-gutter: govuk-spacing(9);
+$email-message-gutter-desktop: govuk-spacing(9);
+$email-message-gutter-mobile: govuk-spacing(3);
 
 // sass-lint:disable no-important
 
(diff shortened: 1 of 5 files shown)


kr8n3r (approved):
this makes sense and looks good to me

quis (approved):
I’ve been through and confirmed that all the tests we need to remove from utils match the ones added here

PR #5267, opened 2024-10-25
Copy common config files from utils in freeze step

quis:
So that every time you bump the utils version you get the correct corresponding files and the build is more reproducible.

***

Duplicates https://github.com/alphagov/notifications-api/pull/4194


kr8n3r (approved):
good to go as the same change had been reviewed

PR #5275, opened 2024-10-30
Make back links work when sending a messages from the inbound SMS flow

klssmith:
When sending from the inbound SMS flow we didn't have a back link on the page to choose a template. Once you had started sending a message, the back links took you to the "standard journey" of sending a message after clicking on a template. This change adds in back links to the "Choose a template page" and makes the back links when sending work for all scenarios. More details in the individual commit messages.

This also fixes a bug with how templates in folders appeared when sending from an inbound SMS:

**Before**
Screenshot 2024-10-29 at 11 30 10

**After**
Screenshot 2024-10-29 at 11 30 31

### DIFF
--- a/app/main/views/conversation.py
+++ b/app/main/views/conversation.py
@@ -42,6 +42,17 @@ def conversation_reply(
     notification_id,
     from_folder=None,
 ):
+    if from_folder:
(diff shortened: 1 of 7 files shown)


kr8n3r (approved):
tested, works as expected. great commit messages for easy review

PR #5277, opened 2024-10-31
notifications-utils: 88.0.1 -> 88.1.0

risicle:
Incorporating https://github.com/alphagov/notifications-utils/pull/1162


tombye (approved):
Makes sense, as far as I understand it from reading the utils PR 👍🏻

PR #5283, opened 2024-11-05
Give checkboxes height again & fix other visual bugs

tombye:
Includes changes to the template list CSS to:
- give the checkbox element has height again
- fix an issue caused by the 'new' design system links styles setting the `box-decoration-break` value to assume all lines a link wraps to inherit the initial padding
- add padding-left to stop content from the template list item wrapping underneath the checkbox

The checkboxes currently have a height of 1px, which interferes with the logic browsers use to scroll them into view when mostly off-screen. Because only their first 1px was considered by the layout, you would often see checkboxes that were
mostly off-screen focused without being scrolled into view.

The `box-decoration-break` change interferred with us having padding at the start of links to folders, to make space for the folder icon. It was only visible when the links were focused.

## Before - unfocused

template_list_path_wrapping_text_before

## Before - focused

template_list_path_wrapping_text_focused_before

## After - unfocused

template_list_path_wrapping_text_after

## After - focused

template_list_path_wrapping_text_focused_after

## Notes for reviewers

It's worth testing with the following pages:
- templates
- templates with a search
- templates > copy an existing template > Choose an existing template
- reply to an inbound sms > Choose a template

### DIFF
--- a/app/assets/stylesheets/components/message.scss
+++ b/app/assets/stylesheets/components/message.scss
@@ -58,18 +58,22 @@ a {
 .template-list {
 
   &-item {
+    display: block;
     margin-bottom: 0;
 
     &-with-checkbox {
+      padding-left: $govuk-touch-target-size + $govuk-checkboxes-label-padding-left-right;
+      min-height: $govuk-touch-target-size;
 
       .template-list-item-hint {
+        width: 100%;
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
great work!

PR #5284, opened 2024-11-06
Update existing join service content

leohemsted:
the main thing here is tweaking a couple of endpoint paths, and some file/folder structure.

also incorporated our new details callout for what to do if you aren't part of that org 
image


kr8n3r (approved):
change make sense to me

PR #5285, opened 2024-11-06
remove nvm

leohemsted:
remove nvm from docker usage as we don't really need to switch between multiple versions of node within docker, so we can just install our chosen lts version of node directly, following instructions as found on https://github.com/nodesource/distributions?tab=readme-ov-file#using-ubuntu-nodejs-20

TODO:

* need to remove from concourse unit-test-image, and then update the build step here, and figure out how to do that nicely in lockstep


risicle:
Do we need an analogous PR for document-download-frontend?

risicle (approved):

PR #5286, opened 2024-11-06
feat: Add service permissions for service join approval update

joybytes:
## Summary
- on the service join request journey, when the user is approved, they have the option of selecting service permissions
- in this PR we also add the ability for them to select folder permissions just like we do on the invite-user journey
- this is the front-end part PR, and the back-end PR is can be found [here](https://github.com/alphagov/notifications-api/pull/4247)

### Ticket:
[Let folder permissions be selected when approving a service join request](https://trello.com/c/70y7hQOW/1021-let-folder-permissions-be-selected-when-approving-a-service-join-request)


klssmith:
This looks good - one small thing to add is clearing the `service-{service_id}-template-folders` key from Redis. That contains a list of folders for the service along with the users who have permission to view each one, so if we've changed any folder permissions that value will be outdated. You can add a decorator like this one to the `update_service_join_requests` method https://github.com/alphagov/notifications-admin/blob/bcd091f465b472bb083aef08897e70895f803c8e/app/notify_client/user_api_client.py#L174

app/models/service.py
```diff
@@ -682,5 +682,5 @@ def is_cancelled(self):
     def from_id(cls, request_id):
         return cls(service_api_client.get_service_join_requests(request_id))
 
-    def update(self, **kwargs):
-        return service_api_client.update_service_join_requests(self.id, self.requester["id"], **kwargs)
+    def update(self, service_id, **kwargs):
+        return service_api_client.update_service_join_requests(self.id, self.requester["id"], service_id, **kwargs)
```
klssmith, line 686: This is an optional change, but if you wanted to you can get the service id with `self.service_id` instead of passing it as an argument to the method.

app/main/views/manage_users.py
```diff
@@ -182,15 +182,24 @@ def service_join_request_choose_permissions(service_id, request_id):
...
-    form = JoinServiceRequestSetPermissionsForm()
+    form = JoinServiceRequestSetPermissionsForm(
+        all_template_folders=current_service.all_template_folders,
+        folder_permissions=[f["id"] for f in current_service.all_template_folders],
+    )
+
     service_join_request = ServiceJoinRequest.from_id(request_id)
     requested_by = service_join_request.requester
 
+    if not current_service.has_permission("email_auth"):
```
klssmith, line 193: We'll need to work out what should happen with the auth radio buttons but that can be a separate story.

klssmith (approved):

klssmith (approved):

PR #5287, opened 2024-11-06
Give all folder icons alt text

tombye:
Their purpose in the visual interface is to show the thing they are next to is a folder. We need this information in the semantic interface too, so users of assistive tech can access it. We made the Jinja SVG macro so the `<svg>` shape can have an accessible name but never set it.

Note: this doesn't change the folder on the team member page because it adds no extra information to the following text ('X of X folders').

## Before

template_list_item_with_folder_before

## After

template_list_item_with_folder_after


kr8n3r (approved):
as expected, "Folder" announced by AT

PR #5288, opened 2024-11-08
Fix text reflow on inbound sms inbox and service api pages

kr8n3r:
## What
Fixes styling on those pages where long content overflows the container and forces horizontal scrolling.

See individual commits for more details

## Visual changes

**inbound sms inbox**

Screenshot 2024-11-08 at 10 10 35
Screenshot 2024-11-08 at 10 01 32

**service api integration page**

Screenshot 2024-11-08 at 10 04 33
Screenshot 2024-11-08 at 10 02 35


tombye (changes requested):
All looks good, I just spotted a stray underline that only appears on smaller / zoomed in screens we could do with fixing:

image

kr8n3r:
@tombye so the underline is set on the entire  govuk-summary-text in govuk-frontend
```
@supports not (-ms-ime-align:auto) {
    .govuk-details__summary-text {
         text-decoration: underline;
```
the underline is there because we put in additional spans into it.
the way to fix that would be to override text-decoration that comes from govuk-frontend and set it ourselves on `.api-notifications-item__recipient`.

What do you think?

kr8n3r:
@tombye see https://github.com/alphagov/notifications-admin/pull/5288/commits/ae46ed3ba72851a152da52efeb95ea15ff88b9d2

Screenshot 2024-11-12 at 16 13 33

kr8n3r:
one more fix :) https://github.com/alphagov/notifications-admin/pull/5288/commits/e402719b23edf1cec790022e89593a73f4d97d19

tombye (approved):
Extra approval to account for the bonus focus fix. Thanks for doing it! 👍🏻

PR #5291, opened 2024-11-08
Conditionally focus template page forms

kr8n3r:
Fixes: https://trello.com/c/Dg0mE6e5/1041-validation-for-new-template-form-doesnt-focus-the-error-summary

When a user clicks say "add a new template" button, we show and focus that form. Should then a user submit that form without any selection, we reload the page, show the error-summary banner, check which form they submitted and then show it and focus it.

This caused an issue for screenreader users, which expected the error summary to be focused.

This PR amends the javascript module logic to check if the rendered page contains an error summary. 
If it does, we don't focus the form and let Design System Javascript handle error-summary focus.

As the error-summary contains the link to the errored form, users can then make changes quickly.

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -289,7 +289,7 @@
       // make sticky JS recalculate its cache of the element's position
       GOVUK.stickAtBottomWhenScrolling.recalculate();
 
-      if (currentStateObj && ('setFocus' in currentStateObj)) {
+      if (currentStateObj && ('setFocus' in currentStateObj) && !this.formHasError()) {
         scrollTop = $(window).scrollTop();
         currentStateObj.setFocus();
         $(window).scrollTop(scrollTop);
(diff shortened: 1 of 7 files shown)


tombye (changes requested):
The code makes sense and it works well. The spacing on smaller screens / is a bit tight between the error summary and the side nav:

image

I think it would be worth seeing if we can fix that but everything else looks good.

kr8n3r:
@tombye in most other pages there's a back-link between the left-nav and error summary. adding bottom spacing to the left nav will affect all the other pages, so should we address that in this PR?

tombye:
I was thinking that might make it a good candidate for just using an override class from design system on the top of the error summary. That way it's just for that page.

kr8n3r:
can we actually pass in a class to the error-summary the way we have it set up as it's automagical?

kr8n3r:
what do you think? https://github.com/alphagov/notifications-admin/pull/5291/commits/f3a1e5754e0f81689980064b471fd804140e4ca6

Screenshot 2024-11-13 at 12 28 53

tombye (approved):
Thanks for addressing the comment. Now looking good to go 👍🏻

PR #5294, opened 2024-11-12
Remove old phonenumber validation

rparke:


tests/app/main/forms/test_service_contact_details_form.py
```diff
@@ -95,30 +95,3 @@ def test_form_phone_number_validation_fails_with_invalid_phone_number_field(noti
         assert not form.validate_on_submit()
         assert len(form.errors) == 1
         assert "Enter a phone number in the correct format" in form.errors["phone_number"]
-
-
-@pytest.mark.parametrize(
-    "short_number, allowed",
-    (
-        ("119", True),
```
CrystalPea, line 103: I think the question should be - are there valid cases where service's contact number is a three-digit number?

It looks like it might be from this list? https://www.inyourarea.co.uk/news/10-three-digit-phone-numbers-and-what-theyre-for

For example Track and Trace had a 3-digit number, and there are some other ones.

I wonder how we previously validated this?

rparke: This is an easy fix, and it's an exception to pure `libphonenumber` validation I'm okay with adding. We can either have an allow list, or just say "any 3 digit number that successfully parses to a "possible" UK number, we are just going to say is valid"

CrystalPea (approved):
Looks great, nicely done with contact phone numbers validation 🙌🏼 

Ready to merge when utils are merged and requirements updated 👍🏼

app/utils/validation.py
```diff
@@ -0,0 +1,11 @@
...
+from notifications_utils.recipient_validation.phone_number import PhoneNumber
+
+
+def format_phone_number_human_readable(number):
+    try:
+        phone_number = PhoneNumber(number)
+    except InvalidPhoneError:
+        # if there was a validation error, we want to shortcut out here, but still display the number on the front end
+        return number
+    return phone_number.get_human_readable_format()
```
quis, line 11: This would live better in https://github.com/alphagov/notifications-admin/blob/main/app/formatters.py

CrystalPea (approved):

PR #5295, opened 2024-11-12
Remove add-or-join-service for user register flow

spatel033:
'Add or Join Service' step added for register journey during 'Join Service Request' prototype. Remove this step and take user to to the 'Your services page'. More info on [this card](https://trello.com/c/9X6eYNA3/1046-remove-add-or-join-service-from-create-new-account-register-flow).


app/main/views/verify.py
```diff
@@ -82,7 +82,7 @@ def activate_user(user_id):
         return redirect(url_for("main.organisation_dashboard", org_id=organisation_id))
 
     if user.default_organisation.can_ask_to_join_a_service:
-        return redirect(url_for("main.add_or_join_service"))
+        return redirect(url_for("main.your_services"))
```
leohemsted, line 85: can you remove the endpoint itself as well, and add a line to the historical redirects?

app/main/views/verify.py
```diff
@@ -82,7 +82,7 @@ def activate_user(user_id):
...
 
     return redirect(url_for("main.add_service", first="first"))
```
leohemsted, line 87: minor, and i know you didnt touch this - but we can remove this `first` parameter, as the `add_service` endpoint doesn't actually appear to do anything with that information.

(we used to use the `first` arg to tell if it was a user's first service, so that we could redirect them to the new user tour, but now we just have that logic separately in `add_service`)

app/main/views/verify.py
leohemsted, line 85: can you remove the endpoint itself as well, and add a line to the historical redirects so that if a user hits the old endpoint, they get redirected to your_services?

leohemsted (approved):
nice :+1:

PR #5297, opened 2024-11-15
Add error summary on error for ask-to-join-a-service

spatel033:
Adding error summary when submitting ask to join a live service form without requester. More details on [this card](https://trello.com/c/kOPFdylE/1053-ask-to-join-a-service-request-form-is-missing-error-summary-on-error).

image

image

### DIFF
--- a/app/main/views/join_service.py
+++ b/app/main/views/join_service.py
@@ -53,6 +53,10 @@ def join_service_ask(service_to_join_id):
         "views/join-a-service/ask.html",
         service=service,
         form=form,
+        error_summary_enabled=True,
+        error_summary_extra_params={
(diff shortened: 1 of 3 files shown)


kr8n3r:
i think the error summary should be 2/3 same as the content. you can achieve that my passing

```
error_summary_extra_params={
    "classes": "govuk-!-width-two-thirds",
 }
```
in `join_service.py` right after you enabled the error summary flag
and then pass it to the error summary macro in `withoutnav_template.htm` with

```
{{ errorSummary(form, error_summary_extra_params) }}
```

Something similar enabled and done in {{ errorSummary(form, error_summary_extra_params) }}

kr8n3r (approved):

PR #5298, opened 2024-11-18
Announce successful letter template PDF attachment upload

kr8n3r:
Currently, when an PDF attachment is successfully uploaded to a template, we redirect to a template preview and anchor to the first page of the attachment.

Visually that shows that the attachment was successfully uploaded (we do show an error if it wasn't), but screenreader users do not get informed.

This PR attaches a `upload_successful="yes"` query string to the URL, based on which we render the Notification banner component -  Success variant - https://design-system.service.gov.uk/components/notification-banner/

The template will then show the notification banner before the H1 with an anchor link to the first page of the attachment that already exists.

This way we indicate that the upload was successful for both visual and assistive tech users.

**Before**

https://github.com/user-attachments/assets/e49c6f1d-c495-4f2a-93da-986617177afe

**After**
(simulating successful upload as my local dev is broken)

https://github.com/user-attachments/assets/14e1c6b7-eb95-4889-a097-2e4df60898ef

### DIFF
--- a/app/assets/javascripts/esm/all-esm.mjs
+++ b/app/assets/javascripts/esm/all-esm.mjs
@@ -1,5 +1,5 @@
 // GOVUK Frontend modules
-import { createAll, Header, Button, Radios, ErrorSummary, SkipLink, Tabs } from 'govuk-frontend';
+import { createAll, Header, Button, Radios, ErrorSummary, SkipLink, Tabs, NotificationBanner } from 'govuk-frontend';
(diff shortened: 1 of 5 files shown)


kr8n3r:
doesn't work reliably on page reload. We'll have to with a Notification banner, but that changes the the current behaviour where we anchor to the first page of attachment

kr8n3r:
changing approaches

PR #5299, opened 2024-11-19
Guidance for adding an unsubscribe link in the email for API users

saimaghafoor:
Adding guidance for API users to add a link at the bottom of their subscription emails.

### DIFF
--- a/app/templates/partials/templates/guidance-links-urls-emails.html
+++ b/app/templates/partials/templates/guidance-links-urls-emails.html
@@ -9,9 +9,31 @@
 </p>
 <pre class="formatting-example"><code class="lang-md">[Apply now](https://www.gov.uk/example)
 </code></pre>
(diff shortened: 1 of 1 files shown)


quis:
I don’t think we should add this info here because we risk non-API users trying to follow it. Let’s keep this in the API documentation only

saimaghafoor:
Thanks for the input @quis - it's defo proving to be a tricky one as I can see how it could confuse users. During the API review, it felt like this guidance didn't fit neatly in the API either. The audience for the docs is mostly different to who would actually add/edit templates (and probably need to add this link). @CrystalPea

CrystalPea:
Hey both, it is a tricky one indeed - maybe let's have a chat about it? Only I see we are all very busy with meetings or off... Let me have a think and get back to you both 🙌🏼

saimaghafoor:
Leaving the content as it currently is. No changes needed.

PR #5302, opened 2024-11-21
Remove CSS for .banner-tour variant of banner

tombye:
This variant was created for emergency alerts. Its HTML was in the .html files kept in:

app/templates/views/broadcast/tour/

These were removed in:

https://github.com/alphagov/notifications-admin/pull/4922

...so this CSS is no longer used.


quis:
This is still used in the onboarding journey. You won’t find it just by grepping for `banner-tour` in the coebase, but you can begin spelunking here: https://github.com/alphagov/notifications-admin/blob/802d6c7a92f61c88b4cb416cce1981d653e1eac0/app/templates/partials/tour.html#L3

PR #5308, opened 2024-11-28
Returned letters guidance for UI and API

saimaghafoor:
Updated 'Undelivered letters and returned mail' content as returned letters callback available in API.

### DIFF
--- a/app/templates/views/guidance/using-notify/delivery-times.html
+++ b/app/templates/views/guidance/using-notify/delivery-times.html
@@ -34,9 +34,12 @@
   <p class="govuk-body">First class letters are delivered one day after they’re dispatched. Second class letters are delivered 2 days after they’re dispatched. Royal Mail delivers from Monday to Saturday, excluding bank holidays.</p>
   <p class="govuk-body">Letters to Europe are delivered 3 to 5 days after they’re dispatched. Letters sent anywhere else in the world take 5 to 7 days to arrive.</p>
 
-  <h3 id="returned-mail" class="heading-small">Undelivered letters and returned mail</h3>
-  <p class="govuk-body">Every letter we send includes our printer’s address on the back of the envelope. It’s not possible to customise the return address.</p>
-  <p class="govuk-body">Returned mail is destroyed by our printer.</p>
-  <p class="govuk-body">Each letter has a unique ID. Our printer sends us the ID of every letter they destroy. We’ll contact you if the IDs match any of your letters. This is a partly manual process, so it can take a few weeks.</p>
+  <h3 id="returned-mail" class="heading-small">Returned letters</h3>
+  <p class="govuk-body">Every letter we send includes our print provider's address on the back of the envelope. It’s not possible to customise the return address.</p>
+  <p class="govuk-body">Undelivered letters are returned to our print provider and are destroyed.</p>
+  <p class="govuk-body">A returned letters report is available on the dashboard. Use this report to help identify the letters returned.</p>
+  <p class="govuk-body">If you’re using the Notify API, you can automate identifying the returned letters. Follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
+    href="https://www.notifications.service.gov.uk/using-notify/api-documentation">API documentation</a>.</p>
+  <p class="govuk-body">You’ll need a developer on your team to set this up for you.</p>
 
 {% endblock %}


app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,12 @@ <h2 id="letters" class="heading-medium">Letters</h2>
   <p class="govuk-body">First class letters are delivered one day after they’re dispatched. Second class letters are delivered 2 days after they’re dispatched. Royal Mail delivers from Monday to Saturday, excluding bank holidays.</p>
   <p class="govuk-body">Letters to Europe are delivered 3 to 5 days after they’re dispatched. Letters sent anywhere else in the world take 5 to 7 days to arrive.</p>
 
-  <h3 id="returned-mail" class="heading-small">Undelivered letters and returned mail</h3>
-  <p class="govuk-body">Every letter we send includes our printer’s address on the back of the envelope. It’s not possible to customise the return address.</p>
-  <p class="govuk-body">Returned mail is destroyed by our printer.</p>
-  <p class="govuk-body">Each letter has a unique ID. Our printer sends us the ID of every letter they destroy. We’ll contact you if the IDs match any of your letters. This is a partly manual process, so it can take a few weeks.</p>
+  <h3 id="returned-mail" class="heading-small">Returned letters</h3>
+  <p class="govuk-body">Every letter we send includes our print provider's address on the back of the envelope. It’s not possible to customise the return address.</p>
```
karlchillmaid, line 38:
```suggestion
  <p class="govuk-body">Every letter we send includes our print provider’s address on the back of the envelope. It’s not possible to customise the return address.</p>
```

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,12 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
+  <p class="govuk-body">Undelivered letters are returned to our print provider and are destroyed.</p>
+  <p class="govuk-body">A returned letters report is available on the dashboard. Use this report to help identify the letters returned.</p>
```
karlchillmaid, line 40: I’m suggesting this change because:

* the first sentence could be more active
* the second sentence kinda repeats the first – or provides detail that is implicit in the first sentence

```suggestion
  <p class="govuk-body">You can see a report of returned letters on the dashboard.</p>
```

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,12 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
+  <p class="govuk-body">If you’re using the Notify API, you can automate identifying the returned letters. Follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
+    href="https://www.notifications.service.gov.uk/using-notify/api-documentation">API documentation</a>.</p>
+  <p class="govuk-body">You’ll need a developer on your team to set this up for you.</p>
```
karlchillmaid, line 43: As we’re describing the UI and API behaviours under a single h3, is there a risk that users could misinterpret this final line/paragraph and think it applies to all of the above?

saimaghafoor: I've made some changes ti help separate the UI stuff from the API in returned letters  @karlchillmaid @quis

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,13 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
 
-  <h3 id="returned-mail" class="heading-small">Undelivered letters and returned mail</h3>
-  <p class="govuk-body">Every letter we send includes our printer’s address on the back of the envelope. It’s not possible to customise the return address.</p>
-  <p class="govuk-body">Returned mail is destroyed by our printer.</p>
-  <p class="govuk-body">Each letter has a unique ID. Our printer sends us the ID of every letter they destroy. We’ll contact you if the IDs match any of your letters. This is a partly manual process, so it can take a few weeks.</p>
+  <h2 id="returned-mail" class="heading-medium">Returned letters</h2>
+  <p class="govuk-body">Every letter we send includes our print provider’s address on the back of the envelope. You cannot customise the return address.</p>
+  <p class="govuk-body">Undelivered letters are returned to our print provider and destroyed.</p>
+  <p class="govuk-body">You can see a report of returned letters on the dashboard.</p>
+  <h3 id="returned-mail" class="heading-small">If you want to identify the returned letters</h3>
```
karlchillmaid, line 41: I understand this to mean that you cannot identify the returned letters from the report on the dashboard – is that right?

saimaghafoor: If you use the API, it automates the whole process - returned letters / identifying them. Non-API users would still need the to view the report and then identify letters

quis: Following the above comment, this section can then just become a line explaining that, if you use the API, you can also get this information via a callback.

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,13 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
   <p class="govuk-body">Letters to Europe are delivered 3 to 5 days after they’re dispatched. Letters sent anywhere else in the world take 5 to 7 days to arrive.</p>
 
-  <h3 id="returned-mail" class="heading-small">Undelivered letters and returned mail</h3>
-  <p class="govuk-body">Every letter we send includes our printer’s address on the back of the envelope. It’s not possible to customise the return address.</p>
-  <p class="govuk-body">Returned mail is destroyed by our printer.</p>
-  <p class="govuk-body">Each letter has a unique ID. Our printer sends us the ID of every letter they destroy. We’ll contact you if the IDs match any of your letters. This is a partly manual process, so it can take a few weeks.</p>
+  <h2 id="returned-mail" class="heading-medium">Returned letters</h2>
+  <p class="govuk-body">Every letter we send includes our print provider’s address on the back of the envelope. You cannot customise the return address.</p>
+  <p class="govuk-body">Undelivered letters are returned to our print provider and destroyed.</p>
+  <p class="govuk-body">You can see a report of returned letters on the dashboard.</p>
```
quis, line 40: I think we should say what information the report contains (or mention some of the most important columns). Because it can take a long time for letters to come back we should explain what information the report doesn’t contain (recipient address etc).

app/templates/views/guidance/using-notify/delivery-times.html
quis: Then following my other comment, we can say that you can get the same information (as in the report) via the API and this is how

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,21 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
+  <h3 id="returned-mail" class="heading-small">Identify the returned letters</h3>
+ <p class="govuk-body">To help you identify what letters are returned, the report will show details such as the:
+ </p>
+ <ul class="govuk-list govuk-list--bullet">
+   <li>template ID</li>
+   <li>date the letter was sent</li>
```
quis, line 46: Should we also talk about the reference here and how it gets populated (from a spreadsheet column or [via the API](https://docs.notifications.service.gov.uk/python.html#send-a-letter-arguments-reference-optional))?

quis: There’s a whole bunch of guidance about this already in the support manual: https://docs.google.com/document/d/1F41AxZdwjebI4yCIhUQvs72ozHfG2aqCnhelSYu_Wi0/edit?tab=t.0#heading=h.y0jci2y8ztjr

saimaghafoor: Thanks for signposting to the manual. I never knew this info on returned letters existed! I'll have a think about how much detail we should get into here...or if it's better elsewhere

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,23 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
+   <li>a reference for each letter if you include this in the list of addresses you upload</li>
```
quis, line 47: I think we should include a bit more about the reference here because it’s not specific to letters that are uploaded 

```suggestion
   <li>a reference for each letter if you include either:
       <ul class="govuk-list govuk-list--bullet">
           <li>a column called ‘reference’ in the list of addresses you upload</li>
           <li>the <code>reference</code> argument when making an API call</li>
       </ul>
   </li>
```

app/templates/views/guidance/using-notify/delivery-times.html
```diff
@@ -34,9 +34,23 @@ <h2 id="letters" class="heading-medium">Letters</h2>
...
+
+ </ul>
+ <p class="govuk-body">For security reasons, the report will not show the recipient postal address. </p>
+  <h3 id="returned-mail" class="heading-small">Automate identifying the returned letters</h3>
+  <p class="govuk-body">Use the Notify API to get the returned letters report and identify the letters. Follow the instructions in our <a class="govuk-link govuk-link--no-visited-state"
+    href="https://www.notifications.service.gov.uk/using-notify/api-documentation">API documentation</a>.</p>
+  <p class="govuk-body">You’ll need a developer on your team to set this up for you.</p>
```
quis, line 54: Should we consider removing this for now and adding it back in once this feature is ready? At the moment we can’t merge this PR because it talks about something which isn’t shipped yet

quis:
Looks like this now:

image

rparke (approved):

PR #5310, opened 2024-11-29
Notifications csv report timeout

spatel033:
There is 30s timeout error set for eventlet which is causing issue for large csv file. Remove stream_with_context and download csv without streaming and raise timeout error if request takes > 30s. More info on [this card](https://trello.com/c/pnDkObaV/1062-streaming-csv-responses-silently-fail-due-to-eventlet-timeout).

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -6,7 +6,6 @@ from flask import (
     redirect,
     render_template,
     request,
-    stream_with_context,
     url_for,
 )
 from markupsafe import Markup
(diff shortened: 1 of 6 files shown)


risicle:
I don't understand why this is catching `EventletTimeout` and turning it into a 504 in the view. This is what the `eventlet_timeout` handler is already doing across the whole app. What difference in behaviour is this making?

risicle:
👍 Yup, I don't think you should have to do anything special for the exception handling.

joybytes (approved):
lgtm!

PR #5311, opened 2024-11-29
Install packages using `uv`

quis:
uv is:
- between 10× and 100× faster than pip
- a drop-in replacement for `pip install`, `pip compile`, etc

***

Replicates what we’ve already done in https://github.com/alphagov/notifications-api/pull/4271


risicle:
> between 10× and 100× faster than pip

To a large extent because it doesn't precompile bytecode.

The Dockerfile needs `UV_COMPILE_BYTECODE=1` set to stop https://github.com/alphagov/notifications-api/pull/4286 happening

PR #5314, opened 2024-12-03
Use `format_thousands` template filter consistently

quis:
This is easier to read and requires less arcane Python string formatting knowledge


kr8n3r (approved):
yeah I was strugling with that earlier

PR #5315, opened 2024-12-04
ITHC recommended headers

spatel033:
ITHC recommended to add and update headers. Some headers were deprecated and incompatible across all browsers. This PR to fix security headers, report was also generated and shared on [this card](https://trello.com/c/EfkvM0wg/1070-ensure-recommended-headers-are-applied-for-admin-app).

There are headers updated related to font, style and script, not sure if this would break anything on server.

image
image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -338,7 +338,6 @@ def load_user_id_before_request():
 
 #  https://www.owasp.org/index.php/List_of_useful_HTTP_headers
 def useful_headers_after_request(response):
-    response.headers.add("X-Frame-Options", "deny")
     response.headers.add("X-Content-Type-Options", "nosniff")
     response.headers.add("X-XSS-Protection", "1; mode=block")
     response.headers.add(
@@ -351,6 +350,8 @@ def useful_headers_after_request(response):
             "font-src 'self' {asset_domain} data:;"
             "img-src 'self' {asset_domain}"
             " *.notifications.service.gov.uk {logo_domain} data:;"
+            "style-src 'self' {asset_domain} 'unsafe-inline';"
+            "frame-ancestors 'self';"
             "frame-src 'self';".format(
                 asset_domain=current_app.config["ASSET_DOMAIN"],
                 logo_domain=current_app.config["LOGO_CDN_DOMAIN"],
@@ -370,6 +371,25 @@ def useful_headers_after_request(response):
     response.headers.add("Cache-Control", "no-store, no-cache, private, must-revalidate")
     for key, value in response.headers:
         response.headers[key] = SanitiseASCII.encode(value)
+    response.headers.add("Strict-Transport-Security", "max-age=31536000; preload")
+    response.headers.add("Referrer-Policy", "strict-origin-when-cross-origin")
+    response.headers.add(
+        "Cross-Origin-Embedder-Policy",
+        "require-corp {asset_domain};".format(asset_domain=current_app.config["ASSET_DOMAIN"]),
+    )
+    response.headers.add(
+        "Cross-Origin-Opener-Policy",
+        "same-origin {asset_domain};".format(asset_domain=current_app.config["ASSET_DOMAIN"]),
+    )
+    response.headers.add(
+        "Cross-Origin-Resource-Policy",
+        "same-origin {asset_domain};".format(asset_domain=current_app.config["ASSET_DOMAIN"]),
+    )
+    response.headers.add(
+        "Permissions-Policy",
+        "geolocation=(), microphone=(), camera=(), autoplay=(), payment=(), sync-xhr=()",
+    )
+    response.headers.add("Server", "Cloudfront")
     return response
 
 
--- a/tests/app/main/views/test_headers.py
+++ b/tests/app/main/views/test_headers.py
@@ -7,7 +7,6 @@ def test_owasp_useful_headers_set(
     client_request.logout()
     response = client_request.get_response(".index")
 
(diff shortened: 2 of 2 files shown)


kr8n3r:
quick glance: this is at minimum breaking JS execution.  we need to get this onto a dev environment to test in something close to production

app/__init__.py
```diff
@@ -370,6 +371,16 @@ def useful_headers_after_request(response):
     response.headers.add("Cache-Control", "no-store, no-cache, private, must-revalidate")
     for key, value in response.headers:
         response.headers[key] = SanitiseASCII.encode(value)
+    response.headers.add("Strict-Transport-Security", "max-age=31536000; preload")
+    response.headers.add("Referrer-Policy", "strict-origin-when-cross-origin")
```
kr8n3r, line 375: this is fine

app/__init__.py
```diff
@@ -370,6 +371,16 @@ def useful_headers_after_request(response):
...
+    response.headers.add("Cross-Origin-Embedder-Policy", "require-corp")
+    response.headers.add("Cross-Origin-Opener-Policy", "same-origin")
+    response.headers.add("Cross-Origin-Resource-Policy", "same-origin")
+    response.headers.add(
+        "Permissions-Policy",
+        "geolocation=(), microphone=(), camera=(), autoplay=(), payment=(), sync-xhr=()",
```
kr8n3r, line 381: fine, to be there as we don't use most of this, though we might need to check synchronous xhr requests

app/__init__.py
```diff
@@ -338,19 +338,20 @@ def load_user_id_before_request():
...
-    response.headers.add("X-Frame-Options", "deny")
     response.headers.add("X-Content-Type-Options", "nosniff")
     response.headers.add("X-XSS-Protection", "1; mode=block")
     response.headers.add(
         "Content-Security-Policy",
         (
-            "default-src 'self' {asset_domain} 'unsafe-inline';"
-            "script-src 'self' {asset_domain} 'unsafe-inline' 'unsafe-eval' data:;"
+            "default-src 'self' {asset_domain};"
+            "script-src 'self' {asset_domain};"
```
kr8n3r, line 347: this breaks JS as a SHA will need to be created / added

kr8n3r:
just the JS errors. some things we can sort out easily, other's not as it's legacy JS
Screenshot 2024-12-05 at 10 37 28

app/__init__.py
```diff
@@ -370,6 +371,16 @@ def useful_headers_after_request(response):
     response.headers.add("Cache-Control", "no-store, no-cache, private, must-revalidate")
     for key, value in response.headers:
         response.headers[key] = SanitiseASCII.encode(value)
+    response.headers.add("Strict-Transport-Security", "max-age=31536000; preload")
+    response.headers.add("Referrer-Policy", "strict-origin-when-cross-origin")
+    response.headers.add("Cross-Origin-Embedder-Policy", "require-corp")
```
kr8n3r, line 376: Given that COEP could block assets from loading, it's a good idea to leverage the Report Only version of the header coupled with reporting via the Reporting API first to ensure that no harm will be caused to your site.

`Cross-Origin-Embedder-Policy-Report-Only: (unsafe-none|require-corp); report-to="default"`

we need to carefully check in a prod-like environemt

app/__init__.py
```diff
@@ -370,6 +371,16 @@ def useful_headers_after_request(response):
...
+    response.headers.add("Cross-Origin-Opener-Policy", "same-origin")
+    response.headers.add("Cross-Origin-Resource-Policy", "same-origin")
```
kr8n3r, line 378: this will likely cause static logos and static asstes on prod not to load as they're on a different origin to the admin app

kr8n3r:
_Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-Nv8QC5KMerDUPUkQ13N3EPtBelVQH+r4Wo3DV9ADatk='), or a nonce ('nonce-...') is required to enable inline execution._

this is for 

```
  <style>
      .govuk-header__container { border-color: #81878b }
  </style>
```

I wonder if we need this @tombye ?

quis:
> I wonder if we need this @tombye ?

It’s still really useful to have a different colour for each environment. We could move this into a stylesheet served from the admin app though.

app/__init__.py
kr8n3r: i've added a SHA for govuk-frontend

kr8n3r:
> > I wonder if we need this @tombye ?
> 
> It’s still really useful to have a different colour for each environment. We could move this into a stylesheet served from the admin app though.

yeah something like that  @quis. quick search shows that inline styles are used in a bunch of places; letter preview and svgs, so need to check if this `style-src` should remain as is

tombye:
I've chatted with @leohemsted and @spatel033 about the points raised by @kr8n3r. We think we need to not block this PR with trying to solve the problems with:
- permissions policy (`sync-xhr`)
- Content Security Policy `insafe-inline` for scripts and CSS, `unsafe-eval` and `data:`

I'll write up a card to spike solutions to these so they don't block this PR.

kr8n3r:
we need to remove that directive or change the way we load the branding so it's not in an iframe
Screenshot 2024-12-06 at 10 07 43

kr8n3r (approved):
I've tested this reduced scope changes as much as possible on local and dev-a. Things appear to work as normal , but we should double check after deploying, first on preview and then on prod

PR #5318, opened 2024-12-05
Use the `Notification` model when viewing a single notification

quis:
This continues the work done in https://github.com/alphagov/notifications-admin/pull/5232

Notifications are one of the last database objects in the admin app that we are still passing around as raw JSON, rather than wrapping them in a proper Python object.


CrystalPea (approved):
:shipit:

PR #5319, opened 2024-12-05
Split breadcrumb HTML off from h1

tombye:
On pages where we currently put the name of the template or folder in with the path to it, this splits the name off from the path.

Intended to make it less confusing to screen reader users. Full details in the story:

https://trello.com/c/YGxJqotD/991-h1-for-template-or-folder-page-is-confusing-for-screen-readers

image

image

image

### Pages to check

This pattern appears on:
- the templates page at the root and in a folder
- manage folder page
- copy template page
- choose template (as part of the 'respond to inbound message' flow)

### Changes to the page title

Because the h1 was confusing when it contained the path, I also changed the page title because it also contained the path. 

Many other sites, including www.gov.uk, use the 'page - parent page - website' pattern because the parent page gives enough context to tell pages with the same name apart. I used a variation of this which also includes the service.

@karlchillmaid will need to sign off on these changes.

### DIFF
--- a/app/assets/stylesheets/components/message.scss
+++ b/app/assets/stylesheets/components/message.scss
@@ -303,25 +303,26 @@ a {
 
   &-breadcrumb,
   & {
-    display: inline-block;
+    display: inline;
   }
 
   &-container {
     margin: govuk-spacing(3) 0 govuk-spacing(4);
-
-    & > .folder-heading {
-      margin: 0;
-    }
   }
 
   &-breadcrumb {
     list-style: none;
-    display: inline-block;
-    min-height: 30px; // set min-height to line-height with folder icons
     @include govuk-font(24, $weight: "bold");
 
-    li {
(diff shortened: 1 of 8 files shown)


kr8n3r:
this looks good.

should manage page also have "Nest 1" in this case without a link?
Screenshot 2024-12-06 at 09 52 11

kr8n3r:
i think some padding/margin got lost somewhere

Screenshot 2024-12-06 at 09 58 30
Screenshot 2024-12-06 at 10 00 21

tombye:
> this looks good.
> 
> should manage page also have "Nest 1" in this case without a link? Screenshot 2024-12-06 at 09 52 11

I think so, assuming I understand. Here's a screenshot of what a similar manage folder page looks like currently:

image

kr8n3r (approved):
happy with the code

PR #5321, opened 2024-12-06
Rewrite ColourPreview JS as ES Module

kr8n3r:
As we want to migrate all of our legacy Javascript files to ESM, this is a small part.
Rewrite follows the same convention we previously used for FocusBanner and Collapsible checkboxes.

In this PR we:
- remove old module JS file


tombye (approved):
Makes sense and works when tested 👍🏻

PR #5322, opened 2024-12-06
Replace HoganJS with template literals

tombye:
https://trello.com/c/mf90DD4c/1080-use-template-literals-instead-of-hoganjs-for-js-templating

Because:
- our supported browsers understand template literals
- HoganJS isn't needed now we have template literals
- HoganJS uses `eval()`, and [we'd like to ban it in our Content Security Policy](https://github.com/alphagov/notifications-admin/pull/5315#pullrequestreview-2481267816)
- we can delete some code

This also makes the JavaScript delivered to our users 8915 smaller before gzipping.

## Notes for reviewers

These changes make use of destructuring to unpack object properties into variables. We use the same feature detection to check if the code should run as design system, checking for `noModule`. Based on caniuse data, this should allow use of destructuring:
- [caniuse for destructuring](https://caniuse.com/?search=destructuring)
- [caniuse for noModule](https://caniuse.com/?search=nomodule)

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -76,7 +76,6 @@ export default [
   // ES5 JS compilation
   {
     input: [
-      paths.npm + 'hogan.js/dist/hogan-3.0.2.js',
       paths.npm + 'jquery/dist/jquery.min.js',
       paths.npm + 'query-command-supported/dist/queryCommandSupported.min.js',
       paths.npm + 'timeago/jquery.timeago.js',
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -124,10 +124,7 @@ export default [
         entryFileName: 'all.js'
       }),
       terser({
-        ecma: '5',
-        mangle: {
-          reserved: ["Hogan"]
-        }
+        ecma: '5'
       }),
     ]
   }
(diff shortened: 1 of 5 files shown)


rollup.config.mjs
kr8n3r, line 1: non blocking: we can also remove (in a separate PR)
```
mangle: {
   reserved: ["Hogan"]
}
```
mangle is `true` by default - https://github.com/terser/terser#minify-options:~:text=mangle%20(default%20true)%20%E2%80%94%20pass%20false%20to%20skip%20mangling%20names%2C%20or%20pass%20an%20object%20to%20specify%20mangle%20options%20(see%20below).

kr8n3r (approved):
love it !

rollup.config.mjs
tombye: Done, in https://github.com/alphagov/notifications-admin/pull/5322/commits/6f6f58cf2f35c7c2e5162ecd5770f00d1fe63d34. For context, that config stopped the compression done by the Terser plugin from obsfucating the 'Hogan' global variable. No longer needed now hogan is gone.

spatel033 (approved):

PR #5325, opened 2024-12-11
Add csp_nonce for inline script and gov_template

spatel033:
Previously we added `unsafe-inline` and `unsafe-eval` directives for `script-src` header. We can remove these  directives as part of ITHC after removing `Hogan`. Additionally to fix CSP errors added `csp_nonce` for request to allow `govuk_frontend_jinja/template.html` and inline `script`. More details on [this card](https://trello.com/c/DTim1BKC/1083-spike-into-solutions-for-ithc-header-changes-which-affect-frontend).

### DIFF
--- a/app/templates/admin_template.html
+++ b/app/templates/admin_template.html
@@ -1,5 +1,7 @@
 {% extends "govuk_frontend_jinja/template.html"%}
 
+{% set cspNonce = request.csp_nonce %}
+
 {% block headIcons %}
   <link rel="icon" sizes="48x48" href="{{ asset_url('images/favicon.ico') }}">
   <link rel="icon" sizes="any" href="{{ asset_url('images/favicon.svg') }}" type="image/svg+xml">
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,7 +9,7 @@ from uuid import UUID, uuid4
 
 import html5lib
 import pytest
-from flask import Flask, url_for
+from flask import Flask, current_app, url_for
 from notifications_python_client.errors import HTTPError
 from notifications_utils.url_safe_token import generate_token
 
@@ -2899,7 +2899,7 @@ def os_environ():
 
 
 @pytest.fixture  # noqa (C901 too complex)
-def client_request(request, _logged_in_client, mocker, service_one):  # noqa (C901 too complex)
+def client_request(request, _logged_in_client, mocker, service_one, fake_nonce):  # noqa (C901 too complex)
     def block_method(object, method_name, preferred_method_name):
         def blocked_method(*args, **kwargs):
             raise AttributeError(
@@ -2934,6 +2934,7 @@ def client_request(request, _logged_in_client, mocker, service_one):  # noqa (C9
             _test_for_elements_without_class=True,
             _test_forms_have_an_action_set=True,
             _test_for_non_smart_quotes=True,
+            _test_for_script_csp_nonce=True,
             _optional_args="",
             **endpoint_kwargs,
         ):
@@ -2946,8 +2947,10 @@ def client_request(request, _logged_in_client, mocker, service_one):  # noqa (C9
                 _test_for_elements_without_class=_test_for_elements_without_class,
                 _test_forms_have_an_action_set=_test_forms_have_an_action_set,
                 _test_for_non_smart_quotes=_test_for_non_smart_quotes,
+                _test_for_script_csp_nonce=_test_for_script_csp_nonce,
             )
 
+        # ruff: noqa: C901
         @staticmethod
         def get_url(
             url,
@@ -2958,9 +2961,12 @@ def client_request(request, _logged_in_client, mocker, service_one):  # noqa (C9
             _test_for_elements_without_class=True,
             _test_forms_have_an_action_set=True,
             _test_for_non_smart_quotes=True,
+            _test_for_script_csp_nonce=True,
             **endpoint_kwargs,
         ):
             from flask.templating import _render
+
+            mocker.patch("secrets.token_urlsafe", return_value=fake_nonce)
 
             with mock.patch("flask.templating._render", wraps=_render) as mock_render:
                 resp = _logged_in_client.get(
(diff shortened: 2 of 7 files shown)


kr8n3r:
missed https://github.com/alphagov/notifications-admin/blob/96d1eb0ae2dff75a8260282fe062c63a420912bb/app/templates/components/webauthn-api-check.html#L2

app/templates/admin_template.html
```diff
@@ -133,6 +135,6 @@
   {% block extra_javascripts %}
   {% endblock %}
 
-  <script type="module" src="{{ asset_url('javascripts/all-esm.mjs') }}"></script>
-  <script type="text/javascript" src="{{ asset_url('javascripts/all.js') }}"></script>
+  <script nonce="{{ request.csp_nonce }}" type="module" src="{{ asset_url('javascripts/all-esm.mjs') }}"></script>
```
tombye, line 138: From my understanding of the use of the `nonce` attribute, we only need to apply it to scripts that are either from external domains or those that contain inline code. These tags reference scripts from our own domain and contain no inline code so shouldn't need it.

tombye (changes requested):
I left a comment but also wanted to suggest this adds a test to [this block in `tests/conftest.py`](https://github.com/alphagov/notifications-admin/blob/96d1eb0ae2dff75a8260282fe062c63a420912bb/tests/conftest.py#L2987) for script tags that either contain inline code or reference a file not hosted on our domain. That would both identify any current scripts without the `nonce` attribute and protect us from adding any in future.

tests/conftest.py
```diff
@@ -3149,6 +3153,16 @@ def test_for_non_smart_quotes(page):
...
+        @staticmethod
+        def test_for_script_csp_nonce(page):
+            for script_tag in page.select("script"):
+                src = script_tag.get("src")
+                nonce = script_tag.get("nonce")
+                if src and "javascripts/all-esm.mjs" in src:
+                    assert nonce is None
+                else:
+                    assert nonce == fake_nonce
+
```
tombye, line 3165: This is great but I think it just checks a specific js file we load and doesn't check inline scripts. I'm guessing at the code but maybe this method could do with a few changes to check script tags for external js files and inline scripts. I don't write much python so the code could be wrong but I was thinking something like:

```python
@staticmethod
def test_for_csp_nonce(page):
    for script_tag in page.select("script"):
        src = script_tag.get("src")
        nonce = script_tag.get("nonce")        
        if src and script_tag.text: # inline script tag
            assert nonce is not None
        if not src.startswith('https://static.example.com'): # script tag for file from external domain
            assert nonce is not None
```

tests/conftest.py
```diff
@@ -3007,6 +3009,8 @@ def get_url(
             if _test_for_non_smart_quotes:
                 ClientRequest.test_for_non_smart_quotes(page)
 
+            ClientRequest.test_for_script_csp_nonce(page)
```
tombye, line 3012: I think this needs a condition testing a boolean with the same name, like the others above it, and for that to be added as a parameter to the following method definitions, so the test can be turned on or off:
- https://github.com/alphagov/notifications-admin/blob/bbeb4fbfc1e6e75d62b1214df5d48192fb94a50c/tests/conftest.py#L2936
- https://github.com/alphagov/notifications-admin/blob/bbeb4fbfc1e6e75d62b1214df5d48192fb94a50c/tests/conftest.py#L2948
- https://github.com/alphagov/notifications-admin/blob/bbeb4fbfc1e6e75d62b1214df5d48192fb94a50c/tests/conftest.py#L2960

Sorry if I didn't mention that in my previous review, I'm less familiar with this code.

tombye (changes requested):
Sorry to be asking for changes again, I'm not sure my last review was clear enough, but I think these changes would mean we could spot and guard against all unsafe script tags.

tests/conftest.py
spatel033: Test does not expect `nonce` for `src="{{ asset_url('javascripts/all.js') }}` , for example
`<script src="https://static.example.com/javascripts/all-esm.mjs?42034a5684d7ffec20704fe4ba8b9543" type="module"></script>`

But it expect `nonce` for all other javascript files, for examples
`<script nonce="TESTs5Vr8v3jgRYLoQuVwA">document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' govuk-frontend-supported' : '');</script>`
`<script src="https://static.example.com/javascripts/all.js?56abf841397f87250b4de8efb35b0826" type="text/javascript"></script>`


The test covers both cases [with `src`](https://github.com/alphagov/notifications-admin/pull/5325/files#diff-96cbd07e702e8ccd28e9d25faef4d3bcff2a45127f89f30c71c71f8820bff060R139) and without `src` (as mentioned above)

kr8n3r: I removed 'nonce` from inline `<script>` tag on app/templates/components/webauthn-api-check.html and got a failed `test_should_show_security_keys_page` on

```
    @staticmethod
    def test_for_script_csp_nonce(page):
        for script_tag in page.select("script"):
            src = script_tag.get("src")
            nonce = script_tag.get("nonce")
            if src and "javascripts/all-esm.mjs" in src:
                assert nonce is None
            else:
>               assert nonce == fake_nonce
E               AssertionError: assert None == 'TESTs5Vr8v3jgRYLoQuVwA'

tests/conftest.py:3164: AssertionError
```

so I gather test is working

kr8n3r (approved):
based on https://github.com/alphagov/notifications-admin/pull/5325#discussion_r1886507980 I'm happy with this but it should wait for Tom as he requested changes

tests/conftest.py
tombye: I might be wrong but I thought this line:
```python
if src and "javascripts/all-esm.mjs" in src:
```
...would catch this script tag:
```html
<script type="module" src="{{ asset_url('javascripts/all-esm.mjs') }}"></script>
```
...but not this script tag:
```html
<script  type="text/javascript" src="{{ asset_url('javascripts/all.js') }}"></script>
```
That also seems to reflect the code in `app/templates/admin_template.html` because the script tag for 'javascripts/all.js' has a nonce but, I think, shouldn't need one because it is served from the same static assets domain and is a file, not an inline script.

spatel033: Thank you for your suggestions. Refactored `app/templates/admin_template.html` and test cases.

tombye (approved):
Thanks for going through all the changes requested. Looks great now 👍🏻

kr8n3r (approved):

PR #5327, opened 2024-12-16
feat: Crete form to remove inbound sms from service

joybytes:
Added two endpoints: `remove_service_inbound_sms` and `get_most_recent_inbound_number_usage_date`
- `get_most_recent_inbound_number_usage_date` responsible for displaying the last date in which the inbound number has been used
- `remove_service_inbound_sms` responsible for updating inbound_sms permission for service, removing service_sms_sender row and releasing or arching inbound_number
- success case: when the platform admin chooses archive/release and the request is successful, we redirect to the `receive-text-messages-stop-success` template
- error case: when we get errors from the remove endpoint, we render `Failed to remove number from service` message
- added delete cache for `service` and `template-*` to update inbound_number for service through the platform

Demo:

https://github.com/user-attachments/assets/fde6622f-8ac1-489f-abbe-c5f5a48cd738

For platform admins
Screenshot 2024-12-19 at 11 49 50

Screenshot 2024-12-19 at 11 51 52

Screenshot 2024-12-17 at 19 13 16

Screenshot 2024-12-17 at 18 52 57

Screenshot 2024-12-17 at 18 53 06

Unchanged for users
Screenshot 2024-12-16 at 09 50 55

### Ticket: 

https://trello.com/c/jC9B8l27/1056-build-platform-admin-endpoint-to-remove-inbound-sms-from-service

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2272,6 +2272,20 @@ class AdminServiceInboundNumberForm(StripWhitespaceForm):
     )
 
 
+class AdminServiceInboundNumberArchive(StripWhitespaceForm):
+    removal_options = GovukRadiosField(
+        "What do you want to do with the number?",
+        choices=[("true", "Archive"), ("false", "Release")],
+        validators=[DataRequired(message="Select an option")],
+        param_extensions={
+            "items": [
+                {"hint": {"text": "No other services can use this phone number"}},
+                {"hint": {"text": "Another service can use this phone number"}},
+            ]
+        },
+    )
+
+
 class CallbackForm(StripWhitespaceForm):
     url = GovukTextInputField(
         "URL",
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -36,6 +36,7 @@ from app.main.forms import (
     AdminPreviewBrandingForm,
     AdminServiceAddDataRetentionForm,
     AdminServiceEditDataRetentionForm,
+    AdminServiceInboundNumberArchive,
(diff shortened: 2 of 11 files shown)


kr8n3r:
on error you should also show the error-summary, otherwise frontend looks ok.
content should be reviewed separately

kr8n3r:
once you select an option, in my case release, i got taken back to this page rendered as below.
I think it needs a success banner we use,  and no empty quote block
Screenshot 2024-12-16 at 10 49 40

joybytes:
> once you select an option, in my case "release", i got taken back to this page rendered as below. I think it needs a success banner we use, no empty quote block and "stop receiving messages" link to longer applies.
> 
> basically, either
> 
> * totally different content once you've done this action,
> * back to settings page or
> * bespoke confirmation page, with a link back to settings for example

I like the idea of back to the settings page: success banner and redirect to settings page, thanks

> On a side note, as platform admins, there's already a page for inbound numbers Screenshot 2024-12-16 at 10 55 01 note (the 2nd number is still listed as active, as we've "released" is from the above journey, so I assume that's correct?)
hm that's interesting, I wonder if it's something related to a cache somewhere. Well spotted!


> if this is a platform admin only feature, is there a technical reason why management isn't part of that page? Perhaps as we look into providing this to end users down the line?

Yes, you guessed it right. Currently we are only exposing this to platform admins as it is in line with the "Reduce manual database queries" goal but the idea is that users will be able to do this themselves.

spatel033 (approved):

PR #5329, opened 2024-12-16
freeze-requirements fixups

risicle:
See https://github.com/alphagov/notifications-api/pull/4311


Makefile
```diff
@@ -67,8 +67,10 @@ fix-imports: ## Fix imports using ruff
 .PHONY: freeze-requirements
 freeze-requirements: ## create static requirements.txt
 	uv pip compile requirements.in -o requirements.txt
+	uv pip sync requirements.txt
```
leohemsted, line 70: can we change these to `pip install` or they'll destroy any additional things in people's development envs (for example, ipython, pdbpp, and other debugging tools that need to be in a venv rather than global)

Makefile
```diff
@@ -67,8 +67,10 @@ fix-imports: ## Fix imports using ruff
...
 	python -c "from notifications_utils.version_tools import copy_config; copy_config()"
 	uv pip compile requirements_for_test.in -o requirements_for_test.txt
+	uv pip sync requirements_for_test.txt
```
leohemsted, line 73: i also wonder if, given that this now doesn't just freeze but also installs the latest reqs, we should just roll this into bootstrap so whenever anyone runs `make bootstrap` it refreezes? 🤔

cos now `freeze-requirements` does more than just create the reqs

risicle: I never run `make bootstrap` because it does crazy unpredictable stuff. I have a good idea of what `make freeze-requirements` does.

It's not entirely deliberate that this installs things into your env, it just needs to to be able to access utils.

It _could_ do something intricate like install to a temp venv from where it does the utils call, but that would require further development, and here I'm just trying to make it not broken.

risicle: Also having `bootstrap` refreeze would cause CI runs to do a refreeze on every run, which would defeat the point.

Makefile
risicle: Also I don't actually think it would be a bad thing to occasionally make people start from a clean venv

leohemsted: hmmm that's fine, but this command shouldn't care about your current venv - it's just setting up a "lock"file right.

if `pip compile` produce consistent results in an empty env/full env/weird dev env, then what this line does shouldn't matter, should it?

this step is "ensure that the correct version of `requirements_for_test_common.in` is installed so that the next compile statement works"

we always pin the version of utils exactly and what subdeps it installs is neither here nor there for the running of this function

a "clean and pure" local venv isn't really an objective we as a dev team are currently trying to reach right now with the existing flaws and quirks of python tooling 🤔

risicle: The problem is less about the version of utils that you get after the first freeze, it's more about the state it leaves your dev env in after it's run (and you presumably continue working on with oblivious...)

The "ideal" thing to do here is probably as I say below, to do the freeze/fetch/freeze with a temporary venv, but some of the details of that make my head hurt.

leohemsted: fair enough - i think ill approve this for now but i might have a poke around over xmas to see if theres a nice clean way to get this to work off a separate freeze venv

leohemsted (approved):

Makefile
risicle: (oh that patch might not apply cleanly because it's for template-preview, just copy-pasta it...)

risicle:
Going to merge just for symmetry with other repos and open the above separate-venv thing as a new PR

PR #5330, opened 2024-12-17
Rewrite Autofocus javascript as an ES Module

kr8n3r:

Rewrite follows the same conventions we previously used for FocusBanner and Collapsible checkboxes, [ColourPreview](https://github.com/alphagov/notifications-admin/pull/5321) and [FileUpload](https://github.com/alphagov/notifications-admin/pull/5323)

In this PR we:

- remove old module JS file


app/assets/javascripts/esm/autofocus.mjs
```diff
@@ -0,0 +1,40 @@
...
+    // if the page loads with a scroll position, we can't assume the item to focus onload
+    // is still where users intend to start
+    if ((window.scrollY > 0) && !forceFocus) { return; }
+
+    // See if the component itself is something we want to send focus to
+    // Otherwise look inside the component to see if there are any elements
+    // we want to send focus to
+    let target = arrayOfTargetElements.includes(this.$component.tagName.toLowerCase())
+      ? this.$component
+      : this.filterByElements(arrayOfTargetElements);
```
tombye, line 30: Could this just use `querySelector` to search within the component using the part of CSS selectors that let you specify a number of different selector patterns?, like:
```js
this.$component.querySelector(arrayOfTargetElements.join(',')) // this.$component.querySelector('input, textarea, select')
```

tests/javascripts/autofocus.test.mjs
```diff
@@ -1,30 +1,33 @@
...
+    });
+
     // set up DOM
     document.body.innerHTML =
       `<div id="wrapper">
         <label class="form-label" for="search">
           ${labelText}
         </label>
-        <input autocomplete="off" class="form-control form-control-1-1" id="search" name="search" type="search" value="" data-notify-module="autofocus">
+        <input autocomplete="off" class="form-control form-control-1-1" id="search" name="search" type="search" value="" data-notify-module="autofocus" >
```
tombye, line 30:
```suggestion
        <input autocomplete="off" class="form-control form-control-1-1" id="search" name="search" type="search" value="" data-notify-module="autofocus">
```

tombye (changes requested):
This is great, thanks for doing 🚀 . Only a few changes to request and one it some kind of pedantry.

app/assets/javascripts/esm/autofocus.mjs
kr8n3r: oh, very clever!

tombye (approved):
Works great, still. Thanks for making the changes, now good to go 👍🏻

PR #5335, opened 2024-12-20
Make success banners alert screen reader users

tombye:
https://trello.com/c/uHOOPsbV/1073-screen-reader-users-arent-notified-when-success-banners-appear

This story was created on the assumption that giving success banners a role of 'alert' would stop the page title being announced. I'm not sure how I got to that assumption but it was wrong.

Testing with NVDA and JAWS showed the banner gets announced when the page loads with the page title following. This means screen reader users will know their action was successful **but also** that the browser has navigated to a new page.


kr8n3r (approved):
always learning

PR #5340, opened 2025-01-02
jobs: make aware of "finished all notifications created" job status

risicle:
This should make admin continue to work properly when https://github.com/alphagov/notifications-api/pull/4323 has been released.

There isn't a huge amount of stuff that's interested in the job status on admin, and most of that that is interested in whether a job is "finished" helpfully uses a model method to determine this.

I don't particularly like the convention for lowercase, space-separated job status values either but it's what we have.

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -20,6 +20,7 @@ from app import (
 from app.formatters import get_time_left, message_count_noun
 from app.main import json_updates, main
 from app.models.job import Job
+from app.notify_client.job_api_client import JobApiClient
 from app.s3_client.s3_csv_client import s3download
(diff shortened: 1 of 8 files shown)


quis (approved):
Looks good. 

Having dug into this a bit the job model is a bit of a mess when it comes to counting things. Wonder if we could clean up the difference between `Job.notification_count` and `Job.notifications_requested` at a later point. But maybe it isn’t causing performance problems like the `missing_rows` check is.

risicle:
I haven't yet tested this on a dev env so will hold off until I've done so.

risicle:
🟢 https://concourse.notify.tools/teams/dev-a/pipelines/deploy-notify/jobs/test/builds/182 🎉

PR #5341, opened 2025-01-02
Use nonce value in Content Security Policy style-src header

kr8n3r:
## What
Continuation of work done in https://github.com/alphagov/notifications-admin/pull/5315 to serve an even more secure site by additionally restricting valid sources for styles.

Restricting execution of majority of our inline styles to only those that contain a `nonce` value, which was implemented in https://github.com/alphagov/notifications-admin/pull/5325

There are two notable exceptions documented in https://github.com/alphagov/notifications-admin/pull/5341/commits/4d149d79d11e5c771b8d7f3fff8fbb3c74852891

## How to test
run locally
- pages to test
  - homepage
  - templates
  - unbranded page (like request unsubscribe)
  - change email branding preview and send an email template preview

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -358,7 +358,7 @@ def useful_headers_after_request(response):
             "font-src 'self' {asset_domain} data:;"
             "img-src 'self' {asset_domain}"
             " *.notifications.service.gov.uk {logo_domain} data:;"
-            "style-src 'self' {asset_domain} 'unsafe-inline';"
+            "style-src 'self' {asset_domain} 'nonce-{csp_nonce}';"
             "frame-ancestors 'self';"
             "frame-src 'self';".format(
                 asset_domain=current_app.config["ASSET_DOMAIN"],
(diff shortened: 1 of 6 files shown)


tombye (commented):
These changes fix the majority of inline styles in our code but the `default-src` directive doesn't work to catch the inlines styles in our email code, as per the comment in `app/__init__.py`. Here's an example of how the inline styles are blocked when viewing an email branding:

image

From my reading of [the `default-src` page on mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src), it acts as a fallback for any absent 'fetch directives' (like `script-src`, `img-src`, etc) rather than catching errors from those declared, which I think the comment suggests.

In other words, I think the approach, of setting a different CSP value for `style-src` for the iframe, is sound but we'll have to find another way, like perhaps setting a different CSP heading for responses to requests from the iframe.

kr8n3r:
hm, yeah I read that and was unsure why I had it working locally. guess I had 'unsafe-inline' there as well.
I'll update the ticket with findings and move it back.

PR #5344, opened 2025-01-03
feat: Redirect to main service dashoard when user is already part of service

joybytes:
## Summary:
- In Join Service Request, when a user is already part of a service, we let them go as far as making the request to join the service they are already part of and then show an error page
- Here, we change this behaviour to redirect the user to the main.service_dashboard instead

**Before:**

https://github.com/user-attachments/assets/e54d78bd-0064-4800-8001-cae415dba43c

**Now:**

https://github.com/user-attachments/assets/08c04dbb-c7b9-4f68-8852-c2a5eb53723f

## Ticket:
[Join a service - Redirect users to the service's dashboard if they are already a team member](https://trello.com/c/bsO3XoDp/1103-join-a-service-redirect-users-to-the-services-dashboard-if-they-are-already-a-team-member)

### DIFF
--- a/app/main/views/join_service.py
+++ b/app/main/views/join_service.py
@@ -32,6 +32,9 @@ def join_service_ask(service_to_join_id):
     if service.organisation != current_user.default_organisation:
         abort(403)
 
+    if current_user.belongs_to_service(service_to_join_id):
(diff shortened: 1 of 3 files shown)


quis (approved):
Nice work adding the videos, makes it really easy to review

PR #5345, opened 2025-01-04
Fix bug - if last page of CSV report empty, return instead of raising

CrystalPea:


joybytes (approved):

leohemsted:
seems fine, but _should_ we return 404 if there's no results? we're still returning `{"data": []}` or whatever right. it's a good request that just doesnt match anything in the database for the filters provided. i'm in two minds about whether we should handle here or change return value in the api

but not strong enough feelings to dictate either way

PR #5346, opened 2025-01-06
Fix copy an existing template radio button bug

BlessedDev:
This PR fixes the bug, described [here](https://trello.com/c/yMwKveap/179-copy-an-existing-template-radio-button-only-appears-if-more-than-one-channel-is-turned-on), where the "copy an existing  template" radio buttonis not displayed when a service has only enabled a single notification channel.

copy an existing template

### DIFF
--- a/app/assets/javascripts/templateFolderForm.js
+++ b/app/assets/javascripts/templateFolderForm.js
@@ -187,19 +187,12 @@
       return changed;
     };
 
-    this.$singleNotificationChannel = (document.querySelector('div[id=add_new_template_form]')).getAttribute("data-channel");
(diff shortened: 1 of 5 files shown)


quis:
Previous discussion and attempt to change this here: https://github.com/alphagov/notifications-admin/pull/3983#issuecomment-887570829

Now we don’t support Emergency Alerts it might be ok to make this change.

kr8n3r (approved):
changes look ok and work locally

BlessedDev:
> this looks like intended behaviour, perhaps original developers have more context.

@kr8n3r Yes it was originally intended to behave this way, however circumstances seem to have changed. Here is a slack discussion on this issue https://gds.slack.com/archives/C037E9EL8RE/p1726829762530589

PR #5350, opened 2025-01-14
Update docker compose to use amd64 unit-tests-image

spatel033:
After recent change tp remove sh file, the new unit-tests-image does not work on mac because of the ghcr image only support amd64.


leohemsted (approved):

leohemsted:
actually just want to ping @risicle / @robinjam to make sure we wont explode concourse/ec2 with these images

PR #5355, opened 2025-01-16
Rewrite homepage JS as an ES Module

kr8n3r:
## What

We want to migrate all of our legacy Javascript
files to ESM, so this is the latest in the series of conversions.

Rewrite follows the same conventions we previously used for FocusBanner, Collapsible checkboxes, ColourPreview, FileUpload and Autofocus.

In this PR we:

- remove old module JS file

## How to test

- run locally
- got to homepage
- click on the top graphic 5 times
- graphic should change


tombye (approved):
I can confirm the...functionality works as expected.

PR #5356, opened 2025-01-17
Remove the announcement about a February letter price increase

karlchillmaid:
New prices come into effect on 6 February, so we can remove the announcement about an upcoming increase.

Do not deploy until Thursday 6 February.


kr8n3r (approved):
good to g on the day

PR #5357, opened 2025-01-20
Rewrite PreviewPane JS as an ES Module

kr8n3r:
## What

Rewrite follows the same conventions we previously used for FocusBanner, Collapsible checkboxes, ColourPreview, FileUpload and Autofocus.

In this PR we:

- remove old module JS file

## How to test

 - run locally
 - check /services/serviceID/service-settings/set-email-branding
 - check /services/serviceID/service-settings/set-letter-branding


app/assets/javascripts/esm/preview-pane.mjs
```diff
@@ -0,0 +1,71 @@
...
+      return this;
+    }
+
+    this.branding_style = $module.value;
+    this.$paneWrapper = document.querySelector('.govuk-grid-column-full');
+    this.$form = document.querySelector('form');
+    this.previewType = this.$form.dataset.previewType;
+    this.letterBrandingPreviewRootPath = `templates/${this.previewType}-preview-image`;
+
+    this.$iframePreviewPane = document.createElement('iframe');
```
tombye, line 25: Checking this in-browser, it looks like setting `this.$iframePreviewPane` and `this.$imagePreviewPane` whatever `this.previewType` is means the element created that isn't used issues a HTTP request that 404s. We should therefore move these inside the conditional that checks (just as-is, or contained in functions, whatever you prefer) so only the element used is created.

As with other comments, this was already a problem with the existing code so an 'upgrade' rather than a fix ;)

tombye (changes requested):
These changes are great but I spotted a few issues that existed in the original code when reviewing so requesting changes. Details in the comments I left.

app/assets/javascripts/esm/preview-pane.mjs
```diff
@@ -0,0 +1,71 @@
...
+// More on ES2015 Classes at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes
+
+class PreviewPane {
+  constructor($module) {
+    if (!isSupported()) {
+      return this;
+    }
+
+    this.branding_style = $module.value;
+    this.$paneWrapper = document.querySelector('.govuk-grid-column-full');
```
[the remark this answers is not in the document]
kr8n3r: 👍 thanks, updated

app/assets/javascripts/esm/preview-pane.mjs
kr8n3r: shuffled the code around to have preview panes created Just-in-time so no extra http request are made

tombye (approved):
All looks good now, thanks for addressing the comments. Top work 👍🏻

PR #5373, opened 2025-02-11
Update rollup-styler plugin and sass versions and silence Dart Sass warnings

kr8n3r:
In v2.0.0 rollup-styler plugin [added the option to silence](https://github.com/plumelo/rollup-plugin-styler/issues/9#issuecomment-2646453249) Dart Sass deprecation warnings that are emitted from govuk-frontend.

This allows us to also upgrade the sass version, which brings in dependency updates, mainly chokidar v3.

These are emitted by govuk-frontend, which can now be silenced by passing in array of specific
deprecations. Documented at https://sass-lang.com/documentation/breaking-changes/import/#silencing-specific-deprecations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,8 +35,8 @@
         "postcss-replace": "2.0.1",
         "rollup": "4.24.0",
         "rollup-plugin-copy": "3.5.0",
-        "rollup-plugin-styler": "1.8.0",
-        "sass": "1.32.7",
+        "rollup-plugin-styler": "2.0.0",
+        "sass": "1.84.0",
         "stylelint": "16.10.0",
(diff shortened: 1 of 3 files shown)


tombye (approved):
The main changes here are really the bump to Sass. From a diff of the output that brings in:
- removal of comments for source maps (without affecting the devtools experience)
- fixing of indentation (doesn't affect us as our compression removes whitespace)
- hex values being converted to rgb (supported by our browsers so only an improvement)

Testing in our supported browsers across the main tasks shows no obvious issues so I think this only adds good stuff. Thanks for raising the related issue with rollup-styler and putting this together, a great improvement 👍🏻

PR #5375, opened 2025-02-11
Show custom error message when test teardown fails

quis:
As of 21ceeb9eac16503b04c81112d96f2985ade9a8a6 test fixtures need to be declared in a certain order.

This is so when pytest tears down the fixtures it does so in a way that avoids instances of the internal API clients losing access to the app context.

If this happens you get a very oblique error message which requires knowledge of the above to resolve.

This commit adds a more helpful error message which tells you how to fix or diagnose what is happening.


risicle (approved):
Haven't tested, but gets my vote - we can remove if it's annoying

PR #5376, opened 2025-02-11
User add returned letters callback

BlessedDev:
This PR adds the returned letter callback feature as described in this [card](https://trello.com/c/SYbHsAz5/212-let-users-add-a-callback-for-returned-letters).
This PR is dependent on https://github.com/alphagov/notifications-api/pull/4378.

A user now has the option to add a returned letters callback if letter notifications are enabled for the service
returned-letters 1
returned-letters 2
returned-letters 3

### DIFF
--- a/app/main/views/api_keys.py
+++ b/app/main/views/api_keys.py
@@ -222,11 +222,6 @@ def delivery_status_callback(service_id):
     )
 
 
-def get_received_text_messages_callback():
-    if current_service.inbound_api:
-        return service_api_client.get_service_inbound_api(current_service.id, current_service.inbound_api[0])
-
-
 @main.route(
     "/services/<uuid:service_id>/api/callbacks/received-text-messages-callback",
     methods=["GET", "POST"],
@@ -236,7 +231,7 @@ def received_text_messages_callback(service_id):
     if not current_service.has_permission("inbound_sms"):
         return redirect(url_for(".api_integration", service_id=service_id))
 
-    received_text_messages_callback = get_received_text_messages_callback()
+    received_text_messages_callback = current_service.inbound_sms_callback_details
     form = CallbackForm(
         url=(received_text_messages_callback.get("url") if received_text_messages_callback else ""),
         bearer_token=dummy_bearer_token if received_text_messages_callback else "",
--- a/app/main/views/api_keys.py
+++ b/app/main/views/api_keys.py
@@ -24,11 +24,8 @@ dummy_bearer_token = "bearer_token_set"
 @main.route("/services/<uuid:service_id>/api")
 @user_has_permissions("manage_api_keys")
 def api_integration(service_id):
-    callbacks_link = (
-        ".api_callbacks"
-        if current_service.has_permission("inbound_sms") or current_service.has_permission("letter")
-        else ".delivery_status_callback"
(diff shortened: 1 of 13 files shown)


app/templates/views/api/callbacks.html
```diff
@@ -31,6 +31,12 @@
           {{ optional_text_field(received_text_messages_callback, truncate=true) }}
 	   	  {{ edit_field('Change', url_for('main.received_text_messages_callback', service_id=current_service.id)) }}
 	    {% endcall %}
+
+		{% call row() %}
+			{{ text_field('Returned letters') }}
+			{{ optional_text_field(returned_letters_callback, truncate=true) }}
+			{{ edit_field('Change', url_for('main.returned_letters_callback', service_id=current_service.id)) }}
+		{% endcall %}
```
quis, line 39: We should make this visible to platform admin users only for now, then we can give it to all users once we’ve tested it end-to-end in production

app/main/views/api_keys.py
```diff
@@ -126,16 +136,25 @@ def revoke_api_key(service_id, key_id):
...
+            elif row["callback_type"] == "returned_letter":
+                returned_letters_callback_api = service_api_client.get_service_callback_api(
+                    current_service.id, row["callback_id"], row["callback_type"]
+                )
+
     if current_service.inbound_api:
         inbound_api = service_api_client.get_service_inbound_api(current_service.id, current_service.inbound_api[0])
 
-    return (callback_api, inbound_api)
+    return delivery_status_callback_api, inbound_api, returned_letters_callback_api
```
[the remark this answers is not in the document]
BlessedDev: That is a good point. I will add `inbound_sms_callback_details` to the service model and get rid of the function entirely

app/main/views/api_keys.py
```diff
@@ -148,54 +145,67 @@ def check_token_against_dummy_bearer(token):
...
-@main.route("/services/<uuid:service_id>/api/callbacks/delivery-status-callback", methods=["GET", "POST"])
+@main.route(
+    "/services/<uuid:service_id>/api/callbacks/delivery-status-callback",
+    methods=["GET", "POST"],
+)
 @user_has_permissions("manage_api_keys")
 def delivery_status_callback(service_id):
-    delivery_status_callback = get_delivery_status_callback_details()
+    delivery_status_callback_details = current_service.delivery_status_callback_details
     back_link = ".api_callbacks" if current_service.has_permission("inbound_sms") else ".api_integration"
```
quis, line 178: This isn’t checking for letters yet.

I’d suggest factoring this into something like this (then it can be reused in the multiple places we do this check):
```python
    def can_have_multiple_callbacks(service, user):
        if service.has_permission("inbound_sms"):
            return True
        if user.platform_admin and service.has_permission("letter"):
            return True
        return False
```

BlessedDev: Good catch, I missed that. I like the  idea of a `can_have_multiple_callbacks` function.

CrystalPea: Added here: https://github.com/alphagov/notifications-admin/pull/5376/commits/1df74a30c8c942a7e716d0b3ad112fb0e3ac5c95

CrystalPea (approved):

PR #5377, opened 2025-02-12
Rewrite ListEntry as ES Module

tombye:
We want to migrate all of our legacy Javascript files to ESM, so this is the latest in the series
of conversions.

Rewrite follows the same conventions we previously used for FocusBanner, Collapsible checkboxes,
ColourPreview, FileUpload, and others:

https://github.com/alphagov/notifications-admin/pulls?q=is%3Apr+is%3Amerged+%27as+an+ES+Module%27

In this PR we:

- remove old module JS file


kr8n3r (approved):
great work!

PR #5380, opened 2025-02-13
Apply z-index to letter template edit text button

kr8n3r:
Edit Welsh or English page text button can appear on any page other then first. In this case the Edit button is attached to the page number container.

As the image is position absolute, it is stacked over the the button, preventing the button to be clickable.

Applying a higher z-index to the button fixes the issue.


tombye (approved):
Make it clickable on mobile again so 👍🏻

PR #5381, opened 2025-02-17
feat: Platform Admin Users List

joybytes:
## Summary:
This PR Introduces a new platform admin reports page to filter and export a list of users based on multiple criteria, including creation date, last login date, permissions, and research opt-in status. The filtered results can be downloaded as a CSV report.

- Users can be filtered by: Creation date (created_from_date, created_to_date), Last login date (logged_from_date, logged_to_date), Permissions (selectable from a predefined list), Research opt-in status (Yes/No)
- Form validation ensures at least one filter is selected before submitting the request and enforces date validation to ensure the "From" date is not later than the "To" date.

- The final report includes: Name, Email, Created At, Research opt-in status, If User is an Organisation Team Member, Number of Live services associated to the user,  Live service permissions for each service name

Walk through:
List of Filters available 
Screenshot 2025-02-17 at 23 31 53

1) Creation Date and Last Login date Filters
Screenshot 2025-02-17 at 23 33 09

2) Permission filters
Screenshot 2025-02-17 at 23 33 35

3) Research opt in
Screenshot 2025-02-17 at 23 33 57

### Related Ticket:
[Add getting list  users and emails for yearly survey to platform admin](https://trello.com/c/qeSQJ5XB/1138-add-getting-list-users-and-emails-for-yearly-survey-to-platform-admin)

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2875,6 +2875,59 @@ class PlatformAdminSearchForm(StripWhitespaceForm):
     )
 
 
+class PlatformAdminUsersListForm(StripWhitespaceForm):
+    created_from_date = GovukDateField("Created (From)", thing="a start date", validators=[Optional()])
+    created_to_date = GovukDateField("Created (To)", thing="an end date", validators=[Optional()])
+    logged_from_date = GovukDateField("Logged In (From)", thing="a start date", validators=[Optional()])
+    logged_to_date = GovukDateField("Logged In (To)", thing="an end date", validators=[Optional()])
+
+    take_part_in_research = GovukRadiosField(
+        label="Research Opt In?",
+        choices=[("yes", "Yes"), ("no", "No")],
+        validators=[Optional()],
+    )
+
+    permissions_field = GovukCheckboxesField(
+        "Permissions",
+        filters=[partial(filter_by_permissions, permissions=permission_options)],
+        choices=list(permission_options),
+        param_extensions={"hint": {"text": "Select the permissions to filter by"}},
+    )
+
+    custom_field_order: tuple = (
+        "permissions_field",
+        "created_from_date",
+        "created_to_date",
+        "logged_from_date",
+        "logged_to_date",
+        "take_part_in_research",
+    )
+
+    def validate(self, extra_validators=None):
+        if not (
+            self.created_from_date.data
+            or self.created_to_date.data
+            or self.logged_from_date.data
+            or self.logged_to_date.data
+            or self.take_part_in_research.data
+            or self.permissions_field.data
+        ):
+            self.form_errors = ("You must provide at least one filter option",)
+            return False
+
+        if self.created_from_date.data and self.created_to_date.data:
+            if self.created_from_date.data > self.created_to_date.data:
+                self.created_to_date.errors = ("The 'Created (To)' date must be after 'Created (From)'",)
+                return False
+
+        if self.logged_from_date.data and self.logged_to_date.data:
+            if self.logged_from_date.data > self.logged_to_date.data:
+                self.logged_to_date.errors = ("The 'Logged In (To)' date must be after 'Logged In (From)'",)
+                return False
+
+        return super().validate(extra_validators=extra_validators)
+
+
 class UniqueServiceForm(StripWhitespaceForm):
     is_unique = GovukRadiosField(
         label="Is the service unique?",
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -25,6 +25,7 @@ from app.main.forms import (
     AdminReturnedLettersForm,
     BillingReportDateFilterForm,
     PlatformAdminSearchForm,
+    PlatformAdminUsersListForm,
     RequiredDateFilterForm,
 )
 from app.notify_client.platform_admin_api_client import admin_api_client
@@ -39,6 +40,7 @@ from app.utils.pagination import (
     get_page_from_request,
 )
 from app.utils.user import user_is_platform_admin
+from app.utils.user_permissions import all_ui_permissions, translate_permissions_from_db_to_ui
 
 COMPLAINT_THRESHOLD = 0.02
 FAILURE_THRESHOLD = 3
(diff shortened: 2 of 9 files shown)


app/main/views/platform_admin.py
```diff
@@ -752,3 +757,95 @@ def format_stats_by_service(services):
...
+    results = admin_api_client.fetch_users_list(
+        created_from_date=extract_date(form.created_from_date),
+        created_to_date=extract_date(form.created_to_date),
+        logged_from_date=extract_date(form.logged_from_date),
+        logged_to_date=extract_date(form.logged_to_date),
+        take_part_in_research=take_part_in_research,
+    ).get("data", [])
+
+    if not results:
+        flash("No results for filters selected")
```
tombye, line 786: Is it possible to use the error summary for this instead, for consistency, and putting this message in `form.form_errors` instead so it mimcs what the flash messaging is doing here?

app/main/forms.py
```diff
@@ -2875,6 +2875,59 @@ class PlatformAdminSearchForm(StripWhitespaceForm):
     )
 
 
+class PlatformAdminUsersListForm(StripWhitespaceForm):
+    created_from_date = GovukDateField("Created (From)", thing="a start date", validators=[Optional()])
+    created_to_date = GovukDateField("Created (To)", thing="an end date", validators=[Optional()])
+    logged_from_date = GovukDateField("Logged In (From)", thing="a start date", validators=[Optional()])
+    logged_to_date = GovukDateField("Logged In (To)", thing="an end date", validators=[Optional()])
+
+    take_part_on_research = GovukRadiosField(
```
tombye, line 2884: Should this be `take_part_in_research`, to match the API and the UI in settings?

app/main/forms.py
```diff
@@ -2875,6 +2875,59 @@ class PlatformAdminSearchForm(StripWhitespaceForm):
...
+    def validate(self, extra_validators=None):
+        if not (
+            self.created_from_date.data
+            or self.created_to_date.data
+            or self.logged_from_date.data
+            or self.logged_to_date.data
+            or self.take_part_on_research.data
+            or self.permissions_field.data
+        ):
+            self.take_part_on_research.errors = ("You must provide at least one filter option",)
```
tombye, line 2915:
```suggestion
            self.form_errors = ("You must provide at least one filter option",)
```
This reads as a form-level error so I'd suggest adding it like so, so that it doesn't have to link to a field in the HTML.

tombye (changes requested):
In this review I've looked at all the code changed but I have focused mainly on the frontend parts.

The Python I read all looks good and, I think, is really clear as it's not my main language but I easily figured out what it all does. What I maybe haven't as strong an opinion on is whether the tests cover all the main variants we care about so that might be worth getting another pair of 👀s on.

I requested a few changes but I think everything else looked good and ready to go. Really nice work 👏🏻.

app/templates/views/platform-admin/users-list.html
```diff
@@ -0,0 +1,107 @@
...
+    </h1>
+
+    <p class="govuk-body">
+        Filter and download the list of users.
+    </p>
+
+    {% call form_wrapper() %}
+        <fieldset class="govuk-fieldset">
+
+            <details class="govuk-details">
```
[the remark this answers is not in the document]
joybytes: Amazing, I didn't quite understand before the "open" functionality so that was great to learn. Thanks!

app/main/forms.py
joybytes: nice catch!

app/main/forms.py
joybytes: yes, nice catch!

app/main/views/platform_admin.py
tombye: Because we don't put the results of successful requests in the page, it's not obvious how to communicate a query returning no results (you'd normally put them where the results are). So I don't think there's an ideal way to do it without re-designing the page.

Thinking about it a bit more, specifically from my experience with support tickets that lead to these queries, I kind of think an empty result set is usually an indication of something wrong, either with the query or the original question. So I don't think it's odd for this to use the validation messaging and, if we're doing that I'd recommend `form.form_errors` for consistency with other validation.

This is all a bit subjective though, what do you think?

app/templates/views/platform-admin/users-list.html
```diff
@@ -0,0 +1,106 @@
...
+        </div>
+        <div class="govuk-grid-column-one-half">
+            {{ form.logged_to_date(param_extensions={"hint": {"text": "YYYY-MM-DD"}}) }}
+        </div>
+    </div>
+    {% endset %}
+    {{ govukDetails({
+                "summaryText": "Filter by last login date",
+                "html": login_date_content,
+                "open": (form.logged_from_date.value or form.logged_to_date.value) or (form.errors | convert_to_boolean)
```
tombye, line 53: Apologies for being pedantic but, as a user, I'd prefer only opening the field with errors when they appear. The `form.errors` dict will map the field name to its errors so we could do something like:
```html
(form.logged_from_date.value or form.logged_to_date.value) or ('logged_from_date' in form.errors) or ('logged_to_date' in form.errors)
```

app/main/views/platform_admin.py
joybytes: Hmm I don't think no_results should be part of form_errors to be honest. That's mainly because there is nothing wrong with the form.
No results is not an error, it just means we don't have values and therefore don't feel the need to create an empty CSV

app/templates/views/platform-admin/users-list.html
joybytes: Nice!
I added the changes and I am only opening the fields when there is a specific error for the filter section

Example
```
                "open": (form.take_part_in_research.value) or ('take_part_in_research' in form.errors)

```

In the case of "No filters selected" we now only show the error message at the top, but no fields will be opened which I think it's a cleaner way of doing it too.

Thanks @tombye

app/templates/views/platform-admin/users-list.html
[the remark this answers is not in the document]
joybytes: No worries, I replied in the comment you put the link for and will close this one.

app/main/views/platform_admin.py
tombye: OK, I have it on my list to replace most of the flash messaging that uses the red version with error summaries but this doesn't seem like it suits a straight swap so happy to leave as is for now.

PR #5382, opened 2025-02-18
Accept basic service callback info both in old and new format

CrystalPea:
So far we have been only getting callback id, but now we want to also be getting callback type, to fit in returned letter callbacks info.

So this commit helps admin accept the information in both formats, as a temporary measure.

Needs to be deployed before this one: https://github.com/alphagov/notifications-api/pull/4378


quis (approved):
Nice

PR #5384, opened 2025-02-19
Send file by email prototype

quis:
image


quis:
Closing this because we’ve wrapped up round 1 of research. Work continues in [round 2](https://github.com/alphagov/notifications-admin/pull/5428)

PR #5386, opened 2025-02-25
Fix overflowing email addresses in template preview

kr8n3r:
When previewing the email template befoire sending, any long email adresses in "Reply to" or "To" fields will overflow the container on mobile or high zoom level.

We already had a fix for the "reply-to" fields.
This adds the same fix to the "To" filed by adding a HTML class to the list item and includes it in the CSS style rule.

Fixes https://trello.com/c/lOf1ttlw/1193-long-email-addresses-for-notifications-dont-reflow

**Before**
Screenshot 2025-02-25 at 14 31 33

**After**
Screenshot 2025-02-25 at 14 32 35

## How to test

Try to send an email and go through the steps to preview the message just before sending. Switch to mobile view or increase text magnification.

### DIFF
--- a/app/assets/stylesheets/components/email-message.scss
+++ b/app/assets/stylesheets/components/email-message.scss
@@ -45,7 +45,8 @@ $email-message-gutter-mobile: govuk-spacing(3);
       }
 
       // deal with long email addresses
-      & > .email-message-meta__reply-to {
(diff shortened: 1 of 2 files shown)


tombye (approved):
Nice work 👍🏻

PR #5393, opened 2025-02-28
Add endpoint for a report request to be downloaded

klssmith:

This adds an endpoint, `/services/<uuid:service_id>/report-request/<uuid:report_request_id>.csv` which calls the API to get the details of a report, then downloads the file from S3 if the results from the API indicate the file can be downloaded.

https://trello.com/c/596NV4VP/1196-get-create-endpoint-to-get-the-report


app/main/views/report_requests.py
```diff
@@ -0,0 +1,27 @@
...
+from flask_login import current_user
+
+from app.constants import REPORT_REQUEST_STORED
+from app.main import main
+from app.models.report_request import ReportRequest
+from app.utils.user import user_is_platform_admin
+
+
+@main.route("/services/<uuid:service_id>/report-request/<uuid:report_request_id>.csv")
+@user_is_platform_admin
```
spatel033, line 11: When sms/email/letter report is requested by service user, then that user should be able to download. So is the permission should be `@user_has_permissions("view_activity")`?

spatel033: Decided to keep `user_is_platform_admin` for now, and we'll update the permission when rolling out this feature to all users.

spatel033 (approved):

PR #5395, opened 2025-03-03
Rename "Your profile" to "Your account" and "Switch service" to "Your services"

klssmith:
This changes the words on the page, and also does step one of renaming the URLs for "Your account" - adding the additional URLs. A later PR will add the redirects in.

Needs to be merged after https://github.com/alphagov/notifications-functional-tests/pull/561


app/main/views/user_profile.py
```diff
@@ -102,6 +106,7 @@ def _check_password(pwd):
 
 
 @main.route("/user-profile/email/confirm/<string:token>", methods=["GET"])
+@main.route("/your-account/email/confirm/<string:token>", methods=["GET"])
```
leohemsted, line 109: ooh we'll need to update emails in api too to start sending this new link

app/main/views/user_profile.py
```diff
@@ -222,6 +233,7 @@ def _check_password(pwd):
 
 
 @main.route("/user-profile/take-part-in-user-research", methods=["GET", "POST"])
+@main.route("/your-account/take-part-in-user-research", methods=["GET", "POST"])
```
leohemsted, line 236: we link to this from the new user survey email: https://www.notifications.service.gov.uk/services/d6aa2c68-a2d9-4437-ab19-3ae8eb202553/templates/55bcb671-4924-46c5-a00d-1a9d48458008

we'll need to update the template on prod. it's probably worth updating the existing template migration to have the new link for when we bring up new envs from scratch, but i'm not sure we need a migration to update existing templates - probably fine just click-opsing it on prod?

app/main/views/user_profile.py
```diff
@@ -34,6 +34,7 @@
 
 
 @main.route("/user-profile")
+@main.route("/your-account")
```
leohemsted, line 37: when we write `url_for('main.user_profile')` on this code, what does it return - the first or the second one in the list?

does it matter which one it does? i think it doesn't matter but not 100% sure on that

🤔

leohemsted (approved):
this itself looks good but we'll need to be careful about deploys due to a couple of API changes. 

i think this is still good to be merged as it only adds new routes and doesnt impact existing ones yet

app/main/views/user_profile.py
klssmith: It's the second one defined that gets returned, so the new "/your-account" URLs. I can't think of a way this would have a bad impact

app/main/views/user_profile.py
klssmith: Good spot! I had forgotten we had this URL there, so will make sure it gets updated

app/main/views/user_profile.py
klssmith: I think just click-opsing on prod is fine. I can change the migration file, though it may not be worth it given I [made a change](https://github.com/alphagov/notifications-api/pull/4377) to only send this email in prod recently - I was worried user research participants would accidentally get sent it from a dev environment if we forgot to delete their account 😅

app/main/views/user_profile.py
klssmith: Updated in https://github.com/alphagov/notifications-api/pull/4395, which can be merged after this change

app/main/views/user_profile.py
klssmith: I've changed this in the migration file since there was an API change to make anyway

PR #5396, opened 2025-03-04
Fix sending box bug

BlessedDev:
This PR is a bug fix for the message displayed in the sending box in in the dashboard for an uploaded contact list as described in this [card](https://trello.com/c/eJwPB79e/105-clicking-0-sending-box-after-uploading-a-file-can-show-confusing-content)

Before:
Screenshot 2025-03-04 at 11 29 31
After:
Screenshot 2025-03-04 at 12 24 08


app/models/job.py
```diff
@@ -205,6 +212,13 @@ def cancel(self):
     def original_file_name_without_extention(self):
         return Path(self.original_file_name).stem
 
+    @property
+    def service_data_retention_days(self):
+        return self._get_service_data_retention_days()
+
+    def _get_service_data_retention_days(self):
+        return service_api_client.get_service_data_retention(self.service)[0]["days_of_retention"]
+
```
quis, line 221:
```suggestion
    @cached_property
    def service_data_retention_days(self):
        return service_api_client.get_service_data_retention(self.service)[0]["days_of_retention"]

```

- you don’t need a separate private method here
- use `cached_property` to prevent hitting the API multiple times for the same data

tests/__init__.py
```diff
@@ -497,6 +499,8 @@ def job_json(
             created_by["email_address"],
         ),
         "processing_started": processing_started,
+        "finished_processing": finished_processing,
+        "service_data_retention_days": 7,
```
quis, line 503: I don’t see where the API is returned this data when serializing the job: https://github.com/alphagov/notifications-api/blob/a1330b51d54304c3fcd9bd470fc1b1461a0b8f31/app/schemas.py#L515-L557

BlessedDev: You are right, I was trying out something and forgot to take them out.

app/models/job.py
```diff
@@ -108,13 +108,20 @@ def finished_processing(self):
         return self.notification_count == self.notifications_sent
 
     @property
-    def awaiting_processing_or_recently_processed(self):
-        if not self.processing_started:
+    def awaiting_processing_or_recently_processed_message(self):
+        data_retention = self.service_data_retention_days
```
quis, line 112: There’s already a method for this on the service object which accounts for different retention periods per channel:
```suggestion
        data_retention = self.service.get_days_of_retention(self.template_type) 
```

BlessedDev: I was aware of this, I had tried to handle this within the `Job` model and complicated things unnecessarily .
I am now taking a different straightforward approach which does not need an adjustment to `awaiting_processing_or_recently_processed`

quis (approved):

PR #5399, opened 2025-03-05
Tweak inbound SMS conversation page for better accessibility

kr8n3r:
Addresses https://trello.com/c/TKEjtzNa/1195-conversation-messages-are-focusable

Tweaking the page by

- adding headings for each message for screenreader users that convey date, time, status and origin of the message
- remove keyboard focus for each message

Modelled on how mobile OS built-in readers announce messages in apps.

Details in each commit. They'll need to be squashed before merge.

**Before**
Screenshot 2025-03-05 at 14 46 03
Screenshot 2025-03-05 at 14 45 36

**After**
Screenshot 2025-03-05 at 13 22 55

Screenshot 2025-03-12 at 13 49 21


kr8n3r:
comment from Karl

> I wonder if the headings should be more like:
> Failed to send at 11:14am on 28 February
> Received at 11:15am on 28 February
> Delivered at 2:30pm yesterday
> Sending since 12:36pm today

app/templates/views/conversations/messages.html
```diff
@@ -1,8 +1,33 @@
...
+    {% set status_text_in_heading = 'sent' %}
+    {% set visual_status_text %}
+      <div class="sms-message-status sms-message-status-outbound" tabindex="-1">
+        {{ message.created_at | format_datetime_relative }}
+      </div>
+    {% endset %}
+  {% elif message.status in ['pending', 'sending', 'created'] %}
+    {% set status_text_in_heading = 'in sending' %}
+    {% set visual_status_text %}
+      <div class="sms-message-status sms-message-status-outbound hint" tabindex="-1">
```
tombye, line 14: Should the datetime bit of this be wrapped in the `<time>` tag? We use it in [a few places around the admin app](https://github.com/search?q=repo%3Aalphagov%2Fnotifications-admin%20%3Ctime&type=code) and this looks like a good place too.

app/templates/views/conversations/messages.html
```diff
@@ -1,8 +1,33 @@
...
+      <div class="sms-message-status sms-message-status-outbound table-field-error-label" tabindex="-1">
+        Failed (sent {{ message.created_at | format_datetime_relative }})
+      </div>
+    {% endset %}
+  {% endif %}
+  
+    <div class="govuk-grid-row sms-message-row" id="n{{ message.id }}">
       {% if message.inbound %}
         <div class="govuk-grid-column-two-thirds sms-message-inbound">
+          <h2 class="govuk-visually-hidden">{{ message.created_at | format_datetime_relative }} message received</h2>
```
tombye, line 30: Trying this out with NVDA, I preferred it with this order of information:

`[status] [datetime]`, so an example would be 'Sent yesterday at 12.03pm'

I think that's because I wanted the status, to tell the type of message (out/inbound) and if it failed and then when it happened. But I'm not sure what you and Karl discussed so leaving this as a suggestion.

tombye (changes requested):
This is already much better without the focusing and each message having a heading. I added a few comments. One (to do with content) is a bit subjective so just interested what you think whereas I think the other (adding `<time>`) I think is worth doing so marking this as 'request changes'.

app/templates/views/conversations/messages.html
kr8n3r: for incoming messages would you still expect to hear 'sent' when the message was 'received'? 
there were suggestions along this line in https://github.com/alphagov/notifications-admin/pull/5399#issuecomment-2703662283 or rough draft of 
```
Failed, 11:45am, 28 February
Sent, 2:30pm, 5 March
Delivered, 4am, 17 April
```
When message is schedule, pending or created we currently say 'sending'.

app/templates/views/conversations/messages.html
kr8n3r: yeah defo can use `<time>`

app/templates/views/conversations/messages.html
tombye: Correct me if I'm wrong but wouldn't all incoming messages have only one status, because we only see them if they're successful? If so, I'd assume it should always say 'received'.

kr8n3r: yes, they would only appear if 'delivered' so we'll always say 'received'.
> for incoming messages would you still expect to hear 'sent' when the message was 'received'?

responded with above as I read you comment 
> [status] [datetime], so an example would be 'Sent yesterday at 12.03pm'

related to inbound messages.

kr8n3r: @tombye rejigged around in https://github.com/alphagov/notifications-admin/pull/5399/commits/0691a437ac22a1653575b98874151ce54fdd1ada ,  keeping it simple

tombye (approved):
Thanks for addressing all the comments. This now looks good and works well with all the main screen readers 👍🏻

PR #5406, opened 2025-03-13
Rewrite RadiosWithImages as an ES Module

kr8n3r:

Rewrite follows the same conventions we previously used for FocusBanner, Collapsible checkboxes, ColourPreview, FileUpload, Autofocus, PreviewPane, CopyToClipboard and others.

In this PR we:

- remove old module JS file

## How to test

- go to service-settings/email-branding and change branding to other to be presented with radios with images


tombye (approved):
Nice work, sorry it took so long to review 👍🏻

PR #5416, opened 2025-03-20
Add announcement of 2025 text message price increase

karlchillmaid:
This PR adds an announcement to the text message pricing page.

It recycles the pattern the @quis used last year – but will need a sense check to make sure I’ve done it correctly!

### DIFF
--- a/app/templates/views/guidance/pricing/text-message-pricing.html
+++ b/app/templates/views/guidance/pricing/text-message-pricing.html
@@ -24,7 +24,13 @@ Text message pricing
       }
     ) }}
 
-  <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
+  {% if sms_rate != 2.33 %}
+   <div class="govuk-inset-text">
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -24,7 +24,13 @@ <h1 class="heading-large">Text message pricing</h1>
       }
     ) }}
 
-  <p class="govuk-body">Each unique service you add has an annual allowance of free text messages.</p>
+  {% if sms_rate != 2.33 %}
```
quis, line 27: Some magic that makes this work:

https://github.com/alphagov/notifications-admin/blob/bf3289fc48e5fc7e970c349d2bbbd84b40730ab7/app/models/sms_rate.py#L21

quis (approved):

quis:
I had to add a commit to update the tests so you’ll need someone else to approve this now

PR #5417, opened 2025-03-21
Add callback type to callback api calls

BlessedDev:
This PR adds `callback_type` to the CRUD service callback API calls. This will facilitate the work being done to consolidate those endpoints as filtering will be done on `callback_type`.
`callback_type` is an attribute in the `service_callback_api` this would help simplify the CRUD callback API endpoints code.
This PR is dependant on https://github.com/alphagov/notifications-api/pull/4405 and should only be merged into main after it has been deployed.


app/main/views/api_keys.py
```diff
@@ -174,18 +174,19 @@ def delivery_status_callback(service_id):
                     bearer_token=check_token_against_dummy_bearer(form.bearer_token.data),
                     user_id=current_user.id,
                     callback_api_id=delivery_status_callback_details.get("id"),
+                    callback_type="delivery_status",
                 )
         elif delivery_status_callback_details and not form.url.data:
             service_api_client.delete_service_callback_api(
-                service_id,
-                delivery_status_callback_details["id"],
+                service_id, delivery_status_callback_details["id"], "delivery_status"
```
quis, line 181: Shouldn’t need to pass in a type here because a callback with a given `id` will only ever have one type

app/models/service.py
```diff
@@ -651,7 +651,7 @@ def can_have_multiple_callbacks(self):
     @property
     def inbound_sms_callback_details(self):
         if self.inbound_api:
-            return service_api_client.get_service_inbound_api(self.id, self.inbound_api[0])
+            return service_api_client.get_service_inbound_api(self.id, self.inbound_api[0], "inbound_sms")
```
quis, line 654: Same here, shouldn’t need to pass in a type if we know that `self.inbound_api[0]` will never be another type of callback

app/main/views/api_keys.py
quis: Although I see why you’ve done this now because otherwise the API won’t know which table to look in

CrystalPea (commented):
Hi, I don't understand the point of big parts of this refactor - there is lots of places where we already have an id for the callback - why would we need the type as well?

Especially since we want to have all types of callbacks in a single table anyway? So finding them by id should be sufficient?

app/notify_client/service_api_client.py
```diff
@@ -348,23 +348,23 @@ def get_most_recent_inbound_number_usage_date(self, service_id):
         )
 
     @cache.delete("service-{service_id}")
-    def create_service_inbound_api(self, service_id, url, bearer_token, user_id):
-        data = {"url": url, "bearer_token": bearer_token, "updated_by_id": user_id}
+    def create_service_inbound_api(self, service_id, url, bearer_token, user_id, callback_type):
+        data = {"url": url, "bearer_token": bearer_token, "updated_by_id": user_id, "callback_type": callback_type}
         return self.post(f"/service/{service_id}/inbound-api", data)
```
quis, line 353: Can start using the new, generic URL as part of this PR:
```suggestion
        return self.post(f"/service/{service_id}/callback", data)
```

Same comment applies to all relevant client methods

CrystalPea: could we though? Since the merge order is:

1. https://github.com/alphagov/notifications-api/pull/4405 (just schema changes)
2. This PR - sending types to API
3. https://github.com/alphagov/notifications-api/pull/4402 - actually changing the endpoints in the API

quis: You’re right can’t start this yet

BlessedDev: It can be done either post table merge or alongside the update to the api to write to both callback tables .

CrystalPea: ok, when [the last commit](https://github.com/alphagov/notifications-admin/pull/5417/commits/85303b3894f6dfe201b0ff36a2b8a9be7d2d5926) is removed, I am happy to approve then

BlessedDev: > ok, when [the last commit](https://github.com/alphagov/notifications-admin/pull/5417/commits/85303b3894f6dfe201b0ff36a2b8a9be7d2d5926) is removed, I am happy to approve then

Done

CrystalPea (approved):

BlessedDev:
@CrystalPea I have removed `callback_type` as an argument to `service_api_client.get_service_inbound_api`. This is because the value is now hardcoded into the endpoint URL.

I have also tested this manually locally.

CrystalPea (approved):

PR #5420, opened 2025-03-28
Move the status update into the main content

tombye:
https://trello.com/c/wckYWWfa/1194-reply-to-success-banner-isnt-announced-to-screen-readers

Fix for successful adding of reply-to addresses not being announced to screen readers. This ended up being a lot more complicated than expected and resulted in me and @karlchillmaid deciding to move the update into the content of the page, removing the success banner in the process.

## Before

|Checking state|Complete state|Failed state|
|---|---|---|
|checking_page_current|success_page_current|failure_page_current|

## After

|Checking state|Complete state|Failed state|
|---|---|---|
|check-reply-to-address-checking-new|reply-to-email-address-check-complete-new|check-reply-to-address-failure-new|

## Notes for reviewers

This is very screen reader-centric but I've tested this in [all the supported screen readers](https://github.com/alphagov/notifications-manuals/wiki/Support-for-browsers,-email-clients-and-assistive-technologies#support-for-assistive-technologies) so reviewers can judge it on whether the logic of these changes make sense and the code is ok, if not comfortable testing with a screen reader.

Local testing can't receive a successful callback from the email we send to check addresses. Because of this, I suggest using this update in your local DB to simulate that change, setting the interval to whatever's best for you:

```sql
UPDATE notifications SET
 notification_status = 'delivered',
 sent_at = NOW(),
 updated_at = NOW()
 WHERE created_at > NOW() - interval '1 minute';
```

### DIFF
--- a/app/templates/views/service-settings/email-reply-to/_verify-updates.html
+++ b/app/templates/views/service-settings/email-reply-to/_verify-updates.html
@@ -1,4 +1,4 @@
-{% from "components/banner.html" import banner, banner_wrapper %}
+{% from "components/banner.html" import banner_wrapper %}
 {% from "components/page-header.html" import page_header %}
 {% from "components/page-footer.html" import page_footer %}
 {% from "components/form.html" import form_wrapper %}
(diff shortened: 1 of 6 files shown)


app/templates/views/service-settings/email-reply-to/_verify-updates.html
```diff
@@ -1,26 +1,35 @@
...
-    </p>
+    <div role="status"></div>
+    {{ page_header('Reply-to email address check') }}
+    <div role="status">
+      <p class="govuk-body">
+        We need to check that ‘{{ reply_to_email_address }}’ is a real email address.
+      </p>
+      <p class="govuk-body">
+        <span class='loading-indicator'>This can take up to a minute</span>
+      </p>
```
karlchillmaid, line 17:
```suggestion
      <p class="govuk-body">
        We need to check that ‘{{ reply_to_email_address }}’ is a real email address.
      </p>
      <p class="govuk-body">
        This can take up to a minute.
      </p>
            <p class="govuk-body">
        <span class='loading-indicator'>Please wait</span>
      </p>
```

tombye: @karlchillmaid removing the `<span class='loading-indicator'>` from line line 16 will stop it having that animated ellipsis so just checking, is that intended? Happy to make that change if so.

tombye: I rolled this into the changes and rebased, only because I couldn't think of another obvious way to fix the spacing on line 26. So marking as resolved.

PR #5427, opened 2025-04-08
Let platform admins remove each other’s permission

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-api/pull/4418

***

Sometimes we need to do this if someone has lost their Yubikey and needs to log back in to add a new one. Other times we need to do it because people have left the team.

It might also be useful if we ever needed to revoke someone’s platform admin access in a hurry (although we’d probably archive their whole account in that case).

Keeping some safeguards in here so that it isn’t possible to write code which _grants_ platform admin access through the UI (that should still be done through the DB).

image

image

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -21,6 +21,7 @@ EVENT_SCHEMAS = {
     "update_email_branding": {"email_branding_id", "updated_by_id", "old_email_branding"},
     "update_letter_branding": {"letter_branding_id", "updated_by_id", "old_letter_branding"},
     "set_inbound_sms_on": {"user_id", "service_id", "inbound_number_id"},
+    "remove_platform_admin": {"user_id", "removed_by_id"},
 }
 
 
@@ -70,6 +71,10 @@ def create_update_letter_branding_event(**kwargs):
 
 def create_set_inbound_sms_on_event(**kwargs):
     _send_event("set_inbound_sms_on", **kwargs)
+
+
+def create_remove_platform_admin_event(**kwargs):
+    _send_event("remove_platform_admin", **kwargs)
 
 
 def _send_event(event_type, **kwargs):
--- a/app/main/views/find_users.py
+++ b/app/main/views/find_users.py
@@ -19,6 +19,17 @@ def user_information(user_id):
     )
 
 
+@main.route("/users/<uuid:user_id>/remove-platform-admin", methods=["GET", "POST"])
+@user_is_platform_admin
+def remove_platform_admin(user_id):
+    if request.method == "POST":
+        User.from_id(user_id).update(platform_admin=False)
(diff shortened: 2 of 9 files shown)


app/main/views/find_users.py
```diff
@@ -19,6 +19,17 @@ def user_information(user_id):
     )
 
 
+@main.route("/users/<uuid:user_id>/remove-platform-admin", methods=["GET", "POST"])
+@user_is_platform_admin
+def remove_platform_admin(user_id):
+    if request.method == "POST":
+        User.from_id(user_id).update(platform_admin=False)
```
klssmith, line 26: We might want to also change their auth type from web_authn to email or sms auth since that's one of the common reasons for removing platform admin access

quis: Good point, changed in d9b3b11

tests/app/main/views/test_find_users.py
```diff
@@ -203,19 +203,30 @@ def test_remove_platform_admin_prompts_for_confirmation(
...
 
 
+@pytest.mark.parametrize(
+    "mobile_number, expected_auth_type",
+    (
+        ("12345", "sms_auth"),
+        (None, "email_auth"),
+    ),
+)
 def test_remove_platform_removes(
```
CrystalPea, line 214: I think there has been a mishap when naming this test? Maybe we could call it `test_remove_platform_admin_removes_user_admin_privilege_and_changes_user_auth`

app/event_handlers.py
```diff
@@ -72,6 +73,10 @@ def create_set_inbound_sms_on_event(**kwargs):
     _send_event("set_inbound_sms_on", **kwargs)
 
 
+def create_remove_platform_admin_event(**kwargs):
```
CrystalPea, line 76: [optional suggestion] `create_remove` sounds confusing, how about we rename this to `create_event_for_remove_platform_admin`?

Actually if all `create_event` functions started with `create_event_for`, they would be easier to search and group - but I appreciate that could be beyond the scope of this PR.

CrystalPea (commented):
Great work, this will save us some db trips 😁 .

A couple comments around naming 🙌🏼

tests/app/main/views/test_find_users.py
quis:
```suggestion
def test_remove_platform_admin_removes_user_admin_privilege_and_changes_user_auth(
```

app/event_handlers.py
quis: Yeah I think it’s better to just follow the pattern here which is `create_{name_of_event}_event`. There’s probably a better way to factor this than all these duplicated functions but that’s definitely beyond the scope of this PR

CrystalPea (approved):

PR #5430, opened 2025-04-11
Hide economy postage rates

spatel033:
We have added the new economy rates to the database in [this PR](https://github.com/alphagov/notifications-api/pull/4426). In some parts of the admin, the new rates were showing and could confuse users, so we added logic to filter out the economy postage options for now.

### DIFF
--- a/app/models/letter_rates.py
+++ b/app/models/letter_rates.py
@@ -35,7 +35,10 @@ class LetterRates(ModelList):
 
     @staticmethod
     def _get_items(*args, **kwargs):
-        return letter_rate_api_client.get_letter_rates(*args, **kwargs)
+        # Once the economy postage is live, we can remove this logic
+        resp = letter_rate_api_client.get_letter_rates(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


app/models/letter_rates.py
```diff
@@ -35,7 +35,10 @@ class LetterRates(ModelList):
 
     @staticmethod
     def _get_items(*args, **kwargs):
-        return letter_rate_api_client.get_letter_rates(*args, **kwargs)
+        # Once the economy postage is live, we can remove this logic
+        resp = letter_rate_api_client.get_letter_rates(*args, **kwargs)
+        filtered_letter_rates = filter(lambda r: r["post_class"] in LetterRates.post_classes.keys(), resp)
+        return list(filtered_letter_rates)
```
quis, line 41: - using a list comprehension is more Pythonic than a lambda
- `x in dict` is functionally the same as `x in dict.keys()` but a bit shorter
```suggestion
        return [rate for rate in resp if rate["post_class"] in LetterRates.post_classes]
```

joybytes (approved):

PR #5432, opened 2025-04-14
New flow for downloading CSV reports

kr8n3r:
## What

New flow for service users to request and download CSV reports.
Users request a CSV report. We create a request and direct them to a holding page while backend generates the report and makes it available. On the holding page we continuously check for the status change. 

Once the report request status is updated to stored, we take them to a download page where they can download it.
If there's an error we show an error page and if the report is no longer available, we show the 'unavailable' page.

Questions:

- do we need to worry about cloudfront caching of the `status.json` file?

---
## Review

- review by commit
- for local UI testing you need to be platform_admin

Uploading the report to s3 and updating the status is not ready yet, you'll need to manually update the status of the report in the db.  You're likely not going to have that table, so you might need to create it as well. Schema in the[ API PR](https://github.com/alphagov/notifications-api/blob/666e8a357b255f3f8009ed80754fe5c94c5a9306/app/service/report_request_schema.py)

```
update report_requests set status='STATUS where id='ID;
```
where STATUS can be 'stored' or 'failed' and ID is the requestID in the URL

## Pages

request-report

Preparing the report
Screenshot 2025-05-21 at 11 03 43

Report status updated notification
Screenshot 2025-05-21 at 11 07 46

Report ready to download
Screenshot 2025-05-21 at 11 05 17

Error generating the report
Screenshot 2025-05-21 at 11 04 13

Report no longer available
report-unavailable

## Screenreader announcements

Screenshot 2025-04-23 at 11 21 38

Screenshot 2025-04-23 at 11 22 09


app/main/views/dashboard.py
```diff
@@ -99,6 +101,30 @@ def redirect_to_main_view_notification(current_service, message_type, search_que
...
+            },
+        )
+        return redirect(
+            url_for(
+                "main.report_request",
+                service_id=current_service.id,
+                report_request_id=report_request_id,
+            )
+        )
+    except HTTPError as e:
```
joybytes, line 124: instead of an error raise here, should we be looking to have a template here instead?
I don't think we handle this error at the consumer at this function `view_notifications`

kr8n3r: we don't, so not sure what the template would be, to make it useful for the user. if the POST fails what can they do?

app/assets/javascripts/esm/check-report-status.mjs
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  runCheck() {
+    this.currentCheck = this.checkStatus();
+  }
+
+  async checkStatus() {
+    try {
+      const response = await fetch(this.reportStatusEndpoint);
+      if (!response) {
```
joybytes, line 36: should this be !response.ok to catch non-2xx responses ?

app/main/views/dashboard.py
joybytes: so what happens if this errors?
do we have a specific page that shows? I am just trying to understand what the user sees if this errors

app/main/views/report_requests.py
```diff
@@ -25,3 +103,10 @@ def report_request_download(service_id, report_request_id):
         download_name=f"{report_request_id}.csv",
         as_attachment=True,
     )
+
+
+@main.route("/services/<uuid:service_id>/download-report/<uuid:report_request_id>/status.json")
+@user_has_permissions()
+def report_request_status_json(service_id, report_request_id):
```
joybytes, line 110: can we add a comment to this function saying it's used on the polling

I think people might try to delete it at some point thinking there are no usages for it

app/main/views/report_requests.py
```diff
@@ -1,10 +1,88 @@
...
+            )
+
+        if report_request.status == REPORT_REQUEST_FAILED:
+            return render_template(
+                "views/csv-report/error.html",
+            )
+
+    return render_template(
+        "views/csv-report/preparing.html",
+        retention_period=current_service.get_days_of_retention("email"),
```
joybytes, line 47:
```suggestion
        retention_period=current_service.get_days_of_retention(report_type),
```
let's keep like the original implementation [here](https://github.com/alphagov/notifications-admin/blob/dcf716a72b14b92074bd495b1e3d34fb2d1fb518/app/main/views/dashboard.py#L265)

app/main/views/report_requests.py
```diff
@@ -1,10 +1,88 @@
...
+    except HTTPError as e:
+        if e.status_code == 404:
+            # if the report is no longer available, show them "No longer available page"
+            return redirect(
+                url_for(
+                    "main.report_request",
+                    service_id=service_id,
+                    report_request_id=report_request_id,
+                )
+            )
```
joybytes, line 66: I think we should include a generic error handling like we do [here](https://github.com/alphagov/notifications-admin/blob/c24a206baa5f669db2187368627ce0617b23d5c6/app/main/views/forgot_password.py#L10)

So the flow should be:

```
try:
            report_request = ReportRequest.from_id(service_id, report_request_id)
        except HTTPError as e:
            if e.status_code == 404:
                [do your logic....]
            else:
                raise e


        if report_request.user_id != current_user.id:
            abort(403)
            [rest of code...]
```


This error handling seems to be how do we with most our python code in the repo, I am not sure about the `try, except, else` pattern you have currently. I might be wrong though - so it might be worth leaving this open for someone with more years of Python experience

app/main/views/report_requests.py
kr8n3r: can make it `message_type`, not `report_type` as report_type is 'notifications_report`.
i think get_days_of_retention logic does fallback as well

app/main/views/dashboard.py
klssmith, line 124: This error handling isn't necessary because we're not doing anything with the error - it's catching it and then just immediately re-raising it. So we can simplify things by removing the `try` and `except...`

app/main/views/report_requests.py
klssmith: The way Joy suggested is good (and the `report_request` view function will need the same change). You do get a generic error page here without that change, but it's giving an error for the wrong reason - because this view function isn't returning anything, not because the API raised an error.

klssmith (commented):
A test that a non-platform admin user won't see the new way of generating reports would be good just to make sure we don't accidentally release this before it's ready

kr8n3r:
> A test that a non-platform admin user won't see the new way of generating reports would be good just to make sure we don't accidentally release this before it's ready

I've added a test for it now

klssmith (approved):
I'm not the best person to comment on the frontend bits so you might want to check with @tombye there, but the Python looks good and I managed to get the end-to-end flow working when trying this locally 🎉

I think we'll need some content changes, but that can be a separate bit of work since this won't be publicly accessible yet

tombye (changes requested):
The frontend code looks good to me, except for the link button having the wrong semantics for the visual design. Once that's fixed, I'd be happy to approve.

app/templates/views/csv-report/_footer-text.html
```diff
@@ -0,0 +1,8 @@
+<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+<h2 class="govuk-heading-m">Your report contains a lot of data</h2>
```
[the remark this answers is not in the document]
kr8n3r: no there isn't, this section is generic. one for @karlchillmaid

app/templates/views/notifications.html
```diff
@@ -70,20 +70,40 @@
...
-    </p>
+    {% set retention_text %}
+      Data available for {{ partials.service_data_retention_days }} days
+    {% endset %}
+    {% if current_user.platform_admin %}
+      <div class="govuk-button-group govuk-button-group--flex-start">
+        {% call form_wrapper(class='csv-request-form form--inline') %}
+          <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+          <input type="hidden" name="report_type" value="notifications_report">
+          {{ govukButton({
```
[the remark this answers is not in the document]
kr8n3r: having checked both with and without 'role=link', Jaws and NVDA accept the same keyboard interaction and activation and announce the element in very similar way: "Download this report"... "button" or "link" (depending if role is added). 
I think we can safely add "role=link"

nvda-button
nvda-link

app/templates/views/csv-report/error.html
```diff
@@ -0,0 +1,21 @@
...
+{% endblock %}
+
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-three-quarters">
+      {% call banner_wrapper(type='dangerous') %}
+      <h1 class='banner-title'>
+        {{ page_title}}
+      </h1>
+      <p class="govuk-body error-text-colour govuk-!-font-weight-bold">The file size is too big. Please <a class="govuk-link govuk-link--no-visited-state error-text-colour" href="{{ url_for('main.feedback', ticket_type='ask-question-give-feedback') }}">contact support</a>.</p>
```
karlchillmaid, line 16:
```suggestion
      <p class="govuk-body error-text-colour govuk-!-font-weight-bold">Please <a class="govuk-link govuk-link--no-visited-state error-text-colour" href="{{ url_for('main.feedback', ticket_type='ask-question-give-feedback') }}">contact support</a>.</p>
```

app/templates/views/csv-report/_footer-text.html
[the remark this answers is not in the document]
karlchillmaid: We designed this content on the basis that only users with large file sizes would go through the 2-step download journey. If every user is going to go through this flow, then the foot content will not be relevant for a majority of users.

We could either:

1. Update the content to try and make it more targeted. For example `If it’s taking a long time to prepare your report`. We’d have to give this some thought though, because ‘a long time’ is pretty subjective.
2. Remove this section altogether.

app/templates/views/csv-report/preparing.html
```diff
@@ -0,0 +1,27 @@
...
+  {{ page_title }}
+{% endblock %}
+
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-three-quarters">
+      {{ page_header(page_title, classes="loading-indicator") }}
+      <div role="status" data-notify-module="check-report-status">
+        <p class="govuk-body">We’re creating a CSV file of {{ report_text(report_request.parameter['notification_type'],report_request.parameter['notification_status']) }} from the last {{ retention_period }} days.
+        <p class="govuk-body">This can take up to 5 minutes.</p>
```
karlchillmaid, line 16: Are we still confident that 5 minutes is the maximum wait time?

kr8n3r:
> I think we'll need some content changes (e.g. for the error message - we probably won't know why the report generation failed), but that can be a separate bit of work since this won't be publicly accessible yet

error message updated

app/templates/views/csv-report/error.html
kr8n3r: updated

app/templates/views/csv-report/preparing.html
kr8n3r: can't tell without testing, but a large size report we had to create on support took 11 minutes

app/templates/views/notifications.html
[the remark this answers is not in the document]
kr8n3r: role="link" added to the button

app/templates/views/csv-report/_footer-text.html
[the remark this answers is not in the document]
kr8n3r: we've decided to not show the footer for now

tombye (approved):
I'm happy with the frontend code now and screen readers are picking up the download link as a link so 👍🏻

joybytes (approved):

PR #5434, opened 2025-04-16
Update the join a service API URLs

klssmith:
The URLs to create, get and update a service join request are changing in the api https://github.com/alphagov/notifications-api/pull/4434

This updates the code to work with the change


app/notify_client/service_api_client.py
```diff
@@ -24,6 +24,11 @@
 
 
 class ServiceAPIClient(NotifyAdminAPIClient):
+    def __init__(self, app):
+        super().__init__(app)
```
joybytes, line 28: why is this needed?

PR #5436, opened 2025-04-17
Refactor event handlers

quis:
Prompted by https://github.com/alphagov/notifications-admin/pull/5427#discussion_r2049100998

***

Organising things like this means:
- things are only named once
- definitions are not split between schemas and (duplicative and very thin) functions
- the amount of code in this file is reduced by almost 50%
- no more horrible function names like `create_remove_user_from_service_event`

Makes use of [metaclasses](https://docs.python.org/3/reference/datamodel.html#metaclasses), which is fun.


CrystalPea (approved):
What a joy to behold 😻

:shipit:

PR #5438, opened 2025-04-22
Update button label

karlchillmaid:
In onboarding usability testing we saw some confusion about new vs existing and trial vs live.

As a result we want to be more careful about keeping these concepts separate until users have had a chance to learn about them (in practice).

Therefore we want to rename the ‘join a service’ button to say ‘existing’ rather than ‘live’.

### DIFF
--- a/app/templates/views/your-services.html
+++ b/app/templates/views/your-services.html
@@ -141,7 +141,7 @@
         {% if current_user.default_organisation.can_ask_to_join_a_service %}
           {{ govukButton({
             "element": "a",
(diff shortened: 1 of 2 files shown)


klssmith:
This will break the functional tests, so we'll need to change those first to work with either the old or new text before this is merged

klssmith (approved):
Looks good. Change also made here to ensure this can be deployed without issues https://github.com/alphagov/notifications-functional-tests/pull/569

PR #5450, opened 2025-05-01
Validate international text message limit when sending through the website

quis:
One-off:

image

***

Spreadsheet upload:

image

***

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/1221

### DIFF
--- a/app/models/service.py
+++ b/app/models/service.py
@@ -618,7 +618,11 @@ class Service(JSONModel):
         return getattr(self, f"{notification_type}_message_limit")
 
     def remaining_messages(self, notification_type):
-        return self.get_message_limit(notification_type) - service_api_client.get_notification_count(self.id, notification_type=notification_type)
+        if notification_type == "international_sms" and not self.has_permission("international_sms"):
+            return 0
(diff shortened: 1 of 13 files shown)


app/models/service.py
```diff
@@ -617,6 +617,13 @@ def can_use_govuk_branding(self):
     def get_message_limit(self, notification_type):
         return getattr(self, f"{notification_type}_message_limit")
 
+    def remaining_messages(self, notification_type):
+        if notification_type == "international_sms" and not self.has_permission("international_sms"):
+            return 0
```
CrystalPea, line 622: suggestion: should we set this to None if the service has international sms permission turned off? Then we would have less checks to make for services that don't use international sms.

CrystalPea (commented):
Looks good - one small comment around performance.

app/models/service.py
quis: If we return `None` then we also have to check the type. If the method always returns an integer it’s simpler. Either way this still avoids a Redis call which is where the performance concern would be

CrystalPea (approved):

PR #5459, opened 2025-05-07
Make test for add_service more useful

CrystalPea:
Recently, a deploy broke on adding a service, because we forgot to add the new argument we are setting to create_service function on ServiceApiClient.

This was not caught by our unit tests, because we mock too early
- we mock the whole client, instead of just mocking the post request it makes.

I changed the test to do mocking later, at the point when we call the API. This way a similar issue will be caught in the future.


quis (commented):
Not convinced this is needed – it’s covered by the test you added here:
https://github.com/alphagov/notifications-admin/commit/07a642c04ed0e2ecf727262dd102d88ba5fe3948#diff-ab9e41205f6010242d62189e6a75cb4ef1d20a1d6bfdf66e5737ea1a1b424a4cR89-L89

PR #5465, opened 2025-05-12
Drop inbound api

BlessedDev:
This PR removes all references to `service_inbound_api` from the admin app.
It also includes refactoring of the callback endpoints to reduce duplication.
This PR is linked to https://github.com/alphagov/notifications-api/pull/4467


CrystalPea (approved):
Looks good to me, I love the refactor you did to de-duplicate the code in the callback views.

I have also run this code on my local machine, both on its own, and then with the API PR, and it works fine.

PR #5466, opened 2025-05-13
Change bulk sending page layout and add additional content for international text messages

saimaghafoor:
Changes made:
- page layout amended to include specific section for daily limits and international text message limits
- added additional information in a details component on AIT

(working on layout of daily limits table)


rparke:
Screenshot 2025-05-13 at 16 29 39
Screenshot 2025-05-13 at 16 29 44

app/templates/views/guidance/using-notify/daily-limits
```diff
@@ -0,0 +1,65 @@
...
+{% endblock %}
+
+{% block content_column_content %}
+
+<h1 class="heading-medium">Daily limits</h1>
+<div class="bottom-gutter-3-2">
+<p class=""gov-uk body>There’s a limit to the number of messages you can send each day. </p>
+  {% call mapping_table(
+    caption='Notification Daily Limits',
+    field_headings=['Message', 'Daily Limit'],
```
kr8n3r, line 20: not sure limit need to be capitalised

saimaghafoor: It definitely should not be capitalised, good spot! Thanks Jani

saimaghafoor:
@rparke please could I have an updated preview of the new daily limits page and the redrafted bulk sending page if possible?

app/templates/views/service-settings/set-international-sms.html
```diff
@@ -23,6 +23,7 @@
         See <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_pricing_text_messages', _anchor='international-numbers') }}">pricing</a> for the list
         of rates.
       </p>
+      <p class="govuk-body">The daily limit for your service is [enter limit]. You can change the daily international text message limit in <b class="govuk-!-font-weight-bold">Settings</b>.</p>
```
quis, line 26: Needs filling in

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -0,0 +1,64 @@
...
+        {{ text_field(limit) }}
+      {% endcall %}
+    {% endfor %}
+  {% endcall %}
+  </div>
+
+<p class="govuk-body">If you need to discuss these limits, <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for("main.support") }}">contact us</a>.</p>
+
+<h3 class="heading-medium">Change the international text message limit</h3>
+<p class="govuk-body">To change the international daily text message limit:</p>
```
quis, line 40: In some places we are saying _international daily_ and in others we are saying _daily international_. I think the latter is better but we should be consistent

karlchillmaid:
```suggestion
<p class="govuk-body">To change the number of international text messages you can send each day:</p>
```

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -0,0 +1,64 @@
...
+
+ <ol class="govuk-list govuk-list--number">
+   <li>Go to <b class="govuk-!-font-weight-bold">Settings</b></li>
+   <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">International text message limit</b> row</li>
+ </ol>
+
+<p class="govuk-body">When changing the limit, you should think about:</p>
+
+ <ul class="govuk-list govuk-list--bullet">
+ <li>how many international messages your service needs to send each day</li>
```
karlchillmaid, line 50:
```suggestion
 <li>how many international text messages your service needs to send each day</li>
```

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -0,0 +1,64 @@
...
+ <li>any costs to your service if there is fraud, such as Artificial Inflation of Traffic (AIT)</li>
+ </ul>
+<p class="govuk-body">If you do not need to send international text messages, you can turn this off in <b class="govuk-!-font-weight-bold">Settings</b>.</p>
+
+
```
karlchillmaid, line 55:
```suggestion
```

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -0,0 +1,64 @@
...
+<h3 class="heading-medium">Help protect your service from AIT</h3>
+   <p class="govuk-body">You should take steps to protect your service from an AIT attack, such as:</p>
```
karlchillmaid, line 57: I know we’ve explained the AIT initialism above, but if users skip straight to this section it might be necessary to explain it here, too.

To avoid explaining it twice:
* what if we only said ‘fraud’ in the previous section – then explain AIT fully here, including mentioning that it is a type of fraud?
* can we do something like the [acronym markdown](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/markdown#acronyms) on pages where we repeatedly use acronyms and initialisms?

app/templates/views/service-settings/set-international-sms.html
saimaghafoor: @CrystalPea could you help with filling the limits here for services?

CrystalPea: Ahh, of course 🙌🏼

CrystalPea: This is done now ☺️

rparke (approved):

PR #5468, opened 2025-05-14
Update 'request to go live' page design and content

kr8n3r:
## What
https://trello.com/c/MpVJikxL/1295-make-your-service-live-1-6-update-current-request-to-go-live-checklist-page

Update existing page with:

- update heading and page title
- add intro paragraphs
- alwayas show the form button, but it should be disabled until all tasks are completed
- add heading to the task list
- replace existing task-list with a Design system component

This is the first change in the series to address the user requests to go live based off the user research.

Firstly, we're replacing our custom task-list component with the Design System one. That comes with a change in the way a list of items and their options are specified. We've already done the same approach when updating the service-settings page where we append conditional tasks to the array of items list.

Secondly, we're now always showing the submit button, but the button is disabled until all
tasks are completed. This change is about setting expectation that once tasks are completed, users still need to submit a requests - it's not automatic.

Lastly, we've added content before the task list to set expectations.

Logic to conditionally show tasks or display text if a request has already been made or a user is not a gov user have not changed.

## How to review
best reviewed by commit

## Visual changes

**Before**

Screenshot 2025-05-15 at 08 23 58
Screenshot 2025-05-15 at 08 24 15

**After**
partial-completed
Screenshot 2025-05-15 at 18 39 40


klssmith (commented):
Backend code looks good and works in the same way as the current list.

Probably worth getting @tombye to check the frontend bits, and double checking the content is as expected for this stage - I wasn't too sure if the "Set up your organisation" heading should be there and noticed some of the task list content is different from in the doc, but that might be set to be changed later.

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -2,80 +2,112 @@
...
+            "title": {
+              "text": "Accept our data processing and financial agreement"
+            },
+            "href": url_for('main.service_agreement', service_id=current_service.id),
+            "status": completed_status if current_service.organisation.agreement_signed else not_completed_status,
+          }
+        )%}
+      {% endif %}
+      <h2 class="govuk-heading-m">Set up your organisation</h2>
+      {{ govukTaskList({
```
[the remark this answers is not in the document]
kr8n3r: nice spot, added those

klssmith:
Just a small thing I hadn't thought of at first is that with the button on the page but disabled, it is possible for someone with a tiny bit of tech knowledge to still request to go live when they shouldn't be able to - they can just remove the "disabled" attribute in dev tools and click it anyway. So we could consider an extra check on the POST endpoint to check that they should actually be able to submit the request - I don't think we'd need to show a nice error message because you can't really do this by accident, so could just use something like `abort(403)`

kr8n3r:
> Just a small thing I hadn't thought of at first is that with the button on the page but disabled, it is possible for someone with a tiny bit of tech knowledge to still request to go live when they shouldn't be able to - they can just remove the "disabled" attribute in dev tools and click it anyway. So we could consider an extra check on the POST endpoint to check that they should actually be able to submit the request - I don't think we'd need to show a nice error message because you can't really do this by accident, so could just use something like `abort(403)`

good shout. added https://github.com/alphagov/notifications-admin/pull/5468/commits/2788e188ee5eb296f07e45f9bcc21b14cfae932e but happy for people with expertise to change it

tombye (commented):
The frontend code looks ok to me. The 2 things that did seem a bit odd were:
- the team members and templates pages not having back links when you click through to them
- the text: Select ‘Send a request to go live’ because I normally think of a selection being from mutiple options but this is just clicking a button

Neither of them are blockers but I'd be interested to know if the first one came up before (and so did we decide not to add backlinks there?) and what @karlchillmaid thinks about the second one.

kr8n3r:
> The frontend code looks ok to me. The 2 things that did seem a bit odd were:
> 
> * the team members and templates pages not having back links when you click through to them
> * the text: Select ‘Send a request to go live’ because I normally think of a selection being from mutiple options but this is just clicking a button
> 
> Neither of them are blockers but I'd be interested to know if the first one came up before (and so did we decide not to add backlinks there?) and what @karlchillmaid thinks about the second one.

I'll leave it to Karl comment on the content,  as for the backlinks on when you're taken to those two pages: as we haven't changed any behaviour as part of this. Can't remember if it came up in UR (through it might have been masked by the RTGL link being in the side nav)

kr8n3r:
we've decided in standup to not do any work to back links on template pages as part of this

app/main/views/service_settings/index.py
```diff
@@ -226,8 +226,10 @@ def request_to_go_live(service_id):
 @user_has_permissions("manage_service")
 @user_is_gov_user
 def submit_request_to_go_live(service_id):
-    ticket_message = render_template("support-tickets/go-live-request.txt") + "\n"
+    if not current_service.go_live_checklist_completed:
```
klssmith, line 229: Just a little suggestion so that this matches the exact logic in the HTML

```suggestion
    if (not current_service.go_live_checklist_completed) or (current_service.able_to_accept_agreement and not current_service.organisation.agreement_signed):
```

kr8n3r: ok, need to mock these two conditions out as well now

kr8n3r: @klssmith services in self-approving orgs won't have the agreement signed (seeing that in failed tests)

klssmith: The requirements for going live are the same for "normal" vs self-approving orgs, so they should still need the agreement signed, it looks like the tests weren't bothering to mock the agreement being signed since we weren't checking it in the endpoint before now.

For the `test_request_to_go_live_is_sent_to_organiation_if_can_be_approved_by_organisation` test, it passes if you add this mock that we use elsewhere:
```py
mocker.patch(
    "app.models.organisation.Organisation.agreement_signed",
    new_callable=PropertyMock,
    return_value=True,
    create=True,
)
```

The `test_request_to_go_live_displays_go_live_notes_in_zendesk_ticket` test should also be changed to have the agreement signed. These changes fix it

```py
# In the second mock
             "Org 1",
+            agreement_signed=True,
             request_to_go_live_notes="This service is not allowed to go live",
```

```py
         "Organisation type: Central government\n"
-        "Agreement signed: No (organisation is Org 1, a crown body). {expected_go_live_notes}\n"
+        "Agreement signed: Yes, for Org 1. {expected_go_live_notes}\n"
         "\n"
```

kr8n3r: thanksfor this. yeah wasn't mocking before, but had failed tests locally when trying to mock.

tombye (commented):
Just a comment to say both the things I raised before were addressed in a discussion we had today so I'm happy with the frontend bits 👍🏻

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -2,80 +2,117 @@
...
+          "title": {
+            "text": "Tell us how many messages you expect to send",
+            "classes": "govuk-link--no-visited-state",
+          },
+          "href": url_for('main.estimate_usage', service_id=current_service.id),
+          "status": completed_status if current_service.has_estimated_usage else not_completed_status,
+        },
+        {
+          "title": {
+            "text": "Add a team member who can manage settings, team and usage",
```
karlchillmaid, line 42: Probably making this content change now, as we observed some confusion over the current wording.

```suggestion
            "text": "Give another team member the ‘manage settings’ permission",
```

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -2,80 +2,117 @@
...
+            "classes": "govuk-link--no-visited-state",
+          },
+          "href": url_for('main.manage_users', service_id=current_service.id),
+          "status": completed_status if current_service.has_team_members_with_manage_service_permission else not_completed_status,
+        },
+        {
+          "title": {
+            "text": "Add templates with examples of the content you plan to send",
```
karlchillmaid, line 50:
```suggestion
            "text": "Add templates with examples of your content",
```

klssmith (approved):

PR #5470, opened 2025-05-16
refactor: Set estimated letter delivery to latest

joybytes:
## Summary
- We have changed the backend to provide the latest delivery date for letters, and this PR is changing the UI to align with the expectation of this date
- Before: Estimated delivery date: [date]
- Now: Estimated delivery by [date]
Screenshot 2025-05-16 at 15 57 01

### Ticket:
- [Update delivery estimates code to accomodate economy postage](https://trello.com/c/JngKcHSO/1272-update-delivery-estimates-code-to-accomodate-economy-postage)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# This file was automatically copied from notifications-utils@99.3.3
+# This file was automatically copied from notifications-utils@99.5.0
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/app/models/notification.py
+++ b/app/models/notification.py
@@ -125,7 +125,7 @@ class Notification(JSONModel):
     @property
     def estimated_letter_delivery_date(self):
         if self.notification_type == "letter":
-            return get_letter_timings(self.created_at.replace(tzinfo=None), postage=self.postage).earliest_delivery
+            return get_letter_timings(self.created_at.replace(tzinfo=None), postage=self.postage).latest_delivery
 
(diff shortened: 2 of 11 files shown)


klssmith:
The estimated delivery date that we use when viewing a single letter isn't coming from the API, it's coming from [this function](https://github.com/alphagov/notifications-admin/blob/fcf7f2ca189a317e5ed43bc3b25fe35fd654e7d5/app/models/notification.py#L128), which will need to be updated.

I've also just noticed another place where we show an estimated delivery date that I hadn't spotted earlier so wasn't discussed in the kick-off thread - when [viewing a job](https://github.com/alphagov/notifications-admin/blob/fcf7f2ca189a317e5ed43bc3b25fe35fd654e7d5/app/templates/partials/jobs/count-letters.html#L16)

joybytes:
> The estimated delivery date that we use when viewing a single letter isn't coming from the API, it's coming from [this function](https://github.com/alphagov/notifications-admin/blob/fcf7f2ca189a317e5ed43bc3b25fe35fd654e7d5/app/models/notification.py#L128), which will need to be updated.
> 
> I've also just noticed another place where we show an estimated delivery date that I hadn't spotted earlier so wasn't discussed in the kick-off thread - when [viewing a job](https://github.com/alphagov/notifications-admin/blob/fcf7f2ca189a317e5ed43bc3b25fe35fd654e7d5/app/templates/partials/jobs/count-letters.html#L16)

@klssmith thanks - changes have been implemented!

PR #5473, opened 2025-05-21
Update guidance for users adding and approving a new service

karlchillmaid:
# What

This PR updates the content on the following pages:

* ‘Enter a service name’
* ‘Make this service live’
* ‘Confirm this service is unique’
* ‘Check the service name’

These changes affect:

* New users creating a new service, after creating a user account
* Existing users creating a new service
* Organisation-level team members with approval permissions, who need to validate a go-live request
* Members of the GOV.UK Notify team on non-tech support, who need to validate a go-live request

# Why

Content on the ‘Enter a service name’ page does not match our style guide.

We need to update the approver content ahead of any changes to the ‘Send a request to go live’ flow.

### DIFF
--- a/app/main/views/make_service_live.py
+++ b/app/main/views/make_service_live.py
@@ -86,7 +86,7 @@ def org_member_make_service_live_service_name(service_id):
         truthy="Yes",
         falsey="No",
         name=f"Will recipients understand the name ‘{current_service.name}’?",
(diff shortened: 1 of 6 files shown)


kr8n3r:
[this test assertion](https://github.com/alphagov/notifications-admin/blob/main/tests/app/main/views/test_make_service_live.py#L248) needs to be updated to match the new error message

kr8n3r (approved):

PR #5476, opened 2025-05-22
Update URL for request to go live page

klssmith:
This changes the request to go live URLs from `/services/<uuid:service_id>/service-settings/request-to-go-live` to
`/services/<uuid:service_id>/make-your-service-live` and adds in redirects for the old routes. We have decided not to change the URLs of the pages in the task list for now, but this may be done later.

See individual commits for details - the first 3 commits deal with renaming or moving code to reflect the fact that the go-live page is no longer nested under service settings, while only the last commit makes a functional change.

https://trello.com/c/dH9XzQZc/1300-make-your-service-live-3-6-update-urls-for-the-task-list-form


kr8n3r (approved):
page still works as expected and redirects from the old url

PR #5487, opened 2025-05-29
feat: Confirm Service Unique - Make your service live

joybytes:
### Summary
- Added a new "Confirm that your service is unique" task to the "Request to go live" checklist

- Created a new view and template for confirming the service name is unique

- When a user changes the service name after confirming uniqueness, the checklist item is reset to "Not completed"

### Ticket
[‘Make your service live’ (5/6) – Add the ‘Confirm your service is unique’ task](https://trello.com/c/vG9OowgU/1302-make-your-service-live-5-6-add-the-confirm-your-service-is-unique-task)


tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -1352,73 +1374,6 @@ def test_should_show_persist_estimated_volumes(
     )
 
 
-def test_should_error_if_bad_estimations_given(
```
kr8n3r, line 1355: reason for removing tests related to volumes page?

joybytes: oh wow thanks, that was definitely a mistake I didn't notice locally

Thank you

spatel033 (approved):

app/main/views/service_settings/index.py
```diff
@@ -98,7 +98,12 @@ def service_name_change(service_id):
 
     if form.validate_on_submit():
         try:
-            current_service.update(name=form.name.data)
+            update_fields = {"name": form.name.data}
+
+            if not current_service.live:
```
klssmith, line 103: I wonder if there'd be some benefit to doing this regardless of whether the service is still in trial mode. It wouldn't affect live services, but would show that the name they are currently using hasn't been confirmed as unique. Could be one to check with UCD

joybytes: I think this might not need UCD input since it doesn’t affect the user experience directly — it’s more of a backend/data design decision.

We’ve talked before about possibly using confirmed_unique to prevent service name changes in the future, so adding this now could help lay the ground-work for that. That said, since it's the only feature using the field at the moment, I think the main question is whether it's worth setting now or if it's unnecessary overhead.

I’m happy to include it — it’s a low-risk change and could be VERY helpful later.

klssmith: Thanks for making the change, agree it's not really a UCD thing actually and might not be something we use for live services. But the main reason I think this is an important change to make regardless is because it ensures that the data stored in that column is always accurate and that we don't have a case where the data for trial services is accurate and for live services can't be relied on.

klssmith (approved):

PR #5488, opened 2025-05-30
Standardise on using Freezegun for freezing time

quis:
We removed `pytest-freezegun` in https://github.com/alphagov/notifications-admin/pull/5480

This meant that uses of `pytest.mark.freeze_time` silently stopped working. This only showed up as failures when the tests were run outside business hours.

### DIFF
--- a/tests/app/main/views/test_feedback.py
+++ b/tests/app/main/views/test_feedback.py
@@ -57,7 +57,6 @@ def test_get_support_index_page_when_signed_out(
     assert normalize_spaces(page.select_one("form button").text) == "Continue"
 
 
-@freeze_time("2016-12-12 12:00:00.000000")
 @pytest.mark.parametrize(
     "support_type, expected_h1",
     [
@@ -66,8 +65,9 @@ def test_get_support_index_page_when_signed_out(
     ],
 )
 def test_choose_support_type(
-    client_request, mock_get_non_empty_organisations_and_services_for_user, support_type, expected_h1
+    client_request, mock_get_non_empty_organisations_and_services_for_user, mocker, support_type, expected_h1
 ):
+    mocker.patch("app.main.views.feedback.in_business_hours", return_value=True)
     page = client_request.post(
         "main.support",
         _data={"support_type": support_type},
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):

tests/app/main/views/test_feedback.py
```diff
@@ -263,23 +263,12 @@ def test_passes_user_details_through_flow(
         user_email="test@user.gov.uk",
         notify_ticket_type=None,
         org_id=None,
-        org_type="central",
-        service_id=SERVICE_ONE_ID,
+        org_type=None,
+        service_id=None,
```
CrystalPea, line 267: We are passing these details on prod though, no? Like when we get a ticket form a logged in user and we see the service name and link?

quis: If they have a service in their session, yes. I’m not sure how the service was getting into the session here, or why changing the decorator means it was no longer in the session

CrystalPea: Wasn't the test getting the values from `mock_get_non_empty_organisations_and_services_for_user`?

quis: That just mocks the API response, it doesn’t control which service gets put in the session

CrystalPea: That's true - it's really weird that it worked before, but it doesn't now 😳

quis: I think I’ve fixed it now. See e974fe8

Other alternative would be to add this to these tests instead of freezing the time – let me know which you think is better:
```python
mocker.patch("app.main.views.feedback.in_business_hours", return_value=True)
```

CrystalPea: Nice one! I'm happy with either 🙌🏼

quis: Think I prefer the mocking approach, changed in [20f8a22](https://github.com/alphagov/notifications-admin/pull/5488/commits/20f8a223c1d059338564695bfab2917ff37aec1c)

PR #5493, opened 2025-06-03
Use new GOV.UK header in homepage screen mocks

quis:
_Should not be merged until 25 June_

***

When the new GOV.UK brand launches we should make sure these fit. I also changed the border colour of the textarea in the first one since it was still using the very old grey


quis:
Closing as this is now part of #5504

PR #5494, opened 2025-06-04
Hide the new crown in the footer

quis:
It takes up space and we don’t think it’s applicable for admin interfaces.

# Before

image

# Before with `govukRebrand = true`

image

# After

image


kr8n3r:
this will also hide it from the homepage which isn't an admin interface

PR #5495, opened 2025-06-04
Set utils to v99.8 & rebrand flag to False

tombye:
Brings in the version of utils with the new GOV.UK branding and sets the flag that turns it on in our email previews to off.

## Notes for reviewers

Please try viewing the email branding preview when set to GOV.UK as is, then changing the `rebrand` argument to `True` and checking the branding preview changes to the new brand.


kr8n3r (approved):
:shipit:

PR #5499, opened 2025-06-05
Dockerfile: Inline contents of `unit-tests-image`

robinjam:
[Trello card](https://trello.com/c/7ReRsnUG/1136-get-rid-of-unit-tests-image-and-db-unit-tests-image)

We want to migrate all of our images from GHCR to ECR, and as part of that we want to remove/retire any images that we no longer need (rather than migrating them to ECR and continuing to build/use them).

`unit-tests-image` is currently used as a base image for "frontend builds" of both `admin` & `document-download-frontend`. This is not ideal because (presumably) both apps have different dependency requirements, but by sharing the same base image they both have the union of each of their requirements, which is a waste of storage. (It's also just confusing)

This PR inlines the contents of `unit-tests-image`'s `Dockerfile` into `admin`'s.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,9 +20,32 @@ WORKDIR /home/vcap/app
 
 ##### Frontend Build Image ###################################################
 ARG NOTIFY_ENVIRONMENT=development
-ARG NODE_VERSION=20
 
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
this build in notifications-local just fine
can't speak about historical reasons behind sharing the unit-test-image across apps

PR #5500, opened 2025-06-05
Update trial mode information

karlchillmaid:
Now that the ‘Make your service live’ tasklist page has launched, we need to update the trial mode page to make sure the guidance there matches the new user journey.

### DIFF
--- a/app/templates/views/guidance/using-notify/trial-mode.html
+++ b/app/templates/views/guidance/using-notify/trial-mode.html
@@ -32,7 +32,7 @@
     <li>confirm your service is the only one of its kind in your organisation</li>
     <li>add examples of the messages you want to send</li>
     <li>update your settings so you’re ready to send and receive messages</li>
-  </ul
+  </ul>
   <p class="govuk-body">If your organisation is new to GOV.UK Notify, we’ll ask you to accept our data processing and financial agreement.</p>
   <p class="govuk-body">It can take up to one working day to make your service live.</p>
   <p class="govuk-body">This is so we can check that you’ve met our terms of use.</p>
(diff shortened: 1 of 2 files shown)


app/templates/views/guidance/using-notify/trial-mode.html
```diff
@@ -11,41 +11,30 @@
...
-  <p class="govuk-body">While your service is in trial mode you can only:</p>
-  <ul class="govuk-list govuk-list--bullet">
-    <li>send messages to yourself and other people in your team</li>
-    <li>send {{ email_and_sms_daily_limit | message_count('email') }} per day</li>
-    <li>send {{ email_and_sms_daily_limit | message_count('sms') }} per day</li>
-    <li>create letter templates, but not send them</li>
-  </ul>
+  <p class="govuk-body">When you add a new service it will start in trial mode.</p>
+  <p class="govuk-body">While your service is in trial mode, you can only send messages to yourself and your team members.</p>
+  <p class="govuk-body">There’s a daily limit of {{ email_and_sms_daily_limit | message_count('email') }} emails and {{ email_and_sms_daily_limit | message_count('sms') }} text messages.</p>
```
klssmith, line 16: The formatter used adds in the word "emails" and "text messages", so the way we had things caused the sentence to read "There’s a daily limit of 50 emails emails... "


```suggestion
  <p class="govuk-body">There’s a daily limit of {{ email_and_sms_daily_limit | message_count('email') }} and {{ email_and_sms_daily_limit | message_count('sms') }}.</p>
```

app/templates/views/guidance/using-notify/trial-mode.html
```diff
@@ -11,41 +11,30 @@
...
+  <p class="govuk-body">You cannot send any letters in trial mode, but you can still create letter templates and upload letters, to see how it works.</p>
 
   <div class="govuk-inset-text"> 
-    <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_who_can_use_notify') }}">GP surgeries</a> cannot send any text messages in trial mode.</p>
+    <p class="govuk-body"><a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_who_can_use_notify') }}">GP surgeries and pharmacies</a> cannot send any text messages in trial mode.</p>
```
klssmith, line 20: There's nothing currently stopping pharmacies from sending text messages in trial mode, since we don't know they're a pharmacy at that point. But this matches what we say on a different page and is what we are going to implement, so isn't necessarily an issue.

klssmith (approved):

PR #5501, opened 2025-06-06
Update the guidance for international text message limits

karlchillmaid:


app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -34,30 +35,38 @@ <h1 class="heading-large">Daily limits</h1>
...
-<p class="govuk-body">To change the number of international text messages you can send each day:</p>
+<h2 class="heading-medium">International text message limits</h2>
 
- <ol class="govuk-list govuk-list--number">
-   <li>Go to <b class="govuk-!-font-weight-bold">Settings</b></li>
-   <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">International text message limit</b> row</li>
- </ol>
+  <p class="govuk-body">By default, you can only send 100 text messages a day to international phone numbers.</p>
+
+  <p class="govuk-body">We set this limit because international text message fraud can be very costly.</p>
```
karlchillmaid, line 44: I changed the heading and introduction because I think we need to set some context before explaining how to change the limits.

The tricky thing is not getting bogged down in over-explaining the nature of the fraud.

I’ve tried to prioritise the ‘why’ (fraud costs you money) over the ‘what’ (AIT).

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -34,30 +35,38 @@ <h1 class="heading-large">Daily limits</h1>
...
 
-<p class="govuk-body">When changing the limit, you should think about:</p>
+  <p class="govuk-body">If you need to change the daily limit, you should take steps to protect yourself from fraud.</p>
```
karlchillmaid, line 46: If we need users to protect themselves before they change the limit, it makes sense to put this above the instructions for changing the limit.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -34,30 +35,38 @@ <h1 class="heading-large">Daily limits</h1>
...
- </ul>
-<p class="govuk-body">If you do not need to send international text messages, you can turn this off in <b class="govuk-!-font-weight-bold">Settings</b>.</p>
+  <p class="govuk-body">For example:</p>
 
-<h3 class="heading-medium">Help protect your service from Artificial Inflation of Traffic (AIT)</h3>
-   <p class="govuk-body">AIT is a type of fraud. You should take steps to protect your service from an AIT attack, such as:</p>
     <ul class="govuk-list govuk-list--bullet">
-      <li>track how many messages you send and set an alert in your system for anything that is not normal</li>
-      <li>stop multiple messages being sent to the same number</li>
+     <li>work out how many international text messages you need to send each day and base your new limit on that</li>
```
karlchillmaid, line 51: This instruction fits better as part of the list, rather than keeping it as a standalone piece of advice. I also made the instruction more explicit – explaining exactly what we want users to do.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -34,30 +35,38 @@ <h1 class="heading-large">Daily limits</h1>
...
+     <li>track how many messages you send and set an alert in your system for anything that is not normal</li>
+     <li>stop your service from repeatedly sending messages to the same number</li>
```
karlchillmaid, line 53: I thought ‘repeatedly’ might work better than ‘multiple’. Sometimes sending multiple messages might be justified.

Hopefully ‘repeatedly’ implies the same thing over-and-over in a short period – something which is a nuisance.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -34,30 +35,38 @@ <h1 class="heading-large">Daily limits</h1>
...
 
- <ul class="govuk-list govuk-list--bullet">
- <li>how many international text messages your service needs to send each day</li>
- <li>any costs to your service if there is fraud</li>
- </ul>
-<p class="govuk-body">If you do not need to send international text messages, you can turn this off in <b class="govuk-!-font-weight-bold">Settings</b>.</p>
+  <p class="govuk-body">For example:</p>
 
-<h3 class="heading-medium">Help protect your service from Artificial Inflation of Traffic (AIT)</h3>
-   <p class="govuk-body">AIT is a type of fraud. You should take steps to protect your service from an AIT attack, such as:</p>
```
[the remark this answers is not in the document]
karlchillmaid: If we do feel that we need to mention AIT, I think it should be more of a ‘for example’ rather than setting it up as the main thing.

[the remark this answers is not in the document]
saimaghafoor: Happy to remove references to AIT - it was specifically for comms around the incident and doesn't add anything more to here.

app/templates/views/guidance/using-notify/daily-limits.html
karlchillmaid: Suggest we change ‘you’ to ‘new services’ – based on the questions we’ve been getting from support.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -23,7 +24,7 @@ <h1 class="heading-large">Daily limits</h1>
   ) %}
     {% for message, limit in  [
     ("Emails", rate_limits.email),
-    ("Text", rate_limits.sms + " text messages including a default 100 international text messages"),
+    ("Text messages", rate_limits.sms + " (including up to 100 international text messages)"),
```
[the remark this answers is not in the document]
karlchillmaid:
```suggestion
    ("Text messages", rate_limits.sms),
```

What about removing the international text message limit from the table?

It feels confusing to say ‘contact us’ about the limits, when international limits are self-service.

If the h2 below the table is ‘International text message limits’ this might be enough to draw the user’s attention if it’s relevant to them.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -36,28 +37,37 @@ <h1 class="heading-large">Daily limits</h1>
...
 
-<h3 class="heading-medium">Change the international text message limit</h3>
-<p class="govuk-body">To change the number of international text messages you can send each day:</p>
+<h2 class="heading-medium">International text message limit</h2>
 
- <ol class="govuk-list govuk-list--number">
-   <li>Go to <b class="govuk-!-font-weight-bold">Settings</b></li>
-   <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">International text message limit</b> row</li>
- </ol>
+  <p class="govuk-body">You can only send 100 text messages a day to international phone numbers. This counts towards your total daily limit of 250,000 text messages.</p>
```
karlchillmaid, line 42: We should make the total SMS limit variable, so it always matches the current daily limit.

app/templates/views/guidance/using-notify/daily-limits.html
```diff
@@ -36,28 +37,37 @@ <h1 class="heading-large">Daily limits</h1>
...
 
-<h3 class="heading-medium">Change the international text message limit</h3>
-<p class="govuk-body">To change the number of international text messages you can send each day:</p>
+<h2 class="heading-medium">International text message limit</h2>
 
- <ol class="govuk-list govuk-list--number">
-   <li>Go to <b class="govuk-!-font-weight-bold">Settings</b></li>
-   <li>Select <b class="govuk-!-font-weight-bold">Change</b> on the <b class="govuk-!-font-weight-bold">International text message limit</b> row</li>
- </ol>
+  <p class="govuk-body">You can only send {{current_service.international_sms_message_limit}} text messages a day to international phone numbers. This counts towards your total daily limit of 250,000 text messages.</p>
```
quis, line 42: This won’t work if the user is signed out. Either need to:
- remove it
- replace it with the default
- have a signed in/signed out version

Personally I feel like guidance pages should be static. If the user has multiple services it’s confusing to give a specific number here

CrystalPea:
Ready for review.

quis (approved):
Checked out locally, looks good both signed in/signed out

PR #5509, opened 2025-06-12
Fix ‘today at midnight’ being labelled ‘tomorrow at midnight’ during British Summer Time

quis:
`formatters.get_human_day` was expecting a UTC datetime but was being given a localized one.

This meant that during British Summer Time it was an hour out.

Our tests only covered winter, so that’s why we didn’t spot this when we retired the `forms.get_human_day` function, which silently expected a localized datetime.

***

In the admin app this showed up as an extra option at the bottom of the list. This then confused the Javascript because ‘Thursday’ wasn’t in its list of days.

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,9 +1,10 @@
 import weakref
 from contextlib import suppress
 from copy import deepcopy
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from functools import partial
 from itertools import chain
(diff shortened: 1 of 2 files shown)


CrystalPea (approved):
It works (on my local), ship it

tombye (approved):
Makes sense, nice fix 👍🏻

CrystalPea (approved):

PR #5510, opened 2025-06-12
Feature flag for request report phase 1a

spatel033:
For phase 1a of the request report feature, a feature flag has been added that allows all users to request and download reports for up to 800,000 notifications. If the number of notifications exceeds 800,000, users will be directed to a support form to request the report. See card for the more details.

Added more test cases and updated existing ones to check the new logic. We would keep an eye on logs and number of requests for a week. If we need to go back to using the CSV download link, we can set `REPORT_REQUEST_MAX_NOTIFICATIONS` to `0` and deploy or roll back the changes.

### DIFF
--- a/app/main/views/report_requests.py
+++ b/app/main/views/report_requests.py
@@ -4,15 +4,22 @@ from notifications_python_client.errors import HTTPError
 from werkzeug.utils import redirect
 
 from app import current_service
-from app.constants import REPORT_REQUEST_FAILED, REPORT_REQUEST_STORED
+from app.constants import REPORT_REQUEST_FAILED, REPORT_REQUEST_MAX_NOTIFICATIONS, REPORT_REQUEST_STORED
 from app.main import main
 from app.models.report_request import ReportRequest
-from app.utils.user import user_is_platform_admin
+from app.utils.user import user_has_permissions
+
+
+def validate_report_request_enabled():
+    if REPORT_REQUEST_MAX_NOTIFICATIONS == 0 and not current_user.platform_admin:
+        abort(403)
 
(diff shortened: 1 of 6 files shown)


app/templates/views/notifications.html
```diff
@@ -73,7 +73,7 @@
     {% set retention_text %}
       Data available for {{ partials.service_data_retention_days }} days
     {% endset %}
-    {% if current_user.platform_admin %}
+    {% if report_request_feature_flag %}
```
klssmith, line 76: What are your thoughts on making this conditional check if the flag is enabled _or_ if the user is a platform admin user to allow us to continue to test with the large services?

app/main/views/report_requests.py
```diff
@@ -4,15 +4,22 @@
...
 from app import current_service
-from app.constants import REPORT_REQUEST_FAILED, REPORT_REQUEST_STORED
+from app.constants import REPORT_REQUEST_FAILED, REPORT_REQUEST_MAX_NOTIFICATIONS, REPORT_REQUEST_STORED
 from app.main import main
 from app.models.report_request import ReportRequest
-from app.utils.user import user_is_platform_admin
+from app.utils.user import user_has_permissions
+
+
+def is_request_report_enabled():
```
klssmith, line 13: Minor, but I would find it easier to work out what is happening when this is called in the view functions if this was named something more like `validate_report_request_enabled` or `abort_if_report_request_disabled` since function names starting with `is_` generally return a boolean.

klssmith (approved):
Thanks for making the tweaks, looks good

PR #5511, opened 2025-06-12
CustomProxyFix: don't trust `x-forwarded-host` header

risicle:
https://trello.com/c/eYCdsDtF/1142-ensure-end-user-can-not-set-x-rewrite-url-x-original-url-x-forwarded-host-to-override-url

This is not used or stripped by cloudfront/ALBs, so not safe to trust.

Tested requesting admin root page in firefox with added garbage `X-Forwarded-Host` header which causes production to 500.

### DIFF
--- a/app/proxy_fix.py
+++ b/app/proxy_fix.py
@@ -3,7 +3,7 @@ from werkzeug.middleware.proxy_fix import ProxyFix
 
 class CustomProxyFix:
     def __init__(self, app, forwarded_proto):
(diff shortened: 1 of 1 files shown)


robinjam (approved):
LGTM

~~Have checked [the CloudFront docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html) and they say it removes `X-Forwarded-Proto` before forwarding the request to us - should we set `x_proto` to 0 as well to be explicit?~~ Ignore me, the LB will set this

PR #5515, opened 2025-06-18
Add fix for downloading report request with no status in query string

klssmith:
The status may not be in the query string when clicking to download a report request - we've seen this happen when clicking the download link after viewing a cached search query. We were handling the case where status in the query string existed like `status=` (and correctly setting it to "all") but not where it didn't exist at all - in these cases it was being passed to the api as `None`, which causes an error.

We now set the status to "all" if it's not in the query string.


kr8n3r (approved):
thanks

spatel033 (approved):

PR #5516, opened 2025-06-18
Send files guidance for UI

saimaghafoor:
Content added for send files by email using the Notify website and the UI

To publish once UI send files by email feature is released.

### DIFF
--- a/app/templates/partials/templates/guidance-send-files-by-email.html
+++ b/app/templates/partials/templates/guidance-send-files-by-email.html
@@ -2,9 +2,8 @@
   Send files by email
 </h2>
 <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


rparke:
> Thanks @rparke. Should I add it to the very top of the page (or you could add it as a suggestion in the right place and I'll accept?).

yeh just add it to the top of the file

quis:
Before | After
---|---
image | image
image | image

quis:
I’ve changed the base of this branch so it will be merged into https://github.com/alphagov/notifications-admin/pull/5858

That means this can be merged now, and when we want to release we only have one PR to merge.

kr8n3r (approved):

PR #5521, opened 2025-06-24
When report does not exist into s3 redirects to report-request unavailable page

spatel033:
Check if the report-request file exists in S3 before rendering the report-request ready page and download option. If the file does not exist, redirect to the report-request unavailable page.

Tested by generating new report and tried to access to old report which does not exit into S3 bucket, both cases handled as expected.

### DIFF
--- a/app/models/report_request.py
+++ b/app/models/report_request.py
@@ -6,6 +6,7 @@ from notifications_utils.s3 import s3download
 
 from app import report_request_api_client
 from app.models import JSONModel
+from app.s3_client import check_s3_object_exists
 
 
 class ReportRequest(JSONModel):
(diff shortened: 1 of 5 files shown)


app/models/report_request.py
```diff
@@ -34,3 +35,10 @@ def download(report_request_id):
             bucket_name=ReportRequest.get_bucket_name(),
             filename=f"notifications_report/{report_request_id}.csv",
         )
+
+    @staticmethod
+    def is_report_exists(report_request_id):
```
klssmith, line 40: Optional: Could perhaps name this something like `exists_in_s3` if we want to be extra clear that it's the bucket and not database that's being checked, but it's up to you

klssmith (approved):

PR #5524, opened 2025-06-26
Allow platform admin users to see the sending journey

quis:
Platform admin users should not be allowed to send messages on behalf of a service (either maliciously or by accident).

To prevent this we restrict any endpoints related to sending using the `restrict_admin_use` flag. However this stops us from seeing these pages as our users do, which makes it harder to debug issues when they occur.

This commit makes the steps of the journey visible to platform admin users, while still leaving a restriction on the final step (either sending a job or sending a one-off message).

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -87,7 +87,7 @@ def get_example_letter_address(key):
 
 
 @main.route("/services/<uuid:service_id>/send/<uuid:template_id>/csv", methods=["GET", "POST"])
-@user_has_permissions("send_messages", restrict_admin_usage=True)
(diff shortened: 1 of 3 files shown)


tombye:
Should this also include showing the 'Get ready to send' button in [_email_or_sms_template.html](https://github.com/alphagov/notifications-admin/blob/ad6f31bec8c47735b28671ac4f7796330cec888f/app/templates/views/templates/_email_or_sms_template.html#L2)?

I know we could hack it but not seeing it is a bit odd. Unless I'm missing something? 🤔 

image

joybytes (approved):

PR #5528, opened 2025-06-30
Update guidance for letter delivery times

karlchillmaid:
Royal Mail is changing the second class service.

We need to update the estimated delivery times for second class.

Economy mail is also affected, because Royal Mail is changing its delivery pattern.

We also need to make clear that Saturday deliveries are only for first class letters. Everything else is only delivered on a working day.


klssmith (approved):
Looks good, though we'll presumably want to merge this on Monday 7th

PR #5530, opened 2025-07-02
Rename the 'sending' status

karlchillmaid:
Users frequently contact support to ask why their emails or text messages are ‘stuck’ in `Sending`.

In our terms, the message is still ‘sending’ until our provider returns a `failed` or `delivered` status. This does not appear to match the users’ mental model of what is happening.

It looks like the word `Sending` implies that the act of sending the messages cannot be completed. Users frequently ask if there is a problem with Notify. When we explain that `Sending` means ‘Notify has sent the message to the provider’, users are immediately reassured.

Hypothesis: changing the `Sending` status to `Delivering` will:

* provide users with reassurance that Notify is working fine
* make clear that the message has neither failed or delivered yet
* reduce the burden on user support

There’s [a trello card for this 🔥 break story](https://trello.com/c/C8zv6w0B/98-change-the-sending-status-to-sent).

---

## Things to think about

What is the scope of the changes?

- [X] UI status
- [X] UI guidance
- [ ] ~App status labels~
- [ ] ~Report download content and status labels~ (we’ll do this separately)

We expect not to be able to change the API status as well. We already have some differences between the UI and API statuses, so the risk here is low.


app/formatters.py
```diff
@@ -158,7 +158,7 @@ def format_notification_status(status, template_type):
             "temporary-failure": "Inbox not accepting messages right now",
             "permanent-failure": "Email address does not exist",
             "delivered": "Delivered",
-            "sending": "Sending",
+            "sending": "Sent",
```
karlchillmaid, line 161:
```suggestion
            "sending": "Delivering",
```

app/formatters.py
```diff
@@ -168,7 +168,7 @@ def format_notification_status(status, template_type):
             "temporary-failure": "Phone not accepting messages right now",
             "permanent-failure": "Not delivered",
             "delivered": "Delivered",
-            "sending": "Sending",
+            "sending": "Sent",
```
karlchillmaid, line 171:
```suggestion
            "sending": "Delivering",
```

app/templates/views/guidance/using-notify/message-status.html
```diff
@@ -63,7 +63,7 @@ <h1 class="heading-large">{{ page_title }}</h1>
       caption_visible=False
     ) %}
       {% for message_status, description in [
-        ('Sending', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
+        ('Sent', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
```
karlchillmaid, line 66:
```suggestion
        ('Delivering', 'Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.'),
```

app/templates/views/conversations/messages.html
```diff
@@ -1,20 +1,27 @@
...
         sending
       </time>
     {% endset %}
+  {% elif message.status in ['pending', 'sending'] %}
+    {% set status_text_in_heading = 'Delivering since' %}
+    {% set visual_status_text %}
+      <time class="sms-message-status sms-message-status-outbound hint" datetime="{{ message.created_at | format_datetime_numeric }}" tabindex="-1">
+        delivering
+      </time>
+    {% endset %}
```
quis, line 24: Do users care about the difference between created and pending/sending?

quis (commented):
Generally happy with the change.

Do we need to update the functional tests to look for both statuses?

app/templates/views/conversations/messages.html
quis: I think it should just say ‘delivering’ in both cases. The four high-level statues we now have are total/delivering/delivered/failed so it should fit into one of those. Having to explain that ‘sending’ is a sub-status of delivering is needless complexity.

***

On the CSV thing I’m not sure. Have we changed statuses in the CSV download before, eg when we changed the ‘Phone not accepting messages’ status?

kr8n3r (approved):
comments addressed and see no obvious ommissions

PR #5534, opened 2025-07-02
Show info about remaining daily limit on Settings pages

CrystalPea:
## Work done:

Step 1: Show info about remaining international text message daily limit.

## Screenshots

Screenshot 2025-07-02 at 14 59 00
Screenshot 2025-07-02 at 14 58 40

### DIFF
--- a/app/url_converters.py
+++ b/app/url_converters.py
@@ -13,11 +13,15 @@ class AgreementTypeConverter(BaseConverter):
 
 
 class TemplateTypeConverter(BaseConverter):
-    regex = f"(?:{'|'.join(Service.TEMPLATE_TYPES)}|international_sms)"
+    regex = f"(?:{'|'.join(Service.TEMPLATE_TYPES)})"
 
 
 class BrandingTypeConverter(BaseConverter):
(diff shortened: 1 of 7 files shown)


app/url_converters.py
```diff
@@ -13,7 +13,7 @@ class AgreementTypeConverter(BaseConverter):
 
 
 class TemplateTypeConverter(BaseConverter):
-    regex = f"(?:{'|'.join(Service.TEMPLATE_TYPES)})"
+    regex = f"(?:{'|'.join(Service.TEMPLATE_TYPES)}|international_sms)"
```
quis, line 16: This should be a new converter like `RateLimitType` – you can’t have an `international_sms` template

If we don’t do this it will cause a `500` should a user ever visit a URL like `/services/1234/templates/international_sms`

quis (changes requested):
One change needed I think

CrystalPea:
Screenshots after the content change:

Screenshot 2025-07-03 at 12 16 36
Screenshot 2025-07-03 at 12 19 33

quis (approved):

PR #5539, opened 2025-07-10
Be stricter with undefined values in Jinja templates

quis:
By default Jinja is very forgiving with undefined variables.

For example:

```jinja
{{ something_not_defined }}
```
…will print an empty string.

Or:
```jinja
{{ something_not_defined|length }}
```
…will print 0.

This is a good way of hiding bugs.

***

Jinja has a different way of handling undefined values called [`StrictUndefined`](https://jinja.palletsprojects.com/en/stable/api/#jinja2.StrictUndefined). This would raise an exception in either of the above examples.

However because there are some cases where we (and the `govuk-frontend-jinja` library) rely on Jinja’s non-strict undefined behaviour. So we can’t use `StrictUndefined` out of the box.

This pull requests implements a custom subclass of `Undefined` which copies [the implementation of `StrictUndefined`](https://github.com/pallets/jinja/blob/5ef70112a1ff19c05324ff889dd30405b1002044/src/jinja2/runtime.py#L1039-L1062) in most cases but is more permissive where it needs to be. It then fixes the places where this would raise an exception.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -633,6 +633,14 @@ def init_jinja(application):
 
 
 class NotifyJinjaUndefined(jinja2.Undefined):
+    """
+    The same as jinja2.StrictUndefined with a few exceptions, noted
+    in specific comments.
+
+    For more context see:
+    https://jinja.palletsprojects.com/en/stable/api/#undefined-types
+    """
+
     __slots__ = ()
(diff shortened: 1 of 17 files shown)


kr8n3r (approved):
happy to have this. like you said there are instances where we rely on current behaviour but this might make it more difficult to introduce bugs

PR #5543, opened 2025-07-16
upgrade utils to 100.1.0

rparke:
adds the latest version of libphonenumber, 9.0.9, to utils. Ensures metadata is kept up to date


klssmith (approved):
Have checked that the breaking change in version 100 of utils that this brings in has already been accounted for ✅

PR #5545, opened 2025-07-17
Add spacing between blue pills on API page

tombye:
Sets the gap between buttons on smaller screens as 15px, to match the existing styles on the template page.

The API page has a gap of 15px below it on larger
screens while the template page has one of 20px.
To make both consistent this makes them use an
override class with a spacing unit of 3, so they
will be 15px on all sizes of screen. Because this
removes 5px below them all on larger screens on
the templates page, this also adds an override
class of 1 to the group.

See the design system docs for details of override classes and their units:

https://design-system.service.gov.uk/styles/spacing/

## Before

image

## After

image

### DIFF
--- a/app/templates/views/api/index.html
+++ b/app/templates/views/api/index.html
@@ -11,14 +11,14 @@
     API integration
   </h1>
 
-  <nav class="govuk-grid-row bottom-gutter-1-2">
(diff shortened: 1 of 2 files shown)


kr8n3r (approved):
🎉

PR #5547, opened 2025-07-17
Bugfix: Do not allow inbound sms number removal when no other sender ids

CrystalPea:
Fixes a bug in the feature to remove the inbound number from a service through the UI. We were not checking that we're not removing the default sender id.

So this PR introduces that check, and a matching unit test.

I also spotted that a few views required "manage_api_keys" permission, where they should have only required "manage_service" permission, so I removed the overly restrictive permission check.

Screenshot of the error message:
Screenshot 2025-07-17 at 16 29 27

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -712,7 +712,7 @@ def service_receive_text_messages_stop(service_id):
     if form.validate_on_submit():
         if current_service.default_sms_sender == current_service.inbound_number:
             form.removal_options.errors.append(
-                "You must change default text message sender ID before you can stop receiving text messages"
+                "You need to change your default text message sender ID before you can continue"
             )
 
         else:
--- a/tests/app/main/views/service_settings/test_service_settings.py
+++ b/tests/app/main/views/service_settings/test_service_settings.py
@@ -5051,10 +5051,7 @@ def test_service_receive_text_messages_stop_fails_when_inbound_number_is_default
     error_summary = page.select_one(".govuk-error-summary")
     assert error_summary is not None
     assert "There is a problem" in error_summary.text
-    assert (
-        "You must change default text message sender ID before you can stop receiving text messages"
-        in error_summary.text
-    )
+    assert "You need to change your default text message sender ID before you can continue" in error_summary.text
 
     radios = page.select("input[type=radio]")
     assert len(radios) == 2


app/main/views/service_settings/index.py
```diff
@@ -710,21 +710,29 @@ def service_receive_text_messages_stop(service_id):
     inbound_number = current_service.inbound_number
 
     if form.validate_on_submit():
-        archive = form.removal_options.data == "true"
+        if len(current_service.sms_senders) == 1:
+            form.removal_options.errors.append(
+                "Can't remove the only text message sender id. Please set up another sender id first."
```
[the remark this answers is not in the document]
karlchillmaid:
```suggestion
                "You need to add a new text message sender ID before you can stop receiving text messages"
```

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -5027,6 +5034,40 @@ def test_service_receive_text_messages_stop_handles_error(
...
+        ".service_receive_text_messages_stop",
+        service_id=SERVICE_ONE_ID,
+        _data={"removal_options": "true"},
+        _expected_status=200,
+    )
+
+    error_summary = page.select_one(".govuk-error-summary")
+    assert error_summary is not None
+    assert "There is a problem" in error_summary.text
+    assert "Can't remove the only text message sender id. Please set up another sender id first." in error_summary.text
```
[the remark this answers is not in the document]
karlchillmaid:
```suggestion
    assert "You must add a new text message sender ID before you can stop receiving text messages" in error_summary.text
```

app/main/views/service_settings/index.py
```diff
@@ -83,7 +83,7 @@
 
 
 @main.route("/services/<uuid:service_id>/service-settings")
-@user_has_permissions("manage_service", "manage_api_keys")
+@user_has_permissions("manage_service")
```
klssmith, line 86: This change isn't doing what was intended. Before this change, the views were accessible to people with _either_  the `manage_api_keys` or `manage_service` permission (or both).

After this change, people with the `manage_api_keys` permission can no longer access it. If you trace the code of the `user_has_permissions` decorator, this is how it's doing the permissions check: https://github.com/alphagov/notifications-admin/blob/a5259314bb16cb9931995151395469a649210601/app/models/user.py#L273

CrystalPea: Ahh, fair, I will revert that commit, thanks a lot!

CrystalPea: I will make a note to add unit test describing this bit of code - as it is not protected by unit tests at the moment.

klssmith:
Works well, so if @karlchillmaid is happy with the content then I think this is good to go (though you might need to force push to get the concourse checks to run)

tests/app/main/views/service_settings/test_service_settings.py
```diff
@@ -5027,6 +5029,39 @@ def test_service_receive_text_messages_stop_handles_error(
...
+        service_id=SERVICE_ONE_ID,
+        _data={"removal_options": "true"},
+        _expected_status=200,
+    )
+
+    error_summary = page.select_one(".govuk-error-summary")
+    assert error_summary is not None
+    assert "There is a problem" in error_summary.text
+    assert (
+        "You must change default text message sender ID before you can stop receiving text messages"
```
karlchillmaid, line 5055:
```suggestion
        "You need to choose another default text message sender ID before you can continue"
```

karlchillmaid:
```suggestion
        "You need to choose a default text message sender ID before you can continue"
```

karlchillmaid:
@CrystalPea I’ve added commits for my preferred error message content

app/main/views/service_settings/index.py
```diff
@@ -710,21 +710,29 @@ def service_receive_text_messages_stop(service_id):
     inbound_number = current_service.inbound_number
 
     if form.validate_on_submit():
-        archive = form.removal_options.data == "true"
+        if current_service.default_sms_sender == current_service.inbound_number:
+            form.removal_options.errors.append(
+                "You need to choose a default text message sender ID before you can continue"
```
karlchillmaid, line 715: Would it be helpful to add a link to the relevant settings page to that error message? So the user can go and fix things quickly rather than navigating to it via the Settings page? Or are we relying on the ‘Add a new text message sender ID’ further down the page to do this?

Are there any usability/accessibility issues with linking to another page from the error summary?

PR #5550, opened 2025-07-17
fix: Test for padded image by avoiding brittle byte comparison

joybytes:
## Summary:
The test `test_pad_adds_transparency_to_top_and_bottom` previously relied on a raw byte-for-byte comparison of PNG files to verify that padding was added with transparency. Minor differences in metadata or encoding could cause the test to fail even when the images were visually identical.

This change updates the test to:

- Normalise the output and expected images using PIL

- Save them into byte streams for a cleaner comparison

### DIFF
--- a/tests/utils/test_image_processing.py
+++ b/tests/utils/test_image_processing.py
@@ -1,4 +1,7 @@
+import io
+
 import pytest
+from PIL import Image
 from werkzeug.datastructures import FileStorage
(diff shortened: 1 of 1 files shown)


CrystalPea (approved):
This test was failing locally for me, too. Now it is passing ✨ . Thanks a lot 🙌🏼

PR #5553, opened 2025-07-18
Rewrite LiveCheckboxControls as ESM module

tombye:

Rewrite follows the same conventions we previously used for:
- FocusBanner
- Collapsible checkboxes
- ColourPreview
- FileUpload
- Autofocus
- PreviewPane
- CopyToClipboard
- LiveSearch
- ListEntry
- RadiosWithImages
- EnhancedTextbox

LiveCheckboxControls is a base class for both AddBrandingOptionsControls and SetAuthTypeForm so I did them all together.

This also includes a rewrite of their tests, not just to allow for them to be a module but also to make sure the states of the page loading with checkboxes selected and that of the page having checkboxes selected afterwards are both represented when needed and in the right place in the test.

Added bonus: removes the use of jQuery as well.

## Notes to reviewers

I refactored the tests so comparing them might be tricky. Sorry. Could be best to run them on `main` and this branch and go by the output rather than working out what's changed in the test code.

Pages to manually test for this are:
- Service settings > Set auth type
- Org settings > Email branding options > Add branding options
- Platform Admin > Email branding


kr8n3r (approved):
great work

kr8n3r (approved):

PR #5559, opened 2025-07-23
Rename the ‘Using Notify’ menu as ‘Guidance’

karlchillmaid:
Change `using-notify` to `guidance`

### DIFF
--- a/app/templates/views/guidance/using-notify/index.html
+++ b/app/templates/views/guidance/using-notify/index.html
@@ -12,7 +12,7 @@
   <h1 class="heading-large">Guidance</h1>
 
   <p class="govuk-body">This guidance is for teams using GOV.UK Notify to send emails, text messages and letters.</p>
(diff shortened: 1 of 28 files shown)


app/templates/views/guidance/using-notify/index.html
```diff
@@ -1,15 +1,15 @@
...
-  Using Notify
+  Guidance
 {% endblock %}
 
 {% block content_column_content %}
 
-  <h1 class="heading-large">Using Notify</h1>
+  <h1 class="heading-large">Guidance</h1>
 
   <p class="govuk-body">This information is for teams using GOV.UK Notify to send emails, text messages and letters.</p>
```
karlchillmaid:
```suggestion
  <p class="govuk-body">This guidance is for teams using GOV.UK Notify to send emails, text messages and letters.</p>
```

PR #5575, opened 2025-08-18
Dockerfile: Base `frontend_build` on `node` image (only)

robinjam:
https://trello.com/c/bY9k4JvU/1311-investigate-image-building-issues-on-admin-document-download-frontend

Previously this image was based on a Python image which had the `/usr/local/bin` and `/usr/local/lib` directories copy-pasted into it from a Node image.

This is brittle because the Node and Python images might be referencing different versions of Debian. It's also unnecessary because we don't use Python at all during frontend builds.

Also includes some drive-by refactors (see commit messages).

Tested in dev-c.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,8 +12,7 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
     libcurl4 \
     curl \
-    && apt-get -y clean && \
-    rm -rf /var/lib/apt/lists/* /tmp/*
+    && rm -rf /var/lib/apt/lists/* /tmp/*
 
 WORKDIR /home/vcap/app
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -21,8 +21,7 @@ WORKDIR /home/vcap/app
 ##### Frontend Build Image ###################################################
 ARG NOTIFY_ENVIRONMENT=development
 
-FROM --platform=linux/amd64 node:22-slim AS node
-FROM --platform=linux/amd64 python:3.11-slim-bookworm AS frontend_build
+FROM --platform=linux/amd64 node:22-slim AS frontend_build
```
risicle, line 23: Should this be fixed to `22-slim-bookworm` to avoid sudden surprises?

PR #5584, opened 2025-09-10
New support form journey, part 1

klssmith:
This implements the first part of the new support form journey, where we ask more questions about the problem faced in order to give different types of Zendesk tickets different subjects and assign them to the person on tech / non-tech support.

This work updates the content of existing pages, adds in the extra pages to triage problems and sends more data to Zendesk about the ticket subject and responder. Some of the existing code can be removed / changed (we no longer have "general" ticket types, the "/triage" page can be updated to work differently), but this has been left for now in case people are using the form as the changes are deployed.

Later work will
- Add in the pages which deal with problems signing in
- Handle in hours-emergencies differently by updating the content relating to our response time and alerting us 

[Trello card](https://trello.com/c/gvrawEAQ/1414-build-new-user-support-form)

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -4,7 +4,11 @@ import pytz
 from flask import current_app, redirect, render_template, request, session, url_for
 from flask_login import current_user
 from notifications_utils.bank_holidays import BankHolidays
-from notifications_utils.clients.zendesk.zendesk_client import NotifySupportTicket, NotifySupportTicketComment
+from notifications_utils.clients.zendesk.zendesk_client import (
+    NotifySupportTicket,
+    NotifySupportTicketComment,
+    NotifyTicketType,
+)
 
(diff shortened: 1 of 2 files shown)


kr8n3r:
had a look from FE perspective
`/ask-question-give-feedback`
`/report-problem?severe=no&issue=something-else`
`/report-problem?severe=no&issue=problem-sending`
`/report-problem?severe=no&issue=technical-error-trial-mode` (and other branched of report-a-problem)
may be missing a textarea label
Screenshot 2025-09-10 at 12 02 00

Error messages and error-summary are there

klssmith:
@kr8n3r thanks, yes it was deliberate not to include it because it was felt that the heading explained what should go in the text box so it wasn't necessary. But if there should be a label there we can add one (cc @karlchillmaid )

PR #5594, opened 2025-09-19
returned letters: remove handling of "old-style" api responses

risicle:
Now that https://github.com/alphagov/notifications-api/pull/4555 has been merged & released.


rparke (approved):

risicle:
Will merge next week when more people with dev envs running will have picked up the required api change.

PR #5604, opened 2025-09-29
Rewrite FullScreenTable as an ES Module

kr8n3r:

In this PR we:

- update the tests to be ES Module and polyfill some Javascript methods that JSDOM doesn't return correct values
- remove old module JS file

## How to test

- this is module is used on the scheduling of notifications on the "send" page and
- upload CSV emergency list preview page


tombye (changes requested):
Really nice work, and so much better to read than the original (I kept thinking this when reviewing ⭐). I did find at least one thing that needs changing but nothing too big.

This review has taken me ages, so sorry for that, but the stuff about changes to `ScreenMock` did need some thinking.

tombye (approved):
All looks good now. Thanks for all the work on this, it's been a lot! 🎉 🏅 👍🏻

PR #5658, opened 2025-11-10
Rewrite RemoveInPresenceOf as an ES Module

kr8n3r:
## What

[Trello card](https://trello.com/c/ElSiCfRh/1539-modernise-our-javascript-removeinpresenceof)

In this PR we:

- remove old module JS file

## How to test

This is module is used on the jobs page `services/<serviceID>/jobs/<jobID`
The simplest way to test JS would be to 
- schedule some fake emails and once sent go to the jobs page

JS should remove the table of recipients in favour actual delivery notifications

Alternatively you can add a dummy container to `app/templates/views/jobs/job.html` page after scheduling messages.
JS should remove the table of recipients in favour of the new container


DilwoarH (changes requested):
Minor change but all looks good otherwise

DilwoarH (approved):

PR #5667, opened 2025-11-14
Rewrite AuthenticateSecurityKey as an ES Module

kr8n3r:
## What

[Trello ticket](https://trello.com/c/YVGCa1fq/1543-modernise-our-javascript-authenticatesecuritykey)

In this PR we:

- remove old module JS file

## Review
- see individual commits
- run locally and test: used in the login flow for Notify team members with admin permissions


tombye (changes requested):
This is SO much nicer to read than the original. Thanks for taking the time to not just rewrite it in ESM but also make the structure match that of the logic. Especially makes it better because of the complexity of the CBOR interactions.

Typical of me, I spotted one thing I think needs some 👀s but happy with whatever's the simplest solution.

tombye (approved):
Thanks for sorting the change, now good to go 👍🏻

tombye (approved):
Still approved 👍🏻

PR #5676, opened 2025-11-24
Rewrite RegisterSecurityKey as an ES Module

kr8n3r:
## What

[Trello ticket](https://trello.com/c/lqEjfroa/1554-modernise-our-javascript-registersecuritykeys)

_Code changes closely resemble that of AuthenticateSecurityKey, rewritten in https://github.com/alphagov/notifications-admin/pull/5667_

In this PR we:

- remove old module JS file

## Review
- see individual commits
- run locally and test: used on the `/your-account` page to add new yubi or pass keys


tombye (approved):
Nice work. Like the other webauthn one, this is also MUCH improved by the new async/await structure, so much kinder on the 👀s.

It all looks good too 👍🏻

PR #5682, opened 2025-11-27
Rewrite ErrorBanner as an ES Module

kr8n3r:
## What

[Trello ticket](https://trello.com/c/UG4DRmaN/1557-modernise-our-javascript-errorbanner)

In this PR we:

- use it `authenticateSecurityKeys` and `registerSecurityKeys` modules and tests
- remove old module JS file
- fix to rollup.config from https://github.com/alphagov/notifications-admin/pull/5366/ where legacy files weren't quite removed

## Review
- see individual commits
- run locally and test: used on the `login` and `/your-account` pages when authenticating and adding new security keys


tests/javascripts/authenticate-security-key.test.mjs
```diff
@@ -68,11 +69,11 @@ describe('Authenticate with security key', () => {
     window.location = mockWindowLocation;
     mockWindowLocation.assign = jest.fn();
 
-    // mock error banner js
-    window.GOVUK.ErrorBanner = {
-      showBanner: jest.fn(),
-      hideBanner: jest.fn(),
-    };
+    // mock error banner class methods
+    new ErrorBanner();
```
tombye: This is a clever way to spy on these methods but why does it create a new instance? Excuse if I'm missing something obvious here.

Might also be worth using jest.spyOn which will just get wiped when jest.restoreAllMocks() is run. That could also happen here but I'm can't see how at a glance.

kr8n3r: I suspect as we wipe all after each test, ErrorBanner gets cleared. I've added https://jestjs.io/docs/jest-object#jestresetmodules afterEach to make it clear.

Updated the comment as we aren't really mocking, we are actually using the ErrorBanner JS module method and then checking if it gets called, so we need to instantiate it. 

Aside: not finding Jest's docs on ES modules that helpful

tombye: Yes, the ES modules docs made me 😞 too.

kr8n3r: resolved by @tombye himself :)

tombye (approved):
Thanks for making the changes, and for doing lots of research to figure out how to test ES modules in Jest 🥇. Now looks good to go 👍🏻

PR #5825, opened 2026-03-04
Update to Design System v6

kr8n3r:
## What 

Update admin to use the latest and greatest version of Design System's [govuk-frontend](https://github.com/alphagov/govuk-frontend/releases).

[Trello card](https://trello.com/c/Eh8gJ5hw/1691-bump-design-system-to-v6-on-admin-app)

## Review

See individual commits for more details

### Green and blue shade/tint changes

Before
Screenshot 2026-03-05 at 11 46 03

After
Screenshot 2026-03-05 at 11 46 10

### Font size on mobile breakpoint is larger
Before
before

After
after

### Footer links are no longer black and have visited state
Confirmed by DS team

Before
Screenshot 2026-03-06 at 09 49 55

After
Screenshot 2026-03-06 at 09 50 04


tombye (approved):
Perfect 🤌

PR #5863, opened 2026-03-31
Wrap window.location methods in utility helper functions

kr8n3r:
Jest 30 updates the version of JSDOM to 26 which no longer allows mocking window.location -
https://jestjs.io/blog/2025/06/04/jest-30#known-issues

We use `window.location` methods in our Javascript functions and tests.
Wrapping window.location methods in utility functions, using them in JS files and then mocking them in Jest tests is our best way forward.

It's explicit and easily removable. We've only added the ones we need in other .mjs files.


README.md
```diff
@@ -85,10 +85,10 @@ make bootstrap
 make test
 
 # run the js tests
-npx jest
```
tombye, line 88: This, and the command for running a single JS test, is also so much nicer. Less cognitive overhead ftw.

package.json
```diff
@@ -6,7 +6,8 @@
     "node": ">=10.15.3"
   },
   "scripts": {
-    "test": "npm run lint:scss && npm run lint:js && node --experimental-vm-modules node_modules/.bin/jest tests/javascripts",
+    "test": "npm run lint:scss && npm run lint:js && npm run test:js tests/javascripts",
+    "test:js": "node --experimental-vm-modules node_modules/.bin/jest",
     "test-watch": "jest --watch tests/javascripts",
     "debug": "node --inspect-brk node_modules/.bin/jest --runInBand $npm_config_test",
```
tombye: Can we also add the `--experimental-vm-modules` flag to the `debug` script? I think turning off transform in the config might have borked this.

```suggestion
    "debug": "node --inspect-brk --experimental-vm-modules node_modules/.bin/jest --runInBand $npm_config_test",
```

tombye (changes requested):
This is really approved but I asked for a change, which would mean I'd have to re-approve when it was done anyway so...

But this is great, and it took ages to work some of the stuff out around mocking modules in Jest to 👏🏻👏🏻👏🏻👏🏻👏🏻 for all the effort and 🧠 power that went into this.

package.json
kr8n3r: great spot!

tombye (approved):
Thanks for making the changes. Now good to go 🚀 👍🏻

PR #5871, opened 2026-04-09
Copy templates also copies files

rparke:
we noticed a bug where we werent' copying over files when copying over a template with files. This results in the new template having fileplaceholders interpreted as "regular" placeholders for normal personalisation. This downloads the old file from s3, creates a new entry in `template_email_files` and then makes it live.

We don't have the api endpoints to make a file live straightaway - we may wish to implement this instead of hitting the create endpoint and then the update endpoint.

Copying templates when contact link is not set:
Screenshot 2026-05-06 at 10 34 17

Screenshot 2026-05-06 at 10 34 47

Screenshot 2026-05-06 at 10 34 56

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -479,6 +479,16 @@ def copy_template(service_id, template_id):
             letter_welsh_content=template.get_raw("letter_welsh_content"),
             has_unsubscribe_link=template.get_raw("has_unsubscribe_link"),
         )["data"]
+        if template.template_type == "email":
+            if new_files := template.email_files:
+                for file in new_files:
+                    file.copy_file(
+                        destination_template_id=new_template["id"],
+                        destination_service_id=current_service.id,
+                        retention_period=file.retention_period,
+                        validate_users_email=file.validate_users_email,
+                        link_text=file.link_text,
+                    )
         if template.template_type == "letter" and template.get_raw("letter_attachment"):
             _copy_letter_attachment(from_template=template, to_template=new_template)
         return redirect(url_for(".view_template", service_id=service_id, template_id=new_template["id"]))
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -17,6 +17,7 @@ from flask import (
     url_for,
 )
 from flask_login import current_user
+from markupsafe import Markup
 from notifications_python_client.errors import HTTPError
 from notifications_utils.formatters import formatted_list
 from notifications_utils.pdf import pdf_page_count
@@ -464,6 +465,15 @@ def copy_template(service_id, template_id):
     form = CopyTemplateForm(template_id=template.id, name=template.name, parent_folder_id=to_folder_id)
 
     if request.method == "POST":
+        files_to_copy = getattr(template, "email_files", False)
+        if template.template_type == "email" and files_to_copy and not current_service.contact_link:
+            flash(
+                Markup(
+                    f"You need to add contact details for your service. "
+                    f'<p class="govuk-body govuk-!-font-weight-bold"><a href="{url_for("main.send_files_by_email_contact_details", service_id=service_id)}" class="govuk-link">Add contact details for your service</a></p>'  # noqa: E501
+                )
+            )
+            return redirect(url_for("main.choose_template_to_copy", service_id=service_id))
         new_template = service_api_client.create_service_template(
             name=form.name.data,
             type_=template.template_type,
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_templates.py
```diff
@@ -2735,6 +2735,81 @@ def test_post_copy_template(
...
+    active_user_with_permissions,
+    mock_get_service,
+    multiple_sms_senders,
+    mock_get_service_email_template_with_file,
+    mock_get_service_templates,
+    mock_get_organisations_and_services_for_user,
+    fake_uuid,
+    mock_create_service_template,
+    mocker,
+):
```
rparke, line 2894: There's a bunch of code duplication in this with `test_post_copy_template_into_folder` which I don't love. And mocking out uuids in a nice way proved tricky, so I'm open to any ideas to make this a bit nicer

quis: Example of getting deterministic UUIDs without mocking:

https://github.com/alphagov/notifications-admin/blob/090b924b88fa1e03a17537e364058f50e5fe0c8c/tests/app/main/views/test_sign_in.py#L74

rparke: Thanks!

rparke: I still need to patch the file_id and the template_id specifically.

app/models/template_email_file.py
```diff
@@ -49,6 +52,40 @@ def create(*, filename, file_contents, template_id):
...
+        destination_service_id,
+        validate_users_email,
+        retention_period,
+        link_text,
+    ):
+        from app import current_user, template_email_file_client
+
+        destination_file_id = str(uuid.uuid4())
+        source_file_location = _get_file_location(self.id, self.service_id)
+        destination_file_bytes = download_template_email_file_from_s3(file_location=source_file_location).read()
```
klssmith, line 68: This line gives an error if you're trying to copy a template with files from a different service.

If copying across services is meant to be supported, it would good to check what happens if you copy a template with an email file to a service which hasn't set up contact details for sending files by email, so shouldn't be able to use emails with files yet.

rparke: Hmm, I'll have a think and discuss it. I think we should just redirect to the view that prompts users to set up contact details for service.

saimaghafoor: the inline validation should be :

'You need to add contact details for your service'
(The 'add contact details for your service' should link to https://www.notifications.service.gov.uk/services/d5637812-377d-4308-ba0a-c76462c04df0/templates/91ad8678-4dc0-4697-bc92-e9ef703d5c33/files/setup - whatever the relevant page is for the service).

app/main/views/templates.py
```diff
@@ -469,6 +470,16 @@ def copy_template(service_id, template_id):
     form = CopyTemplateForm(template_id=template.id, name=template.name, parent_folder_id=to_folder_id)
 
     if request.method == "POST":
+        if template.template_type == "email":
+            files_to_copy = template.email_files
+            if files_to_copy and not current_service.contact_link:
+                flash(
+                    Markup(
+                        f"You need to add contact details for your service. "
+                        f'<a href="{url_for("main.send_files_by_email_contact_details", service_id=service_id)}" class="govuk-link">Add contact details for your service</a>'  # noqa: E501
```
kr8n3r:
```suggestion
                        f'<p class="govuk-body govuk-!-font-weight-bold"><a href="{url_for("main.send_files_by_email_contact_details", service_id=service_id)}" class="govuk-link">Add contact details for your service</a></p>'  # noqa: E501
```
i would wrap the link in a paragraph

Image

joybytes (approved):

app/main/views/templates.py
```diff
@@ -465,16 +465,15 @@ def copy_template(service_id, template_id):  # noqa: C901
...
-        if template.template_type == "email":
-            files_to_copy = template.email_files
-            if files_to_copy and not current_service.contact_link:
-                flash(
-                    Markup(
-                        f"You need to add contact details for your service. "
-                        f'<p class="govuk-body govuk-!-font-weight-bold"><a href="{url_for("main.send_files_by_email_contact_details", service_id=service_id)}" class="govuk-link">Add contact details for your service</a></p>'  # noqa: E501
-                    )
+        files_to_copy = getattr(template, "email_files", False)
+        if template.template_type == "email" and files_to_copy and not current_service.contact_link:
```
BlessedDev, line 469: How about using the walrus operator here?
   `if template.template_type == "email" and (files_to_copy:=getattr(template, "email_files", False)):`

This refactor loses the pre `email template ` check the original code had before checking for any present files

rparke: The files_to_copy argument is used in multiple places. It's used here to abort the copy before doing anything else if a service doesn't have contact details set, so that's why I wanted to keep it out of the if clause, to make that clear.

BlessedDev (approved):

PR #5872, opened 2026-04-09
Encryption for tokens

whpearson:
This turns on the feature 

https://github.com/alphagov/notifications-utils/pull/1336

By passing in environment variables.

We start checking for encrypted tokens before we encrypt them.

The plan is to keep the signing code around for at least a year,
so that we don't have eny tokens in things like unsubscribe links.

Encryption has been added everywhere for simplicities sake.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -6,6 +6,7 @@ class Config:
     ADMIN_CLIENT_SECRET = os.environ.get("ADMIN_CLIENT_SECRET")
     API_HOST_NAME = os.environ.get("API_HOST_NAME")
     SECRET_KEY = os.environ.get("SECRET_KEY")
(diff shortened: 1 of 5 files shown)


pytest.ini
```diff
@@ -8,6 +8,7 @@ env =
     API_HOST_NAME=test
     DANGEROUS_SALT=dev-notify-salt
     SECRET_KEY=dev-notify-secret-key
+    TOKEN_SECRET_KEY="5YNWU0e_pN5ZyaSZvBd5uZb_sZlrVDFeOjiea6dq4zQ="
```
quis: You don’t need to add this here – see https://github.com/alphagov/notifications-admin/pull/5879

spatel033 (approved):

PR #5879, opened 2026-04-14
Remove environment variables from `pytest.ini`

quis:
Most of these are ignored because they are hard coded in `Config.Test`.

Let’s be consistent so people who are adding or changing them only have to do so in one place.


whpearson (approved):
Makes sense

PR #5880, opened 2026-04-14
`TemplateList`: add `interruptible_iter` property, use in jinja iteration

risicle:
For services with a very large number of templates, the `main.choose_template` view can spend a long time rendering the list, during which time it will not yield to the event loop. This may adversely affect other requests being handled by the same process.

To avoid this, give `TemplateList` an `interruptible_iter` property that produces an iterator that will automatically inject `sleep(0)`s every 32 iterations.

### DIFF
--- a/app/models/template_list.py
+++ b/app/models/template_list.py
@@ -1,6 +1,7 @@
 from werkzeug.utils import cached_property
 
 from app import format_notification_type
+from app.utils import interruptible_io
 
 
 class TemplateList:
@@ -22,6 +23,8 @@ class TemplateList:
     comments on those classes for more details.
     """
 
+    INTERRUPTIBLE_ITER_INTERRUPTIBLE_EVERY = 32
+
     def __init__(
         self,
         *,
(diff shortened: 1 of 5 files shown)


app/models/template_list.py
```diff
@@ -36,6 +39,10 @@ def __init__(
     def __iter__(self):
         yield from self.items
 
+    @property
+    def interruptible_iter(self):
+        return interruptible_io.interruptible_iter(self, self.INTERRUPTIBLE_ITER_INTERRUPTIBLE_EVERY)
+
```
quis, line 45: I’d consider replacing the body of `__iter__` rather than calling through to it. There is more than one place where the template list is rendered. Authors either need to always remember to call `interruptible_iter`, or it could be the default behaviour.

Example: https://github.com/alphagov/notifications-admin/blob/38670ad4711f8c4c7a624e7e6a5c073d3d18241e/app/main/views/conversation.py#L58-L60

Which is iterated over here: https://github.com/alphagov/notifications-admin/blob/38670ad4711f8c4c7a624e7e6a5c073d3d18241e/app/templates/views/templates/choose-reply.html#L48

risicle: I'm just a little too nervous to make it the default behaviour. It has what I would call surprising side effects. But yes I should probably use `interruptible_iter` there too.

quis (approved):

PR #5883, opened 2026-04-15
Add `min-release-age=3` config to npmrc and use it in CI as well

kr8n3r:
This tells NPM to reject any package version published less than 3 days ago.
Part of the recommendation from the last round of npm supply chain attack.
Available with Node 24 which we already run the app on.

We also want the CI `npm ci` command to have access to

```
ignore-scripts=true
min-release-age=3
````

protections we set in the npm config. To do that, we need to copy the .npmrc file from the repo.

[Tested the config being picked up by CI by adding a "NPM WITH NPRMC into a heading string](https://concourse.notify.tools/teams/pull-requests/pipelines/pull-requests/jobs/build-notifications-admin-pr/builds/1297#L69df6f23:233)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 ignore-scripts=true
+min-release-age=3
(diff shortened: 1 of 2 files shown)


tombye (approved):
Makes sense, and we tested the `npmrc` is being picked up by all the calls to `npm` on a dev env so 👍🏻

PR #5887, opened 2026-04-17
Let users download a CSV of their team members

quis:
Kat reckons this comes up fairly often in Zendesk when people want to audit a team with lots of members.

It’s something we’d also find useful when we are trying to contact the ‘admins’ of a service for some reason. We can download this CSV, filter it, then re-upload it into Notify.

## For users who have the manage settings permission (or platform admins)

image

## For users who are members of the service but not the organisation

image

## The downloaded file

image


kr8n3r:
something to fix on smaller viewports

Screenshot 2026-04-20 at 08 04 32

the css adjustment also breaks the check letter preview page

Screenshot 2026-04-20 at 08 14 47

perhaps we need our own custom button-group wrapper like https://design-system.service.gov.uk/components/button/button-group/

that will:
-stack items on smaller viewports
- allow us to right-align the 2nd or last item on larger viewports

saying that, just adding the override to it
```
    .govuk-button-group .govuk-link {
        margin-left: auto;
     }
```
gives us below if useful

Screenshot 2026-04-20 at 08 27 09

tablet and down arrangement
Screenshot 2026-04-20 at 08 27 16


---- 
in the report we show 'invited' or '(cancelled invite)' in the `name` column. feels weird but i see we don't know their name so only thing we can show is this

quis:
Tweaked the margins so it looks consistent with other similar pages:

image

image

image

kr8n3r (approved):
looks good now 👏

PR #5895, opened 2026-04-21
Update Rollup ecosystem

kr8n3r:
Update
 - rollup to 4.60.1
 - @rollup/plugin-node-resolve to 16.0.3

 Remove
 - @rollup/plugin-multi-entry

 Replaced by own simple concatenation plugin.
 Plugin-multi-entry introduced a breaking change in v7 of always creating a deterministic output which uses `sort()` to bundle the inputs in order
- https://github.com/rollup/plugins/pull/1879

 For the legacy bundle, we need the order of dependencies to be fixed.

 Instead of continuing to use this dependency, we have our own "plugin" which concatenates those files. In the build, we make sure that `this` means `window` context and that the file as before.

**How to test**

Compare the compiled `all.js` on `main` vs this branch.
Only difference is in the textarea-caret-position plugin, the difference in whitespace character.
Screenshot 2026-04-21 at 12 12 25

Manually testing this in browser shows no difference in functionality

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -1,9 +1,29 @@
 import { nodeResolve } from '@rollup/plugin-node-resolve';
-import multi from '@rollup/plugin-multi-entry';
 import terser from '@rollup/plugin-terser';
 import copy from 'rollup-plugin-copy';
 import styles from "rollup-plugin-styler";
 import postCSSReplace from 'postcss-replace';
+import fs from 'node:fs/promises';
+ 
+
+const LEGACY_BUNDLE_ID = 'legacy-bundle'
+// Simple file concatenation plugin
+const concatenateFiles = (files) => ({
+  // checking ID as it's a virtual module
+  // https://rollupjs.org/plugin-development/#a-simple-example
+  resolveId(id) {
+    if (id === LEGACY_BUNDLE_ID) return id;
+  },
+  async load(id) {
+    if (id === LEGACY_BUNDLE_ID) {
+      const contents = await Promise.all(
+        files.map(file => fs.readFile(file, 'utf-8'))
+      );
+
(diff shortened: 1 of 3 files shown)


tombye (approved):
The bump just seems to bring in fixes so 👍🏻, and the shim rollup plugin is a nice solution to an annoying situation. I put in a code change suggestion, from reading up on rollup's plugin system but it's a nice-to-have tbh and shouldn't block approval.

rollup.config.mjs
```diff
@@ -1,9 +1,32 @@
...
+
+const LEGACY_BUNDLE_ID = 'legacy-bundle'
+// Simple file concatenation plugin
+const concatenateFiles = (files) => ({
+  // checking ID as it's a virtual module
+  // https://rollupjs.org/plugin-development/#a-simple-example
+  resolveId(id) {
+    if (id === LEGACY_BUNDLE_ID) return id;
+  },
+  load(id) {
```
tombye: Sorry, thought I'd added this to the review 🤦🏻.

Rollup's docs suggest [plugins should use asynchronous methods where possible](https://rollupjs.org/plugin-development/#conventions), so could this use Node's promise-based file access methods instead? Like:
```js
  async load(id) {
    if (id === LEGACY_BUNDLE_ID) {
      const contents = await Promise.all(
        files.map(file => fs.readFile(file, 'utf-8'))
      );

      return contents.join('\n');
    }
  }
```

Note: if you added these changes, line 6 would need to be changed to:
```js
import fs from 'node:fs/promises';
```

kr8n3r: nice. implemented

tombye (approved):
Thanks for making the changes. This looks good to go now 👍🏻

PR #5896, opened 2026-04-21
Remove `download` attribute from anchor tags

quis:
The `download` attribute forces the browsers to download the file, rather than try to display it.

This is fine and good, unless you get signed out in between viewing the page and clicking the link. In this case your browser will download the sign in page as `sign-in.html`. That’s confusing.

Files should still be forced to download because we are also setting the `Content-Disposition: attachment` HTTP header, which is [widely supported](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#browser_compatibility) and deferred to if the `download` attribute is not present.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -471,7 +471,7 @@ def inbox_download(service_id):
         mimetype="text/csv",
         headers={
             "Content-Disposition": (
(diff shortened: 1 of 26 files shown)


kr8n3r (approved):
good find.
described is likely a niche case, but changes work as expected

PR #5898, opened 2026-04-22
Don’t loop through all checkbox options to find `id`

quis:
Looks like we use the `id` of the first checkbox or radio in a set in order to work out where the error summary should link to.

Because we only ever care about the first item, we shouldn’t go through all the work of building a list of dictionaries for all the other items:
https://github.com/alphagov/notifications-admin/blob/f5ce6bc44d8cb8b3f8756c7fea7efdb4ef76b6df/app/main/forms.py#L689-L699

…items whose number can be many in the case of the templates page.

No idea if this will be a meaningful performance improvement but the code isn’t any worse so 🤷🏻‍♂️

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -700,10 +700,9 @@ class GovukCheckboxesField(GovukFrontendWidgetMixin, SelectMultipleField):
 
     @property
     def error_summary_id(self):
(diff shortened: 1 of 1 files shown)


tombye (approved):
Nice tidy up, and all still working fine 👍🏻

PR #5899, opened 2026-04-23
Refactor rendering of template checkboxes to build data structures in Python, not Jinja

quis:
This should be faster because:
- it’s in a lower-level language
- we can add logic to only compute the data for the checkboxes if the user has the `manage_templates` permission 

That said, I can’t produce a test case which shows this is faster.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -8,7 +8,7 @@ from math import ceil
 from numbers import Number
 from zipfile import BadZipFile
 
-from flask import request
+from flask import render_template, request
 from flask_login import current_user
 from flask_wtf import FlaskForm as Form
 from flask_wtf.file import FileAllowed, FileSize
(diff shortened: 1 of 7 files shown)


quis:
Test used for attempted profiling:

```python
def test_lots_of_templates_and_folders(
    client_request,
    mock_get_template_folders,
    mock_get_no_api_keys,
    service_one,
    mocker,
    active_user_view_permissions
):
    mocker.patch(
        "app.template_folder_api_client.get_template_folders",
        return_value=[
            _folder("folder", uuid.uuid4()) for _ in range(100)
        ]
    )
    mocker.patch(
        "app.service_api_client.get_service_templates",
        return_value={
            "data": [
                _template("sms", "sms_template_one") for _ in range(10_000)
            ]
        },
    )
    client_request.login(active_user_view_permissions)
    page = client_request.get("main.choose_template", service_id=SERVICE_ONE_ID, _test_page_title=False)
    assert len(page.select("#template-list a")) == 10_100
```

quis:
Closing this because it didn’t make the page load any faster. Mind boggling, but I guess Jinja is doing some dark magic…

PR #5901, opened 2026-04-24
Don’t allow filenames longer than 100 characters

quis:
Document download restricts filenames to 100 characters or fewer: https://github.com/alphagov/document-download-api/blob/cf8a33016d5184b743bc53cf6a200f29140a44d6/app/utils/validation.py#L40-L44

If we allow a file whose name is great than 100 characters, this will result in an error from document download at the time of sending.

This commit adds some form validation to prevent this.

Longer term, we should look to align/share all these rules between the apps.

***

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -3256,6 +3256,11 @@ class TemplateEmailFilesUploadForm(StripWhitespaceForm):
         if field.errors:
             return
 
+        if (length := len(field.data.filename)) > 100:
+            raise ValidationError(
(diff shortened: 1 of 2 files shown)


kr8n3r:
I would say 'File name ...' which is what we use in other instances, but non-blocking

kr8n3r (approved):

quis:
@kr8n3r good point, that’s consistent with https://github.com/alphagov/notifications-admin/blob/f5ce6bc44d8cb8b3f8756c7fea7efdb4ef76b6df/app/main/validators.py#L393

PR #5902, opened 2026-04-24
Updates with npm audit fix

kr8n3r:
Updates with npm audit fix


tombye (approved):
Same as with [doc download frontend for the dependencies](https://github.com/alphagov/document-download-frontend/pull/361/changes), all the bumps look sensible, and not too recent 👍🏻

PR #5903, opened 2026-04-24
Update the error messages for errors archiving a user or removing them from a service

klssmith:
We're changing the logic around when a user can be removed from a service or archived. Previously, we required there to be at least one active user with the `manage_settings` permission, but we're changing this requirement to two users with the permission, unless the service is in trial mode and hasn't made a go-live request.

This updates the code to have the new error messages we want to display. The error messages that the API will raise if a user can't be removed or archived will change, so we support both API error messages while the change is rolled out. When the API change is made, we can simplify the code here to only check for the new message.

[Trello card](https://trello.com/c/ZVJ7qdu1/1692-error-for-removing-a-penultimate-user-with-manage-settings-permissions)

---
# Screenshots

## Error when trying to archive a user
This will only be seen by platform admin users, since only they have the permission to archive a user
Screenshot 2026-04-28 at 11 24 35

## Error when trying to remove a user from a service - platform admin user version
Screenshot 2026-04-28 at 16 31 22

## Error when trying to remove a user from a service - standard user version
Screenshot 2026-04-28 at 16 30 57

### DIFF
--- a/app/main/views/find_users.py
+++ b/app/main/views/find_users.py
@@ -41,7 +41,8 @@ def archive_user(user_id):
         try:
             user_api_client.archive_user(user_id)
         except HTTPError as e:
-            if e.status_code == 400 and "manage_settings" in e.message:
+            new_msg = "User cannot be removed from a service"
+            if e.status_code == 400 and ("manage_settings" in e.message or new_msg in e.message):
                 flash(
                     "User can’t be removed from a service - "
                     "check all services have another team member with manage_settings"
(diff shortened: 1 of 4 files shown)


kr8n3r (approved):
follows a pattern we've used before for creating error messages that use this style

spatel033 (approved):

klssmith:
I've updated the content for the error messages for removing a user from a service, so there have been some small changes since this was approved. The screenshots have been updated and content should now match the latest versions @karlchillmaid

kr8n3r (approved):

PR #5904, opened 2026-04-25
`TemplateAndFoldersSelectionForm`: make rendering of `templates_and_folders` and `move_to` fields interruptible

risicle:
Using two new `Field` mixins - `InterruptibleChildRenderingNestedFieldMixin` which effectively wraps the `render_children` method to yield interruptibly every N calls. Not done with the `@interruptible_every` decorator as this way we can keep a separate counter for each instance and customise the `child_rendering_interruptible_every` at construction time.

`InterruptibleItemsFieldMixin` simply causes the `get_items_from_options` method to return an `InterruptibleIterableList` - a `list` subclass whose iterator is interruptible. This is not 100% foolproof because of how list objects can get replaced with vanilla ones by later processing steps, but it does seem to work correctly most the time.

I tracked this down and analyzed this by extending the sleep calls to print timestamps and used the `setitimer` signal to breakpoint long pauses between sleeps. Tested on a local machine with realistic data, the only remaining ~300ms pause appears to be during the escaping & re-indenting process performed on the final massive html generated.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -114,7 +114,7 @@ from app.utils.govuk_frontend_field import (
     render_govuk_frontend_macro,
 )
 from app.utils.image_processing import CorruptImage, ImageProcessor, WrongImageFormat
-from app.utils.interruptible_io import interruptible_iter
+from app.utils.interruptible_io import InterruptibleIterableList, interruptible_iter
 from app.utils.user_permissions import (
     all_ui_permissions,
     organisation_user_permission_names,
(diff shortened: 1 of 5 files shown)


quis (commented):
I think this should have a test which renders the template page with ~100 templates and a bunch of nested folders, and checks that `sleep` is called the expected number of times.

I would be so easy for a change in either our code or the `govuk-frontend-jinja` package, or GOV.UK Frontend itself to nullify this work.

risicle:
That's a nice idea... let me see how easy it is.

quis (approved):

PR #5905, opened 2026-04-27
Remove jQuery from focusBanner

kr8n3r:
When we rewrote it to ESM we still left jQuery in there. As the goal is to remove it entirely, this PR replaces it with vanilla Javascript.

If the `update-content` module inserts html that has or contains an element with a target class, this script will focus it.

We listen for the `blur` event (as it looses focus) The `blur` event listener automatically gets removed when invoked with `once: true` optional parameter.

FocusBanner is triggered by when `updateContent.onafterupdate` event is dispatched by the `update-content` script.

This is (was?) used when verifying reply-to email address.

### DIFF
--- a/app/assets/javascripts/esm/all-esm.mjs
+++ b/app/assets/javascripts/esm/all-esm.mjs
@@ -148,4 +148,4 @@ if ($updateContentBlocks.length > 0) {
   });
 };
 
-const focusBanner = new FocusBanner();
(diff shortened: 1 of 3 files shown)


tombye (approved):
Nice work, and I think more readable. I like how you find any instances of focus banner nodes and the `once: true` on the event listener is 🤌. 👍🏻

PR #5916, opened 2026-05-06
Replace timeago jquery plugin with our own UpdateRelativeTime module 

kr8n3r:
## Why
Created as a replacement for [timeago jquery plugin](https://timeago.yarp.com/) we used before as we are moving to ES Modules and removing jQuery as a dependency.

## What
Purpose of this plugin is to update targeted `<time>` elements with how much relative time has passed since an event occurred.

It reads the element's `datetime` attribute and

- creates a title attribute with readable date and time of the event
- updates element's text with language-sensitive relative time format, like '5 minutes ago'.
- runs on a 60 second interval and keeps updating the text if required.

For relative time text format it uses [`Intl.RelativeTimeFormat` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) - a native Javascript feature for localised text.

`Intl.RelativeTimeFormat handles plurals` ('day(s)'), past vs future prefix/suffix ('in'...'ago') as well as the style ('1 day ago' vs 'yesterday').

It contains a set of thresholds that determine which units to display in what scenario - seconds vs minutes, minutes vs hours, etc.

## Visual changes

When hovering over the element it now accurately shows the human readable date and time of the event

**Before**
before-tag-hover

**After**
after-tag-hover

## Test
Used on the following pages

- /users/<userID> - last logged in
- /services/<serviceID>/service-settings/receive-text-messages/stop  - if inbound number present
- /services/<serviceID>/templates/<templateID> - if template was edited before
- /services/<serviceID>/api

### DIFF
--- a/app/assets/javascripts/esm/update-relative-time.mjs
+++ b/app/assets/javascripts/esm/update-relative-time.mjs
@@ -1,0 +1,117 @@
+import { isSupported } from 'govuk-frontend';
+
+// This new way of writing Javascript components is based on the GOV.UK Frontend skeleton Javascript coding standard
+// that uses ES2015 Classes -
+// https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/js.md#skeleton
+//
+// It replaces the previously used way of setting methods on the component's `prototype`.
+// We use a class declaration way of defining classes -
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/class
+//
+// More on ES2015 Classes at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes
+
+// inspired by https://polypane.app/blog/the-intl-api-the-best-browser-api-youre-not-using/
+class UpdateRelativeTime {
+  // creating formatters is expensive so we create them once and reuse
+  static relativeFormatter = new Intl.RelativeTimeFormat('en-GB', { numeric: 'auto' });
+  static fullDateFormatter = new Intl.DateTimeFormat('en-GB', { dateStyle: 'long', timeStyle: 'short' });
+
+  constructor(selector) {
+    if (!isSupported()) {
+      return this;  
+    }
+
+    this.selector = selector;
+
+    this.init();
+  }
+
+  init() {
+    this.updateExistingElements();
+    
+    // start global 60s interval
+    setInterval(() => this.updateExistingElements(), 60000);
+  }
+
+  updateExistingElements() {
+    const $elements = document.querySelectorAll(this.selector);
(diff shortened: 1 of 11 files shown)


quis:
Would be nice to harmonise with this code for short intervals (0–1 mins):
https://github.com/alphagov/notifications-admin/blob/5b9eff02b8548a237984ec5b465f5b80dbb59f6f/app/formatters.py#L124-L127

I worry if the seconds are ticking away it might be a bit distracting.

kr8n3r:
> Would be nice to harmonise with this code for short intervals (0–1 mins):
> 
> https://github.com/alphagov/notifications-admin/blob/5b9eff02b8548a237984ec5b465f5b80dbb59f6f/app/formatters.py#L124-L127
> 
> I worry if the seconds are ticking away it might be a bit distracting.

update script runs on every 60 seconds so seconds don't tick away. `RelativeTimeFormat` doesn't do strings like 'in the last minute" but It does "now". Current threshold for when it switches text from "seconds" to "minute" is 45s

if we want to match the formatters, then I can look at custom code

EDIT: Done


Screenshot 2026-05-08 at 11 18 46
Screenshot 2026-05-08 at 11 19 01
Screenshot 2026-05-08 at 11 21 17

app/assets/javascripts/esm/update-relative-time.mjs
```diff
@@ -0,0 +1,117 @@
...
+    const now = new Date();
+    $elements.forEach(($el) => this.updateTextString($el, now));
+  }
+
+  updateTextString(element, currentTime) {
+    const elementDatetimeStamp = element.getAttribute('datetime');
+    // don't run if we forgot to include the attribute
+    if (!elementDatetimeStamp) return;
+
+    const date = new Date(elementDatetimeStamp);
```
tombye: Very minor point but I found this a bit tricky to read as `date`, because it didn't tell me anything about its role in the logic. Once I read it as `elementTime` in `calculateRelativeTimeString`, it all made much more sense.

tombye (approved):
Really enjoyed reading this code! Learnt loads about the problem space and this really does show how clear it models the logic. I had one tiny comment on a variable name, because it threw me out for a sec but this is top work and ready to go 👍🏻

app/assets/javascripts/esm/update-relative-time.mjs
kr8n3r: yeah, always in two mind naming the internal variable the same as the function argument, but you're right it doesn't immediately scan what it is so renamed it

tombye (approved):
Thanks for making the change 👍🏻

PR #5917, opened 2026-05-06
Show "Trial mode services" page to users who can self-approve

klssmith:
The "Trial mode services" organisation page was platform admin only, but we want to show this to organisation team members who have the permission to approve go-live requests. Organisation users without this specific permission won't see the page or any links to it. As well as changing permissions for who can view the page, the permissions for who can see the link to the page in the left hand nav have also been changed.

[Trello card](https://trello.com/c/7VpLeDV5/1782-make-a-version-of-trial-services-page-visible-to-organisations-with-self-approve-enabled-6)

### DIFF
--- a/app/main/views/organisations/index.py
+++ b/app/main/views/organisations/index.py
@@ -12,6 +12,7 @@ from app import (
     org_invite_api_client,
     organisations_client,
 )
+from app.constants import PERMISSION_CAN_MAKE_SERVICES_LIVE
 from app.main import main
 from app.main.forms import (
     AddGPOrganisationForm,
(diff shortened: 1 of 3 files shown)


app/main/views/organisations/index.py
```diff
@@ -200,8 +201,13 @@ def download_organisation_usage_report(org_id):
 
 
 @main.route("/organisations/<uuid:org_id>/trial-services", methods=["GET"])
-@user_is_platform_admin
+@user_has_permissions()
```
quis, line 204: Might be nicer if this decorator took an `org_permissions` argument. Could do that as a refactoring PR though

quis (approved):

app/main/views/organisations/index.py
klssmith: Agree, I'll look at doing that after this is merged

klssmith: Now done in https://github.com/alphagov/notifications-admin/pull/5927

PR #5920, opened 2026-05-12
Add canEncode validator for use in callback urls and bearer tokens

whpearson:
To validate that we can encode with latin-1.

This is needed to stop unicode errors we are seeing in notifications-api when attempting
to send to these callback urls

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -75,6 +75,7 @@ from app.formatters import (
     sentence_case,
 )
 from app.main.validators import (
+    CanEncode,
     CannotContainURLsOrLinks,
     CharactersNotAllowed,
     CommonlyUsedPassword,
@@ -2532,11 +2533,16 @@ class CallbackForm(StripWhitespaceForm):
                 r"(?:#[\w\-._~%!$&'()*+,;=:@/?]*)?$",
                 message="Must be a valid https URL",
             ),
+            CanEncode(field_type="a web address"),
         ],
     )
     bearer_token = GovukPasswordField(
         "Bearer token",
-        validators=[DataRequired(message="Cannot be empty"), Length(min=10, thing="the bearer token")],
+        validators=[
+            DataRequired(message="Cannot be empty"),
+            Length(min=10, thing="the bearer token"),
+            CanEncode(field_type="a bearer token"),
+        ],
     )
 
     def validate(self, *args, **kwargs):
--- a/tests/app/main/test_validators.py
+++ b/tests/app/main/test_validators.py
@@ -5,6 +5,7 @@ from notifications_utils.recipient_validation.errors import InvalidPhoneError
 from wtforms import ValidationError
 
 from app.main.validators import (
+    CanEncode,
     CharactersNotAllowed,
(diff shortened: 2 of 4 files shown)


app/main/forms.py
```diff
@@ -2533,6 +2533,7 @@ class CallbackForm(StripWhitespaceForm):
                 r"(?:#[\w\-._~%!$&'()*+,;=:@/?]*)?$",
                 message="Must be a valid https URL",
             ),
+            CanEncode(message="Urls have to be capable of being encoded in latin-1"),
```
karlchillmaid:
```suggestion
            CanEncode(message="URL cannot include Unicode characters"),
```

quis: I don‘t know how much we care about technical correctness here but URLs can contain Unicode characters. Unicode contains basically all characters, including boring ones like `a` or `1`. 

When people try to use emoji in text messages we don’t get into the weeds about character sets. Instead we tell them which characters are the problem:
https://github.com/alphagov/notifications-admin/blob/06fe9c311e469d8b8a063c4e4726142b1de5c061/app/main/validators.py#L166-L171

whpearson: Unicode characters have to be percent encoded, so perhaps that should be the correct error message? Happy to highlight the characters that should be.

quis: Yeah, if the error message was something like

> 🤪 and ŵ must be percent-encoded in URLs

that would be better

whpearson: "You cannot use the following characters in URIs. These characters, ∆ or 📲, might be mis-encoded."

I kept close to the sms error. What do you think @quis @karlchillmaid ?

karlchillmaid: How about this?

> You cannot use ∆ in a web address. You must use percent encoding if you want to include this character in a URL.

> You cannot use ∆ or 📲 in a web address. You must use percent encoding if you want to include these characters in a URL.

whpearson: I went with calling it a web address in in both places (for ease of coding/consistency)

whpearson: Removed the repetition now. So it is consistent with @karlchillmaid 's advice (in slack)

tests/app/main/test_validators.py
```diff
@@ -217,3 +218,39 @@ def test_string_cannot_contain_string_with_custom_error_message():
 
     assert str(error.value) == "No sequences please"
     assert mock_field.error_summary_messages == ["No sequences in %s please"]
+
+
+@pytest.mark.parametrize(
```
quis, line 223: These tests are good, but they only check that the `CanEncode` validator works as expected. They don’t test that it’s being used on the bearer token field.

whpearson: Pushed up a version that checks whether CallbackForms have the CanEncode validators attached to the fields. I'd rather not test the functionality twice (to avoid duplication of effort if it changed).

I couldn't find examples of this kind of code (our forms tests are quite small), so pointers on how to do it properly appreciated.

quis (approved):

PR #5924, opened 2026-05-15
Rewrite sticky JS as an ES Module

tombye:
## What

Trello ticket: https://trello.com/c/lGwCAXFu/1789-modernise-our-javascript-sticktowindowwhenscrolling

Rewrite does not follow the conventions laid down by previous work like this, which operate on DOM nodes sent in as parameters. It does a lot of work to find and cache elements, but also their position and dimension so seemed like a less suitable fit for those conventions.

image

## How to review

First, read up on how it works. There's [a presentation explaining it](https://docs.google.com/presentation/d/1h0OKdRjAOBTUlPB8viPSafUTxQ5RQEw54DLXEh6OsSk/edit?usp=sharing) and some diagrams in [this folder](https://drive.google.com/drive/folders/1Q1tD7A2F6ayBMQ7ojtRLCiPRnzHUJPdf?usp=drive_link).

Go commit-by-commit.


rollup.config.mjs
```diff
@@ -1,10 +1,11 @@
 import { nodeResolve } from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
```
kr8n3r, line 2: sorry, I took this out not long ago :(

kr8n3r (changes requested):
this code does a lot, but the way this is now structured, making use of Classes, private fields and methods it makes it much easier to read.

Works in the browser as expected.

What's missing is the `isSupported` check from govuk-frontend we use to prevent the script from running in non-supported browsers.  I _think_ this will need to go into the 
`stickAtTopWhenScrolling` and `stickAtTopWhenScrolling`

rollup.config.mjs
tombye: We didn't need it before so makes sense!

kr8n3r (approved):
thank you 🎉 
🚢

PR #5925, opened 2026-05-15
error handlers: use "failsafe" mechanism to avoid risky operations rendering error pages

risicle:
Specifically, this prevents a request being made (either to redis or the api) by flask-login's template context processor.

When we're rendering error pages, we've got to remember we're doing so in an error handler, and any further exceptions won't get caught and handled properly - resulting in an ugly and difficult to decipher error (both for the user and in our logs).

Here, we set a request-local flag to indicate this, and make our `user_loader` reference this before continuing to any possible requests. This is the only place I've found such a request being made in the basic rendering path, but there could be more.

Currently only using this for error pages that are likely to have arisen from availability problems as some error pages want to be able to show a full nav bar with service- and user-contextual information. They'll just have to risk it for now.

This fell out of testing performed for https://trello.com/c/PEpdGAPE/1518-implement-solution-for-failed-writes-deletes-to-redis

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -304,6 +304,11 @@ def reset_memos():
 
 @login_manager.user_loader
 def load_user(user_id):
+    if g.get("failsafe"):
+        # avoid doing anything that may result in a network request and therefore another exception
+        # (sadly we're not in charge of flask-login's context processor so can't stop such attempts
+        # there)
+        return None
     return User.from_id(user_id)
 
 
@@ -420,7 +425,7 @@ def useful_headers_after_request(response):
 
 
 def register_errorhandlers(application):  # noqa (C901 too complex)
-    def _error_response(error_code, error_page_template=None):
+    def _error_response(error_code, *, error_page_template=None, failsafe=False):
         template_file = f"error/{error_code}.html"
 
         if error_code == 404 and request.view_args and "service_id" in request.view_args and g.current_service:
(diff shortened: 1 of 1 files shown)


risicle:
I would add a test for this, but it appears we have no unit tests that exercise our `user_loader` function.

app/__init__.py
```diff
@@ -304,6 +304,11 @@ def reset_memos():
 
 @login_manager.user_loader
 def load_user(user_id):
+    if g.get("failsafe"):
+        # avoid doing anything that may result in a network request and therefore another exception
+        # (sadly we're not in charge of flask-login's context processor so can't stop such attempts
+        # there)
+        return None
```
quis, line 311: [Docs for `user_loader`](https://flask-login.readthedocs.io/en/latest/#how-it-works) say it should:
> return `None` (not raise an exception) if the ID is not valid

So I think this will correctly set `current_user` to an instance of [`AnonymousUser`](https://github.com/alphagov/notifications-admin/blob/16fef2fcd7c61cf924f6f12330c5423148da960c/app/models/user.py#L689-L703), which implements `.is_authenticated`, `.platform_admin`, etc.

quis (approved):
Almost all the tests hit `user_loader` in passing because the [test client](https://flask-login.readthedocs.io/en/latest/#flask_login.FlaskLoginClient) sets up the session cookie with the current user’s ID.

That said, I also couldn’t find a way to write a test which raised an exception in such a way that it would cause the problem you are trying to mitigate.

risicle:
Yeah I tried dropping an exception in `user_loader` but nothing in the tests raised it.

PR #5931, opened 2026-05-20
Update the team members and permissions guidance

karlchillmaid:
Changes:

* state that 2 team members need the ‘manage settings’ permission
* explain the role and responsibilities of the ‘manage settings’ permission


tombye:
This is what it looks like:
image

tombye (approved):
Really nice work. So good to see this all made more public and accessible.

PR #5934, opened 2026-05-21
Enable uploaded files to use document download api file checks

BlessedDev:
This PR enables the admin app to use the document download API's `/antivirus-and-mimetype-check` endpoint.
This was a better option than creating a new document download API's endpoint and unpicking bits of that apps current structure.
As a result the admin will perform filename validation checks and then hand off the file contents for secondary validation checks and antivirus scans by the endpoint. This ensures that the same validation is carried on all files sent via email wether it is via the ui or the API.
For now the preference is to use the logic in the admin app to determine MIME Types even though it is now been returned by the endpoint.
It will be helpful to review this PR commit by commit and I have been testing on `dev-d`.

🚨 This PR is dependant on https://github.com/alphagov/notifications-aws/pull/2984 being merged in first. 🚨


app/s3_client/s3_template_email_file_upload_client.py
```diff
@@ -11,4 +11,5 @@ def upload_template_email_file_to_s3(data, file_location):
         bucket_name=current_app.config["S3_BUCKET_TEMPLATE_EMAIL_FILES"],
         file_location=file_location,
         metadata=metadata,
+        content_type=mimetype,
```
quis: Are we going to read this value anywhere?

app/main/forms.py
```diff
@@ -3279,15 +3280,11 @@ def __init__(self, *args, template, **kwargs):
...
-    file = VirusScannedFileField(
+    file = FileField(
         "Add a file",
         validators=[
             DataRequired(message="You need to upload a file to submit"),
-            DocumentDownloadFileValidator(),
-            FileSize(
-                max_size=2 * 1_024 * 1_024,
-                message="The file must be smaller than 2MB",
-            ),
```
quis, line 3290: I think it’s worth keeping these so we can skip sending known-bad files to document download

tests/app/main/views/test_template_email_files.py
```diff
@@ -1041,12 +1043,9 @@ def test_upload_file_returns_error_if_filename_is_too_long(
 
     if expected_status < 300:
         assert normalize_spaces(page.select_one(".govuk-error-message").text) == (
-            "File name cannot be longer than 100 characters (‘"
-            "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.pdf’ "
-            "is 101 characters)"
+            "`filename` cannot be longer than 100 characters"
```
quis: This error message is less useful

BlessedDev: I had looked at it to synchronise with the document download api. However I take your point and will also add a card to fix the bug where a filename longer than 100 characters is displayed beyond the boundaries of o the error message box.

app/s3_client/s3_template_email_file_upload_client.py
BlessedDev: As discussed, I will roll this back in favour of using the current logic for determining the mime type in admin.
We can revisit this in the future if we want a tighter integration with the document downloads api.

tests/app/main/views/test_template_email_files.py
BlessedDev: @kr8n3r has already fixed the bug https://github.com/alphagov/notifications-admin/pull/5940

tests/app/notify_client/test_document_download_api_client.py
```diff
@@ -0,0 +1,51 @@
...
+import responses
+
+from app.notify_client.document_download_api_client import DocumentDownloadError, document_download_api_client
+
+
+def test_document_download_api_file_check(client_request, requests_mock):
+    service_id = str(uuid.uuid4())
+    client = document_download_api_client
+    file_content = b"%PDF-1.4 test content"
+    expected_url = f"{client.base_url}/services/{service_id}/antivirus-and-mimetype-check"
```
quis: Better to hard-code the domain here so if something breaks the client looking up the value from the config we will know about it:
```suggestion
    expected_url = f"http://localhost:7000/services/{service_id}/antivirus-and-mimetype-check"
```

tests/app/notify_client/test_document_download_api_client.py
```diff
@@ -0,0 +1,51 @@
...
+            "this_is_a_very_long_filename_that_exceeds_one_hundred_characters_for_testing_validation_boundaries_ok.pdf",
+            "`filename` cannot be longer than 100 characters",
+            400,
+        ),
+        (
+            "test_file.pdf",
+            "The file must be smaller than 2MB",
+            413,
+        ),
+    ],
```
quis, line 38: Can we add a test (as a param of this test or a separate test) for when:
- `response` is `None`
- `response.status` is 500

In other words a test which makes sure we are only catching `400` and `413` errors

quis (commented):
Generally looks good, have left a couple of comments with suggested improvements

tests/app/notify_client/test_document_download_api_client.py
BlessedDev: Yes, that is a good point. I will add tests for those cases.

quis (approved):
Good to go once the AWS PR is merged

PR #5941, opened 2026-05-28
Refactor character counting message into Jinja

quis:
Jinja is a better place to be composing and formatting strings into output text.

It means if we want to make this output richer in the future we have a more flexible place to do so.

***

Also makes use of the new `SMSPreviewTemplate.count_of_characters_above_limit` property from https://github.com/alphagov/notifications-utils/pull/1362


CrystalPea (approved):
Looks good  👍🏼

PR #5944, opened 2026-05-29
Add prompts telling editors how to reduce the cost of their text messages

quis:
## One fragment

image

## Multiple fragments

image

## One fragment with a non-GSM-7 character

image

## One fragment with multiple non-GSM-7 characters

image

## Multiple fragments with multiple non-GSM-7 characters

image

### DIFF
--- a/app/templates/partials/templates/content-count-message.html
+++ b/app/templates/partials/templates/content-count-message.html
@@ -6,6 +6,20 @@
     {% if template.placeholders %}
       (not including personalisation)
     {% endif %}
+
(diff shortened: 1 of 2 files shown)


kr8n3r:
I know this is still in draft, but are we removing "(not including personalisation)" as we can't tell how many chars personalisation will take up, so the prompt on the edit screen might be a bit misleading.

I like the the prompt changes 👍

quis:
@kr8n3r example when the message is personalised:

image

kr8n3r (approved):
great work 🎆

PR #5946, opened 2026-06-01
Add custom 404 for notification not found

whpearson:
This adds a custom 404 page if the page erroring has a notification_id request.

We are adding this because we might often get this if a user has bookmarked or remembered a notification.

So we want to be clear about the likely cause of this.

Screenshot 2026-06-02 at 14 03 30

### DIFF
--- a/app/templates/error/404-notifications-page.html
+++ b/app/templates/error/404-notifications-page.html
@@ -1,0 +1,26 @@
+{% extends "withnav_template.html" %}
+{% block per_page_title %}Page not found{% endblock %}
+{% block maincolumn_content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-five-sixths">
+      <h1 class="heading-large">
+        Page not found
+      </h1>
+      <p class="govuk-body">
+        If you typed the web address, check it is correct.
+      </p>
+      <p class="govuk-body">
+        If you pasted the web address, check you copied the entire URL.
+      </p>
+      <p class="govuk-body">
+       If the web address is correct, the page you’re looking for may no longer exist.
+       </p>
+       <p class="govuk-body">
+       For security, GOV.UK Notify only keeps a temporary record of the emails, text messages and letters you send.
+       </p>
+       <p class="govuk-body">
+       For more information, see: <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_data_retention_period') }}">Data retention period</a>
+      </p>
+    </div>
+  </div>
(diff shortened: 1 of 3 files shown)


spatel033 (approved):

quis:
Has this had a content designer look at it? The new message is quite technical (we don’t tend to refer to ‘notifications’ when talking to users of the website)

whpearson:
> Has this had a content designer look at it? The new message is quite technical (we don’t tend to refer to ‘notifications’ when talking to users of the website)

I'm going to ask @karlchillmaid to have a look.

app/templates/error/404-notifications-page.html
```diff
@@ -0,0 +1,20 @@
...
+      </h1>
+      <p class="govuk-body">
+        If you typed the web address, check it is correct.
+      </p>
+      <p class="govuk-body">
+        If you pasted the web address, check you copied the entire address.
+      </p>
+      <p class="govuk-body">
+        This might be a web address for an expired notification. We have a default <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_data_retention_period') }}">data retention period</a> of 7 days. If you think it shouldn’t be expired <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.support') }}">contact us</a>.
+      </p>
```
karlchillmaid:
```suggestion
        If you pasted the web address, check you copied the entire URL.
      </p>
      <p class="govuk-body">
       If the web address is correct, the page you’re looking for may no longer exist.
       </p>
       <p class="govuk-body">
       For security, GOV.UK Notify only keeps a temporary record of the emails, text messages and letters you send.
       </p>
       <p class="govuk-body">
       For more information, see: <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.guidance_data_retention_period') }}">Data retention period</a>
      </p>
```

karlchillmaid: Is the screenshot accurate in showing the main and left-hand navigation menus?

If so, that’s great, as it will give users a way to reorientate themselves.

whpearson:
It looks like this 
Screenshot 2026-06-04 at 12 45 28
Yup. The navigation is still there. I borrowed from @quis 's work on errors in services

quis:
I think this will cause a `500` if the service doesn’t exist (for example it’s been archived)

whpearson:
> I think this will cause a `500` if the service doesn’t exist (for example it’s been archived)

Good spot

quis (approved):

PR #5947, opened 2026-06-03
Replace window.GOVUK.stickAtBottomWhenScrolling.recalculate with module import

kr8n3r:
We have modernised `stickAtBottomWhenScrolling` into ESM so we can now directly import it where it's needed and no longer have to rely on global scope.

This updates instance in other ES modules and updates the tests to mock this dependency


tombye (approved):
Nice work. Very clear to read and understand 👍🏻

PR #5948, opened 2026-06-03
Fix DOM text reinterpreted as HTML in collapsible-checkboxes

kr8n3r:
Seeing if this will resolve https://github.com/alphagov/notifications-admin/security/code-scanning/1


tombye (approved):
Really nice solution 🤌

PR #5950, opened 2026-06-04
Fix DOM text reinterpreted as HTML in file-upload script

kr8n3r:
Resolve https://github.com/alphagov/notifications-admin/security/code-scanning/2

by replacing the use of `insertAdjacentHTML` with element creation and use of `insertAdjacentElement` and `appendChild`

### DIFF
--- a/app/assets/javascripts/esm/file-upload.mjs
+++ b/app/assets/javascripts/esm/file-upload.mjs
@@ -59,28 +59,38 @@ class FileUpload {
     const buttonText = this.$field.dataset.buttonText;
     const fieldId = this.$field.getAttribute('id'); // copy the id across so error links work
     const oldFieldId = `hidden-${fieldId}`;
-    let buttonHTMLStr = `
-        <button type="button" class="file-upload-button govuk-button govuk-!-margin-right-1" id="${fieldId}">
-          ${buttonText}
-        </button>`; // Styled as a submit button to raise prominence. The type shouldn't change.
+    
+    const $button = document.createElement('button');
+    $button.setAttribute('type', 'button');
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/esm/file-upload.mjs
```diff
@@ -59,28 +59,38 @@ class FileUpload {
...
+    this.$field.insertAdjacentElement('afterend', $button);
+
+    if (formErrors) {
+      const $label = document.createElement('label');
+      $label.classList.add('file-upload-button-label', 'govuk-error-message');
+      $label.setAttribute('for', fieldId);
+
+      const $span = document.createElement('span');
+      $span.classList.add('govuk-visually-hidden');
+      $span.textContent = `${buttonText} `;
```
tombye: Does this need to be in a template literal? If so, does it need the trailing space?

tombye (commented):
One tiny thing I spotted. Otherwise this is a clever solution and nice to read. Happy to approve once that's addressed.

app/assets/javascripts/esm/file-upload.mjs
kr8n3r: Edit: it does need it. original template literal had space between the text and span text that gets added after. https://github.com/alphagov/notifications-admin/pull/5950/changes#diff-a2ce320ab588c6ab30a85287cc52af1ca37e43a41ece2bd0cb84ccc7865beebdL74

resulting in

```
  Expected: "Upload logo The logo must be a PNG file"
  Received: "Upload logoThe logo must be a PNG file"
```

tombye: Thanks for explaining.

tombye (approved):
Thanks for answering my comment. Looks good to go now (I mean, it was before but still) 👍🏻

PR #5951, opened 2026-06-04
Tidy up after improvements to fragment counter

quis:
## Use `max_items_shown` argument of `formatted_list` 

Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/1365

## Remove prompt about non-GSM characters if message is already 1 fragment 

There’s no way to make the message cheaper at this point.


kr8n3r (approved):
nice

PR #5953, opened 2026-06-05
Fix DOM text reinterpreted as HTML in enhanced-textbox

kr8n3r:
This is to avoid any potential XSS vectors of attack.

Resolves
https://github.com/alphagov/notifications-admin/security/code-scanning/3
https://github.com/alphagov/notifications-admin/security/code-scanning/4

`contentReplaced` method was updated to now create a `DocumentFragment` look through string of characters in the textarea and wrap any placeholders in `span` depending on their type (regular vs optional). It then returns original text (without placeholders) and placeholder values.

`contentEscaped` method can be removed as we now apply text that doesn't contain any placeholders directly in the `update` method.

`update` method was updated to use `.textContent` to apply raw text as the browser substitues < and > characters to harmless &lt; and &gt;

For text with placeholders we use `.append` which works in the same way as `.textContent`.

### DIFF
--- a/app/assets/javascripts/esm/enhanced-textbox.mjs
+++ b/app/assets/javascripts/esm/enhanced-textbox.mjs
@@ -73,27 +73,51 @@ class EnhancedTextbox {
     stickAtBottomWhenScrolling.recalculate();
   }
 
-  contentEscaped () {
-    const $el = document.createElement('div');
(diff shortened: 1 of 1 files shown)


tombye (approved):
Nice changes, clear to read and bonus performance points for using a document fragment 👍🏻

PR #5956, opened 2026-06-08
Fix DOM text reinterpreted as HTML in radio-select

kr8n3r:
Fixes https://github.com/alphagov/notifications-admin/security/code-scanning/5

To avoid any potential XSS attack we're replacing use of `innerHTML` with `replaceChildren` and `appendChild`.

To achieve that we're creating each node separately with `createElement` instead of using a block of Template literal.

`getInitialHTML` was renamed to `buildInitialHTML` as it's more descriptive which also means that obtaining an existing `componentLabelText` was moved out of this method into the constructor.

### DIFF
--- a/app/assets/javascripts/esm/radio-select.mjs
+++ b/app/assets/javascripts/esm/radio-select.mjs
@@ -31,6 +31,7 @@ class RadioSelect {
       };
     });
     this.componentName = this.$module.querySelector('input[type=radio]').getAttribute('name');
+    this.componentLabelText = this.$module.previousElementSibling.textContent.trim();
     this.$module.querySelectorAll('label').forEach(function (label, idx) {
       let labelText = label.textContent.trim();
       let relatedRadio = label.previousElementSibling;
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/esm/radio-select.mjs
```diff
@@ -55,14 +56,14 @@ class RadioSelect {
...
-      'componentLabel': this.$module.previousElementSibling.textContent.trim(),
+    const initialHTML = this.buildInitialHTML({
       'componentName': this.componentName,
       'selectedTime': this.selectedTime,
       'days': this.days,
       'times': this.timesByDay
-    });
+    }, this.componentLabelText);
+
+    this.$module.replaceChildren(initialHTML);
```
tombye: I think [replaceChildren](https://caniuse.com/mdn-api_element_replacechildren) is newer than [noModule](https://caniuse.com/mdn-api_htmlscriptelement_nomodule) (the current minimum browser version test we use). But `initialHTML` will always be a document fragment so could we instead do something like:
```js
this.$module.innerHTML = '';
this.$module.append(initialHTML);
```
?

tombye (changes requested):
Spotted one small thing but otherwise this looks good and is nice to read. Happy to approve once that comment is addressed.

app/assets/javascripts/esm/radio-select.mjs
kr8n3r: yeah, though support is good, happy to extend it further back with 
```
this.$module.textContent = ''; 
this.$module.append(initialHTML)
```

tombye (approved):
Thanks for making the change. Now good to go 👍🏻

PR #5958, opened 2026-06-09
Remove jQuery from ListEntry

kr8n3r:
- Replaces 2 instances of jQuery methods with native Javascript methods.
- For consistency, rename ListEntry to list-entry to match other test file names.


tombye (approved):
Looks good to me, and works as expected 👍🏻

PR #5960, opened 2026-06-10
Consolidate Javascript utils

kr8n3r:
Consolidate into a single file and update imports

Check individual commits for grouped changes.


tests/javascripts/update-content.test.mjs
```diff
@@ -21,12 +21,12 @@ describe('Update content', () => {
...
     jest.resetModules();
 
-    jest.unstable_mockModule('../../app/assets/javascripts/utils/location.mjs', () => ({
+    jest.unstable_mockModule('../../app/assets/javascripts/esm/utils.mjs', () => ({
       locationReload: jest.fn()
     }));
 
     const updateContentModule = await import('../../app/assets/javascripts/esm/update-content.mjs');
-    const locationUtilModule = await import('../../app/assets/javascripts/utils/location.mjs');
+    const locationUtilModule = await import('../../app/assets/javascripts/esm/utils.mjs');
```
quis: Should probably rename these variables to match (here and in other instances)
```suggestion
    const utilsModule = await import('../../app/assets/javascripts/esm/utils.mjs');
```

quis (approved):

PR #5962, opened 2026-06-11
Announce upcoming letter price changes

saimaghafoor:
Add announcement to letter pricing page for upcoming economy mail pricing changes.  Some prices are going down and some going up which is why the banner mentions a 'change'

### DIFF
--- a/app/templates/views/guidance/pricing/letter-pricing.html
+++ b/app/templates/views/guidance/pricing/letter-pricing.html
@@ -20,7 +20,7 @@
     }
   ) }}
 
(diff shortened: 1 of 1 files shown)


app/templates/views/guidance/pricing/letter-pricing.html
quis: Set this to 30 June to make the message show:
```suggestion
  {% set date_new_letter_rates_go_live = "2026-06-30" %}
```

It will automatically hide if the rates in the database are updated any time after 30 June (doesn’t need to be an exact match).

quis (approved):

quis (approved):

PR #5964, opened 2026-06-15
Update govuk-frontend to v6.2.0

kr8n3r:
[govuk-frontend 6.2.0 release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v6.2.0)

## What and why

This updates the library which brings in the use of [@use and @forward in Sass](https://sass-lang.com/documentation/at-rules/use/) as a replacement for `@import`. 

As `@import` will be deprecated, we need to migrate to `@use` ourselves in our codebase, which is being done here.

Changes here are minimal, but they do move some things around.  Personal choices, so happy to be challenged on them.

## Reviewing

Changes have been broken down by logical commits, so hopefully reviewing each commit will be easier.
I've checked major user flows for any visual glitches.

[Diff between 6.1 with @import and v6.20 with @use unminified](https://www.diffchecker.com/0hAYLNZT/)


tombye (approved):
It takes a while to get your head round `@use` and `@forward` but the commits helped and this all looks good to me. I also checked all the core journeys and didn't find anything wrong 👍🏻

PR #5965, opened 2026-06-16
Add DOCUMENT_DOWNLOAD_API_HOST_NAME_INTERNAL to the Dockerfile

spatel033:
Add missing `DOCUMENT_DOWNLOAD_API_HOST_NAME_INTERNAL` to the dockerfile


quis (commented):
Do you also need `DOCUMENT_DOWNLOAD_API_HOST`, per:
https://github.com/alphagov/notifications-admin/blob/9d64dfa33030592e7840b9d28c3667f464cdc07e/app/config.py#L147-L150
?

Or if that variable is unused, can we delete it?

PR #5967, opened 2026-06-17
Update the ‘Confirm your service is unique’ guidance

karlchillmaid:
One of the mandatory tasks we make users complete before they go live is ‘Confirm your service is unique’.

We introduced this page as part of the ‘onboarding’ epic. It was good enough for MVP, but now it needs to change.

It’s biggest weaknesses are:

* the concept of a ‘unique service’ is abstract
* we don’t explain why we want your service to be ‘unique’ (the free allowance)
* there’s a disconnect between the title/h1 and the interaction (text field, green button)

The new page design:

* makes a stronger connection between the title/h1 and the interaction
* explains that this is all about the free allowance

It’s not perfect, but it’s a reasonable first step and sets us up well for the work we’ll do in Q2.

This PR also changes the link on the ‘Make your service live’ task list page, so it matches the new h1.

### DIFF
--- a/app/templates/views/confirm-your-service-is-unique.html
+++ b/app/templates/views/confirm-your-service-is-unique.html
@@ -5,7 +5,7 @@
 {% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
 
 
-{% set title = "Confirm that your service is unique" %}
+{% set title = "Confirm your service name" %}
 
 {% block service_page_title %}
     {{ title }}
--- a/app/templates/views/confirm-your-service-is-unique.html
+++ b/app/templates/views/confirm-your-service-is-unique.html
@@ -22,25 +22,38 @@
         <div class="govuk-grid-column-five-sixths">
 
             <p class="govuk-body">
-                Our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.terms_of_use') }}">terms of use</a> say that your service must be the only one of its kind in your organisation.
+                Your service name helps us to decide if you’re eligible for an annual allowance of free text messages.
             </p>
 
             <p class="govuk-body">
-                To help us check this, choose a service name that’s easy to understand.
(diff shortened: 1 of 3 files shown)


app/templates/views/confirm-your-service-is-unique.html
karlchillmaid:
```suggestion
{% set title = "Confirm your service name" %}
```

app/templates/views/confirm-your-service-is-unique.html
```diff
@@ -22,25 +22,38 @@
         <div class="govuk-grid-column-five-sixths">
 
             <p class="govuk-body">
-                Our <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('main.terms_of_use') }}">terms of use</a> say that your service must be the only one of its kind in your organisation.
+                Your service name helps us to decide if you’re eligible for an annual allowance of free text messages.
             </p>
 
             <p class="govuk-body">
-                To help us check this, choose a service name that’s easy to understand.
+                The free allowance is discretionary. To qualify for it, your service must be the only one of its kind in your organisation.
```
saimaghafoor, line 29: Could we list the services in the org or signpost the user to where this list exists?

saimaghafoor, line 29: Could we list the services in the org or signpost the user to where this list exists?

saimaghafoor (commented):
reviewed  - comment on service being one of its kind

app/templates/views/confirm-your-service-is-unique.html
karlchillmaid: Not for this iteration. That stuff comes later.

karlchillmaid: Not for this iteration. That stuff comes later.

saimaghafoor (approved):

quis (approved):
Definite improvement

PR #5970, opened 2026-06-19
Adding reason to provider changes

whpearson:
Form:
Screenshot 2026-06-19 at 14 57 51

View:
Screenshot 2026-06-19 at 14 57 37

This adds a single field to the provider ratio form for a reason for the change.

This is propagated to the call to the notifications-api.

Tests have been updated

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1945,11 +1945,11 @@ class EstimateUsageForm(StripWhitespaceForm):
 
 
 class AdminProviderRatioForm(OrderableFieldsForm):
-    def __init__(self, providers):
+    def __init__(self, providers, *args, **kwargs):
         self._providers = providers
 
         # hack: https://github.com/wtforms/wtforms/issues/736
-        self._unbound_fields = [
+        fields = [
             (
                 provider["identifier"],
                 GovukIntegerField(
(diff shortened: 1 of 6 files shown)


whpearson:
Needs https://github.com/alphagov/notifications-api/pull/4887 merging first

app/main/forms.py
```diff
@@ -1963,8 +1965,17 @@ def __init__(self, providers):
             )
             for provider in providers
         ]
+        fields += [
+            (
+                "reason",
+                GovukTextInputField(
+                    "Reason", validators=[NotifyDataRequired(thing="your reason for changing the priorities")]
```
klssmith: Wondering if it would be helpful not to make this a required field - ideally people should put a reason down, but we've had an incident in the past where we manually changed the percentages at very frequent intervals. If that ever happened again, needing to enter the same reason each time might be time consuming

app/templates/views/providers/provider.html
```diff
@@ -46,10 +46,18 @@
...
+
         {{ text_field(item.priority) }}
 
         {{ text_field(item.active) }}
 
+        {% if item.reason %}
+            {{ text_field(item.reason) }}
+        {% else %}
+            {{ text_field('None') }}
+        {% endif %}
```
quis: This will show a default and grey-out the text if the value is falsey:
```suggestion
        {{ optional_text_field(item.reason, default='Not given') }}
```

app/main/forms.py
```diff
@@ -1945,11 +1945,13 @@ def validate(self, *args, **kwargs):
 
 
 class AdminProviderRatioForm(OrderableFieldsForm):
-    def __init__(self, providers):
+    def __init__(self, providers, *args, **kwargs):
         self._providers = providers
 
+        super().__init__(*args, **kwargs)
```
klssmith: Is this needed if we're also calling `super().__init__(...)` later in the function?

klssmith (commented):
Just left 1 minor question, but I think this looks good to go after https://github.com/alphagov/notifications-api/pull/4893 is merged

klssmith (approved):

PR #5972, opened 2026-06-22
Rewrite TemplateFolderForm as an ES Module

kr8n3r:
## What 
Final javascript file to be rewritten with Class syntax into an ES Module, following an established pattern.

As this also replaces jQuery with native JavaScript methods, there's also additional cleanup.

This PR:
- rewrites JS
- updates the tests
- removes jQuery package
- removes and updated javascript Jest helper
- removes legacy JS bundle pipeline
- removes loading of legacy JS in the template

## Review

Best done by commit

## Impact

As this is the final conversion and we've removed legacy jquery, the impact is a follows

**Before**

all-js-and-all-esm

**After**

all-esm


app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  setupDOM() {
+    // remove the hidden unknown button - if you've got JS enabled then the action you want to do is implied by 
+    // which field is visible.
+    this.$form.querySelector('button[value=unknown]').remove();
+
+    this.$liveRegionCounter = this.$form.querySelector('.selection-counter');
+
+    this.nothingSelectedButtons = this.createNothingSelectedButtons();
```
tombye: Should this be `this.$nothingSelectedButtons`? Same for `this.itemsSelectedButtons` below.

app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+        cancellable: true,
+        setFocus: () => document.getElementById('add_new_template_form').focus(),
+        action: 'new template',
+        description: 'Press continue to confirm selection or cancel to close'
+      }
+    ];
+  }
+
+  setupStateModifiers() {
+    this.states.forEach(state => {
```
tombye, line 83: Much more efficient, and readable, version of this. Really nice.

app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+    $btn.setAttribute('href', '');
+    $btn.innerHTML = text;
+    $btn.classList.add('govuk-link', 'govuk-link--no-visited-state', 'js-cancel');
+    // isn't set if cancelSelector is undefined
+    if (opts.cancelSelector) {
+      $btn.dataset.target = opts.cancelSelector;
+    }
+    $btn.setAttribute('tabindex', '0');
+
+    const handleTrigger = (event) => {
```
tombye: Should this be something more specific, like 'onClick'? Reading it, I assumed it was used in more than one place, due to the name.

app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  render() {
+    let mode = 'default';
+    let currentStateObj = this.states.filter(state => { return (state.key === this.currentState); })[0];
+    let scrollTop;
+
+    // detach everything, unless they are the currentState
+    this.states.forEach(
+      state => (state.key === this.currentState ? (this.$liveRegionCounter && state.$el && this.$liveRegionCounter.before(state.$el)) : (state.$el && state.$el.remove()))
```
tombye: This is a bit 😬 to read on one line, could we have it as the equivalent if/else please?

app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+    );
+
+    // use dialog mode for states which contain more than one form control
+    if (['move-to-existing-folder', 'add-new-template'].indexOf(this.currentState) !== -1) {
```
tombye: Could we use the `includes` method here instead of `indexOf`? Just to read nicer.

app/assets/javascripts/esm/template-folder-form.mjs
```diff
@@ -0,0 +1,431 @@
...
+      mode = 'dialog';
+    }
+    stickAtBottomWhenScrolling.setMode(mode);
+    // make sticky JS recalculate its cache of the element's position
+    stickAtBottomWhenScrolling.recalculate();
+
+    if (currentStateObj && ('setFocus' in currentStateObj) && !this.formHasError()) {
```
tombye, line 347: Can we put a comment above this line, explaining:
- why to ignore the block if the form has errors
- that we cache and reset the scrollTop to negate scrolling the browser does when elements inside the new state are focused

I know the first point is hidden in the history but the last one isn't, because I didn't mention it in commits and both provide useful context.

tombye (changes requested):
This is great work, and kind of amazing to see the commits clearing up the old JS and related pipeline code.

I also really liked your rewriting multiple 'filtering' operations into a single loop in places and overall it just reads better.

I put in a few suggested changes but they're all to do with readability rather than any issues with the code or logic. Once they're dealt with (and feel free to argue against my suggestions), I'm happy to merge.

tombye (approved):
Thanks for addressing the comments. Now looking, and still working great 👍🏻

PR #5974, opened 2026-06-23
Add a new ‘Fonts and typefaces’ guidance page

karlchillmaid:
Users sometimes ask:

* which fonts we use for emails, text messages and letters
* whether they can change the default fonts

The number of enquiries is not high, but this still feels like valuable information to include in the Using Notify pages.

This PR:

* adds a new guidance page
* adds that page to the left-hand navigation
* adds a link to the new page from the ‘Using Notify’ page
* adds a link to the new page from the ‘Formatting’ guidance page

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -238,10 +238,14 @@ def guidance_email_branding():
 
 
 @main.route("/using-notify/fonts-typefaces")
-def guidance_fonts_typefaces():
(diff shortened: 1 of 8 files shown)


saimaghafoor (commented):
Looks good, no typos spotted

tombye:
I pushed a commit to fix a few things. Now looking like this:

image
image
image

spatel033 (approved):

PR #5978, opened 2026-06-29
Bump utils to 121.0.0

quis:
Breaking change is replacing the use of `InsensitiveDict.from_keys()` with `InsensitiveSet`, which enables some nice refactors.

***

## 121.0.0

* Removes `InsensitiveDict.from_keys`
* Removes the `override_duplicates` argument to `InsensitiveDict`

## 120.2.0

* Adds the ability to specify per-project config for uv by creating an optional `uv-overrides.toml` file

***

Complete changes: https://github.com/alphagov/notifications-utils/compare/120.1.0...121.0.0

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -135,7 +135,7 @@ def send_messages(service_id, template_id):
             )
         )
 
-    column_headings = get_spreadsheet_column_headings_from_template(template)
+    column_headings = fields_to_fill_in(template)
 
     return render_template(
         "views/send.html",
@@ -152,9 +152,7 @@ def send_messages(service_id, template_id):
 def get_example_csv(service_id, template_id):
     template = current_service.get_template(template_id)
     return (
-        Spreadsheet.from_rows(
-            [get_spreadsheet_column_headings_from_template(template), get_example_csv_rows(template)]
-        ).as_csv_data,
+        Spreadsheet.from_rows([fields_to_fill_in(template), get_example_csv_rows(template)]).as_csv_data,
         200,
         {
             "Content-Type": "text/csv; charset=utf-8",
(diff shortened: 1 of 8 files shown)


app/main/views/send.py
```diff
@@ -152,9 +148,7 @@ def send_messages(service_id, template_id):
 def get_example_csv(service_id, template_id):
     template = current_service.get_template(template_id)
     return (
-        Spreadsheet.from_rows(
-            [get_spreadsheet_column_headings_from_template(template), get_example_csv_rows(template)]
-        ).as_csv_data,
+        Spreadsheet.from_rows([fields_to_fill_in(template), get_example_csv_rows(template)]).as_csv_data,
```
ipdigital24, line 151: i can't verify the correctness of replacing `get_spreadsheet_column_headings_from_template` with `fields_to_fill_in`

quis: **`get_spreadsheet_column_headings_from_template`** returned:
- all the recipient columns in a spreadsheet (for example `email address` or `address line 1`, `address line 2`, etc)
- combined with all the placeholders in a template (for example `name`, `day of week`, `colour`, etc)

It did this by using a loop to build up a unique list, item-by-item:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/app/main/views/send.py#L1061-L1065

**`fields_to_fill_in`** does the same thing, but achieves uniqueness with a `Set` type instead of a loop:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/app/main/views/send.py#L781-L786

ipdigital24: i'm with you now on that aspect.

`fields_to_fill_in` seems to be doing something else to session if template is for sms and email while `get_spreadsheet_column_headings_from_template` does not

are these compatible https://github.com/alphagov/notifications-admin/blob/main/app/main/views/send.py#L788-L793
and https://github.com/alphagov/notifications-admin/blob/main/app/main/views/send.py#L1058-L1059

quis: `fields_to_fill_in` only does stuff to the session if it’s called with `prefil_current_user=True`, which we are not doing here.

Otherwise the functions both return the same thing, which is why I think it’s ok to delete one of them.

***

There is a test which checks the correct example column headers are shown for text message templates here:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/tests/app/main/views/test_send.py#L317

…and here:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/tests/app/main/views/test_send.py#L2912

And then for letters here:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/tests/app/main/views/test_send.py#L333-L352

…and here:
https://github.com/alphagov/notifications-admin/blob/a66ecbf4f1ac746cb5f553626965dd64ce4457e8/tests/app/main/views/test_send.py#L2942-L2946

ipdigital24: gotcha. thanks

ipdigital24 (approved):

PR #5979, opened 2026-06-29
Show cost of text message on the ‘preview sending’ page

quis:
We show an estimate of the cost of a text message when editing a template. However this will always be an estimate because:
- we don’t know how the message will be personalised
- we don’t know if the message will be sent to an international phone number

To better help users understand how text messages pricing works this commit adds the same cost information to the page the user sees before sending a message. This number will be exact because we know all the information that impacts the cost at this point.

We’re only going to show this information to people who have either the edit templates or manage service permission. This upholds a long-standing Notify principle that people who only have permission to send messages shouldn’t be using cost as a basis for whether or not to send the message – either it needs to be sent or it doesn’t.

We’re also not showing it in the tour because we don’t want to overload people with the cost information while they are trying to understand personalisation. Plus they will always be within the free allowance at this point.

# Before 

image

# After 

image

### DIFF
--- a/tests/app/main/views/test_send.py
+++ b/tests/app/main/views/test_send.py
@@ -32,6 +32,7 @@ from tests.conftest import (
     create_multiple_email_reply_to_addresses,
     create_multiple_sms_senders,
     create_service_one_admin,
+    create_service_one_user,
     create_template,
     do_mock_get_page_counts_for_letter,
     mock_get_service_email_template,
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_send.py
```diff
@@ -4358,6 +4359,86 @@ def test_check_notification_shows_back_link(client_request, service_one, fake_uu
...
+            "ŵ" * 80,
+            {},
+            "07900900123",
+            "Will be charged as 2 text messages",
+        ),
+        (
+            "Sending abroad",
+            {},
+            "+225 01 01 01 01 01",  # Côte d'Ivoire
+            "Will be charged as 10 text messages",
```
rparke, line 4387: I didn't realise we have a 10 x multiplier!

rparke (approved):

PR #5981, opened 2026-06-30
Sanitise placeholder content

saimaghafoor:
Added content telling users they can use notify to sanitise placeholders with the API:

Content added to:

- perosnalisation guidance
- edit email template page

To deploy when sanitise placeholder content is added to all api docs


kr8n3r:
fixed duplicate header id

CrystalPea (commented):
Nice work!
The only thing I wish we could do better is link straight to the section in the docs - but sadly since we have multiple docs it's not possible. They all share the section title name though - should we tell that to the users maybe to help them navigate?

Or at least we should tell them it's in "Send email message" section, and that it's an email-only feature.

kr8n3r:
fixed the failing test

BlessedDev (approved):

PR #5982, opened 2026-07-01
fix rounding error with fractional pence to display pence to 1 decimal point

rparke:
https://github.com/alphagov/notifications-api/commit/ad0b987c97fc0b0788bc16058ad17626cb9b1f14 updated letter pricing for economy mail to fractional pence. While billing etc. is expected to work fine with this, it doesn't display correctly on the pricing page in admin because we were rounding pence to 0 decimal points. THis now rounds to 1 decimal point to display the new economy pricing correctly.

### DIFF
--- a/app/formatters.py
+++ b/app/formatters.py
@@ -263,7 +263,7 @@ def nl2br(value):
 
 
 def format_pounds_as_currency(number: float):
-    return format_pennies_as_currency(round(number * 100), long=False)
+    return format_pennies_as_currency(round(number * 100, ndigits=0), long=False)
 
 
 def format_pennies_as_currency(pennies: int | float, long: bool) -> str:
@@ -272,7 +272,11 @@ def format_pennies_as_currency(pennies: int | float, long: bool) -> str:
     if pennies >= 100:
         pennies = round(pennies)
         return f"£{pennies // 100:,}.{pennies % 100:02}"
-    elif long:
+    # non-fractional pence should be reported to 2 sf 82 p not 82.0 p
+    if float(pennies) == float(int(pennies)):
+        pennies = round(pennies)
+
+    if long:
         return f"{pennies} pence"
 
     return f"{pennies}p"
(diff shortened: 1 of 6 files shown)


app/formatters.py
```diff
@@ -263,7 +263,7 @@ def nl2br(value):
 
 
 def format_pounds_as_currency(number: float):
-    return format_pennies_as_currency(round(number * 100), long=False)
+    return format_pennies_as_currency(round(number * 100, ndigits=0), long=False)
```
ipdigital24, line 266: any reason it is not rounded to 1 dp?

```
>>> round(1.923*100,1)
192.3
>>> round(1.923*100,0)
192.0
```

app/formatters.py
```diff
@@ -272,7 +272,10 @@ def format_pennies_as_currency(pennies: int | float, long: bool) -> str:
     if pennies >= 100:
         pennies = round(pennies)
         return f"£{pennies // 100:,}.{pennies % 100:02}"
-    elif long:
+    if float(pennies) == float(int(pennies)):
```
ipdigital24, line 276: it is difficult to understand the purpose of this condition just by looking at the code. it would be good to explain the intended logic

app/formatters.py
rparke: I've asked Saima and Karl for their content opinion.

rparke: Karl says to only have actually fractional pence to 1 dp in line with gov.uk style.

app/formatters.py
rparke: added a comment

app/formatters.py
ipdigital24: does that mean the input number in pound will never have more than 2dp? or even if it has more than 3dp e.g. 1.923, we do drop anything after 2dp

app/formatters.py
ipdigital24: thanks, another option could have been `is_non_fractional(pennies)` function

ipdigital24 (approved):

PR #5984, opened 2026-07-01
Update dependencies and apply fixes

kr8n3r:
## What

- Update dependencies - should [resolve these open issues](https://github.com/alphagov/notifications-admin/security/dependabot?q=is%3Aopen+ecosystem%3Anpm)
- Apply Stylint Sass fixes and remove some rules overrides. Simple ones only.
- Update NPM overrides

## Review
See individual commits 

[See diff for Sass update](https://www.diffchecker.com/XewyBE1C/)


tombye (approved):
Thanks for making the commits so clear. Really like the removal of stylelint overrides too 👍🏻

PR #5992, opened 2026-07-16
Add "Trial mode" tag next to the service name

klssmith:
This adds a "Trial mode" tag next to the service name for services which
are in trial mode. We've made this grey so it doesn't look like it can be
clicked.

We had a "suspended" tag in the service navigation for archived
services. This used custom CSS but has been updated to use the tag
component for consistency and renamed to "Archived" since suspending a
service is no longer a feature. Only platform admin users can see this
tag, since we don't show archived services to normal users.

[Trello card](https://trello.com/c/cxiSyPzL/1847-onboarding-build-trial-mode-tag)

**Screenshots**
Showing mobile and desktop view. Only platform admins will see both tags, other users
never see the "archived" tag.

Screenshot 2026-07-22 at 17 26 22

Screenshot 2026-07-22 at 16 57 26

Screenshot 2026-07-22 at 16 57 40

Screenshot 2026-07-22 at 16 57 32

### DIFF
--- a/app/assets/stylesheets/components/navigation.scss
+++ b/app/assets/stylesheets/components/navigation.scss
@@ -9,26 +9,15 @@
   $padding-bottom: 17px;
 
   &-service-name,
-  &-organisation-link {
+  &-organisation-link,
+  &-status-tag {
     float: left;
   }
 
-  &-service-type {
-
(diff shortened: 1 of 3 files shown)


quis (approved):
Looks good, thanks

PR #5994, opened 2026-07-17
block sms sending to china

rparke:
Removes China from our list of allowed countries after issues successfully sending to China were discovered:

Screenshot 2026-07-17 at 12 37 01

Adds Banner to the top of the pricing page:
Screenshot 2026-07-17 at 13 15 26

### DIFF
--- a/app/templates/views/guidance/pricing/text-message-pricing.html
+++ b/app/templates/views/guidance/pricing/text-message-pricing.html
@@ -3,6 +3,7 @@
 {% from "components/table.html" import mapping_table, row, text_field, field %}
 {% from "components/live-search.html" import live_search %}
 {% from "govuk_frontend_jinja/components/details/macro.html" import govukDetails %}
+{% from "govuk_frontend_jinja/components/notification-banner/macro.html" import govukNotificationBanner %}
 
 {# Used by the content_template.html layout, prefixes the "navigation" accessible name #}
 {% set navigation_label_prefix = 'Pricing information' %}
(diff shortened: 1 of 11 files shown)


app/templates/views/guidance/pricing/text-message-pricing.html
```diff
@@ -13,6 +14,17 @@
 
 {% block content_column_content %}
 
+{% call govukNotificationBanner({
+    "titleText": "You can no longer send text messages to China",
+    "classes": "govuk-!-margin-bottom-4"
```
kr8n3r: damn our legacy page headings containing top margin

add `govuk-!-margin-top-9` into the 'classes`

rparke: I was like "this looks to high up but MAYBE it's right?"

rparke: Much nicer:

Screenshot 2026-07-17 at 12 56 08

kr8n3r:
have to rebuild docker but `test_guidance_pricing_sms` will need to be updated, likely to `page.select("main > .govuk-body")` so it targets direct children of main, ignoring the one in the banner

rparke:
> have to rebuild docker but `test_guidance_pricing_sms` will need to be updated, likely to `page.select("main > .govuk-body")` so it targets direct children of main, ignoring the one in the banner

done

kr8n3r (dismissed):

kr8n3r (approved):

spatel033 (approved):

PR #5995, opened 2026-07-17
Convert tables with single fields to unordered lists

kr8n3r:
## What 

Convert tables with single fields to use an unordered list.
- app/templates/views/uploads/contact-list/contact-list.html
- app/templates/views/returned-letter-summary.html

utilising the Design system's `govuk-list` and extending it to create a border list and list with metadata

## Review

To test the contact-list, upload a CSV or phone number or email addresses.

To test the list of returned letters you can use the below script to insert some dummy data in your local DB
```
INSERT INTO returned_letters (
    id, 
    reported_at, 
    service_id, 
    notification_id, 
    created_at, 
    updated_at
) VALUES 
(
    '8517c43e-63fa-4921-bdd0-eef86683f23d', 
    '2026-07-01', 
    'yourserviceid', 
    'cd44b76e-bdf3-44e2-94ad-88e86c75fd59', 
    '2026-07-01 12:51:28.011264', 
    NULL
),
(
    '06f55df0-5d51-4fce-a7c2-c50331027dd2', 
    '2026-07-01', 
    'yourserviceid', 
    '39b579c0-ec0e-4365-86e3-8ad9e34ee02f', 
    '2026-07-01 08:02:42.989609', 
    NULL
);
``` 

and then view the list on `services/<serviceID>/returned-letters` 

## Visuals

**Before**
list-before

**After**
list-after

**Before**
metadatalist-before

**After**
metadatalist-after

### DIFF
--- a/app/assets/stylesheets/govuk-frontend/extensions.scss
+++ b/app/assets/stylesheets/govuk-frontend/extensions.scss
@@ -332,3 +332,18 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
 .govuk-error-summary__list a {
   word-break: break-word;
 }
+
+// extending the list component
+// border between items
+.govuk-list--bordered > li {
+  border-top: 1px solid govuk-functional-colour(border);
+  padding-top: govuk-spacing(2);
+  padding-bottom: govuk-spacing(2);
+  // with borders we're using padding to control spacing,
+  // so this removes the additional margin added by the Design System styles
+  margin-bottom: 0; 
+}
+.govuk-list--bordered > li:last-of-type {
+  border-bottom: 1px solid govuk-functional-colour(border);
+  padding-bottom: govuk-spacing(2);
+}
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk-frontend/extensions.scss
```diff
@@ -331,4 +331,28 @@ $govuk-touch-target-size: ($govuk-checkboxes-size + $govuk-touch-target-gutter);
...
-}
\ No newline at end of file
+}
+
+// extending the list component
+// border between items
+.govuk-list--bordered > li {
+  border-top: 1px solid govuk-functional-colour(border);
+  padding-top: govuk-spacing(2);
+  padding-bottom: govuk-spacing(2);
```
tombye, line 341: From offline discussions with Jani, we talked about the 5px margin the design system styles add to list items introduces a bit of extra space underneath each item, that looks odd with the border style, because the borders make it look like that extra space is inside each item. So this is a note to say we agreed to add some styles to negate it here, so the space above and below are the same.

tombye (changes requested):
Nice to see this being tackled finally, and it reads a lot easier with a screen reader. Just one change suggested, and we discussed offline already so the approach should be clear enough. Happy to approve once in.

app/assets/stylesheets/govuk-frontend/extensions.scss
kr8n3r: agreed and changes made - https://github.com/alphagov/notifications-admin/pull/5995/changes#diff-9d5f04c6b0f11ed757c81f10574f3c7d3cca79dd293680f1f63b85455797abf4R342-R344

tombye (approved):
Thanks for making the change. Now looking good to go 👍🏻

PR #5997, opened 2026-07-17
Update banner content

karlchillmaid:
Update banner content to match supporting comms.


rparke (approved):

karlchillmaid:
Incident closed. We rolled back the block on +86 numbers. This banner update is no longer needed.

PR #5999, opened 2026-07-17
TemplatePreviewClient: ensure raw dicts are used for json encoding

risicle:
Tests for this mocked out actual json encoding.

I can't prove this will fix the issue, because I'm not about to rewrite the tests for this in a way that would catch problems.


quis:
Do we still need this with the new encoder?

risicle:
Nope

PR #6000, opened 2026-07-20
Check agreeement_signed as well as agreeement_signed_by_id

whpearson:
If we don't do something like this we can get into a bad state.

This is where they have signed the agreement and then someone sets them not to have signed the agreement (due to a change of government or similar)

This sets agreement_signed to false but leaves agreement_signed_by_id, so it shows up in the UI as signed but still doesn't actually allow them to go live.

### DIFF
--- a/app/templates/views/request-to-go-live.html
+++ b/app/templates/views/request-to-go-live.html
@@ -106,7 +106,7 @@
         {# the current user signed the org-level agreement #}
         {# this is because for other services in that org that want to go live #}
         {# we want to show a different text if org-level agreement already accepted #}
-        {% if current_service.able_to_accept_agreement and (current_service.organisation.agreement_signed_by_id == current_user.id or not current_service.organisation.agreement_signed_by_id) %}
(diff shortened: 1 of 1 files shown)


spatel033:
`able_to_accept_agreement` checks `self.organisation.agreement_signed is not None`, it would be better to fix this logic and check if `self.organisation.agreement_signed is not True`.

whpearson:
> `able_to_accept_agreement` checks `self.organisation.agreement_signed is not None`, it would be better to fix this logic and check if `self.organisation.agreement_signed is not True`.

Looking into this self.agreement_signed is None is a special case, it means that they have a service specific agreement.

From ef384d4675e2299a767eaf4c7d25ce47cf98a21c

Note we still have the concept of a service-specific agreement (defined
    as `agreement_signed=None`) which this continues to support. Argument
    for another day whether we should remove that concept entirely.

spatel033 (approved):

spatel033 (approved):

PR #6001, opened 2026-07-21
Remove custom padding from notify-summary list

kr8n3r:
In https://github.com/alphagov/govuk-frontend/pull/6679 a change was made to align text in list and so our custom padding throws text into misalignment.

Removing custom padding and deferring to Design System styling resolves that.

**Before**

before

**After**
after


tombye (approved):
Much cleaner! Also, nice sleuthing 🕵🏻‍♂️ 👍🏻

PR #6002, opened 2026-07-21
Update template statistics "dashboard" tables to summary lists

kr8n3r:
## What 

Replace template stats "dashboard" tables with a [summary list component](https://design-system.service.gov.uk/components/summary-list/). These stats are key:value pairs so from an accessibility standpoint, a summary list that's built on [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl) element is more appropriate.

Key value pairs each take 50% of the available space and we no longer truncate the template name.

A new summary list modifier class introduced to contain needed styling.
Styles and jinja macro moved out of `table` as they're now n longer used in the table context.

## Review
- see individual commits for details.
- send some email with different templates to see stats.
- urls:
  - dashboard page: /services/<serviceId>
  - template usage: /services/<serviceId>/template-usage

## Visuals
**Dashboard page: lazy load and stats**
dashboard-page
lazy-load

**Template usage page**
template-usage

### DIFF
--- a/app/templates/views/dashboard/template-statistics-lazy.html
+++ b/app/templates/views/dashboard/template-statistics-lazy.html
@@ -1,46 +1,39 @@
-{% from "components/table.html" import list_table, row_heading, field %}
 {% from "components/show-more.html" import show_more %}
+{% from "govuk_frontend_jinja/components/summary-list/macro.html" import govukSummaryList %}
 
 <div class="ajax-block-container">
-    <div class='template-statistics-table'>
-      {% call(item, row_number) list_table(
-        [1],
-        caption="By template",
-        caption_visible=True,
-        empty_message='',
-        field_headings=[
-          'Template',
-          'Messages sent'
-        ],
-        field_headings_visible=False
-      ) %}
-
-        {% call row_heading() %}
-          <span class="template-statistics-table-template-name" aria-label="Loading data for templates">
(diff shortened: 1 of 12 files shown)


app/templates/views/dashboard/template-statistics-lazy.html
```diff
@@ -1,46 +1,39 @@
...
-        {% endcall %}
-
-      {% endcall %}
-      {{ show_more(
-        url_for('main.template_usage', service_id=current_service.id),
-        'See templates used by month',
-        with_border=False
-      ) }}
-    </div>
+    <h3>By template</h3>
```
tombye: This is a bit 'just my opinion' but I think this should probably be a h2, based on the overall page headings:

Image

tombye (changes requested):
Thanks for doing this, it makes a big difference reading the template code and for users of screen readers. I also really appreciate the work done on good class names in the CSS.

app/templates/views/dashboard/template-statistics-lazy.html
kr8n3r: yeah, agree with that. pushed the change up

tombye (approved):
Thanks for addressing the comment about headings. Looks good to go now 👍🏻

PR #6010, opened 2026-07-27
typing: fix imports of `current_user`

risicle:
No stubs for the flask-login package + complexity of `LocalProxy` use + complexity of custom `User` class means the most straightforward thing is to cheat and lie about the type, which also means we have to do a weird reassignment thing. ~Avoid unnecessarily introducing new potential cyclical imports through use of `TYPE_CHECKING`.~

This reduces the number of mypy errors from 1522 to 1317.

### DIFF
--- a/app/main/views/organisations/index.py
+++ b/app/main/views/organisations/index.py
@@ -2,13 +2,13 @@ from datetime import datetime
 from functools import partial
 
 from flask import flash, redirect, render_template, request, send_file, url_for
-from flask_login import current_user
 from notifications_python_client.errors import HTTPError
 from werkzeug.exceptions import abort
 
 from app import (
     current_organisation,
     current_service,
+    current_user,
     org_invite_api_client,
     organisations_client,
 )
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -18,7 +18,8 @@ from flask import (
     session,
     url_for,
 )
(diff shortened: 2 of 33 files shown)


app/main/views/organisations/index.py
```diff
@@ -37,12 +37,14 @@
...
 )
 from app.models.organisation import AllOrganisations, Organisation
-from app.models.user import InvitedOrgUser
+from app.models.user import InvitedOrgUser, User
 from app.s3_client.s3_mou_client import get_mou
 from app.utils.csv import Spreadsheet
 from app.utils.user import user_has_permissions, user_is_platform_admin
 from app.utils.user_permissions import organisation_user_permission_options
 
+current_user: User = _current_user  # type: ignore[assignment]
```
quis: Is it possible to define this once (for example in `app/__init__.py`) and import it from there?

risicle: If it works, that's a genius idea - I wonder if I can do it while avoiding circular imports though..

risicle: There we go, much better.

risicle:
Wait, missed a couple..

risicle:
Nope, this will need more work...

risicle:
All good now

quis (approved):
Tidy

PR #6012, opened 2026-07-28
Correct type hints on letter attachment form

quis:
Some possible combinations were missing

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -94,12 +94,12 @@ def get_template_form(
 
 
 class LetterAttachmentFormError(Exception):
-    def __init__(self, *, title: str = None, detail: str, attachment_page_count: int = 0) -> None:
+    def __init__(self, *, title: str | None = None, detail: str, attachment_page_count: int = 0) -> None:
         self.title = title or "There is a problem"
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
I see this aligns with how we raise this error, but should it not always have a title?

quis:
@kr8n3r looks like we have an banner without a title here:
https://github.com/alphagov/notifications-admin/blob/0a944bb01314fd9267e2f0c772c439b32ba598e5/app/main/views/templates.py#L1243-L1250

However that is rendered in this template, which expects `title` to be set:
https://github.com/alphagov/notifications-admin/blob/0a944bb01314fd9267e2f0c772c439b32ba598e5/app/templates/views/templates/attach-pages.html#L21-L26

Not sure what the correct thing to do is here…

kr8n3r:
these banners need work, so let's address that at some point. I'd guess "There is a problem" would the a default title string

quis:
Done some more digging and I realise it does set that as the default title here:
https://github.com/alphagov/notifications-admin/blob/0a944bb01314fd9267e2f0c772c439b32ba598e5/app/main/views/templates.py#L98

PR #6013, opened 2026-07-28
Fix template-preview error cache

spatel033:
When the admin requests a template-preview to get the page count for a letter, a connection error or another exception may occur. In this case, the template-preview returns an error, and the admin stores that error (for example, `{"message": "This is a test exception", "result": "error"}`) in Redis instead of the page count. Later, when the admin tries to use the cached page count, it fails because an error response was stored instead of a valid page count.
To fix this issue, we want to make sure the template-preview response always includes the required fields: `count`, `attachment_page_count`, and `welsh_page_count`. More details on this [card](https://trello.com/c/aOj7vNP2/1836-dont-cache-errors-from-template-preview).

### DIFF
--- a/tests/app/utils/test_templates.py
+++ b/tests/app/utils/test_templates.py
@@ -1,4 +1,5 @@
 from datetime import date
+from functools import partial
 from unittest import mock
 from uuid import UUID
 
@@ -8,9 +9,11 @@ from freezegun import freeze_time
 from markupsafe import Markup
 from notifications_utils.template import Template
 from ordered_set import OrderedSet
+from requests.exceptions import HTTPError
 
 from app import load_service_before_request
 from app.models.template_email_file import TemplateEmailFile
+from app.template_previews import TemplatePreviewClient, template_preview_client
 from app.utils.templates import EmailPreviewTemplate, TemplateChange, TemplatedLetterImageTemplate, get_sample_template
 from tests import ConcreteTemplate, template_json
 from tests.conftest import SERVICE_ONE_ID, do_mock_get_page_counts_for_letter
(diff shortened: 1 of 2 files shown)


app/utils/templates.py
quis, line 179: Would it be better if `template_preview_client.get_page_counts_for_letter` raised an exception for a non-`200` HTTP status?

spatel033: Tested with [these](https://github.com/alphagov/notifications-template-preview/pull/1012) changes, but the admin still caches the error response: `{"message": "400 Bad Request: Unable to calculate page count", "result": "error"}`

quis: I meant something like 1e61dbc05407dc7a5b814574bfb5b38fa38c0361

spatel033: Refactored to raise non 200 errors

tests/app/utils/test_templates.py
```diff
@@ -68,6 +69,44 @@ def test_get_page_counts_for_letter_caches(
...
+        return_value=None,
+    )
+    mock_redis_set = mocker.patch(
+        "app.extensions.RedisClient.set",
+    )
+
+    mocker.patch(
+        "app.template_preview_client.get_page_counts_for_letter",
+        side_effect=HTTPError("Simulated 400 Bad Request"),
+    )
```
quis, line 96: You could remove the `raise_for_status()` and this test would still pass. Good tests prevent the undoing of the corresponding code change.

This is what I had to do to fix that:

```suggestion
    from functools import partial
    from app.template_previews import TemplatePreviewClient, template_preview_client
    # requests_mock is a fixture
    requests_mock.post("http://localhost:9999/get-page-count", status_code=500)
    mocker.patch(
        "app.template_preview_client.get_page_counts_for_letter",
        wraps=partial(TemplatePreviewClient.get_page_counts_for_letter, template_preview_client)
    )
```

I do not like it but there’s something very weird going on with the mocking in this file where mocking `get_page_counts_for_letter` in one test seems to carry over to other tests. Overwriting with a new, wrapping mock gets around this. But it would be worth investigating why this strange mocking behaviour is happening

spatel033: Great catch! and thanks for the suggestion. I have created [card](https://trello.com/c/myrsTx5g/551-investigate-mocking-behaviour-for-admin) to investigate this.

quis (approved):

PR #6014, opened 2026-07-28
Remove `InvitedUser.has_permission_for_organisation`

quis:
`InvitedUser`s (users invited to a service) do not have any relationship to organisations. They do not have an `organisation` attribute and it’s not a meaningful question to ask of an `InvitedUser`. There is a separate `InvitedOrgUser` class where this is relevant.

Long and short of it is: anything calling this code is going to encounter an error at some point. Better to remove the method and raise an error early, where a test is more likely to catch it.

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -583,11 +583,6 @@ class InvitedUser(BaseUser):
             return False
         return self.service == service_id and permission in self.permissions
 
-    def has_permission_for_organisation(self, organisation_id, permission):
-        if self.status == "cancelled":
-            return False
-        return self.organisation == organisation_id and permission in self.organisation_permissions
-
     def __eq__(self, other):
         if not isinstance(other, InvitedUser):
             return False


DilwoarH:
Should we not just return False in the function?

It looks like a bunch of templates call it: https://github.com/search?q=repo%3Aalphagov%2Fnotifications-admin%20has_permission_for_organisation&type=code

Would this type of user ever reach these templates?

quis:
I those cases it’s a `User` object, which has its own, separate version of the method here:
https://github.com/alphagov/notifications-admin/blob/75bf0ed11609cc05a59095fa20258a7de8f7e136/app/models/user.py#L299-L300

`InvitedUser` would never reach those templates.

DilwoarH (approved):

DilwoarH:
> I those cases it’s a `User` object, which has its own, separate version of the method here:
> 
> https://github.com/alphagov/notifications-admin/blob/75bf0ed11609cc05a59095fa20258a7de8f7e136/app/models/user.py#L299-L300
> 
> `InvitedUser` would never reach those templates.

cool sounds good - approved

PR #6023, opened 2026-07-31
Beef up update-relative-time tests

kr8n3r:
Revert quick fix from https://github.com/alphagov/notifications-admin/pull/6022 as the issue was with the test helper.

Remove getCalendarOffset helper to avoid JS  arithmetic around months that have 31 days.
In its place, use fixed dates instead.

Also adding some calendar month boundry and  leap year tests cases.

Javascript code functioning as expected

**Current month**
Screenshot 2026-07-31 at 12 04 04

**Last month**
Screenshot 2026-07-31 at 12 04 52

### DIFF
--- a/app/assets/javascripts/esm/update-relative-time.mjs
+++ b/app/assets/javascripts/esm/update-relative-time.mjs
@@ -102,14 +102,8 @@ class UpdateRelativeTime {
     }
 
     // for months and years, we use the calendar date logic as lengths are different
-    let monthDiff = (elementTime.getFullYear() - currentTime.getFullYear()) * 12 + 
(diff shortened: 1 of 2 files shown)


tombye (approved):
We chatted outside of this PR, about whether it's clear what the new logic does and why it needs to. Thanks for making the changes 👍🏻

kr8n3r:
> We chatted outside of this PR, about whether it's clear what the new logic does and why it needs to. Thanks for making the changes 👍🏻

have since agreed to simplify further and remove that helper and replace with fixed dates

tombye (approved):
Changes since my last review are 🤌, so 👍🏻 again

quis (approved):

tombye (approved):
Nice extra changes. Makes the tests much easier to read 👍🏻

PR #6025, opened 2026-08-03
Add white-space: no-wrap to spark-bar to correctly display text and % value bar

kr8n3r:
Fix spark-bar display

When a service has a large spread of template-usage,  the calculated percentage width is too small to contain the value.

Because govuk-summary list uses table-cell display we need to cell the spark-bar_bar to no wrap.

This is caused by the summary-list__value container having `display: table-cell;`

Screenshot 2026-08-03 at 13 47 35

### DIFF
--- a/app/assets/stylesheets/components/spark-bar.scss
+++ b/app/assets/stylesheets/components/spark-bar.scss
@@ -23,4 +23,5 @@
   text-align: right;
   margin: 2px 0 1px 0;
   transition: width 0.6s ease-in-out;
+  white-space: nowrap;
 }


quis:
This is by design, don’t think we should change it. Otherwise, if the first number in your example was 6,000,000 the proportions would suggest that 14,000 is about 1/3 of 5,000,000

kr8n3r:
> This is by design, don’t think we should change it. Otherwise, if the first number in your example was 6,000,000 the proportions would suggest that 14,000 is about 1/3 of 5,000,000

yeah fixed now

for context, original approach was to use `min-width:fit-content` to expand the bar width, but this approach restores original design

quis (approved):

PR #6037, opened 2026-08-17
Allow a full range of unicode characters in text messages

quis:
Depends on:
- [x] https://github.com/alphagov/notifications-utils/pull/1443
- [x] https://github.com/alphagov/notifications-api/pull/4937

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -96,7 +96,6 @@ from app.main.validators import (
     NotifyDataRequired,
     NotifyInputRequired,
     NotifyUrlValidator,
-    OnlySMSCharacters,
     StringsNotAllowed,
     ValidEmail,
     ValidGovEmail,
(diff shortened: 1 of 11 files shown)


app/main/forms.py
```diff
@@ -1481,8 +1480,7 @@ def new_template_data(self):
 
 
 class SMSTemplateForm(BaseTemplateForm, TemplateNameMixin):
```
CrystalPea, line 1482: do we need this class if it's empty?

quis: Yes because it it is combining two other classes (`BaseTemplateForm` and `TemplateNameMixin`)

quis: There’s another example of this usage here:
https://github.com/alphagov/notifications-admin/blob/9ea0d68c4731d59ebab7236a4567c42c79fc517d/app/main/forms.py#L1453-L1454

CrystalPea (approved):

PR #6048, opened 2026-09-01
Update Home Office brand colour hex

kr8n3r:
Based on https://github.com/alphagov/govuk-frontend/blob/a3b356a3db7358d828d020da84b8ae9158b6ab07/packages/govuk-frontend/src/govuk/settings/_colours-organisations.scss#L92 and https://www.gov.uk/government/organisations/home-office

This is for our branding identity creation tool. We won't update existing Home office brand hex


quis (approved):
Matches what’s on GOV.UK 👍🏻

PR #6049, opened 2026-09-02
Fix list of contents on text message pricing page

quis:
The ‘symbols’ section doesn’t exist anymore 

Without that section we can remove the duplicated _is_ from the start of the other bullet points, and use the _it’s_ contraction per GOV.UK style

# Before

image

# After

image


saimaghafoor (approved):
Happy with content

PR #6053, opened 2026-09-04
Truncate SMS body to 1 line in notification list 

kr8n3r:
We can't easily split

```
SMSBodyPreviewTemplate(
    self.template,
    self.personalisation,
)
```
to show 1 line, so we'll use CSS to truncate it.

Fully message text is available when you view the notification.

**Before**
Screenshot 2026-09-04 at 15 22 18

**After**
Screenshot 2026-09-04 at 15 21 56


quis (approved):
Thanks for the quick fix

PR #6055, opened 2026-09-07
event handlers: reimplement using decorators instead of a metaclass

risicle:
This approach should work properly with static type hints, because we specify the accepted arguments via the call signature of the original method. The decorator sniffs its signature using the inspect module and assigns arguments to parameters in the same way a regular call would, but forwards this information to `events_api_client.create_event` instead. The original method is never actually called.

Should we choose to, this approach should also work with type constraints, and positional parameters - although for now I've used a leading `*` argument to retain the same keyword-only behaviour.

This reduces the number of `mypy` errors by 11.

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -1,54 +1,89 @@
+from collections.abc import Callable
+from functools import wraps
+from inspect import signature
+from typing import cast
+
 from flask import request
 
 from app.notify_client.events_api_client import events_api_client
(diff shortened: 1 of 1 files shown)


quis (approved):
Very nice

risicle:
Ohh I should have made these `@staticmethod`

risicle:
#6082

PR #6056, opened 2026-09-07
service settings views: simple typing fixes

risicle:
And some ignore markers where we're reaching in to add extra errors to form fields, which we should probably find better ways to do.

This PR makes one important decision - allowing `assert` statements in production code, but *only* for the purposes of type-narrowing. In these cases it's to tell the type checker something it can't figure out itself - that a valid form won't have `None` values for certain fields. These are all cases where a `None` being present would have thrown an exception anyway because of what the following code was trying to do with it.

This seems preferable to me than wrapping each use of these `.data` attributes in `cast(...)` calls, which would also have not given certainty about correctness in the way `assert` statements do.

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -104,7 +104,7 @@ def service_name_change(service_id):
             if http_error.status_code == 400 and (
                 error_message := service_api_client.parse_edit_service_http_error(http_error)
             ):
-                form.name.errors.append(error_message)
+                form.name.errors.append(error_message)  # type: ignore[attr-defined]  # is there a better way?
             else:
                 raise http_error
         else:
@@ -723,7 +723,9 @@ def service_receive_text_messages_stop(service_id):
                 e,
                 extra={"inbound_number": inbound_number, "service_id": service_id},
             )
-            form.removal_options.errors.append("Failed to remove number from service")
+            form.removal_options.errors.append(  # type: ignore[attr-defined]  # is there a better way?
+                "Failed to remove number from service"
+            )
 
     recent_use_date = None
 
@@ -884,6 +886,7 @@ def service_set_auth_type_for_users(service_id):
     )
 
     if form.validate_on_submit():
+        assert form.users.data is not None  # type narrowing
         for user in all_service_users:
             should_use_email_auth = user.id in form.users.data
             new_auth_type = "email_auth" if should_use_email_auth else "sms_auth"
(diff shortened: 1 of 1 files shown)


risicle:
Oops sorry @CrystalPea, was trying to click @quis for review (this is a strawman for something we were already discussing).. but take a look if you like anyway.

app/main/views/service_settings/index.py
```diff
@@ -704,7 +704,7 @@ def service_receive_text_messages_stop(service_id):
 
     if form.validate_on_submit():
         if current_service.default_sms_sender == current_service.inbound_number:
-            form.removal_options.errors.append(
+            form.removal_options.errors.append(  # type: ignore[attr-defined]  # there must be a better way
```
quis: This one can definitely be in a custom `post_validate` method on the `Form` subclass

app/main/views/service_settings/index.py
```diff
@@ -729,7 +729,9 @@ def service_receive_text_messages_stop(service_id):
                     e,
                     extra={"inbound_number": inbound_number, "service_id": service_id},
                 )
-                form.removal_options.errors.append("Failed to remove number from service")
+                form.removal_options.errors.append(  # type: ignore[attr-defined]  # is there a better way?
+                    "Failed to remove number from service"
+                )
```
quis: This is catching `Exception` all the way down 😨 
https://github.com/alphagov/notifications-api/blob/675e3b976b16dff347a2c20df83de4844ff81f10/app/inbound_sms/rest.py#L94

https://github.com/alphagov/notifications-api/blob/675e3b976b16dff347a2c20df83de4844ff81f10/app/dao/inbound_numbers_dao.py#L77

app/main/views/service_settings/index.py
```diff
@@ -1012,6 +1017,7 @@ def service_add_sms_sender(service_id):
     form = ServiceSmsSenderForm()
     first_sms_sender = current_service.count_sms_senders == 0
     if form.validate_on_submit():
+        assert form.sms_sender.data is not None  # type narrowing
         service_api_client.add_sms_sender(
             current_service.id,
             sms_sender=form.sms_sender.data.replace("\r", "") or None,
```
quis, line 1017: I don’t think we need `replace` here, the field shouldn’t allow carriage returns (and why disallow `\r` but allow `\n` 🤔 )

app/main/views/service_settings/index.py
quis: https://github.com/alphagov/notifications-admin/pull/6059

app/main/views/service_settings/index.py
quis: https://github.com/alphagov/notifications-admin/pull/6060

risicle: Well, still, if/when we add type annotations to `add_sms_sender(...)` do we want to make it accept `None` for `sms_sender` just to appease this? The `replace` issue is just a symptom.

quis (approved):
_Accidental comment_

PR #6060, opened 2026-09-07
Remove calls to `replace` on sender form data

quis:
Our `Form` classes should be handling this kind of data santitisation.

I suspect this has mostly entered through copy/paste and before we had stuff like `StripWhitespaceForm` but need a bit more digging to:
- confirm the API will do the right thing here
- write tests for the expected behaviour (since this change doesn’t cause any tests to fail)

### DIFF
--- a/app/main/views/service_settings/index.py
+++ b/app/main/views/service_settings/index.py
@@ -920,7 +920,7 @@ def service_add_letter_contact(service_id):
     if form.validate_on_submit():
         new_letter_contact = service_api_client.add_letter_contact(
             current_service.id,
-            contact_block=form.letter_contact_block.data.replace("\r", "") or None,
+            contact_block=form.letter_contact_block.data,
             is_default=first_contact_block if first_contact_block else form.is_default.data,
         )
         if from_template:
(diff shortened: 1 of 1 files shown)


app/main/views/service_settings/index.py
```diff
@@ -920,7 +920,7 @@ def service_add_letter_contact(service_id):
     if form.validate_on_submit():
         new_letter_contact = service_api_client.add_letter_contact(
             current_service.id,
-            contact_block=form.letter_contact_block.data.replace("\r", "") or None,
+            contact_block=form.letter_contact_block.data,
```
quis, line 923: Will never be `None` or `""` because the form:
- calls `strip_all_whitespace` on the input here: https://github.com/alphagov/notifications-admin/blob/b4a6ef6636e504d21856352b82f0eaa7c3efcc55/app/main/forms.py#L576
- validates against empty values here: https://github.com/alphagov/notifications-admin/blob/b4a6ef6636e504d21856352b82f0eaa7c3efcc55/app/main/forms.py#L2126-L2128

risicle: Sadly a _static_ analyzer will never be able to determine that, so without some type narrowing, the `contact_block` argument will have to accept `None`.

quis: Fair. I think we can contend with that when we come to adding type hints to `service_api_client` et al though.

risicle (approved):

PR #6065, opened 2026-09-08
Multilingual sms template guidance

saimaghafoor:
Added some high level guidance on how to write a text message in a language other than English.

This was the most asked question once we'd released comms. 

edit

### DIFF
--- a/app/templates/partials/templates/guidance-other-language-sms
+++ b/app/templates/partials/templates/guidance-other-language-sms
@@ -1,0 +1,5 @@
+<h2 class="heading-medium">Send a message in a language other than English</h2>
+
+<p class="govuk-body">Write the message in the language you want to send directly in the text message template.</p>
+
+<p class="govuk-body">If you cannot type in that language, write the message outside of Notify and copy and paste it into the template. </p>
--- a/app/templates/views/edit-sms-template.html
+++ b/app/templates/views/edit-sms-template.html
@@ -53,6 +53,7 @@
           {% include "partials/templates/guidance-personalisation.html" %}
           {% include "partials/templates/guidance-optional-content.html" %}
           {% include "partials/templates/guidance-links-urls-sms.html" %}
+          {% include "partials/templates/guidance-other-language-sms" %}
           {% include "partials/templates/guidance-character-count.html" %}
         </div>
       </div>


app/templates/partials/templates/guidance-other-language-sms
```diff
@@ -0,0 +1,5 @@
+<h2 class="heading-medium">Send a message in a language other than English</h2>
+
+<p class="govuk-body">Write the message in the language you want to send directly in the text message template.</p>
+
+<p class="govuk-body">If you cannot type in that language, write the message outside of Notify and copy and paste it into the template. </p>
```
CrystalPea, line 5: I don't understand that bit - why wouldn't they be able to type the message in Notify?

app/templates/partials/templates/guidance-other-language-sms
```diff
@@ -0,0 +1,5 @@
+<h2 class="heading-medium">Send a message in a language other than English</h2>
+
+<p class="govuk-body">Write the message in the language you want to send directly in the text message template.</p>
```
CrystalPea, line 3: I am not sure how helpful this will be - I think the confusion, and the Zendesk queries stemmed directly from the comms title, and will die down once we clarified with everyone who misunderstood the comms email.

Being able to use all characters to send a message is like a default for most people.

CrystalPea (commented):
Hi there,

Sorry to be a malcontent, but I don't think this guidance will be helpful - it states the obvious - for most users, being able to send a message in a broad range of characters is a default.

The confusion was because users misunderstood the comms title, and we weren't super clear that the new feature is just access to unicode characters (something that most users wouldn't even see as a feature, but more a default state of how sms behave on their phone). To clear up the confusion, we can either respond to users when they write to us, letting them know that no, we do not provide translation service. Or we could send another mailout with an errata.

Adding a guidance in a place that is not directly linked from the comms is unlikely to reduce the confusion.

saimaghafoor:
thanks for the points you raised. I agree it won't help with the few users who interpreted the comms as meaning we translate messages. 
My understanding from the support tickets was that most users were asking _how_ to send messages in another language, for example whether there was a template or a specific way to do it.
I thought a short piece of guidance could help answer those questions without getting into the technical details of unicode characters or introducing this term.
We also probably shouldn't assume everyone knows they simply need to write or paste their message in the language they want . Some users may not have that language enabled on their device/keyboard, or have limitations especially on work devices. In this instance they'd need to copy/paste from elsewhere into template. 
That said, I'm happy to remove it if we don't think the value justifies the additional guidance. Or we can wait until we look at this again (metrics, support ticket analysis).
I definitely think we should not send another email, considering there's a handful of responses out of 25k users that received the comms.

CrystalPea:
> My understanding from the support tickets was that most users were asking how to send messages in another language, for example whether there was a template or a specific way to do it.

I looked through all of the tickets I could find - I found 12, excluding the bots. I think most of these were asks if we translate the message for them, a couple asked about pricing, and I could see a couple which were "how do I use this feature/ is there a template", which to me also looked like users expect ready-made content in another language (but hard to be sure without asking them what they meant).

Look how we never had much issue with people asking how to write an email message or a letter in another language, and we do not have a guidance for that, even though users have been able to use all characters for emails and letters since the start. I reckon it is the same for text messages.

Edit template page is quite busy, and we saw in user research that users struggled to find guidance they need because there is too much of it - that's why I am reluctant to add this new guidance without any solid evidence that it's needed.

quis:
I agree with everything @CrystalPea has said, especially
> we never had much issue with people asking how to write an email message or a letter in another language, and we do not have a guidance for that

saimaghafoor:
Closing this PR - w can revisit this if/when there's evidence users are struggling.

PR #6076, opened 2026-09-09
Refactor letter and attachment upload views to do more work in the forms

quis:


quis:
Think this is slightly better but still horrible. Wouldn’t want to put anyone through the pain of trying to review it.

PR #6077, opened 2026-09-10
Fix test environments base urls

DilwoarH:
The base URLs were expected to be localhost but this is not the case in docker containers.

This change fixes it by using the actual environment variable instead of hardcoded values.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,7 +4,8 @@ testpaths = tests
 log_level = 999
 env =
     NOTIFY_ENVIRONMENT=test
-
+    ADMIN_BASE_URL=http://localhost
+    DOCUMENT_DOWNLOAD_API_HOST=http://localhost:7000
 filterwarnings =
     error:Applying marks directly:pytest.RemovedInPytest4Warning
 addopts = -p no:warnings


quis:
This makes the test assertions weaker because you could mess up setting `ADMIN_BASE_URL` and the tests wouldn’t catch it. I don’t like referring to config in tests for this reason. 

I think it would be better to investigate why `ADMIN_BASE_URL` is getting set to the wrong thing when running the tests in Docker.

DilwoarH:
> This makes the test assertions weaker because you could mess up setting `ADMIN_BASE_URL` and the tests wouldn’t catch it. I don’t like referring to config in tests for this reason.
> 
> I think it would be better to investigate why `ADMIN_BASE_URL` is getting set to the wrong thing when running the tests in Docker.

The config comes from the environment variables set using this template:
https://github.com/alphagov/notifications-local/blob/main/notify-admin.env.tmpl#L12

When the application loads up, it will set the base url via the following:
https://github.com/alphagov/notifications-admin/blob/main/app/config.py#L15

When the test runs, it instantiates the application using the environment variables.

quis:
I understand how environment variables work. 

Looks like `notify-local` is setting the wrong values when running the tests in that case. For example this should be `test`:
https://github.com/alphagov/notifications-local/blob/7ce44d66ac3524fc525261cdfdeb747242b8e7de/notify-admin.env.tmpl#L1

I don’t know enough about how `notify-local` works to know how you’d set different environment variables for local development and when running the tests

DilwoarH:
> I understand how environment variables work.
> 
> Looks like `notify-local` is setting the wrong values when running the tests in that case. For example this should be `test`: https://github.com/alphagov/notifications-local/blob/7ce44d66ac3524fc525261cdfdeb747242b8e7de/notify-admin.env.tmpl#L1
> 
> I don’t know enough about how `notify-local` works to know how you’d set different environment variables for local development and when running the tests

I believe the test environment overwrites this here: https://github.com/alphagov/notifications-admin/blob/main/pytest.ini#L6

If we want to hardcode the tests, we should pass in the variables otherwise it will just use the ones on the host machine.

quis:
Adding the variables to `pytest.ini` is a better idea, let’s do that

klssmith:
I think hardcoding environment variables to be the Docker versions in `pytest.ini` is going to break my local test set-up. But since no one has mentioned an issue with tests failing locally before, maybe we could check how everyone is running them - e.g. are they running `make test`, or `make test-with-docker`? I didn't think people used notifications-local for unit testing a single app - only as an easy way to run all the apps together

DilwoarH:
> I think hardcoding environment variables to be the Docker versions in `pytest.ini` is going to break my local test set-up. But since no one has mentioned an issue with tests failing locally before, maybe we could check how everyone is running them - e.g. are they running `make test`, or `make test-with-docker`? I didn't think people used notifications-local for unit testing a single app - only as an easy way to run all the apps together

The behaviour should be same in the test environment whether someone runs it locally or docker.

I have pushed the required changes, these should fix everything.

klssmith (approved):

PR #6078, opened 2026-09-10
Update API keys page

kr8n3r:
Update api key page:

- add created_at and created_by data to each key
- convert the keys list from a table to an unordered list
- update design based on https://github.com/alphagov/notify-design-archive/issues/18 

**Before**
Screenshot 2026-09-10 at 12 28 18

**After**
Screenshot 2026-09-10 at 11 42 13

### DIFF
--- a/app/templates/views/api/keys.html
+++ b/app/templates/views/api/keys.html
@@ -1,8 +1,8 @@
 {% extends "withnav_template.html" %}
-{% from "components/table.html" import list_table, field %}
 {% from "components/page-header.html" import page_header %}
 {% from "govuk_frontend_jinja/components/button/macro.html" import govukButton %}
 {% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/tag/macro.html" import govukTag %}
 
 {% block service_page_title %}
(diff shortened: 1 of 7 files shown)


app/templates/views/api/keys.html
```diff
@@ -16,46 +15,40 @@
...
-          <a class="govuk-link govuk-link--destructive" href='{{ url_for('main.revoke_api_key', service_id=current_service.id, key_id=item.id) }}'>
-            Revoke<span class="govuk-visually-hidden"> {{ item.name }}</span>
-          </a>
-        {% endcall %}
-      {% endif %}
-    {% endcall %}
-  </div>
+
+          {% if item.key_type == 'normal' %}
+            <strong class="govuk-tag govuk-tag--turquoise api-key__type">Live – sends to anyone</strong>
```
tombye: Should we use [the design system tag component](https://design-system.service.gov.uk/components/tag/) here instead of just the classes?

tombye (approved):
Nice work, all looks good to me, just one comment about how the tag component comes in but should be good to go once that's been addressed.

tombye (approved):
Thanks for addressing the comment. All good to go now 👍🏻

PR #6079, opened 2026-09-10
Add a new page for agreeing to the free allowance terms

klssmith:
This adds a new page where users can tick that they agree to the terms of the text message free allowance. Ticking the box saves the data in the `is_unique` column. This page is not linked to from anywhere yet, however if people do come across this page it isn't an issue - there's nothing they shouldn't see, and clicking the box will save the data correctly.

If the terms have already been accepted, the content on the page is the same, but the form is hidden.

**Screenshots**
Screenshot 2026-09-10 at 13 43 27

Screenshot 2026-09-10 at 14 14 19

Screenshot 2026-09-10 at 13 52 36

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -2103,6 +2103,13 @@ class ServiceSmsSenderForm(StripWhitespaceForm):
     is_default = GovukCheckboxField("Make this text message sender ID the default")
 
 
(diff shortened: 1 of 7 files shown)


kr8n3r (approved):
looks all looks good to me

PR #6081, opened 2026-09-11
Convert security keys table to list

kr8n3r:
Utilising the Summary List with a tweak as we use "manage" link  as the list value.

Variant of the notify-list style class sets the width of the value as well as align the text to the right.

Screenshot 2026-09-11 at 11 42 05


tombye (approved):
Looks good to me, and suits a summary list well 👍🏻

PR #6087, opened 2026-09-14
Improve display of datetimes and usernames on API keys page

quis:
I reckon what people mostly care about is how old the key is, especially when it comes to knowing if it should be revoked. In that case saying 2 weeks or 10 years is more helpful than an exact date and time.

For showing when the key was revoked: the exact time is useful to know if investigating an incident, for example when did things break? When was the vulnerability mitigated? We can still be a bit more friendly by displaying ‘today’ and ‘yesterday’ instead of the calendar date.

We don’t need to pass `date_prefix` to `format_datetime_human` because it defaults to `"on"`.

***

_Created by Archived user_ is also bit clunky, grammar-and-capitalisation-wise. We can be a bit better at handling this special case too.


kr8n3r (approved):
nice

PR #6090, opened 2026-09-15
Add an extra go live check for having read the SMS terms, and conditions for turning SMS on

klssmith:
## Background 
We are adding a new check before a service goes live - they need to have read the terms of the free allowance if they intend to send SMS. If a service goes live and has indicated they won't send SMS, we want to turn the SMS permission off and to require certain conditions to be met if it is turned on in future. This is something we recently started doing for emails.

## Where we're up to
A [new database column](https://github.com/alphagov/notifications-api/pull/4956) was created to store whether the service name has been confirmed, so that we can use the existing `is_unique` column solely for the new page to confirm the terms of the agreement have been read. The [page to confirm the terms of the agreement exists](https://github.com/alphagov/notifications-admin/pull/6079), but is not linked to from anywhere yet.

Screenshot 2026-09-15 at 12 14 23

## This PR
This PR implements everything else, including:
- Updating the go-live checklist to add in a check that the free allowance terms have been agreed (if necessary)
- Updating content on the "Check your service name" page so that it no longer refers to free text message allowance
- ... which means that the "Check your service name" page can solely update the `confirmed_service_name` column
- Always keeping some SMS settings showing on the settings page, even when SMS are turned offf
- Setting requirements for turning SMS on if turned off

Screenshot 2026-09-15 at 12 23 44

**See the individual commit messages for the full details**

---

### Screenshots
Screenshot 2026-09-15 at 12 26 39

Screenshot 2026-09-15 at 12 26 49

Screenshot 2026-09-15 at 12 28 45


app/templates/views/service-settings/set-sms.html
```diff
@@ -29,24 +30,66 @@
...
-        </h2>
-      <p class="govuk-body">
-          GOV.UK Notify will hide:
+
+      {% if current_service.has_permission("sms") %}
+        {% call form_wrapper() %}
+          {{ form.enabled }}
+          {{ page_footer('Save') }}
+        {% endcall %}
+          <h2 class="govuk-heading-m">
```
tombye, line 39: Having the following content below the submit button might mean some users miss it. 

Image

This kind of issue was mentioned in previous accessibility audits on the 'Edit template' page, where we put all formatting guidance in the same place.

Edit: discussed outside of this pull request, the page already has this content so this is outside the scope of these changes

tombye (commented):
This all makes sense to me, and the frontend changes work well.

I had 1 comment about the placement of some guidance on the 'Turn off SMS' page but otherwise this is good to go.

whpearson (approved):

PR #6092, opened 2026-09-15
Fix a couple `None`-defaults on `request.args.get()`

quis:
Assuming `request.args.get("foo")` will always be a string is incorrect when no such query string parameter exists (it will be `None`). We should specify `""` as the default so downstream code only has to handle strings.


quis:
Closing in favour of https://github.com/alphagov/notifications-admin/pull/6105 and https://github.com/alphagov/notifications-admin/pull/6106

PR #6097, opened 2026-09-16
fix(formatters): compare scheduled dates against London time, not UTC

### LINKED ISSUE #6096: Scheduled day labels compare against UTC instead of London time
Hey! Found that get_human_day in app/formatters.py compares the message date (London) against a plain UTC now, so around midnight a message sent today shows as tomorrow. The rest of the file already uses the London-aware helper. Happy to PR the one-line fix if that helps.

soroush5:
get_human_day compared the message date (London) against a plain UTC now, so around midnight a message sent today showed as tomorrow. One-line fix using the London-aware helper the rest of the file uses.

Verified: old says tomorrow, new says today at 23:30 UTC; full cases green.

Fixes #6096


kr8n3r:
Hi, that's but this is not something we are considering at the moment

PR #6099, opened 2026-09-17
fix(invites): show friendly page when inviting an existing team member

### LINKED ISSUE #6098: Inviting someone already on the team fires a duplicate invite
Hey! Found that typing the email of someone already on the team into the invite form fires a duplicate invite and flashes Invite sent, instead of showing the friendly already-a-team-member page the user list path shows. Re-inviting or mistyping a colleague's email hits this. Happy to PR a fix that checks the typed email against current team members first if that helps.

soroush5:
Typing the email of someone already on the team fired a duplicate invite with a misleading Invite sent flash. The invite form now checks the typed email against current team members first and shows the same already-a-team-member page the user list path shows.

Verified: new regression test fails before and passes after, full manage-users suite green.

Fixes #6098


kr8n3r:
Hi, that's but this is not something we are considering at the moment

PR #6100, opened 2026-09-18
Make a single API call when a service goes live

klssmith:
When making a service live, we currently turn off emails in some scenarios
and will soon turn off text messages too if certain criteria are met. This
was being done as separate requests to update the service - a request to
the API to make it live, followed by a request to the API to remove the
email permission. However, making multiple updates to a service in the
same Flask request can cause issues because `current_service` is only
loaded once - before the request. This can lead to a situation where the
first API request updates the service but `current_service` doesn't have
those changes (because we're still in the same request, so it hasn't been
reloaded). We can then use outdated values from `current_service` when
making the second API request, which leads to the first change to the
service being undone.

This changes the `update_status` methods of the Service model and the
service_api_client to optionally take a list of permissions to change at
the same time as changing the service status. We then make use of this
when making a service live to ensure that there is only one API request when
updating the service.


whpearson (approved):
Looks good. Tested it locally

PR #6107, opened 2026-09-20
`add_service` view: typing fix

risicle:
Reduces `mypy` errors by 2.


quis (approved):

app/main/views/add_service.py
```diff
@@ -59,6 +59,8 @@ def add_service():
 @user_is_gov_user
 def name_service():
     default_organisation_type = current_user.default_organisation_type
+
+    form: CreateNhsNotifyServiceForm | CreateServiceForm
```
klssmith, line 63: I realise I'm commenting too late for this PR, but did we also want the third type of form here too, `CreateNhsServiceForm`?

risicle: `CreateNhsServiceForm` is a subtype of `CreateServiceForm`

PR #6109, opened 2026-09-21
templates views: typing fixes

risicle:
Reduces `mypy` errors by 12.


app/main/views/templates.py
```diff
@@ -1314,7 +1327,7 @@ def _save_letter_attachment(*, service_id, template_id, upload_id, original_file
 @user_has_permissions("manage_templates")
 def view_invalid_letter_attachment_as_preview(service_id, file_id):
     try:
-        page = int(request.args.get("page"))
+        page = int(request.args.get("page"))  # type: ignore[arg-type]  # we're prepared for ValueError
```
quis:
```suggestion
        page = int(request.args.get("page", ""))
```

because
```python
int("any str")
>>> ValueError

int(None)
>>> TypeError
```

risicle: Ah phooey

quis (approved):

PR #6110, opened 2026-09-21
 Replace table-field-error-label with sms-message-status--error

kr8n3r:
'table-field-error-label'  is part of legacy table styles, which we're replacing with design system equivalent so these style will no longer be avaialble / wil lbe rewritten for the new tables specifically.

For this case (which is not a table) we're replacing this with modifier class of the 'sms-message-status'


quis (commented):
Probably needs bold too, per 
https://github.com/alphagov/notifications-admin/blob/43833fb4f8b64e116dfc347b594ceeeae97539e6/app/assets/stylesheets/components/table.scss#L209-L213

kr8n3r:
> Probably needs bold too, per
> 
> https://github.com/alphagov/notifications-admin/blob/43833fb4f8b64e116dfc347b594ceeeae97539e6/app/assets/stylesheets/components/table.scss#L209-L213

it does indeed. thanks, updated

quis (approved):

tombye (approved):
Already approved but looks good by me too 👍🏻

PR #6114, opened 2026-09-21
Validate template email files for email addresses

quis:
If a user uploads a spreadsheet which has a column full of email addresses it probably means they are trying to send the contents of the template to those email addresses.

This is a problem if the route they have gone down is the ‘Attach files’ journey, because then what will happen is that spreadsheet gets attached to the template as a downloadable link. If the template is later sent, each recipient will have access to that list of email addresses.

This commit adds a validation check which tries to catch something which looks a bit like a spreadsheet of recipient data.

It’s quite broad. It looks for any column which has at least 5 email addresses anywhere in the first 100 rows.

If this results in too many false positives we should hear from users on support, and we can dial it back.

Other algorithms I thought about and discarded were:
- email addresses in more than _n_ rows (broader)
- email addresses in more than _n_ cells (even broader still)
- some kind of density check (would be more complex, for example any column where more than _n_% of the values are email addresses based on a weighted average)
- scanning the whole sheet and looking for runs of email addresses (this could get quite slow given we support up to 100,000 rows × 1,000 columns = 100,000,000 cells)
- AWS Macie (setting this up would be a [much more involved piece of work](https://trello.com/c/JdRGDSzU/569-aws-macie-for-identifying-pii))

***

image


rparke:
I wonder, do we want to be slightly stricter and disallow .txt files too? Might be added complexity for no real benefit.

quis:
By disallow do you mean:
- stopping them from being uploaded?
- stopping them from containing email addresses?

rparke:
> By disallow do you mean:
> 
> - stopping them from being uploaded?
> 
> - stopping them from containing email addresses?

Sorry, from containing email addresses.

quis:
If we did I’d do that as a follow up because of the extra complexity it will introduce

rparke:
> If we did I’d do that as a follow up because of the extra complexity it will introduce 

No that's fair

PR #6115, opened 2026-09-23
block phone numbers based on block_ofcom_protected_block service permission.

rparke:
We want to block sending to ofcom protected blocks only if a service asks to opt in to mitigate AIT (we don't want this to be a blanket setting until we are confident in the data OFCOM are providing, and are confident in a strategy to keep it up to date)

This needs to be merged before https://github.com/alphagov/notifications-api/pull/4977


risicle:
How does this interact with jobs?

rparke:
> How does this interact with jobs?

good spot

app/main/forms.py
```diff
@@ -249,8 +249,10 @@ class PhoneNumber(GovukTextInputFieldMixin, TelField):
     input_type = "tel"
 
 
-def valid_phone_number(label="Mobile number", international=False, sms_to_uk_landline=False):
-    if not (sms_to_uk_landline or international):
+def valid_phone_number(
+    label="Mobile number", international=False, sms_to_uk_landline=False, block_ofcom_protected_blocks=False
```
risicle: Not your fault but what on earth are the semantics of these other arguments? Looks like someone got rid of the `allow_` from their names for brevity but in the process made them incomprehensible.

rparke: They're quite inconsistent all over the place. Sometimes the argument matches the permission name, sometimes it doesn't sometimes it's a pluralised version, sometimes it has the allow verb etc.

In this case we seem to have all 3 going on at once 😢. Could attempt to fix that here?

risicle: No not today

